PR #28, opened 2019-06-30
Alice-Faber demo update

andrewwhitehead:


swcurran:
@ianco @nrempel - can one or both of you review this PR?  Like it before we transfer the repo.

PR #41, opened 2019-07-04
Readme updates

nrempel:
Contains changes from #40 

I haven't been able to make the table formatting better in the readme. If anyone knows a trick to making the table horizontally scrollable, let me know.


nrempel:
> Looks good. Assume we'll be able to make the creation of the distro package and publishing to PyPi a "post-create-release" triggered action? Not a hold up, but I assume that's the approach.

That's the goal. We will need to do some investigation on the approach for this. Ideally, creating a new github release also publishes a pypi build for that release.

PR #46, opened 2019-07-08
Check for readiness instead of activeness in credential admin routes 

nrempel:


andrewwhitehead:
Should we apply the same change for presentation routes? It looks like the basicmessage handler also checks the active state.

nrempel:
@andrewwhitehead was just going to ask about protecting other routes using this. I'll add that.

PR #56, opened 2019-07-10
Aries 0036

sklump:
Aries#0036 support toward JIRA IS-1281 contribution, work in progress.


sklump:
Will resubmit from more modern base

PR #60, opened 2019-07-11
Aries#0036 support as JIRA IS-1281 contribution, based vs. 2019-07 hy…

sklump:
…perledger aries-cloud-agent


sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/130

supercedes and augments with Aries#0037.

PR #63, opened 2019-07-11
Minor fix to enable performance demo on play-with-docker

andrewwhitehead:


WadeBarnes:
@swcurran, I just tested Play With Von and it’s working.

swcurran:
Thanks - Weird.  I'll try it in the office.  Still not working for me -
multiple browsers.

I just get the yellow banner across the top.

andrewwhitehead:
Merged as part of #68

PR #75, opened 2019-07-16
Use a common base class for storage-based data records

andrewwhitehead:
This changeset adds the BaseRecord class with support for webhooks and logging of state changes. It removes duplicate code from the ConnectionRecord, CredentialExchange, and PresentationExchange classes and moves webhook handling out of the corresponding Manager classes.

### DIFF
--- a/aries_cloudagent/messaging/models/base_record.py
+++ b/aries_cloudagent/messaging/models/base_record.py
@@ -25,14 +25,16 @@ class BaseRecord(BaseModel):
     class Meta:
         """BaseRecord metadata."""
 
-    RECORD_ID_NAME = "record_id"
+    RECORD_ID_NAME = "id"
     RECORD_TYPE = None
     WEBHOOK_TOPIC = None
     LOG_STATE_FLAG = None
+    CACHE_TTL = 60
+    CACHE_DISABLED = False
 
     def __init__(
         self,
-        record_id: str = None,
+        id: str = None,
         state: str = None,
         *,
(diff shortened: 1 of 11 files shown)


aries_cloudagent/messaging/models/base_record.py
```diff
@@ -0,0 +1,367 @@
...
+from ..util import datetime_to_str, time_now
+
+
+class BaseRecord(BaseModel):
+    """Represents a single storage record."""
+
+    class Meta:
+        """BaseRecord metadata."""
+
+    RECORD_ID_NAME = "record_id"
```
nrempel: This is a nit - but do you think we could just call this "id"? Is there a reason to name it explicitly?

andrewwhitehead: It's mainly for compatibility with the existing stored records where it's named connection_id, credential_exchange_id, etc. The ID isn't stored that way but it would still be a lot of references to update.

andrewwhitehead: Also `id` is a python keyword..

nrempel: I think `id()` is a built-in function but it should be safe to use as a property on an object.

Right, forgot about the current stuff. However - arguably, id makes more sense in a vacuum and we're still early enough that we can introduce breaking changes.

What do you think about defaulting to using `id` and allowing it to be overridden?

PR #76, opened 2019-07-16
Support direct response for connection response message

andrewwhitehead:
These changes allow the initial connection response message (or problem report) sent after a connection request is received to be delivered by direct response, in cases where the sender has provided a transport decorator and is using a supported transport.


nrempel:
@andrewwhitehead the logic looks sound to me! Can you update the PR description with a description of the use-case for this change?

PR #82, opened 2019-07-17
Connection invitation fixes and improvements; support DID-based invitations

andrewwhitehead:
- Return a problem report when a connection invitation is submitted as an agent message
- Remove support for connection invitations as HTTP GET requests
- Rename parameters to `--auto-accept-invites` and `--auto-accept-requests`
- Allow the creation and reception of public (DID-based) connection invitations when `--public-invites` is enabled
- Publish the endpoint to the ledger when a public DID is used and the ledger is available


nrempel:
Looks good to me!

PR #84, opened 2019-07-17
Allow demo scripts to run outside of Docker; add command line parsing

andrewwhitehead:


andrewwhitehead:
For native Python run with `GENESIS_URL=.. python3 -m runners.alice` (for example) from the `demo` directory.

demo/runners/acme.py
```diff
@@ -99,18 +104,24 @@ def connection_ready(self):
...
-                    random.randint(1, 101),
-                    random.randint(1, 101),
-                )
-            )
+            pass
             # TODO define schema
-            #(schema_id, credential_definition_id) = await agent.register_schema_and_creddef(
-            #    "employee id schema", version, ["employee_id", "name", "date", "position"]
-            #    )
+            # version = format(
```
nrempel: Are these comments safe to delete?

andrewwhitehead: @ianco ?

swcurran:
@andrewwhitehead the instructions are not updated (and presumably I still need to run the `pip3 install...`) command, which I can do after.  We will also have to make sure the instructions are adjusted to where the code is now placed.  All works nicely in Docker and Play with Docker.

When I do run it using python, I get an python3-indy module not installed error message. I assume we need a pointer to somewhere in indy-sdk land for how to build the indy module and put it into our instructions.    I'd say we go with this and I'll add a line that says "You have to build indy-sdk `indy` library" and ask for help from someone (Sam?) to provide the instructions.

However, even if we do that, we still won't have it working, because Sam has fixed this issue (the folders) but is speculating that the startup parameters are not right.  Don't know if that is correct...

Perhaps that fastest thing we can do is get on a session with Sam to figure out why it is not working on his system and thus, with python generally?  Any volunteers?

PR #85, opened 2019-07-17
Don't extract decorators with names matching the 'data_key' of defined schema fields

andrewwhitehead:
@sklump Re: Aries#36 support

### DIFF
--- a/aries_cloudagent/messaging/agent_message.py
+++ b/aries_cloudagent/messaging/agent_message.py
@@ -61,7 +61,9 @@ class AgentMessage(BaseModel):
         else:
             self._message_id = str(uuid.uuid4())
             self._message_new_id = True
-        self._message_decorators = _decorators or DecoratorSet()
+        self._message_decorators = (
+            _decorators if _decorators is not None else DecoratorSet()
+        )
         if not self.Meta.message_type:
             raise TypeError(
                 "Can't instantiate abstract class {} with no message_type".format(
--- a/aries_cloudagent/messaging/agent_message.py
+++ b/aries_cloudagent/messaging/agent_message.py
@@ -340,7 +340,7 @@ class AgentMessageSchema(BaseModelSchema):
             ValidationError: If there is a missing field signature
 
         """
-        processed = self._decorators.extract_decorators(data)
+        processed = self._decorators.extract_decorators(data, self.__class__)
 
         expect_fields = resolve_meta_property(self, "signed_fields") or ()
         found_signatures = {}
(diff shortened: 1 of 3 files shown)


sklump:
agent_message.py line 64 check is still wrong: should test against non, not falsiness. All DecoratorSet descendants passed in are empty (false-ish), since they are instances not classes.

sklump:
Unless I've misread the intent, this approach still needs about 20 lines of boilerplate per descendant to replace one parameter, skip_attrs, in AgentMessageSchema.extract_decorators() call to self._decorators.extract_decorators().

Could taking skip_attrs into extract_decorators itself, like this:
@pre_load
def extract_decorators(self, data, skip_attrs: Sequence[str] = None):
    """..."""
    processed = self._decorators.extract_decorators(data, self.__class__, skip_attrs if skip_attrs else None)

then in descendant
@pre_load
def extract_decorators(self, data):
    return super().extract_decorators(data, ["skip", "these", "attrs"])

work OK? Apart from the one line delta, I think the rest of the processing ought to be identical in descendants.

sklump:
I will also propagate this change into my Aries#0036 PR by tomorrow morning.

nrempel:
> Unless I've misread the intent, this approach still needs about 20 lines of boilerplate per descendant to replace one parameter, skip_attrs, in `AgentMessageSchema.extract_decorators()` call to `self._decorators.extract_decorators()`.
> 
> Could taking skip_attrs into extract_decorators itself, like this:
> 
> ```
> @pre_load
> def extract_decorators(self, data, skip_attrs: Sequence[str] = None):
>     """..."""
>     processed = self._decorators.extract_decorators(data, self.__class__, skip_attrs if skip_attrs else None)
>     ...
>     return processed
> ```
> 
> then in descendant
> 
> ```
> @pre_load
> def extract_decorators(self, data):
>     return super().extract_decorators(data, ["skip", "these", "attrs"])
> ```
> 
> work OK? Apart from the one line delta, I think the rest of the processing ought to be identical in descendants.

@sklump I didn't mean to clobber your comment -

I do like the idea of keeping things declarative. @andrewwhitehead right now, the intent is that the sub-class of agent_message would override `@pre_load` and pass in `skip_attrs`?

What about something declarative like Stephen's approach, but checking `skip_attrs` in a `Meta` class  variable?

andrewwhitehead:
It's not meant to be necessary to override `extract_decorators`, I just added `skip_attrs` as an optional parameter there in case somebody's using it on non-Schema'd content. It should be sufficient to declare fields like the one in the test case: `handled_decorator = fields.Str(required=False, data_key="handled~decorator")` and have them automatically skipped when the decorators are loaded, to be parsed by the Schema itself.

sklump:
OK, I will try to get it to work and adapt accordingly by tomorrow morning.

PR #89, opened 2019-07-19
Document listing the Aries RFCs supported by ACA-Py and reference to …

swcurran:
…the list in the README


swcurran:
@andrewwhitehead - let me know if I have that mostly right, and especially if there are any caveats to add.  I removed from the list you gave me originally the l10n/i18n reference is I don't think that is well enough defined to implement yet.  Am I wrong?

swcurran:
@andrewwhitehead can you push this one ahead. I'll use branches from now on. Promise (mostly).

PR #100, opened 2019-07-25
Detailed Documentation of HowTo run demo agents locally on macOS

SmithSamuelM:
This adds a markdown document to the demos folder that details the install, configuration, and execution instructions for running local versions fo the Faber and Alice demos on macOS. It includes the missing information discovered during the process of getting it working.


swcurran:
Hi Sam,

Thanks!  Tons of information.  I've gone through it and took out the salient pieces and added it to the core document, including removing the references to using von-network as a method for running the ledger.  Note as well, that instead of using LEDGER_URL, I've suggested using the GENESIS_FILE=/path/to/file environment variable.  Glad to have you look at that PR, if you have a chance.

What is the purpose of the `faber_run.py` that was included in the PR?  I'm thinking it's not needed, as we did a fix to use the `runners*` file.

In merging this, I'll likely add a caveat at the top that the file was accurate and provides a good summary of the process at the time it was done. We are going to be changing some things - e.g. separating out the controller from the agent and launching them separately to better align with the architecture documents (as you point out is needed).  We are not going to be able to maintain this level of detail - particularly at the level of the versions of things.  Does that make sense?

Thanks again!  Have a great weekend!

SmithSamuelM:
I think that the docs for the tutorials should closely track the actual tutorials not the future intended tutorials. Much of the didactic value of the tutorial is lost if one spends time trying to run something that can’t yet work as documented.  This means more work updating the docs to match the code as the code evolves but it defeats the purpose of more community engagement if you frustrate devs otherwise. =).   The Ledger_URl was at the suggestion of IAN.  If you only set Genesis_URl I am not sure it will work.  I overlooked faber_run. My bad it’s not needed anymore

PR #117, opened 2019-08-02
Fix relative imports so can run as main entry point when debugging

SmithSamuelM:
From python docs:
   modules intended for use as the main module of a Python application
   must always use absolute imports.

In order to use a debugger with this agent we want to invoke it as a
main entry point but the relative import of .support then breaks.

A fix is to add  the path to the runners directory to the python path
so that we may use it with an absolute import instead.
e.g. from runners.support.

I cleaned up my branch to get rid of spurious commits that were reversed.

### DIFF
--- a/demo/runners/acme.py
+++ b/demo/runners/acme.py
@@ -4,8 +4,23 @@ import logging
 import os
 import sys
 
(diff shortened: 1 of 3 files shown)


andrewwhitehead:
Resolved via PR #129

PR #131, opened 2019-08-09
Fixes for TAA and public invite

dbluhm:
We ran into a couple of small issues recently; hopefully these are acceptable fixes.

### DIFF
--- a/aries_cloudagent/config/ledger.py
+++ b/aries_cloudagent/config/ledger.py
@@ -79,7 +79,7 @@ async def ledger_config(
         taa_info = await ledger.get_txn_author_agreement()
         if taa_info["taa_required"] and public_did:
             taa_accepted = await ledger.get_latest_txn_author_acceptance()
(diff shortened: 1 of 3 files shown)


andrewwhitehead:
@dbluhm What is the goal in adding the endpoint to the public invitation? It's not clear to me that doing that is even supported in the spec. Always adding it certainly doesn't seem necessary as it can be looked up on the ledger.

dbluhm:
Ah, that's mostly a hack so that the invite URL will be correctly formatted. Without it, it comes out as `?c_i=<invite b64>`. I can remove that commit as it's obviously not a permanent fix.

dbluhm:
There's another issue that I discovered. I fixed it but need to push (hopefully sometime this weekend or monday morning) for TAA. The current timestamps added to the TAA are too accurate and are rejected by the ledger.

dbluhm:
Removing some of the extra bits and reopening as a separate PR. Thanks for the feedback!

PR #133, opened 2019-08-10
TAA Fixes

dbluhm:
Fixed a couple of KeyErrors and fixed the timing of acceptance which was formerly too precise and was being rejected by the Indy ledger.


andrewwhitehead:
@dbluhm The syntax checker is failing on a couple long lines, I'd suggest running flake8 on those two files: `pip install flake8`; `python -m flake8 {path}`

PR #138, opened 2019-08-12
Cleanup RTD - add import mocks, fix some comment formatting in code and markdown

swcurran:

@andrewwhitehead - the code changes are all in comments.  Note how embedded JSON is to be defined for extraction in RTD:

- blank line before
- a line of "::"
- blank line
- the JSON indented further
- blank line after

Also, notice that I had to add some config mock imports, but they are internal to the aries-cloudagent-python code base.  Is there something in the code - perhaps imported, but not defined?

Thanks!


andrewwhitehead:
@swcurran The markdown and prompt-toolkit libraries were only recently added to requirements.txt so your environment might just need those dependencies to be added, instead of mocked. Can we filter out the tests directories completely? That seems like it would remove most of the additional mocks, plus I don’t really think we want them included.

swcurran:
I'll take a look at the tests thing. Your right that would be a big help.

My environment is up to date I think, but I'll remove those two and try
again.  Nto the end of the world either way.

swcurran:
@andrewwhitehead - please take another look. Should be OK for now.  The extra mocks are still there, but I don't think they matter and it makes it easier for me to test locally.

PR #158, opened 2019-08-28
Feature/142 outbound queue

TelegramSam:
There are some open questions:

1. There is a mechanism to clean out the queue for expired messages. I couldn't find any existing 'scheduled task' code within the project, so wanted to discuss where it should go.
2. I've set a 1-week message retention policy by default but we may need smarter defaults in place.
3. This does not yet retry message transmission, but that should be added as we determine how to do scheduled tasks per question 1.

I welcome feedback and anticipate making some changes before this PR is merged.


aries_cloudagent/messaging/socket.py
```diff
@@ -116,6 +116,12 @@ def select_outgoing(self, message: OutboundMessage) -> bool:
                 and message.reply_to_verkey in self.reply_verkeys
             ):
                 return True
+            if (
+                mode == self.REPLY_MODE_ALL
+                and any(True for k in message.target.recipient_keys
```
andrewwhitehead: I'm not sure that `message.target` is guaranteed to be set.

aries_cloudagent/conductor.py
```diff
@@ -277,9 +284,40 @@ def __init__(self, context_builder: ContextBuilder) -> None:
...
         )
         if socket:
+            # if a reply mode is present then a response is allowed over this connection
+            # wait till dispatch is complete,
+            # then check to see if any queued responses can be sent.
+            if socket.reply_mode:
+                await complete
+                await self.queue_processing(socket)
+
+            # close a single_response socket if no answer given by now.
```
andrewwhitehead: This method returns the async task so that it can be cancelled, which is why it uses a done callback instead of awaiting the task. The queue processing call could be added to the done callback.

aries_cloudagent/conductor.py
```diff
@@ -277,9 +284,40 @@ def __init__(self, context_builder: ContextBuilder) -> None:
...
         return complete
 
+    async def queue_processing(self, socket):
+        """
+        Interact with undelivered queue to find applicable messages.
+
+        Args:
+            socket: The incoming socket connection
+        """
+
```
andrewwhitehead: I think this method should check whether the socket is complete first (has `reply_mode` and is not `closed`) rather than checking in `inbound_message_router`.

aries_cloudagent/conductor.py
TelegramSam: I didn't use an additional done callback because the docs were inspecific about execution order and the queue processing must be complete before the dispatch_complete() is called. 

Are you proposing that `queue_processing()` be called from inside `dispatch_complete()`? That feels like an odd flow to me.

aries_cloudagent/conductor.py
TelegramSam: I think this suggested change depends on where we call `queue_processing()` per the previous comment. I'll wait for that resolution before resolving this.

TelegramSam: Interestingly, `socket.select_outgoing()` performs both of these checks, so it may not be needed in either place.

aries_cloudagent/messaging/socket.py
TelegramSam: fixed with additional checks

aries_cloudagent/conductor.py
andrewwhitehead: The method could return one task which awaits the dispatch then awaits `queue_processing`

aries_cloudagent/conductor.py
andrewwhitehead: I still think it would be best to check the state of the socket before going through the items in the queue, as most of the time we're using a single-response HTTP transport

aries_cloudagent/conductor.py
TelegramSam: I'm not sure I follow what you are proposing here. Can you elaborate a little more?

aries_cloudagent/conductor.py
TelegramSam: I had placed it above this method for the same reason, but maybe that's not the right place.

TelegramSam:
Thanks for finding the errant await.

PR #206, opened 2019-10-06
Remove old exchange records

nrempel:
This pull request automatically removes old credential exchange records after the exchange completes in order to keep down database bloat.

On the issuer agent, records are removed automatically after the exchange completes taking into account parents thread records that need to stick around.

On the holder side, parent exchange records are kept around for an hour before removing them. Child thread exchanges are removed immediately. There is no way to identify records that can be immediately deleted so we set an expiration of an hour for parent records.


andrewwhitehead:
@nrempel Looks good so far, some code duplication of course

PR #208, opened 2019-10-08
Admin api richer hints

sklump:
Richer form hints for requests and results in admin API swagger UI. Tighter marshmallow validation where obvious.


sklump:
Sorry Andrew, I see some problems still. I will touch it up and commit within the hour.

PR #213, opened 2019-10-11
Fix signing when not needed

dbluhm:
Avoid signing ledger requests in cred_def_id2schema_id method. This makes it so potential holders are not required to have a public DID to send a proposal to an issuer.


dbluhm:
Closing in favor of Andrew's more complete solution in #215

PR #222, opened 2019-10-15
Multiple proof referents

esune:
Added support for providing multiple proof referents to the present-proof API endpoint.

No unit tests were available, so some basic test were added for the functionality that was updated (alongside a couple of "bonus" ones 😉 ).

Tagging @nrempel for reviewing the change and the tests.


esune:
Also tagging @andrewwhitehead as possible reviewer.

PR #223, opened 2019-10-16
Cred def 2 schema

sklump:
Allow cred-def-id tagging via admin API
Accept seq# as schema-id for get-schema() via admin API
Cache schema by seq# and by schema-id


Sarangc1203:
Querying about a schema using its seq# still doesn't work.
The SwaggerUI doesn't accept an integer input.
I also tried using the curl request:
```bash
curl -X GET "http://__IP__:11021/schemas/42" -H "accept: application/json"
```
It gives me this error:
```bash
{"schema_id": ["Value 42 is not an indy schema identifier"]}
```

sklump:
You're right, good catch. I will get to it soon.

PR #226, opened 2019-10-21
Attach deco signature

sklump:
Allow for sign/verify on attachment decorator data (base64 variant only for the moment, can look at further support as required).

Details may need adjustment as RFC work solidifies.


sklump:
LGTM reports a false positive. I've moved, not added, wallet.util dependencies. I don't know what to do about this, if anything.

PR #239, opened 2019-10-28
Align DID and DIDDoc fileds with RFC 0160: must be did and did_doc

jovfer:
As defined in the RFC these fields must be `did` and `did_doc`.

https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol#attributes

### DIFF
--- a/aries_cloudagent/messaging/connections/models/connection_detail.py
+++ b/aries_cloudagent/messaging/connections/models/connection_detail.py
@@ -96,13 +96,13 @@ class ConnectionDetailSchema(BaseModelSchema):
         model_class = "ConnectionDetail"
 
     did = fields.Str(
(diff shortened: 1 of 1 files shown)


swcurran:
@andrewwhitehead @tmarkovski - should this be a fix to the code or to the RFC?  We need to look back at the indy-hipe and see if the text was changed in that transition.  We don't want to break existing implementations.

swcurran:
did and did_doc is definitely right.  But implementations... :-)

jovfer:
@dhh1128 made an investigation and recognize the problem in 0160 History for 1.0 version. It should has `DID` and `DIDDoc` fields.

andrewwhitehead:
RFC updated

PR #241, opened 2019-10-28
Align Signature decorator with RFC 0234: verkey must be in signers field

jovfer:
`signer` -> `signers`

https://github.com/hyperledger/aries-rfcs/tree/master/features/0234-signature-decorator#tutorial
https://github.com/hyperledger/aries-rfcs/blob/master/features/0234-signature-decorator/ed25519sha256_single.md#application

### DIFF
--- a/aries_cloudagent/messaging/agent_message.py
+++ b/aries_cloudagent/messaging/agent_message.py
@@ -218,11 +218,11 @@ class AgentMessage(BaseModel):
             raise BaseModelError(
                 "Field signature verification failed: {}".format(field_name)
             )
-        if signer_verkey is not None and sig.signer != signer_verkey:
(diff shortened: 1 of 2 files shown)


sklump:
Correct, but note that signature decorator is itself on the way out.

swcurran:
This change will break compatibility with aries-framework-dotnet unless implemented and deployed together.  Need to think about this, especially since ~sig is going away.

sklump:
The solution may be to change the RFC to fit the implementations, at the risk of a Lighthouse situation: https://en.wikipedia.org/wiki/Lighthouse_and_naval_vessel_urban_legend

jovfer:
The RFCs is updated to avoid not so motivated changes in the implementations
https://github.com/hyperledger/aries-rfcs/pull/298

PR #242, opened 2019-10-28
Reuse credential offers and requests when possible

andrewwhitehead:
Also adds tests and type info to the CredentialManager class, and allows create_offer to be called without a credential proposal.

### DIFF
--- a/aries_cloudagent/messaging/issue_credential/v1_0/manager.py
+++ b/aries_cloudagent/messaging/issue_credential/v1_0/manager.py
@@ -304,7 +304,7 @@ class CredentialManager:
 
     async def create_request(
         self, credential_exchange_record: V10CredentialExchange, holder_did: str
-    ):
+    ) -> Tuple[V10CredentialExchange, CredentialRequest]:
         """
         Create a credential request.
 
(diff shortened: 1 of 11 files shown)


aries_cloudagent/messaging/issue_credential/v1_0/messages/credential_issue.py
```diff
@@ -57,6 +57,13 @@ def indy_credential(self, index: int = 0):
         """
         return self.credentials_attach[index].indy_dict
 
+    @classmethod
+    def wrap_indy_credential(clss, indy_cred: dict) -> AttachDecorator:
```
sklump: cls is conventional in place of clss

aries_cloudagent/messaging/issue_credential/v1_0/manager.py
```diff
@@ -73,6 +63,12 @@ def context(self) -> InjectionContext:
 
         """
 
+        credential_definition_id = credential_proposal.cred_def_id
```
sklump: The proposal will not include a cred def id in future - rather, an issuer and schema id - so this will need some touching up at that time. Until then, it's fine.

PR #250, opened 2019-11-05
Refactor protocol support from aries_cloudagent.messaging to aries_cloudagent.protocols

andrewwhitehead:


andrewwhitehead:
@nrempel I think this might be a good first step, saving more refactoring for future PRs

PR #254, opened 2019-11-06
Output timing information to log file when --timing-log is provided

andrewwhitehead:


swcurran:
Weird.  On linux when the ../logs folder does not exist, it creates it owned by root:root and gets a permission error.  If it exists and is owned by my account, it still doesn't work. If I change it to be chmod'd to uga+w (adding the all), it works.  That seems like more effort than necessary.  Not sure how to fix it, but there you go...

nrempel:
@andrewwhitehead can you add a timing collector for the response time of a webhook in this PR? I'll work on doing the work in the webhook handler asynchronously.

andrewwhitehead:
@nrempel That info should be collected as webhook:POST

swcurran:
@andrewwhitehead I tried to push this to your branch, but it didn't work.  Ah well...

How about this addition to `./run_demo`

```bash
elif [ "$AGENT" = "performance" ]; then
	AGENT_MODULE="performance"
	AGENT_PORT=8030
	AGENT_PORT_RANGE=8030-8038
	if [ "$1" == "--timing" ]; then
		if [ ! -d "../logs" ] || [ "$(ls -ldh ../logs | grep dr..r..rwx)" == "" ]; then
			echo "Error: To use the --timing parameter, the directory '../logs' must exist and all users must be able to write to it."
			echo "For example, to create the directory and then set the permissions use: 'mkdir ../logs; chmod uga+rws ../logs'"
			exit 1
		fi
	fi	   
```

To make it work nicely, the user has to precreate the folder and set the permissions.  The "s" on the permissions is so that files below are inherited so that the user can delete the files after creation.

The problems are because docker is creating the folder when the `-v` option is specified - so root owns it, and the -s is needed because otherwise the docker user in the script (likely UID 1000) owns the file and the linux user running the script can't delete it.

PR #256, opened 2019-11-07
Add support for locking a cache key, to avoid overlapping expensive operations

andrewwhitehead:
Now done in a standard way instead of using polling or other methods


swcurran:
Can you get someone else to review/approve this? I'm on a plane...

PR #259, opened 2019-11-08
broaden aries-rfc-36 cred proposal to underspecify cred def, allow is…

sklump:
Allow for cred proposal underspecification of cred def id, only lock down cred def id at issuer on offer. Sync up api requests to Aries RFC-36 verbiage.


sklump:
Oh! I totally forgot about the demo. It may need light re-touching. I will patch that up Tuesday.

PR #272, opened 2019-11-20
minor update to docker agent start command

smohan-dw:
```PORTS="5000:5000 10000:10000" ./scripts/run_docker ``` -> ```PORTS="5000:5000 10000:10000" ./scripts/run_docker start```


andrewwhitehead:
Thanks!

PR #276, opened 2019-11-21
allow additional endpoints to be specified

TelegramSam:
Signed-off-by: Sam Curren <telegramsam@gmail.com>


aries_cloudagent/protocols/connections/manager.py
```diff
@@ -838,13 +840,19 @@ def context(self) -> InjectionContext:
                     True,
                 )
                 routing_keys.append(rk)
-                svc_endpoint = service.endpoint
+                svc_endpoint = [service.endpoint]
```
andrewwhitehead: Should be `svc_endpoints` according to LGTM

TelegramSam: good catch. fixed.

PR #278, opened 2019-11-26
issuer includes please-ack on cred storage at issue; holder acks storage

sklump:
I've reintroduced credential_received, credential_acked and credential_stored as distinct states. Only the issuer gets credential_acked in an exchange record; the record is gone at the holder side at ack in any case; the holder's last web hook update is credential_saved.

The please-ack from the issuer side necessarily asks for ack on credential_stored (accepted) as per Aries RFC 0036 (issue-credential) state transition diagram (text disagrees for the moment - patience will provide convergence) and 0317 (please-ack) semantics (not the example - again, I hope to convince the example to converge with all germane Aries RFC pics and text).

For the moment this impl is a stake in the ground to show the concept working as I interpret the Aries RFC design intent.

### DIFF
--- a/aries_cloudagent/messaging/decorators/please_ack_decorator.py
+++ b/aries_cloudagent/messaging/decorators/please_ack_decorator.py
@@ -1,4 +1,4 @@
-"""The please-ack decorator to request acknowledgement."""
+"""The please-ack decorator (~please_ack) to request acknowledgement."""
 
 from typing import Sequence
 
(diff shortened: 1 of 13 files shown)


swcurran:
The "cred-stored" state is not a known state in the protocol.  On the other hand, I don't think any are, and the RFC should be updated to include that.  My recommendation is that the please_ack decorator should have a default behaviour, and the ability for the sender to specify "receipt" or "outcome", and that the "ack" should be sent either on receipt of the message or when the processing of the message is complete.  It is most useful for "end of protocol" explicit "acks", so an "outcome" type sent on the last step of the process. In this case, that would be on the "issue-credential" message.

The other question is whether in this protocol that the ACK is intended to be explicit and ALWAYS sent at the end of the processing of the "issue-credential" message.

To be resolved in the aries-rfcs process.

PR #285, opened 2019-11-29
Align problem-report schema with current RFC

andrewwhitehead:


swcurran:
Per Andrew, this is likely a breaking change that needs to be managed appropriately.

andrewwhitehead:
Out of date

Issue #280, opened 2019-11-28
Agents can get into a bad state and not recover


WadeBarnes:
## Summary

It would appear agents can get into a bad state and continue to run in that state, stopping traffic from reaching the receiving controller.

When agents get into a bad state, they either need to recover or stop processing traffic. If it's possible to implement a health check on the agents that monitors such conditions we could wire it up in OpenShift and have the pods taken out of service.

We can monitor:
- **readiness**; the pod is ready to receive traffic. OpenShift will only route traffic to the pod when this check passes.
- **liveness**; health condition of the pod (running, etc). OpenShift will kill the pod if this check fails a number of times (all configurable).

## Details

Using `bcgovimages/aries-cloudagent:py36-1.11-1_0.4.0-alpha`

Pod Configuration:
- 1 - ICIA (Indy Catalyst Issuer Agent) Controller
- 21 - ICIA Agents
- 17 - ICOB (Indy Catalyst OrgBook) Agents
- 16 - ICOB Controllers
- *This configuration has been confirmed to support a maximum issuance rate of ~2100 credentials per minute.*
- Data set is 2,090,092‬ credentials.

The following issues appear after several hours of sustained load.

Seeing a lot of these errors in the ICIA (Indy Catalyst Issuer Agent) Agents;
```
ModuleNotFoundError: No module named 'aries_cloudagent.protocols.problem_report.messages'
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/transport/inbound/http.py", line 98, in inbound_message_handler
    await self.message_router(body, self._scheme, single_response=response)
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/conductor.py", line 318, in inbound_message_router
    parsed_msg, delivery, connection, self.outbound_message_router
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/dispatcher.py", line 65, in dispatch
    message = await self.make_message(parsed_msg)
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/dispatcher.py", line 132, in make_message
    message_cls = registry.resolve_message_class(message_type)
[...]

PR #287, opened 2019-11-29
Dispatcher queue improvements

andrewwhitehead:
- Add environment variable for adjusting maximum active dispatcher tasks
- Track dispatcher timing via its task queue and log time spent in queue
- Fix issue with empty TaskQueue evaluating as false


WadeBarnes:
I'm still running performance (scaling) test on this, but it's been stable and has resolved the issues I reported regarding agents getting into a bad state; https://github.com/hyperledger/aries-cloudagent-python/issues/280

PR #288, opened 2019-11-29
Add create-proposal endpoint to the present-proof API namespace

esune:
Adding a new API endpoint `/present-proof/create-request` to support creating a Proof request without needing an active connection with an agent, as previously supported by the deprecated endpoint `/presentation_exchange/create_request`.

tests have been added following the ones that were already available for `presentation_exchange_send_free_request` and removing the test cases that are not required due to the missing `connection-id` requirement.

Please note my formatter has made changes to the files I edited (currently using black version `19.3b0`): let me know if this is expected, or what I should be doing to have a formatter behaviour that is consistent with the rest of the community.

Tagging @andrewwhitehead for review.

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v1_0/routes.py
+++ b/aries_cloudagent/protocols/present_proof/v1_0/routes.py
@@ -32,7 +32,6 @@ from .models.presentation_exchange import (
     V10PresentationExchange,
     V10PresentationExchangeSchema,
 )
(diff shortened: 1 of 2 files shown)


esune:
Changes in this PR depend on [this aca-py pull request](https://github.com/hyperledger/aries-cloudagent-python/pull/288).

PR #296, opened 2019-12-06
Add support for connection-less presentation requests

esune:
The pull request includes changes necessary to support presentation requests that do not rely on an existing agent-to-agent connection.

A use case for this is when [vc-authn-oidc](https://github.com/bcgov/vc-authn-oidc) creates a presentation request to allow users that hold credentials that satisfy it to access a protected resource.
In this case, the presentation request does not require a connection between the vc-authn-oidc agent and the user's agent to exist and be active: any user holding appropriate credentials should be able to receive the same presentation request and be authenticated when they respond with the correct credential claims.

Tagging @andrewwhitehead for review.

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v1_0/handlers/presentation_request_handler.py
+++ b/aries_cloudagent/protocols/present_proof/v1_0/handlers/presentation_request_handler.py
@@ -3,7 +3,6 @@
 from .....messaging.base_handler import (
     BaseHandler,
     BaseResponder,
-    HandlerException,
     RequestContext,
 )
 from .....holder.base import BaseHolder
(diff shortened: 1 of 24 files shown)


aries_cloudagent/protocols/present_proof/v1_0/handlers/presentation_request_handler.py
```diff
@@ -34,9 +33,6 @@ class PresentationRequestHandler(BaseHandler):
             context.message.serialize(as_string=True)
         )
```
andrewwhitehead: I'm not sure why the connection-ready check is being removed for this handler? There's also references to the connection record later which look like they would fail if there wasn't one identified.

esune: You are right about the references further down the code, good catch. It is very odd that an exception is not being thrown...

I am pretty sure the presentation request was failing because of this connection readiness check, and that is why I had to remove it. We could re-test the flow after reverting the change to double check if it ireally is necessary.

esune: You were correct. I re-tested everything after reverting this change and it seems to work just fine - not sure why the error was coming up in the logs (maybe I got confused between this and the presentation_handler).

PR #299, opened 2019-12-10
Restore changes from #289

esune:
Restore changes lost during conflict resolution between #288  and #289

@nrempel can you please check and confirm everything is there?


nrempel:
Looks good

PR #307, opened 2019-12-13
Update the RTD generated files and the README about how to generate; fix comment format in code

swcurran:


swcurran:
@andrewwhitehead - I updated a comment in a code module that should be checked to be sure I didn't break anything.  Pretty sure not :-).  The rest of this probably has to be just accepted and I'll test on the RTD site to make sure we are happy.

PR #309, opened 2019-12-16
Enable GENESIS_URL option for demo agents

andrewwhitehead:


andrewwhitehead:
@swcurran It's failing on trying to register the DID there, because it hasn't been given a LEDGER_URL to do the registration. I think this setting only works with the Alice agent because it doesn't need a public DID.

PR #316, opened 2020-01-02
Unit tests proto discovery

sklump:
Extend unit test coverage through discovery protocol


sklump:
Aw heck, somehow I've mashed ledger content into this imperfect PR. I will resubmit.

PR #322, opened 2020-01-07
Allow side loading alternate inbound transports

dbluhm:
I ran into a use case where there was a need to run both the "http" and "ws" transport on the same port. I wasn't sure a transport that combined both of these into one "http+ws" transport was something that was wanted upstream (if it is, I'm happy to address that separately) but I noticed that a couple of small tweaks would make it easy enough to side load my combined transport.

I'm curious how others feel about this approach.


aries_cloudagent/transport/inbound/manager.py
```diff
@@ -77,8 +77,15 @@ def register(self, config: InboundTransportConfiguration) -> str:
 
         """
         try:
+            if '.' in config.module:
+                package, *mod_parts = config.module.split('.')
```
andrewwhitehead: Maybe `package, module = config.module.rsplit(".", 1)`?

dbluhm: Thanks for the suggestion!

PR #323, opened 2020-01-09
revisit send-cred-def and associated non-secret record write

sklump:

### DIFF
--- a/aries_cloudagent/ledger/indy.py
+++ b/aries_cloudagent/ledger/indy.py
@@ -3,7 +3,6 @@
 import asyncio
 import json
 import logging
-import re
 import tempfile
 
 from hashlib import sha256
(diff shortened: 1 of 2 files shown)


aries_cloudagent/ledger/indy.py
```diff
@@ -486,95 +485,108 @@ def __init__(
 
         """
 
+        async def cred_def_in_wallet(wallet_handle, cred_def_id) -> bool:
```
nrempel: Could this function be useful on its own? Does it make sense to move it out to the class level?

aries_cloudagent/ledger/indy.py
```diff
@@ -486,95 +485,108 @@ def __init__(
...
+            """Create an offer to check whether cred def id is in wallet."""
+            try:
+                await indy.anoncreds.issuer_create_credential_offer(
+                    wallet_handle,
+                    cred_def_id
+                )
+                return True
+            except IndyError as error:
+                if error.error_code != ErrorCode.CommonInvalidStructure:
```
nrempel: Do we also need to check the content of the error message string? i.e., is it possible for `ErrorCode.CommonInvalidStructure` to be raised in other scenarios which we should allow to bubble up?

sklump:
> 
> 
> Added some general questions - may not be an issue.
> 
> In the case of cred def on ledger but not on wallet - this is a "broken" state? Do we need the ability to override this and publish a new cred def, or should we never be in this state?
The user is attempting to send a def to the ledger that already exists, but some other issuer has already sent - it's at best a misunderstanding and at worst an attempt to wreck the existing cred def. In any case, it must not proceed any further.

> 
> Does this approach have any issues with revocation?

No. Revocation registries are not cred defs.

aries_cloudagent/ledger/indy.py
sklump: No: it is useful only in establishing whether this issuer can use this cred def already. If it were useful at class level, it would go in the indy wallet, not the indy ledger. Once the cred defs per issuer are established, the non-secrets records already serve to tell what cred defs are available, but this tiny call helps bootstrap those into place.

aries_cloudagent/ledger/indy.py
sklump: The only other case would be a cred def id that did not parse as a cred def id, which would have been kicked out many layers above.

PR #325, opened 2020-01-14
Fix broken assignment in IndyProofReqPredSpecSchema

dbluhm:
Ran into this recently. Seems to be breaking requesting a presentation through a controller, resulting in a schema validation failure if `p_type` and `p_value` are specified.


swcurran:
@andrewwhitehead - please take a look at this ASAP.  Looks like it might be significant?

andrewwhitehead:
@dbluhm Could you update the PR branch? It won't let me do it automatically.

PR #328, opened 2020-01-15
Add GitHub action to publish releases to PyPI

ryjones:
Signed-off-by: Ry Jones <ry@linux.com>


ryjones:
As it turns out, this is one of the default workflows for GHA. I have already added the secrets.

PR #330, opened 2020-01-17
accommodate proofs with repeat cred defs, check for bait-and-switch b…

sklump:
…etween presentation proposal and request

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v1_0/routes.py
+++ b/aries_cloudagent/protocols/present_proof/v1_0/routes.py
@@ -429,61 +429,6 @@ async def presentation_exchange_send_proposal(request: web.BaseRequest):
 
 @docs(
     tags=["present-proof"],
-    summary="""
-    Creates a presentation request not bound to any proposal or existing connection
-    """,
-)
-@request_schema(V10PresentationRequestRequestSchema())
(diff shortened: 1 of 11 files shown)


aries_cloudagent/protocols/present_proof/v1_0/routes.py
```diff
@@ -427,61 +427,6 @@ class V10PresentationRequestSchema(Schema):
     return web.json_response(presentation_exchange_record.serialize())
```
andrewwhitehead: I'm not sure if `presentation_exchange_create_request` was removed intentionally, but I believe we use that in identity kit to create presentation exchanges that can be responded to without a connection.

swcurran:
We definitely need to be able to support that use case.  @sklump - if something needs to be added to or clarified in the aries-RFC 0037, please let me know.

sklump:
Sorry, I figured it was redundant with send...(). I will put it back and amend the PR Monday.

aries_cloudagent/protocols/present_proof/v1_0/routes.py
sklump: My mistake, I had thought it redundant vs. ...send...(). I am fixing and amending the PR.

PR #333, opened 2020-01-20
revisit send-cred-def and associated non-secret record write

sklump:


WadeBarnes:
This is a re-submission of https://github.com/hyperledger/aries-cloudagent-python/pull/323

PR #344, opened 2020-01-27
Verify raw vs encoding

sklump:
Visit raw vs. encoded attribute values, predicate threshold values in proof request/proof to fail proofs with tamper evidence before submitting to indy-sdk for verification.


sklump:
Please refrain from merging this just yet. I have to frig with endianism.

Issue #341, opened 2020-01-23
Evaluate effort and impact of implementing 0036 Credential Proposal V1.1


swcurran:
Because of the intricacies of the Aries Interop Profile versioning, I'd like to know what the impact would be of adding support for version 1.1 of [credential proposal in 0036](https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential#propose-credential). The change is outlined [here](https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential#11propose-credential). I would assume we would have to make the receipt of that message support the optional parameters, and handle them appropriately. Please confirm or correct my understanding, and the effort and impact of making that change.

Needs also to address this [PR](https://github.com/hyperledger/aries-rfcs/pull/393) to the 0036 RFC.

swcurran:
@andrewwhitehead Can please take a quick look at this issue and reply with a comment that answers my questions?

Thanks

sklump:
The business logic is already in place.

What's left is to change the `PROTOCOL_URI` to read `1.1` instead of `1.0` in `message_types.py`, and/or to change the V10 prefixes all over everything, which prevents duplicate names with the bronto-protocol, then finally to rename the subdirectory (subpackage) from `v1_0` to `v1_1`. I don't know whom that affects though.

Since any v1.0 proposal, other than message type, is a v1.1 proposal, whom does it hurt if we quietly change over on some evening?

sklump:
What I can do is duplicate the v1_0 subpackage into a v1_1 subdirectory, which I'll submit in a PR for evaluation.

PR #347, opened 2020-01-29
Support (Aries RFC 36) issue-credential protocol v1.1

sklump:
Address https://github.com/hyperledger/aries-cloudagent-python/issues/341


swcurran:
It turns out that the v1.1 upgrade was inappropriately handled for this. The version is on the protocol, not on the messages, so what should have happened (and what will be corrected) is that ALL of the messages will have the v1.1 tag on them.

A further complication is that the change that was made to go from v1.0 was to add Indy specific data to the proposal message.  However, we took pains to prevent Indy specific data in the other credential exchange messages -- through the use of attachments.  We need to relook at the change in that light.

More discussion in the community...

sklump:
What is indy-specific in the proposal message?

swcurran:
That's good about the message types. The RFC needs to be updated.

The Indy-specific fields are these ones:

-    "schema_issuer_did": "DID of the proposed schema issuer",
-    "schema_id": "Schema ID string",
-    "schema_name": "Schema name string",
-    "schema_version": "Schema version string",
-    "cred_def_id": "Credential Definition ID string"

sklump:
I guess the way forward here would be to base64-encode and wrap the five-or-six* criteria into a "filter~attach" attachment.

*  in or out: issuer_did.
IN:
+ one place for all filter criteria
- not backward-compatible with v1.0
OUT:
+ backward-compatible with v1.0
- need to visit two places to assemble filtration criteria: we're only on v1.1 and it's already got a tumour
An a

PR #355, opened 2020-01-31
Add CLI arg for preserving exchange records

dbluhm:
The toolbox taps into ACA-Py's credential exchange records to get a record of credentials issued; deleting these records once the exchange completes makes it a little harder to accurately show a list of issued credentials. The added command line argument gives the option to preserve those records after the exchange has completed.


swcurran:
@andrewwhitehead @nrempel - I leave it to you to decide on this one.  We had agreed that agent storage should NOT be used for protocol state objects except while in flight. If the controller wants to keep the state objects, that should be done by the controller.

That said, we used to support this, and it was useful for some demo use cases.

If we do add this want to make it clear that this should probably not be used in any production use cases and developers of controllers should plan accordingly to have the controller manage such data.

Your thoughts?

andrewwhitehead:
`--preserve-exchange-records` would seem to apply to presentation exchange records as well, maybe it should?

dbluhm:
> `--preserve-exchange-records` would seem to apply to presentation exchange records as well, maybe it should?

That would make sense to me; this actually points out one thing that seemed inconsistent to me. At least at the point that I branched from master, presentation exchange records are not deleted in the built in handlers for present-proof. I can add that and then similarly gate it with this setting but I also didn't want to introduce deletion by default when it wasn't already present.

swcurran:
@dbluhm - are you going to add the preservation of presentations that was discussed above?  I think there was confusion on that, which is why this was held up.

We'd definitely like that added before this feature is in a tagged release (e.g. 0.4.3).

PR #357, opened 2020-01-31
Tweaks to run_docker script for windows bash compatibility

esune:
Fixes compatibility issues with the `run_docker` script when running in bash under windows.

Tested with:
- GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)
- GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu) (WSL in Windows 10)


andrewwhitehead:
Please update `demo/run_demo` as well

esune:
@andrewwhitehead done! Sorry I missed that.

I also added `run_demo` to `.gitattributes` as the line endings were being switched to crlf in Windows (everything in the `scripts` folder is already handled correctly).

PR #373, opened 2020-02-11
add request schema to post /wallet/did/create

steveLiuu:
The request for seed isn't defined in `/wallet/did/create`. But I wanna generate did with the specific seed so I made some modifications.
Thx!

### DIFF
--- a/aries_cloudagent/wallet/routes.py
+++ b/aries_cloudagent/wallet/routes.py
@@ -71,6 +71,14 @@ class SetTagPolicyRequestSchema(Schema):
         description="List of attributes to set taggable for credential search",
     )
 
+class SetDidCreateRequestSchema(Schema):
+    """Request did seed."""
+
+    seed = fields.Str(
+        required=False,
+        description="seed for creating did",
+        example="00000000000000000000000000sample"
+    )
 
 def format_did_info(info: DIDInfo):
     """Serialize a DIDInfo object."""
@@ -159,6 +167,7 @@ async def wallet_did_list(request: web.BaseRequest):
 
 
 @docs(tags=["wallet"], summary="Create a local DID")
+@request_schema(SetDidCreateRequestSchema)
 @response_schema(DIDResultSchema, 200)
 async def wallet_create_did(request: web.BaseRequest):
     """
@@ -172,10 +181,12 @@ async def wallet_create_did(request: web.BaseRequest):
 
     """
     context = request.app["request_context"]
+    body = await request.json()
+    seed = body.get("seed")
     wallet: BaseWallet = await context.inject(BaseWallet, required=False)
     if not wallet:
         raise web.HTTPForbidden()
-    info = await wallet.create_local_did()
+    info = await wallet.create_local_did(seed)
     return web.json_response({"result": format_did_info(info)})
 
 


aries_cloudagent/wallet/routes.py
```diff
@@ -71,6 +71,14 @@ class SetTagPolicyRequestSchema(Schema):
...
     )
 
+class SetDidCreateRequestSchema(Schema):
+    """Request did seed."""
+
+    seed = fields.Str(
+        required=False,
+        description="seed for creating did",
+        example="00000000000000000000000000sample"
+    )
```
sklump: This should have a regex to check validity

sklump: I think it's good thinking though; I can plumb this in with appropriate unit tests and submit.

andrewwhitehead:
I didn't add this because the DID can be provisioned on the command line using a seed, and I didn't want to encourage people to pass private keys over the REST interface. What's the use case, supporting multiple identities?

sklump:
> 
> 
> I didn't add this because the DID can be provisioned on the command line using a seed, and I didn't want to encourage people to pass private keys over the REST interface. What's the use case, supporting multiple identities?

That's true - seeds are secrets

steveLiuu:
> I didn't add this because the DID can be provisioned on the command line using a seed, and I didn't want to encourage people to pass private keys over the REST interface. What's the use case, supporting multiple identities?

Oh, I didn't notice this security issue. The use case is when I need to operate the ledger with the `STEWARD` identity, there is no place to import DID or create the same steward DID. Or there is another way to reach my needs?
Thx!

aries_cloudagent/wallet/routes.py
steveLiuu: Thx! :)

swcurran:
I agree that this should not be part of the framework because of the danger of misuse.

Our belief on the right pattern for the deployment of an app in production (and also to be used for testing) is that there be a "provision" agent and "operational" agent, as follows:

- "provision" agent gets a config list of objects that must exist in the wallet and on the ledger, checks if they exist and if not creates them, and then exits.  This must be a higher security app, as it is creating things from secrets (e.g. seeds), access to DIDs that can write to the ledger, etc. It might also have to create the wallet, and need higher level access database credentials.
- "operational" agent gets a config list of objects that should be in the wallet and on the ledger, checks if they exist and if not, terminates with an error.  If the objects do exist, it runs the full agent logic with a "--read-only-ledger" setting to prevent creating anything else on the ledger. It is a lower security app and should not be creating things from secrets (and so need not know the secrets).

To manage this in testing, we recommend a wrapper script that has entry points to invoke either of the agents, and an entry point for testing that invokes both.

In production, with a production ledger, the "provision" script is only run when you need to create new objects on the ledger.  The wallet that is created MUST be properly guarded with an excellent backup and recovery mechanism. And, of course, the ledger is immutable.  No rewrites! :-)

swcurran:
Second comment - we cannot accept this PR as the DCO has not been done on the commits. Please see the contributors guidelines in the root folder.  All Hyperledger repos require all commits be signed for DCO.

PR #378, opened 2020-02-18
Pg scheme support

ianco:
Add support for additional postgres wallet schemes, such as "MultiWalletDatabase"

### DIFF
--- a/aries_cloudagent/wallet/plugin.py
+++ b/aries_cloudagent/wallet/plugin.py
@@ -2,7 +2,7 @@
 
 import logging
 import platform
-from ctypes import cdll
+from ctypes import cdll, c_char_p
 
 EXTENSION = {"darwin": ".dylib", "linux": ".so", "win32": ".dll", "windows": ".dll"}
 LOADED = False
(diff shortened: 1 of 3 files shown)


aries_cloudagent/wallet/plugin.py
```diff
@@ -29,6 +29,16 @@ def load_postgres_plugin(raise_exc=False):
...
+        if "wallet_scheme" in storage_config:
+            c_config = c_char_p(storage_config.encode('utf-8'))
+            c_credentials = c_char_p(storage_creds.encode('utf-8'))
+            result = stg_lib.init_storagetype(c_config, c_credentials)
+            if result != 0:
+                LOGGER.error("Error unable to configure postgres stg: %s", result)
+                if raise_exc:
+                    raise OSError(f"Error unable to configure postgres stg: {result}")
+                else:
+                    raise SystemExit(1)
```
andrewwhitehead: This code is only run once, the first time a PG wallet is opened. Does it need to run more than once if there's different storage types?

ianco: There should only be one storage type per postgres instance, so the code should only be run once.

Issue #244, opened 2019-11-01
Support new message type prefix


nrempel:
Per [aries-rc issue 225](https://github.com/hyperledger/aries-rfcs/issues/225#issuecomment-545645456), the prefix of all core Aries protocols will be "https://didcomm.org", instead of the current "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec". The transition approach to this change across the community has been proposed in [this PR](https://github.com/hyperledger/aries-rfcs/pull/348), which will eventually be merged as [this RFC](https://github.com/hyperledger/aries-rfcs/tree/master/features/0348-transition-msg-type-to-https/README.md).

An example of the change in the strings can be seen in the transition document linked above.

Per discussions internally there are three impacts:

1. Support receiving either prefix such that the correct handler is invoked.
2. Persist which prefix was used on receipt of a message and respond with that same prefix. That will support agents the ONLY support one format or the other.
3. Enable us to switch (when ready) to use the the new format on protocol instances we initiate.

On (3), existing deployments should work with the old format without any config changes, and a config is needed to use the new prefix.  Eventually, we'll probably make another change (and another ticket) to reverse that - to use the new prefix by default and use the old prefix only when configured.

nrempel:
Pausing this for now since the implementation gets a bit messy.

Draft: https://github.com/hyperledger/aries-cloudagent-python/pull/253

Handling multi-step protocols require us to track which prefix to use and persist this information.

PR #379, opened 2020-02-18
Support new message @type prefix

nrempel:
This pull request adds support for Step 1 of [RFC 0348](https://github.com/hyperledger/aries-rfcs/blob/master/features/0348-transition-msg-type-to-https/README.md)


nrempel:
addresses #244

PR #382, opened 2020-02-22
Updated AriesOpenAPIDemo.md

shaktishenoy:
Added screenshots from @sklump 
Signed-off-by: shaktishenoy <shaktishenoy@yahoo.com>


swcurran:
This is cool!  Thanks.  Will go through this shortly.

I don't know if you would be interested in doing this, but I'd like to get this OpenAPI demo so that instead of using the `/send` endpoint for issuing, it uses the full offer/request/response protocol.  That would give a new person using this more of a feel of what a controller needs to do.

We'll be updating it shortly to use the `--events` and `--no-auto` parameters so that the developer learns about what the webhook messages look like and how data can be pulled from those to populate the requests sent to the endpoints.

shaktishenoy:
> This is cool! Thanks. Will go through this shortly.
> 
> I don't know if you would be interested in doing this, but I'd like to get this OpenAPI demo so that instead of using the `/send` endpoint for issuing, it uses the full offer/request/response protocol. That would give a new person using this more of a feel of what a controller needs to do.
> 
> We'll be updating it shortly to use the `--events` and `--no-auto` parameters so that the developer learns about what the webhook messages look like and how data can be pulled from those to populate the requests sent to the endpoints.

Sure! Let me know when the changes are done.

PR #385, opened 2020-02-25
taa_accept default to True when sign is True

dbluhm:


andrewwhitehead:
FYI @Alexis-Falquier

PR #390, opened 2020-02-28
Fix pre-verify check

dbluhm:
This appears to be a typo that is causing `pre_verify` to return `None` when it should be returning `OK`


swcurran:
@dbluhm - how did that manifest?  We're trying to decide if we need an immediate release because of this.  Thanks!

dbluhm:
Working through our toolbox demos and I tried a presentation with only a predicate on an age attribute. It showed up as this error:
Screenshot from 2020-02-28 16-12-01

swcurran:
Sorry to hear about that.  We need to figure out how to prevent that in the future.

We want to have trust in the updates we release.  Something for us to ponder.

dbluhm:
No worries; thanks for responding quickly and getting a release out!

PR #412, opened 2020-03-11
Check err_value is instance of IndyError

dbluhm:
Rather than checking that err_type is IndyError. The Indy SDK wrapper now returns Errors that are a subclass of IndyError but not of type IndyError, failing the check in the handler. This causes the Error Handler to not properly wrap raised errors.


dbluhm:
Fix already included in #406

PR #414, opened 2020-03-12
Update revocation support to the alice/faber demo

ianco:
Add support for publishing tails files in an alternate location (e.g. github) so we can service via https

Update api docs (mobile scenario) to describe how to publish tails files and enable revocation


swcurran:
For the tails file piece - in PWD, how do you get to the command line to do the copying of the tails file, etc.?

ianco:
@swcurran  it gets printed out in the faber console

swcurran:
The instructions are good.  There are a few cleanups to be done:

- mention revocation earlier, before starting the agent the first time.
- Put in an environment variable in the instructions for copying and pasting for the server URL and the github user so that they can be set once and then reused by the copy paste.
- provide instructions for getting the URL for 8021 port on PWD (right click on the link - copy link)

We can do those cleanups after.

However, could not connect to Streetcred - the connection hung at "response" - did not go to active.

PR #415, opened 2020-03-13
Update instructions for revocation so that they can be tested

swcurran:


swcurran:
Some clean up on the instructions after going through it.  I'd like to clean up the copy paste of the curl command a little more, but that's OK.  Once merged (if OK) I'll send for testing with mobile agents.

PR #422, opened 2020-03-16
Allow faber to continue when connection is in 'response' state

ianco:

Makes the mobile demo a bit cleaner, as the connection will only be in "response" state when accepted by the mobile agent.

Tested and we can send a credential when the connection state is "response".


sukalpomitra:
@ianco @swcurran @andrewwhitehead I observed with hyperledger version of the code I can see the OSMA in connected state. But if i use ianco's repo, i can see the faber console but OSMA connection is always at negotiating. state

sukalpomitra:
@ianco @swcurran - This was an issue on my side, where I mistakenly was using an old consumer id of the mediator. I can see OSMA moving to connected and faber console also coming

PR #423, opened 2020-03-16
Update the run_demo script to add background mode and usage info

swcurran:


demo/run_demo
```diff
@@ -46,6 +50,38 @@ do
 			echo "For example, to create the directory and then set the permissions use: 'mkdir ../logs; chmod uga+rws ../logs'"
 			exit 1
 		fi
+		continue
```
andrewwhitehead: Adding continue here means that the argument gets stripped off, which I don't think is intended.

demo/run_demo
```diff
@@ -46,6 +50,38 @@ do
...
+		;;
+	--bg)
+		if [[ "${AGENT}" == "alice" ]] || [[ "${AGENT}" == "faber" ]] || [[ "${AGENT}" == "acme" ]]; then
+			DOCKER_OPTS="-d"
+			echo -e "\nRunning in ${AGENT} in the background. Note that you cannot use the comand line console in this mode."
```
andrewwhitehead: *command :)

demo/run_demo
swcurran: Thanks - wasn't sure about that.

swcurran: Fixed

demo/run_demo
swcurran: Fixed

demo/run_demo
swcurran: Confirmed that --timing works.

PR #424, opened 2020-03-17
More updates to the OpenAPI demo after more feedback

swcurran:


swcurran:
Hey Ian, more updates. Didn't get further on the proof. Interesting state this time:

- Faber: present-proof/send-request; rechecked - still in the "request_sent" state
- Alice: Received proof, fetched the credentials, stopped processing (shows on terminal)
- Alice: "state": "request_received",
- From that state, how does one get to completing the proof to send the response?  I looked at the present-proof call and for it I need a proof. How does that get constructed via API calls?

Not sure how it got into this state...ran it twice with the same result.

In the instructions, it says that "--auto..." flags are set on ACA-Py, but with "--no-auto" flag set on ./run_demo I don't think that is the case.

swcurran:
Cancelled - changes made in later PR.

PR #431, opened 2020-03-23
Use recommend indy generate_nonce

nrempel:


swcurran:
@nrempel -- can you please clean up the LGTM issue that was found and then we can merge?

nrempel:
that should do it

PR #440, opened 2020-03-27
Event tracing

ianco:
Current - can be configured to write events (json format) to standard log file or http endpoint.

(Future - append to exchange and message records)

Sample docker-based EFK stack under demo/EFK-stack - runs Elasticsearch, Fluentd and Kibana - for testing writing events to monitoring and reporting tool.

Currently includes event tracing on credential exchange only.


ianco:
Please let me know if this PR can me merged.  I'll work next on adding the trace events to the messages, but I'd like to do that in a separate PR if possible.

Thanks.

PR #441, opened 2020-03-31
Revo cleanup to controller

sklump:
Revoke by rev reg id + cred rev id. Leave rev regs out of credential exchange records until issue. Unit test coverage for indy-issuer.


andrewwhitehead:
@swcurran This gets rid of the revocation by credential exchange ID so the only way to revoke is by rev reg ID + cred rev ID. Is that the intention? I think it's probably best to only have one way to do it.

sklump:
Via cred ex record is janky, as we knew from the start - it was a way to get the thing running,  nothing more. It injects revocation into the process way before it needs to be there. I took out the former process on purpose.

I've also taken out the revocation registry cache from the indy Revocation context, since it was indexed by cred def id and there can be multiple active rev regs per cred def id (relying on only the most recent will orphan revocation capacity). Since it's going to the wallet and not the ledger, I don't think it will present a significant loss of amenity. If need by I can put it back in.

swcurran:
I agree - this should be the only way.  The cred_exchange_id approach was a a short cut but this is the right/long term way to do it.

Issue #409, opened 2020-03-10
Add support for minor version handling


swcurran:
In aries-rfcs issue [407](https://github.com/hyperledger/aries-rfcs/issues/407), it was agreed that there were rules that frameworks should support around the handling of minor versions of a protocol:

1. Minor versions can only include the addition of new fields and the continued semantic meaning of existing fields. When a change violates those conditions, a new major version is required.

2. An Aries agent should be able to handle the receipt of any minor version within a single major version above the one they implement and any minor versions below that they want to support. E.g. an agent that handles the 1.3 message should be able to handle any 1.3+ message, and 1.0, 1,1 and 1.2 if they want to. On backwards compatibility - generally they would support the earlier minor versions unless there is a good reason not to, e.g. a feature in HTTP 1.1 (keep-alive) that is not supported in HTTP 1.0, and hence unusable for that use case.

  1. We'd like by default that ACA-Py support all minor versions of a major protocol version. We should add a controller mechanism that tells ACA-Py to respond with a "Not Supported" problem-report.

3. Any fields not recognized in a message must be silently ignored. Thus, an agent that supports version 1.3 can receive a 1.5 message and process it using the 1.3 handling, ignoring any added v1.4 and v1.5 data items. Of course, that also means ignoring fields incorrectly put into the message.

4. If an agent chooses to respond to a discover feature message, the agent should respond with the latest version that the agent fully supports (e.g. 1.3 in the previous example).

  1. Likely already done in ACA-Py, but this should be confirmed.

5. When sending a response, an Aries agent should always respond with the latest version of the message that it supports.

  1. Again, likely already done, but should be confirmed.

6. ACA-Py should be capable of sending back a problem-report if a message of an unsupported major version is received, with options for the controller to activate (or not) that protocol. For now, we should make this an all or nothing feature, but it if it is easy, implement a protocol-by-protocol config, including an "unknown protocol" configuration.

This issue will be converted to an epic at some point with specific issues added to address smaller chunks of the work.

PR #443, opened 2020-04-02
Add support for major/minor protocol version routing

nrempel:
Addresses: #409, https://github.com/hyperledger/aries-rfcs/issues/407 

This pull requests major/minor version routing for aries protocols.

The implementation follows these rules:

- Existing protocols (internal and external) should still work. This work is designed to not introduce any breaking changes.
- If a protocol uses the old style, it can't receive other minor version messages. For them, the `@type` parameter must match exactly
- New protocols must include a top level [definition module](https://github.com/hyperledger/aries-cloudagent-python/pull/443/files#diff-c8ea657f1116ef5e00980dae3e252b19) and include references to protocol implementation as children.
- There can only be any number of  major version implementations per protocol. Each major version protocol is effectively treated as a separate protocol and is referenced by the definition module.
- There can only be one minor version implementation per protocol major version.
- Minimum minor version can be specified in `definition.py`. If a message is received for a protocol and major version with a minor version *less than* the minimum supported version, the agent will respond with a problem report with text describing the violation.
- Any extra fields are now ignored when deserializing _any_ agent messages. This ensures that protocols can support minor versions greater than the current implementation.

Remaining todo:

- [x] Documentation page on creating a new protocol
- [x] Even more manual testing?

Docs: https://github.com/nrempel/aries-cloudagent-python/blob/proto-versions/aries_cloudagent/protocols/README.md


swcurran:
@nrempel - curious about this constraint "If a protocol uses the old style, it can't receive other minor version messages. For them, the @type parameter must match exactly".  By old style, I assume you mean "did:sov:...".  Not that I think it's a particular issue, but wondering why that was added?

nrempel:
> @nrempel - curious about this constraint "If a protocol uses the old style, it can't receive other minor version messages. For them, the @type parameter must match exactly". By old style, I assume you mean "did:sov:...". Not that I think it's a particular issue, but wondering why that was added?

In this context by old style I mean the structure of the protocol directory. If it doesn't define it's supported major version and minimum minor version, it uses the current functionality of an exact match. Basically, it's for backwards compatibility since that is the current behaviour.

nrempel:
@TelegramSam do you have an opinion on these changes at all? I know you've been active in the spec discussion.

nrempel:
@andrewwhitehead I've opened up this PR

aries_cloudagent/core/protocol_registry.py
```diff
@@ -50,18 +56,55 @@ def protocols_matching_query(self, query: str) -> Sequence[str]:
...
             self._typemap.update(typeset)
 
-    def register_controllers(self, *controller_sets):
+        # Track versioned modules for version routing
+        if version_definition:
+            for typeset in typesets:
+                for message_type_string, module_path in typeset.items():
+                    parsed_type_string = self.parse_type_string(message_type_string)
+
+                    if not version_definition["major_version"] in self._versionmap:
```
andrewwhitehead, line 96: Should be `x not in y` to avoid ambiguity I think

aries_cloudagent/core/plugin_registry.py
```diff
@@ -30,23 +32,145 @@ def plugins(self) -> Sequence[ModuleType]:
...
+            raise ProtocolDefinitionValidationError(
+                "Versions list must define at least one version module"
+            )
+
+        for version_dict in version_list:
+            # Dicts must have correct format
+            is_dict = type(version_dict) is dict
+            if not is_dict:
+                raise ProtocolDefinitionValidationError(
+                    "Element of versions definition list is not of type obj"
```
andrewwhitehead, line 57: What's an obj?

andrewwhitehead:
Added a couple comments but it looks good otherwise

nrempel:
@andrewwhitehead thanks, fixed

PR #445, opened 2020-04-02
Format code

nrempel:
Run auto-formatter to bring entire codebase up to spec


nrempel:
This should also clean up my other PR since my editor ran a full format with our rules.

PR #450, opened 2020-04-05
Updates to the OpenAPI demo and the Mobile demo; rename Mobile Demo

swcurran:


swcurran:
@ianco -- reworked both scripts.  The mobile one is a little cleaner now - flows better, I think.

Please review as I'd like to get these to the mobile folks.

PR #456, opened 2020-04-14
Signed attachments

sklump:


aries_cloudagent/messaging/decorators/attach_decorator.py
```diff
@@ -24,15 +23,190 @@
...
+def did_key(verkey: str) -> str:
+    """Qualify verkey into DID key if need be."""
+
+    return verkey if verkey.startswith("did:key:z") else f"did:key:z{verkey}"
+
+
+def raw_key(verkey: str) -> str:
+    """Strip qualified key to raw key if need be."""
+
+    return verkey[9:] if verkey.startswith("did:key:z") else verkey
```
andrewwhitehead, line 207: I don't think decoding a did:key is as simple as stripping off the prefix, I believe it adds a couple bytes to the binary representation and re-encodes as base58

PR #458, opened 2020-04-15
Add Black format check github action

nrempel:
Addresses #459 

This pull request adds a new github action to run the Black formatter with `--check` as a linting step. We already use the Black formatter along with custom linting rules. This check will enforce running Black on all work before being merge into master.

Ensuring that all code uses the same auto-formatter is beneficial because it eliminates git conflicts on formatting changes.


sklump:
*Item:*
This check fails code that passes in /scripts/run_tests_indy. One or the other has to change. At the very least, test_*.py must be exempt.

PR #464, opened 2020-04-17
Correct format of the timestamp in the JSON

swcurran:


swcurran:
@ianco - can you do a quick review?  This was reported as a problem on rocketchat.

PR #491, opened 2020-05-05
instrument pre-verification to reject proof bypassing revocation check

sklump:


andrewwhitehead:
@sklump Looks like this one might need a manual update

PR #497, opened 2020-05-08
do not crash on connection request/response creation if no endpoints …

sklump:
…at all in config

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/manager.py
+++ b/aries_cloudagent/protocols/connections/v1_0/manager.py
@@ -296,7 +296,8 @@ class ConnectionManager:
 
         # Create connection request message
         if not my_endpoint:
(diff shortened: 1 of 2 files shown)


aries_cloudagent/protocols/connections/v1_0/manager.py
```diff
@@ -296,7 +296,8 @@ async def create_request(
 
         # Create connection request message
         if not my_endpoint:
-            my_endpoints = [self.context.settings.get("default_endpoint")]
+            my_endpoints = []
+            my_endpoints.extend(self.context.settings.get("default_endpoint", []))
```
andrewwhitehead, line 300: That doesn't seem right, `default_endpoint` is a string

sklump: You're right, will fix

PR #499, opened 2020-05-08
Demo tails server

nrempel:
This PR updates the phone demo to use the new tails server available here: https://github.com/bcgov/von-tails-server

### DIFF
--- a/demo/AliceGetsAPhone.md
+++ b/demo/AliceGetsAPhone.md
@@ -1,6 +1,6 @@
 # Alice Gets a Mobile Agent! <!-- omit in toc -->
 
-In this demo, we'll again use our familiar Faber ACA-Py agent to issue credentials to Alice, but this time Alice will use a mobile wallet. To do this we need to run the Faber agent on a publicly accessible port (we'll use Play With Docker), and Alice will need a compatible mobile wallet. We'll provide pointers to where you can get them. As well, we'll also show how you can run through this sequence with credential revocation activated, so you can see how the mobile wallets handle that.
+In this demo, we'll again use our familiar Faber ACA-Py agent to issue credentials to Alice, but this time Alice will use a mobile wallet. To do this we need to run the Faber agent on a publicly accessible port, and Alice will need a compatible mobile wallet. We'll provide pointers to where you can get them. This demo also introduces revocation of credentials.
 
 # Contents <!-- omit in toc -->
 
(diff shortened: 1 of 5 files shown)


swcurran:
Some things to do, I think.

1. Please rename von-tails-server to indy-tails-server.  That's a more appropriate name for this component.
2. On the free account, I was not able to run two ngroks.  Is it possible to start up two in one session?  I think @esune is doing that with identity-kit, but I could be wrong.
3. The same Emiliano chap was able to start ngrok in the background and automatically determine the appropriate URLs created by nrgok.  It would be really nice to use that same technique here to do that.

I'm stuck on 2 and can't proceed without getting a higher level ngrok account. I'm fine with that but it won't work very well for a public demo of this.  Any other ideas?

sklump:
Please remove unused imports so my phone doesn't go off on push?

esune:
> 
> 
> Some things to do, I think.
> 
>     1. Please rename von-tails-server to indy-tails-server.  That's a more appropriate name for this component.
> 
>     2. On the free account, I was not able to run two ngroks.  Is it possible to start up two in one session?  I think @esune is doing that with identity-kit, but I could be wrong.
> 
>     3. The same Emiliano chap was able to start ngrok in the background and automatically determine the appropriate URLs created by nrgok.  It would be really nice to use that same technique here to do that.
> 
> 
> I'm stuck on 2 and can't proceed without getting a higher level ngrok account. I'm fine with that but it won't work very well for a public demo of this. Any other ideas?

I think the free ngrok account limits you to two tunnels unfortunately. One way around this (not sure) could be to use the dockerized version of ngrok like the one used in aries-toolbox, but haven't tried yet so it needs to be confirmed.

nrempel:
I've tested opening 6 concurrent ngrok tunnels without issue. Version 2.3.35. What version do you have installed?

esune:
Same version. How are you starting `ngrok`? If you are starting one tunnel at a time it may be possible to have more than two, this is what I get when I try to start more than two tunnels at the same time using a configuration file:
```
ngrok.io https://d24b48ef.ngrok.io]\n\r\n\r\nERR_NGROK_324\r\n"
Your account may not run more than 2 tunnels over a single ngrok client session.
The tunnels already running on this session are:
[https://968e7ecf.ngrok.io https://d24b48ef.ngrok.io]
```

nrempel:
I've confirmed with Stephen that this limitation only exists if you are logged in and not paying for an account. If you are not logged-in, there is no limit.

esune:
> 
> 
> I've confirmed with Stephen that this limitation only exists if you are logged in and not paying for an account. If you are not logged-in, there is no limit.

How do you check whether you are logged-in or not? I have never logged-in to `ngrok` (I do not have an account) and still get the error...

swcurran:
I was able to get this to work after a little playing around.

- getting the ngrok unlinked from my account (by remove ~/.ngrok folder) got me passed the ngrok account limitation.
- Switching from a light theme to a dark on my terminal (vscode, in my case) solved the QR code being unusable. 

After that, I could get the demo to work with Streetcred and eSatus quite happily.

Nick - anything we can do with the QR code generator to generate in the correct colours always?  Ideally an autodetect.

It would be great if we could do what Emiliano did to auto-parse the ngrok URLs and invoke them automatically.  Getting them copied and pasted was a bit of a pain.  Manageable, but what Emiliano did was way slicker :-).

esune:
Using a configuration file like [this one](https://github.com/bcgov/vc-authn-oidc/blob/master/demo/ngrok.yml) the tunnels can be named.

Then a script like [this one](https://github.com/bcgov/vc-authn-oidc/blob/master/demo/docker/manage#L345) can be used to pull the endpoint out of the running ngrok instance (using `jq`, that needs to be installed).

esune:
> 
> 
> I was able to get this to work after a little playing around.
> 
>     * getting the ngrok unlinked from my account (by remove ~/.ngrok folder) got me passed the ngrok account limitation.
> 

FYI I do **NOT** have an `ngrok` account (or a `.ngrok` folder in my home) and I still can't start more than two tunnels.

swcurran:
For this a user only needs two tunnels.  My issue was that I couldn't create more than 1.

esune:
> 
> 
> For this a user only needs two tunnels.  My issue was that I couldn't create more than 1.

Oh, I see. I was still going with Nick's ability of creating more than two tunnels. If this is the case I guess we're good.

PR #501, opened 2020-05-11
Maximum try count for outbound messages

denizgunseli:
Maximum try count for outbound messages was defined as hardcoded in manager. For any reason, it might be necessary to increase or decrease this parameter. 

This value was set as 4 as a defult value in manager and set as a a parameter in config and changes has been made for using this value from settings.


andrewwhitehead:
Looks good, but I can't sync it up with master to merge it.

PR #502, opened 2020-05-12
Get tails file response type annotation

sklump:
More formal implementation allows for any module to define free function `set_openapi_file_responses()` to set JSON in swagger (OpenAPI) for registered methods returning files. Admin server picks these up and executes them via plugin registry after initial aiohttp Web application setup.

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -19,6 +19,7 @@ from marshmallow import fields, Schema
 from ..config.injection_context import InjectionContext
 from ..core.plugin_registry import PluginRegistry
 from ..messaging.responder import BaseResponder
+from ..revocation.routes import set_tails_file_binary_response_schema
 from ..transport.queue.basic import BasicMessageQueue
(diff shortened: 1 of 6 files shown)


andrewwhitehead:
Is there a reason why `PluginRegistry.register_admin_routes` can't pick up these hooks as well?

Also, might it make sense to just make it a generic route post-processing method, rather than calling it `set_openapi_file_responses`?

sklump:
> 
> 
> Is there a reason why `PluginRegistry.register_admin_routes` can't pick up these hooks as well?
> 
> Also, might it make sense to just make it a generic route post-processing method, rather than calling it `set_openapi_file_responses`?

When register() is called, the openapi structure hanging off the app is not guaranteed to be there - in fact it isn't there on typical startup. The server builds the routes, then generates the openapi out of them. But there is no way to specify a file response with aiohttp_apispec via its polite interface, so the fixup has to be post-hoc.

I could rename this, what do you want to see?

sklump:
Bugger all this for a lark. New PR.

PR #503, opened 2020-05-12
Post registration openapi hook

sklump:


sklump:
That's it. Nuking it from space.

PR #513, opened 2020-05-20
Fix #427 websocket --admin-api-key connect issue

### LINKED ISSUE #427: Websocket webhook not working with --admin-api-key enabled
Hi all,

I'm using the websocket webhook to receive changes from aca-py, e.g. new connection invitations and I pretty much like this feature :-)

Unfortunately after activating ```--admin-api-key```, websocket can't connect any more. Reason: the ```server.py``` middleware code correctly checks for the http header ```x-api-key```
Problem is, that with (builtin) Javascript ```WebSocket()``` there is no way to set custom http-headers. Only way (I found) is when switching to "polling" protocol, which is the fallback and probably not the reason why to use websockets in the beginning.

I think we somehow should enable the admin-api-key also for websockets and I see 2 options at the moment:

1. Allow query params as an additional method to transfer ```x-api-key``` to aca-py.
One-liner change on the server in [server.py](https://github.com/hyperledger/aries-cloudagent-python/blob/master/aries_cloudagent/admin/server.py#L169)
```
header_admin_api_key = request.headers.get("x-api-key") or request.query.get("x-api-key")
```
Client:
```
new WebSocket('wss://localhost:10000/ws?x-api-key=12345
```
pro: Rather simple fix on both sides
con: URL with the key in the browser cache and server logs? (Still, client can decide not to use this feature...)

2. Send a message with the api-key to the server after websocket connection is established.
Pro: Security?
Con:
Need to define message.
Keep state of connection and only publish messages when allowed to do so.
Maybe more likely to accidentally publish data on websocket connections with wrong state (before api-key message) when implementing new features?

3. any other ideas?

Thanks for your help,
Matthias

matgnt:
This change is based on the 2 proposals in #427

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -171,7 +171,9 @@ class AdminServer(BaseAdminServer):
 
             @web.middleware
             async def check_token(request, handler):
-                header_admin_api_key = request.headers.get("x-api-key")
+                header_admin_api_key = request.headers.get(
+                    "x-api-key"
+                ) or request.query.get("x-api-key")
                 if not header_admin_api_key:
                     raise web.HTTPUnauthorized()
 


matgnt:
Alternative implementation in pull request #528

PR #514, opened 2020-05-20
Protocol version links

sklump:


swcurran:
@WadeBarnes -- can you please see if symlinks used like this work on Windows?

WadeBarnes:
The symlinks don't work on Windows.  They show up as regular text files.  When using `git clone --config core.symlinks=true` the symlinks don't show up at all, not even as text files.

WadeBarnes:
Scratch that, I got it working;
image

WadeBarnes:
You have to run the `git` commands using a shell with Administrative rights, and use the `core.symlinks=true` setting; `git clone -c core.symlinks=true https://github.com/sklump/aries-cloudagent-python.git`

WadeBarnes:
It looks like it is possible to allow non-administrators on Windows to create symlinks; https://github.com/git-for-windows/git/wiki/Symbolic-Links#allowing-non-administrators-to-create-symbolic-links

WadeBarnes:
This is the articular that lead to the solution; https://stackoverflow.com/questions/5917249/git-symlinks-in-windows/42137273#42137273

WadeBarnes:
I've confirmed the permissions for allowing non-administrators on Windows to create symlinks.  You need to reboot, or at least logout and back on again, but afterward I'm able to switch to the branch with the symlinks and they get created without error or issue.

swcurran:
Can we do a git config on the repo to set `core.symlinks=true `?

WadeBarnes:
It looks like you could add a `.gitconfig` to the repo, but it then becomes a manual process on the developer side to include it in their working copy's configuration.  It seems easier to do the `-c core.symlinks=true` up front or add it to the confg manually after the fact (in the case you've already cloned the repo).

WadeBarnes:
I should also try a docker build to see if the Windows symlinks cause any issues.

WadeBarnes:
I ran the `scripts/run_tests` which does a docker build and runs the tests.  The build completed successfully and the tests ran.

sklump:
Python's build into wheel does not preserve symbolic links: it copies directories, which will bloat the archive trebly, then worse as protocol versions emerge.

PR #515, opened 2020-05-20
Minor bug in performance.py

wip-abramson:
Just a minor bug I ran into, figured I may aswell push a fix


andrewwhitehead:
Merged in another PR, thanks!

Issue #522, opened 2020-05-22
 demo/runners/performance.py fails with --routing flag


Howlla:
the error can be found in line 299
 ```
 File "/home/indy/demo/runners/performance.py", line 299, in main
    invite, auto_accept=json_dumps(False)
NameError: name 'json_dumps' is not defined
```
It can reproduced by running

$ LEDGER_URL=http://greenlight.bcovrin.vonx.io ./run_demo performance --routing

I think it should be replaced with json.dumps

sklump:
Sorry, my mistake. I'm on it.

PR #523, opened 2020-05-22
json.dumps not json_dumps

sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/522

### DIFF
--- a/demo/runners/performance.py
+++ b/demo/runners/performance.py
@@ -295,9 +295,7 @@ async def main(
             invite = await faber.get_invite()
 
             if routing:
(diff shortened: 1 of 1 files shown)


demo/runners/performance.py
```diff
@@ -296,7 +296,7 @@ async def main(
 
             if routing:
                 conn_id = await alice.receive_invite(
-                    invite, auto_accept=json_dumps(False)
+                    invite, auto_accept=json.dumps(False)
```
andrewwhitehead, line 299: receive_invite actually takes a boolean, not JSON

PR #525, opened 2020-05-26
Key rotation operations

sklump:
Please have a look. I think:
* the public DID key rotation should hang off the ledger API - it's defensive design against an off-line agent trying to the rotate key pair for a DID that's public.  Ledgers need wallets but wallets don't need ledgers; the operation initializes at the wallet, updates the ledger, then returns to the wallet to apply the update, and it uses the same wallet operations as for local DID key pair rotation below.
* the local DID key pair rotation should hang off the wallet because it only needs the wallet. There are two calls: start and apply; the `routes.py` calls one and then the other. The wallet needs to retain both calls to accommodate public DID key pair rotation as above.
* the wallet master encryption key rotation is a provision operation. Could you look this over in particular? On some level I can't believe it's this simple.

I'm open to the idea of renaming stuff in `routes.py` for ledger, wallet. I was toying with the idea of 'nym' instead of 'public-did' for symmetry, but we've used public-did everywhere else to refer to public DIDs, and properly the keys belong to the DID not the NYM.

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -744,7 +744,14 @@ class WalletGroup(ArgumentGroup):
             "--wallet-key",
             type=str,
             metavar="<wallet-key>",
-            help="Specifies the master key value to use for opening the wallet.",
+            help="Specifies the master key value to use to open the wallet.",
+        )
+        parser.add_argument(
+            "--wallet-rekey",
+            type=str,
+            metavar="<wallet-rekey>",
+            help="Specifies a new master key value to which to rotate and to\
+            open the wallet next time.",
         )
         parser.add_argument(
             "--wallet-name",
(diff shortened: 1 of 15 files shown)


andrewwhitehead:
The wallet rekey isn't working for me so far, here's what I'm trying:

`docker run --rm -ti --entrypoint bash aries-cloudagent-run`

```sh
aca-py provision --wallet-type indy --wallet-key ORIG_KEY --wallet-name test --seed 0000000000000000000000000000Test

aca-py start --wallet-type indy --wallet-key ORIG_KEY --wallet-name test -it http 0.0.0.0 8000 -ot http

aca-py provision --wallet-type indy --wallet-key ORIG_KEY --wallet-name test --wallet-rekey NEW_KEY

# fails
aca-py start --wallet-type indy --wallet-key NEW_KEY --wallet-name test -it http 0.0.0.0 8000 -ot http

# succeeds
aca-py start --wallet-type indy --wallet-key ORIG_KEY --wallet-name test -it http 0.0.0.0 8000 -ot http
```

I think there should be some output if it has performed a wallet rekey, but I'm not seeing anything.

sklump:
Thanks - I'm working on it

Issue #524, opened 2020-05-22
ACA-Py Admin API throwing 500 internal server error when issue-credential send-offer is sent with an incorrect cred_ex_id


nodlesh:
### Issue

When using the Issue Credential Protocol, the Admin API calls that take a cred_ex_id return a 400 internal server error when the id is not proper, that is either ill formed, or properly formed but invalid. 

### Steps to Reproduce
Just simply run the issue-credential send-offer though the Admin API in code or even in the Swagger web interface

### Outcome

500 Internal Server Error\n\nServer got itself in trouble

### Expected Outcome

A problem report more closely aligned to the actual problem of the id.

### Workaround

Use the correct cred_ex_id

### Test Cases Affected

None

### Severity

Medium

### Business Priority

Medium

swcurran:
@sklump - could you take a look at this when you get a chance?  Given the work you did on the API, I'm thinking that you would be best to understand the scope of this issue.  Is this a one-off or at least a relatively rare issue, or is there a need for a systemic handling of these types of issues?  What's the best approach?

sklump:
I'll get to it later this week. It fits more generally into Nick's omnibus error audit.

PR #526, opened 2020-05-26
respond with 404 Not Found for bad cred ex id in issue-cred

sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/524 and cousins throughout issue-credential protocol

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v1_0/routes.py
+++ b/aries_cloudagent/protocols/issue_credential/v1_0/routes.py
@@ -284,7 +284,9 @@ async def credential_exchange_retrieve(request: web.BaseRequest):
             context, credential_exchange_id
         )
     except StorageNotFoundError:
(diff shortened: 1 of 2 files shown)


swcurran:
Cool - I'm going to see if I can use the AATH to test this...  I'll use Stephen's branch, so please keep that in place for a little bit.

swcurran:
Good news and bad.  Super easy to use a another branch in the AATH, so that's neat.  I think we can easily make this parameterized so that it's trivial to run these tests with a branch,.

Bad is that there were no tests that exercised this, so I couldn't tell if it did anything.

sklump:
Easy: spin up an instance, get cred Ex record from Admin API specifying bogus cred def id (pick a UUID from the sample black box). Observe the error message, 404 not found.

sklump:
Also, note that there are unit tests for all of these cases in test_routes.py

PR #531, opened 2020-05-27
Revisit issue cred vs rfc 36

sklump:
I hope this doesn't blow anyone's use case up: aries offers always have previews (pure-indy offers don't - they're just a handshake to prove issuer's bona fides).


sklump:
Hold on, let me address https://github.com/hyperledger/aries-cloudagent-python/issues/532 as well here thanks.

PR #537, opened 2020-06-01
Feature/jsonld

TelegramSam:
This adds JSO-LD support for signing and verifying credentials. Currently supported in the administrative REST api.

CC: @victormartinezjurado


TelegramSam:
GADFQ#@$R@#RT@#W

PR #547, opened 2020-06-08
move mime-types request into holder

sklump:

### DIFF
--- a/aries_cloudagent/holder/routes.py
+++ b/aries_cloudagent/holder/routes.py
@@ -19,6 +19,10 @@ from ..messaging.valid import (
 from ..wallet.error import WalletNotFoundError
 
 
+class AttributeMimeTypesResultSchema(Schema):
+    """Result schema for credential attribute MIME type."""
+
+
 class RawEncCredAttrSchema(Schema):
     """Credential attribute schema."""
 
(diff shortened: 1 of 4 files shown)


aries_cloudagent/holder/routes.py
```diff
@@ -190,6 +215,11 @@ async def register(app: web.Application):
     app.add_routes(
         [
             web.get("/credential/{credential_id}", credentials_get, allow_head=False),
+            web.get(
+                "/issue-credential/mime-types/{credential_id}",
+                credentials_attr_mime_types_get,
+                allow_head=False,
+            ),
```
andrewwhitehead, line 222: Should it move to /credential/ instead of /issue-credential/? I'm not sure where the mime type info comes from.

sklump: I think it's natural in holder/, since it's in the holder wallet.

The mime type info comes from the credential issue. If an attribute is a PNG for example, then it has the corresponding mime-type in the preview. The holder stores mime-type info per attribute (if any are non-null for straight text) in a non-secret record when he saves the credential. Thereafter, he can use mime-types to figure out how to interpret the (base64-encoded) value. Similarly, he can present this mime-type in presentation previews so that the verifier can tell what he's looking at.

andrewwhitehead: I'm just talking about the prefix on the URL of the route, since it doesn't seem to be specific to the issue-credential protocol. I think in future the field types will be XML schema types derived from the RDF relations in the credential itself. Changing the prefix would be a breaking change for some controllers but not a major one.

sklump:
Oh heck, you are right. I will fix and commit tomorrow first thing.

PR #559, opened 2020-06-16
allow provision of local-only DID in wallet

sklump:
Also, ledger must be open on receiving presentation preview to check if revocation applies


aries_cloudagent/protocols/routing/v1_0/handlers/route_query_request_handler.py
```diff
@@ -25,8 +25,6 @@ async def handle(self, context: RequestContext, responder: BaseResponder):
         if not context.connection_ready:
             raise HandlerException("Cannot query routes: no active connection")
 
-        # TODO implement pagination
```
sklump, line 28: Note that the storage query uses pagination: comment was obsolete

PR #564, opened 2020-06-18
No hopeless offers

sklump:


sklump:
Hold on! Please leave this out of 0.5.2.

It will cause mayhem on schemas with attribute names with capital letters or spaces. I will fix it tomorrow morning.

PR #579, opened 2020-06-26
Revocation hotfixes

sklump:
Publish any pending revocations for rev reg id on any revocation with publish=True.
Recover from revocation failures gracefully.
Allow clearing of pending revocations by rev reg id (null/empty for all) and by cred rev ids (null/empty for all)
Allow for granularity in publication of pending revocations by rev reg id (null/empty for all) and by cred rev ids (null/empty for all).
Demo touch-ups to match.


aries_cloudagent/issuer/indy.py
```diff
@@ -246,26 +246,46 @@ async def revoke_credentials(
...
-                    self.wallet.handle, tails_reader_handle, revoc_reg_id, cred_revoc_id
-                )
+                try:
+                    delta_json = await indy.anoncreds.issuer_revoke_credential(
+                        self.wallet.handle,
+                        tails_reader_handle,
+                        revoc_reg_id,
+                        cred_revoc_id,
+                    )
+                except IndyError as error:
```
nrempel, line 265: Nice, this is great.

Also note @esune I believe this should cure the problem you were seeing without any changes on your end.

nrempel:
Looks good to me

aries_cloudagent/issuer/indy.py
sklump: ... then take the troublesome cred_rev_ids on the rev_reg_id and purge them via clear_pending_revocations API as per issue_credentials/v1_0/routes.py, submitting body
```
{
    "rrid2crid": {
        "<rev-reg-id>": ["<cred-rev-id>", "<cred-rev-id>", ...]
    }
}
```

Issue #584, opened 2020-06-30
Don't rely on local tails file location when revoking credential


nrempel:
When revoking a credential, we rely on the tails file location previously written to a temporary directory: https://github.com/hyperledger/aries-cloudagent-python/blob/master/aries_cloudagent/protocols/issue_credential/v1_0/manager.py#L734

This is bad because the directory is temporary and the tails file has likely already been stored at a permanent location (possibly on an external server).

We should rely on the tailsLocation from the revoc_reg_def (either in wallet or on ledger). If the path is a valid URL, the agent should download, cache, and use the tails file from the URL. If the path is a local path, it should assume that the local path is not temporary and that the file was moved to a permanent location when the revoc_reg_def was published to the ledger.

sklump:
I can take this one

sklump:
What would you think of copying/moving the tails file from the temporary location to a permanent location? They are potentially huge and I'd prefer not to have to wait on downloading them if possible.

nrempel:
@sklump I think that's a good approach. I had a thought that we could move the file to a configured "cache" location from the temporary location. Then when revoking, if the cache is empty, the tails file could be downloaded from the tails server.

esune:
> 
> 
> What would you think of copying/moving the tails file from the temporary location to a permanent location? They are potentially huge and I'd prefer not to have to wait on downloading them if possible.

+1 to this. Thinking about our deployed agents (e.g.: OpenShift) would it make sense to have the location (optionally) configured using an environment variable? Or were you thinking of establishing a path that will always be used for this type of stuff (tails files)?

sklump:
I was thinking of adapting the von-anchor approach of putting them in a subdirectory (by rev-reg-id) within ~/.indy_client ($INDY_HOME). Von-anchor separates first by cred def id, but in the end I don't think the extra tree depth bought us anything.

esune:
As long as it is a directory that is well-defined/documented and can be independently mounted to a persistent volume when running agents in the cloud I think it will be okay 👍

PR #590, opened 2020-07-08
Fix reliance on temp tails file

sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/584


sklump:
Note that this gets rid of configuration for holder tails file storage location: every tails file always goes to $INDY_HOME/tails/<rrid>/<hash> locally.

PR #597, opened 2020-07-09
Add some docs on using tracing

ianco:


demo/AgentTracing.md
```diff
@@ -0,0 +1,98 @@
...
+--trace-log            Enables tracing to the standard log output
+                       (sets TRACE_ENABLED, TRACE_TARGET, TRACE_TAG)
+
+--trace-http           Enables tracing to an HTTP endpoint (specified by TRACE_TARGET_URL)
+                       (sets TRACE_ENABLED, TRACE_TARGET, TRACE_TAG)
+```
+
+When running the Faber controller, tracing can be enabled using the `T` menu option:
+
+```bash
```
andrewwhitehead, line 56: The red operators because it expects bash syntax are a little distracting

PR #613, opened 2020-07-17
Revocation changes

nrempel:
This pull request modifies the behaviour of managing revocation registries so that controllers need to track and manage less.

When a credential definition is created, if revocation is enabled, the agent will create and publish revocation registry entry, definition, and tails file. It also publishes an additional set of registry and tails file to be swapped out when the current registry is full.

When issuing credentials, the correct revocation information is used automatically and rotates registries when needed creating a new ready-to-swap one in the background.

Demo has been updated.

With a tails [server](https://github.com/bcgov/indy-tails-server) running, run the demo like so:

```
# Linux
./run_demo performance --revocation --tails-server-base-url http://172.17.0.1:6543
# Mac/Windows
./run_demo performance --revocation --tails-server-base-url http://host.docker.internal:6543
```


nrempel:
I'm closing this PR and re-opening in #619 due to DCO signoff issues.

PR #618, opened 2020-07-21
Oob unit tests

sklump:
unit test coverage for out-of-band protocol


aries_cloudagent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -89,10 +85,13 @@ async def create_invitation(
         """
 
         connection_mgr = ConnectionManager(self.context)
-        connection, connection_invitation = await connection_mgr.create_invitation(
-            auto_accept=True, public=use_public_did, multi_use=multi_use
+        (connection, connection_invitation) = await connection_mgr.create_invitation(
```
sklump, line 88: @nrempel 
Is this correct? Should these parameters flow on into the connection-manager call?

nrempel: That is the intent for the first phase of the migration. create/send and receive are both wrappers for existing functionality.

The work roughly aligns with step 1 of https://github.com/hyperledger/aries-rfcs/blob/4afa0a941d1943f67116c9aa1529d0e073a0c426/features/0496-transition-to-oob/README.md

sklump: In particular, though, they hadn't been in the `connection_mgr.create_invitation(...)` call and I put them in. Is that right or wrong, as per design intent? I will be happy to revert that change if you say the word.

nrempel: Ah, yes I missed that part of the diff. Yeah, I think that's a good change. I think I missed that scenario when running through my own testing. I say leave it in :+1:

PR #619, opened 2020-07-21
Revocation changes

nrempel:
This pull request modifies the behaviour of managing revocation registries so that controllers need to track and manage less.

When a credential definition is created, if revocation is enabled, the agent will create and publish revocation registry entry, definition, and tails file. It also publishes an additional set of registry and tails file to be swapped out when the current registry is full.

When issuing credentials, the correct revocation information is used automatically and rotates registries when needed creating a new ready-to-swap one in the background.

Demo has been updated.

With a tails [server](https://github.com/bcgov/indy-tails-server) running, run the demo like so:

```
# Linux
./run_demo performance --revocation --tails-server-base-url http://172.17.0.1:6543
# Mac/Windows
./run_demo performance --revocation --tails-server-base-url http://host.docker.internal:6543
```


nrempel:
As for migration path to using this new stuff: if a controller is currently managing revocation stuff manually it will continue to work without changes.

Any new credential definitions published will use the new behaviour.

demo/runners/faber.py
```diff
@@ -184,16 +192,17 @@ async def main(
...
 
-        if revocation:
-            with log_timer("Publish revocation registry duration:"):
-                log_status(
-                    "#5/6 Create and publish the revocation registry on the ledger"
-                )
-                await agent.create_and_publish_revocation_registry(
-                    credential_definition_id, TAILS_FILE_COUNT
-                )
+        # if revocation:
```
sklump, line 198: if it's obsolete, delete it: git can remember old code for us if need be

aries_cloudagent/config/wallet.py
```diff
@@ -79,4 +79,7 @@ async def wallet_config(context: InjectionContext, provision: bool = False):
             seed=test_seed, metadata={"endpoint": "1.2.3.4:8021"}
         )
 
+    if public_did:
+        context.settings["wallet.public_did"] = public_did
```
andrewwhitehead, line 83: There's a method on the wallet called `get_public_did` which does a dynamic lookup for this, because there are admin endpoints to dynamically change the public DID. If it's stored in the context then there isn't a reliable way to update it everywhere

aries_cloudagent/revocation/models/issuer_rev_reg_record.py
```diff
@@ -144,7 +148,7 @@ async def generate_registry(self, context: InjectionContext):
         issuer: BaseIssuer = await context.inject(BaseIssuer)
         tails_hopper_dir = indy_client_dir(join("tails", ".hopper"), create=True)
 
-        LOGGER.debug("create revocation registry with size:", self.max_cred_num)
+        LOGGER.debug(f"create revocation registry with size: {self.max_cred_num}")
```
andrewwhitehead, line 151: I think this change throws warnings in some linters, the logger functions are meant to take the format arguments separately so the string formatting can be skipped if the log level is higher

sklump:
LOGGER.debug("... %d", self.max_cred_num) is typical here I believe

PR #624, opened 2020-07-22
Multitenent Functionality 

LukasSMueller:
This PR enables ACA-Py to handle multiple wallets for issuing, holding and verifying. 

## Running ACA-Py as Multi Wallet Agent

to start the agent in a multi wallet setup you need to pass the following flags during start:

`-it http_custodial`

`--plugin aries_cloudagent.wallet_handler`

### Admin API

To be able to use the admin API with multiple wallets each request might need to use a different wallet. To prevent that one requests overwrites the wallet inside another request which is waiting for some async coroutine each request needs it's individual local context with a different wallet requested in the `DynamicProvider` for the wallet class. To accomplish that a middleware checks the authorisation in the header of the request and obtains (somehow - best with a plugable authorization module -) the corresponding wallet for that request. Than the global context is copied and the `requested_instance` of the wallet provider is set according to the authorization. The request can than be handled as for the one-wallet-agent.

### Inbound Messages

To enable the agent to associate an inbound message with the correct wallet, for communication a new inbound path `default_endpoint/sha256(wallet_name+num(connections))` is created and passed to the agent which participates in that communication. If the agent receives a message he uses the endpoint to which the message is posted to check (by querying the `wallet_handler`) with what wallet the message can be received. In `create_session()` of the inbound manager the global context is copied, adapted for the correct wallet and passed to the session which handles the incoming message.

### Outbound Messages

Outbound messages are always associated to a specific connection. The `wallet_handler` keeps a mapping of connections to wallets so that it can return the correct wallet for any connection. Similar as for inbound messages the `outbound_manager` uses this functionallity to adapt a copied context with the correct wallet inside `perform_encode()` and pass it to the `queued` object. The following functionallity than picks up the correct wallet during injection and works as always.

### Persistance

If postgres is used as storage everything is persistent (credentials, connection, etc). However to get the data back from a wallet stored in the storage after the agent is restarted it needs to be opened again with `POST /wallet` by providing the same information as for the initial creation. This is because currently the list of handled wallets is not persisted by the agent.

## ToDo
- [ ] Make inbound message handling use key in message instead of endpoint
- [ ] Create larger test setup
- [ ] Production ready authorization setup
- [ ] Persist wallet  list

### DIFF
--- a/aries_cloudagent/config/default_context.py
+++ b/aries_cloudagent/config/default_context.py
@@ -2,7 +2,7 @@
 
 from .base_context import ContextBuilder
 from .injection_context import InjectionContext
-from .provider import CachedProvider, ClassProvider, StatsProvider
+from .provider import CachedProvider, ClassProvider, StatsProvider, DynamicProvider
 
 from ..cache.base import BaseCache
 from ..cache.basic import BasicCache
@@ -56,15 +56,15 @@ class DefaultContextBuilder(ContextBuilder):
 
         context.injector.bind_provider(
             BaseStorage,
-            CachedProvider(
-                StatsProvider(
-                    StorageProvider(), ("add_record", "get_record", "search_records")
-                )
-            ),
+            #CachedProvider(
+            StatsProvider(
+                StorageProvider(), ("add_record", "get_record", "search_records")
+            )
+            #),
         )
         context.injector.bind_provider(
             BaseWallet,
-            CachedProvider(
+            DynamicProvider(
                 StatsProvider(
                     WalletProvider(),
                     (
@@ -74,23 +74,24 @@ class DefaultContextBuilder(ContextBuilder):
                         # "unpack_message",
                         "get_local_did",
                     ),
-                )
+                ),
+                'wallet.name'
             ),
         )
 
         context.injector.bind_provider(
             BaseLedger,
-            CachedProvider(
-                StatsProvider(
-                    LedgerProvider(),
-                    (
-                        "create_and_send_credential_definition",
-                        "create_and_send_schema",
-                        "get_credential_definition",
-                        "get_schema",
-                    ),
-                )
-            ),
+            #CachedProvider(
+            StatsProvider(
+                LedgerProvider(),
+                (
+                    "create_and_send_credential_definition",
+                    "create_and_send_schema",
+                    "get_credential_definition",
+                    "get_schema",
+                ),
+            )
+            #),
         )
         context.injector.bind_provider(
             BaseIssuer,
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -24,6 +24,8 @@ from ..transport.outbound.message import OutboundMessage
 from ..utils.stats import Collector
 from ..utils.task_queue import TaskQueue
 from ..version import __version__
+from ..wallet_handler import WalletHandler
+from ..wallet_handler.error import WalletNotFoundError
 
 from .base_server import BaseAdminServer
 from .error import AdminSetupError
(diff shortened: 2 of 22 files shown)


swcurran:
Nice work on the DCO!  :-)

kukgini:
Thanks for the great work. As one suggestion, If It want to enable fail-over or scale-out,  It seems that the mapping part needs to be shared with multiple instances through an external storage. (like some kind of session cluster using redis or rdbms)

LukasSMueller:
Good point @kukgini! the mapping for each wallet is stored inside the wallet itself, maybe that can be used. There needs to be some synchronization if new mappings are created.

nrempel:
Thank you @LukasSMueller!

nrempel:
I've merged this into a working branch while we continue to work on this and collaborate on changes:

https://github.com/hyperledger/aries-cloudagent-python/tree/multitenancy

aries_cloudagent/config/default_context.py
```diff
@@ -74,23 +74,24 @@ async def bind_providers(self, context: InjectionContext):
...
-                    LedgerProvider(),
-                    (
-                        "create_and_send_credential_definition",
-                        "create_and_send_schema",
-                        "get_credential_definition",
-                        "get_schema",
-                    ),
-                )
-            ),
+            #CachedProvider(
```
baegjae, line 84: @LukasSMueller  This change results in an error when requesting "POST /credential-definitions" immediately after requesting "POST /schema".
This change seems to support multiple ledgers.
I appreciate for the multi tenant functionality, by the way.

Error log is below.
```
2020-08-04 13:41:38,885 indy.libindy.native.indy.commands.pool INFO 	src/commands/pool.rs:75 | OpenAck handle 14, pool_id 14, result Ok(())
2020-08-04 13:41:38,901 indy.libindy.native.indy.services.ledger INFO 	src/services/ledger/mod.rs:149 | build_get_schema_request() => Ok("{\"reqId\":1596516098900375000,\"identifier\":\"Th7MpTaRZVRYnPiabds81Y\",\"operation\":{\"type\":\"107\",\"dest\":\"Th7MpTaRZVRYnPiabds81Y\",\"data\":{\"name\":\"degree_schema\",\"version\":\"75.21.61\"}},\"protocolVersion\":2}")
2020-08-04 13:41:39,225 indy.libindy.native.indy.services.ledger INFO 	src/services/ledger/mod.rs:143 | build_schema_request() => Ok("{\"reqId\":1596516099225387000,\"identifier\":\"Th7MpTaRZVRYnPiabds81Y\",\"operation\":{\"type\":\"101\",\"data\":{\"name\":\"degree_schema\",\"version\":\"75.21.61\",\"attr_names\":[\"age\",\"name\",\"degree\",\"date\"]}},\"protocolVersion\":2}")
2020-08-04 13:41:39,379 aiohttp.access INFO 127.0.0.1 [04/Aug/2020:04:41:38 +0000] "POST /schemas HTTP/1.1" 200 415 "-" "Go-http-client/1.1"
2020-08-04 13:41:39,384 indy.libindy WARNING _indy_loop_callback: Function returned error 
2020-08-04 13:41:39,385 aries_cloudagent.core.dispatcher ERROR Handler error: AdminServer.make_application.<locals>.activate_wallet
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/ledger/indy.py", line 178, in open
    self.pool_handle = await indy.pool.open_pool_ledger(self.pool_name, "{}")
  File "/Users/baegjae/Library/Python/3.8/lib/python/site-packages/indy/pool.py", line 85, in open_pool_ledger
    res = await do_call('indy_open_pool_ledger',
indy.error.PoolLedgerInvalidPoolHandle
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/admin/server.py", line 287, in activate_wallet
    response = await handler(request)
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/messaging/credential_definitions/routes.py", line 133, in credential_definitions_send_credential_definition
    async with ledger:
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/ledger/indy.py", line 244, in __aenter__
    await self._context_open()
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/ledger/indy.py", line 213, in _context_open
    await self.open()
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/ledger/indy.py", line 178, in open
    self.pool_handle = await indy.pool.open_pool_ledger(self.pool_name, "{}")
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/indy/error.py", line 25, in __exit__
    raise IndyErrorHandler.wrap_error(
  File "/Users/baegjae/src/aries-cloudagent-python/aries_cloudagent/ledger/indy.py", line 178, in open
    self.pool_handle = await indy.pool.open_pool_ledger(self.pool_name, "{}")
  File "/Users/baegjae/Library/Python/3.8/lib/python/site-packages/indy/pool.py", line 85, in open_pool_ledger
    res = await do_call('indy_open_pool_ledger',
aries_cloudagent.ledger.error.LedgerConfigError: Exception when opening pool ledger: Error: Invalid pool handle
  Caused by: Pool with the same name is already opened
```

LukasSMueller: Hi @baegjae, thanks for reporting. I just quickly tested it with postman (creating schema -> creating cred_def) that worked. How are you hitting the endpoints with a script, postman, etc? And how did you start the agent?

baegjae: @LukasSMueller Thanks for reply. I used a script of controller for this. But I can replay error using postman. 1. create schema. 2. get schema from ledger (GET /schemas/{schema_id}) 3. get schema again right after response of 2. Maybe you can see 500 Error at postman and error log at agent.

baegjae: @LukasSMueller One solution I've found is to set ledger.keepalive to 0. It solves those errors. I'm not sure this solution can be final because it could degrade performance.

https://github.com/hyperledger/aries-cloudagent-python/blob/0a034b8f462ca10b7c809e2f5c4995503fe07b5b/aries_cloudagent/ledger/provider.py#L22

PR #628, opened 2020-07-23
unit test coverage for utils, remove vestigial out-of-band context param

sklump:
... and `utils/temp.py`, no longer in play for revocation and now dead code


aries_cloudagent/utils/task_queue.py
```diff
@@ -186,9 +186,9 @@ def __bool__(self) -> bool:
         """
         Support for the bool() builtin.
 
-        Otherwise, evaluates as false when there are no tasks.
+        Otherwise, evaluates as False when there are no tasks.
         """
-        return True
+        return bool(self.current_size)
```
andrewwhitehead, line 191: Isn't that what the method is trying to avoid?

sklump: * it strikes me as un-pythonic to have len() evaluate to zero and bool() come out True?
* this change matches the comments; the original does not, so something was wrong before?
* what is the point of a method that always returns True? Might as well not call it or, better, remove it?

andrewwhitehead: The comment was a bit confusing, but the issue was that the entire task queue would evaluate to false if it didn't currently have any tasks running, which was confusing when checking for its existence elsewhere.

sklump: OK, let me fix it and push the change tomorrow morning, thanks

PR #631, opened 2020-07-23
allow forwarded messages to use existing connections and the outbound queue

TelegramSam:
This adjustment allows the mediator forward function to work with endpoint-less connections, either via active connections with return route, or via the outbound queue. (Sending to an endpoint was already supported.)


aries_cloudagent/protocols/routing/v1_0/handlers/forward_handler.py
```diff
@@ -38,8 +38,20 @@ async def handle(self, context: RequestContext, responder: BaseResponder):
...
+        connection_verkey = connection_targets[0].recipient_keys[0]
+
         # Note: not currently vetting the state of the connection here
         self._logger.info(
             f"Forwarding message to connection: {recipient.connection_id}"
         )
-        await responder.send(packed, connection_id=recipient.connection_id)
+        await responder.send(
+            packed,
+            connection_id=recipient.connection_id,
```
andrewwhitehead, line 55: Looking at the responder code, I think you can pass `connection_targets` as `target_list`, instead of passing `connection_id`, and it would avoid the need to look up the connection information twice.

swcurran:
@andrewwhitehead -- good to go?  This should be it for 0.5.4, I think.

PR #634, opened 2020-07-29
feature set/get endpoints of different types on ledger (resolves #589)

### LINKED ISSUE #589: Different types of endpoints (services) for public DIDs
ACA-PY currently only uses one type of endpoint for local, as well as public DIDs. If we use the DIF indy resolver driver to resolve an Indy DID, then it generates a DID doc with one entry in the service array of type 'endpoint'. (See: https://uniresolver.io/1.0/identifiers/did:sov:HR6vs6GEZ8rHaVgjg2WodM)

We'd like to define multiple services/endpoints for a public DID (published on the ledger). The use case is to provide public profile/master data information about the entity represented by the DID. *

I started working on integrating this functionality in ACA-PY (setting and getting an endpoint of a different type) and I'd like to know if this would be a feature that might get accepted upstream. 

I'd be also interested in getting some feedback concerning the implementation:

Getting an endpoint seems straight forward to add to the /ledger/did-endpoint route:
image
 
However setting an endpoint is currently done with the /wallet/set-did-endpoint route. I've implemented setting an endpoint of a different type here, however, I only update the endpoint in the wallet if its of type "endpoint".

The alternative approaches would be 
- Add a /ledger/set-did-endpoint that allows to write endpoints of type != endpoint
- Carry the endpoints of type "other than endpoint" also in the wallet.

WiP can be found here: https://github.com/BoschDigitalSolutions/aries-cloudagent-python/tree/feat-endpoint-type

Looking forward to your thoughts.

* The public profile is represented as a JSON-LD Verifiable Presentation signed with the /jsonld/sign endpoint. The aim is to have a base interoperability for public profile information using only the W3C DID & VC spec without reliance on a particular ledger or DIDcomm.

domwoe:

### DIFF
--- a/aries_cloudagent/ledger/base.py
+++ b/aries_cloudagent/ledger/base.py
@@ -6,6 +6,8 @@ from typing import Tuple, Sequence
 
 from ..issuer.base import BaseIssuer
 
+from .util import EndpointType
+
 
 class BaseLedger(ABC, metaclass=ABCMeta):
     """Base class for ledger."""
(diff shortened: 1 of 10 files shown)


domwoe:
@sklump any feedback on how to move that forward?

sklump:
Is there a need to allow for provisioning with different endpoint types?

sklump:
I've merged master and completed unit test coverate in https://github.com/hyperledger/aries-cloudagent-python/pull/667.

Is this good to go? Please comment there.

PR #636, opened 2020-07-29
Add a pause when only retries are in the queue

ianco:

Added a sleep() when there are *only* retries in the queue.

Was causing problems with aries-vcr on startup - when the request was sent to establish a self-connection, aca-py was hogging cpu and preventing the aries-vcr startup from continuing; aries-vcr was unable to respond to callbacks until aca-py retries had all failed and timed out.

### DIFF
--- a/aries_cloudagent/transport/outbound/manager.py
+++ b/aries_cloudagent/transport/outbound/manager.py
@@ -392,6 +392,9 @@ class OutboundTransportManager:
             if self.outbound_buffer:
                 if (not new_pending) and (not retry_count):
                     await self.outbound_event.wait()
+                elif retry_count:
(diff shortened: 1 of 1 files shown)


nrempel:
Good catch. `await asyncio.sleep(0)` should (I think?) have essentially the same effect and avoid the use of magic numbers. But perhaps we want a slightly longer delay in any case.

ianco:
> Good catch. `await asyncio.sleep(0)` should (I think?) have essentially the same effect and avoid the use of magic numbers. But perhaps we want a slightly longer delay in any case.

I think `sleep(0.05)` is ok, a short delay doesn't hurt (the retry interval is much longer anyways), and the magic number is used in other delays elsewhere in aca-py

PR #640, opened 2020-07-30
Add please-ack to the message

baegjae:
This PR adds `~please-ack` to the message in below three cases.
- POST /connections/{conn_id}/accept-request
- POST /issue-credential/records/{cred_ex_id}/issue
- POST /present-proof​/records​/{pres_ex_id}​/send-presentation

**Reason:**
- I was testing the communication between faber (**aca-py**) and alice (**LibVCX**).
- I encountered the alice not sending an acknowledgement to the faber, and as a result the faber's status remains `credential_issued` after issuing.
- I checked the RFC of issue credential (https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential#issue-credential), and I found that `~please-ack` is needed if the issuer wants an acknowledgement.
- The faber's status was changed `credential_acked` correctly after adding `~please-ack`.
- I think `~please-ack` is needed to explicitly request an acknowledgement in above three cases.

### DIFF
--- a/aries_cloudagent/messaging/agent_message.py
+++ b/aries_cloudagent/messaging/agent_message.py
@@ -1,7 +1,7 @@
 """Agent message base class and schema."""
 
 from collections import OrderedDict
-from typing import Mapping, Union
+from typing import Mapping, Union, Sequence
 import uuid
 
 from marshmallow import (
@@ -18,6 +18,7 @@ from ..wallet.base import BaseWallet
 from .decorators.base import BaseDecoratorSet
 from .decorators.default import DecoratorSet
 from .decorators.signature_decorator import SignatureDecorator
+from .decorators.please_ack_decorator import PleaseAckDecorator
 from .decorators.thread_decorator import ThreadDecorator
 from .decorators.trace_decorator import (
     TraceDecorator,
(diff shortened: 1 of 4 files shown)


sklump:
I think this is correct - @swcurran can you imagine why we might have missed this? Is there a reason why we don't include it? I get the sense there may have been but I don't recall exactly why.

sklump:
It does need unit test coverage and I don't mind backfilling that once it's in master.

swcurran:
This needs to be optional.  I've not looked at the code, but it needs to align with the RFCs.  We have been talking about this on the V2 of present-proof and the idea we have settled on for whether there is a ACK at the end is that we'll be adding a field for the verifier to say as part of the request "will_confirm" (Yes/No), and the holder can add a ~please_ack if they want.

Note that a ~please_ack is a polite request, not a requirement, so the verifier may ignore it.  As such, there should be some timeout period and clean up done if the ACK never comes.

baegjae:
@swcurran Ok. I understand the current status. Thanks.

Quick question: Is there any way to change `credential_issued` to `credential_acked` at controller side ? or configuration of aca-py ?

sklump:
There is no API to manipulate exchange record states directly. Note that from the issuer point of view, the state is `credential_issued` at the end of the credential exchange because the issuer does not send an ack. The issuer issues a credential; the holder sends an ack.

Recall that the issuer should purge the cred ex record once it receives the ack. This is the default behaviour; `--preserve-exchange-records` is for troubleshooting/debugging.

baegjae:
Now, I understand the usage of credential exchange record clearly.
Many thanks for your kind explanation :)

I close this PR.

Issue #638, opened 2020-07-29
Update health/liveliness endpoints to check agent execution status 


ianco:
The health/liveliness checks should detect if the agent has reached a state where it cannot process requests.

One example is as follows - when the agent is in this state *every* requests returns this error:

```
Traceback (most recent call last):
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/core/dispatcher.py", line 178, in handle_message
    await handler(context, responder)
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/protocols/issue_credential/v1_0/handlers/credential_offer_handler.py", line 55, in handle
    holder_did=context.connection_record.my_did,
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/protocols/issue_credential/v1_0/manager.py", line 400, in create_request
    cred_req_result = await _create()
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/protocols/issue_credential/v1_0/manager.py", line 367, in _create
    async with ledger:
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/ledger/indy.py", line 244, in __aenter__
    await self._context_open()
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/ledger/indy.py", line 213, in _context_open
    await self.open()
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/ledger/indy.py", line 178, in open
    self.pool_handle = await indy.pool.open_pool_ledger(self.pool_name, "{}")
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/indy/error.py", line 27, in __exit__
    ) from err_value
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aries_cloudagent/ledger/indy.py", line 178, in open
    self.pool_handle = await indy.pool.open_pool_ledger(self.pool_name, "{}")
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/indy/pool.py", line 88, in open_pool_ledger
    open_pool_ledger.cb)
[...]

PR #643, opened 2020-07-31
Set liveliness status on fatal ledger error

ianco:

issue #638 

Set live check to False when we get a fatal ledger error

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -18,6 +18,7 @@ from marshmallow import fields, Schema
 
 from ..config.injection_context import InjectionContext
 from ..core.plugin_registry import PluginRegistry
+from ..ledger.error import LedgerConfigError
 from ..messaging.responder import BaseResponder
 from ..transport.queue.basic import BasicMessageQueue
 from ..transport.outbound.message import OutboundMessage
@@ -131,7 +132,12 @@ async def ready_middleware(request: web.BaseRequest, handler: Coroutine):
         "/status/live",
         "/status/ready",
     ) or request.app._state.get("ready"):
-        return await handler(request)
+        try:
+            return await handler(request)
+        except LedgerConfigError:
+            # fatal, signal server shutdown
+            request.app.notify_fatal_error()
+            raise
 
     raise web.HTTPServiceUnavailable(reason="Shutdown in progress")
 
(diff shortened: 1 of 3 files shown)


ianco:
Found some issues, so pulling the PR for further testing

ianco:
Fixed, added support for resetting liveliness on admin api's

aries_cloudagent/admin/server.py
```diff
@@ -338,6 +362,7 @@ async def start(self) -> None:
     async def stop(self) -> None:
         """Stop the webserver."""
         self.app._state["ready"] = False  # in case call does not come through OpenAPI
+        self.app._state["alive"] = False
```
WadeBarnes, line 365: I don't think we should unset `alive` on shutdown.  At this point aca-py is trying to shutdown gracefully and process any remaining work before voluntarily shutting down.  Setting `alive` to false provides k8s with a false indication that could cause it to kill the pod before it gracefully shuts down on it's own.

ianco: Good catch!

aries_cloudagent/core/conductor.py
```diff
@@ -255,12 +256,19 @@ def inbound_message_router(
...
-            lambda completed: self.dispatch_complete(message, completed),
-        )
+        try:
+            self.dispatcher.queue_message(
+                message,
+                self.outbound_message_router,
+                self.admin_server and self.admin_server.send_webhook,
+                lambda completed: self.dispatch_complete(message, completed),
+            )
+        except LedgerConfigError:
```
andrewwhitehead, line 266: Redundancy can be reduced with `except (LedgerConfigError, LedgerTransactionError):`. All the aca-py errors also derive from `BaseError` in case that's useful.

aries_cloudagent/admin/server.py
```diff
@@ -131,7 +132,24 @@ async def ready_middleware(request: web.BaseRequest, handler: Coroutine):
...
+            raise
+        except LedgerTransactionError:
+            # fatal, signal server shutdown
+            LOGGER.error(f"Shutdown with LedgerTransactionError")
+            request.app._state["ready"] = False
+            request.app._state["alive"] = False
+            raise
+        except Exception as e:
+            # some other error?
+            LOGGER.error("Handler error with exception: " + str(e))
```
andrewwhitehead, line 151: Log statements should use this format: `LOGGER.error("msg: %s", arg)` (also the f-strings above are not necessary)

PR #644, opened 2020-07-31
Update revocation demo

nrempel:
Update revocation demo to support old tails server environment variable for backwards compatibility and documentation alignment


ianco:
Is `--revocation` supposed to work *without* an external tail server?  When I run `./run_demo faber --revocation` without any other parameters/options I get an error:

```
  File "/home/indy/demo/runners/support/agent.py", line 452, in admin_request
    resp.raise_for_status()
  File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aiohttp/client_reqrep.py", line 946, in raise_for_status
    headers=self.headers)
aiohttp.client_exceptions.ClientResponseError: 400, message='tails_server_base_url not configured', url=URL('http://192.168.65.3:8021/credential-definitions')
```

andrewwhitehead:
Up to @ianco to approve as it's waiting on his changes requested

demo/AliceGetsAPhone.md
```diff
@@ -111,38 +104,37 @@ Open a new bash shell, and in a project directory, run:
...
 
-If you are running in *Play with Docker*, navigate to [The demo direcory](/demo) and run:
+If you are running in _Play with Docker_, navigate to [The demo direcory](/demo) and run:
 
 ```bash
 PUBLIC_TAILS_URL=https://def456.ngrok.io LEDGER_URL=http://test.bcovrin.vonx.io ./run_demo faber --revocation --events
 ```
 
-The `PUBLIC_TAILS_URL` parameter lets the demo script know how to connect to the tails server.  This can be running in another PWD session, or even on your local machine - the ngrok endpoint is public and will map to the correct location.
+The `PUBLIC_TAILS_URL` parameter lets the demo script know how to connect to the tails server. This can be running in another PWD session, or even on your local machine - the ngrok endpoint is public and will map to the correct location.
```
ianco, line 137: Need to remove use of PUBLIC_TAILS_URL here.  Code that references this env var is no longer used.

Note that TAILS_NETWORK still works based on updates to ngrok_wait.sh

demo/runners/support/agent.py
```diff
@@ -38,6 +36,8 @@
 
 AGENT_ENDPOINT = os.getenv("AGENT_ENDPOINT")
 
+PUBLIC_TAILS_URL = os.getenv("PUBLIC_TAILS_URL")
```
ianco, line 39: Remove references to PUBLIC_TAILS_URL in agent.py?  Note that this code will never get reached due to the validation check for `if args.revocation and not args.tails_server_base_url` in faber.py

OR move use of this env var into faber.py?

demo/AliceGetsAPhone.md
nrempel: I updated the demo to function with either `--tails-server-base-url` or `PUBLIC_TAILS_URL` so this should still work

PR #656, opened 2020-08-07
allow (but exclude) extra fields in protocols and messages where conc…

sklump:
…eivably OK

I (used to) fundamentally disagree with this approach: I believe standards should be exclusive to ensure the tightest possible profile for interop, but I concede that embrace-and-extend is the zeitgeist of the day. Also, allowing extra fields and ignoring them by default puts us on a more error-resistant posture when it comes time to jockey multiple versions of a protocol.

This work allows extra fields (which aca-py discards) on protocols and messages without blowing up.

~I'm also working on a less invasive approach which might not suffice in the end. I'd rather not see this one merged until I fully exhaust that possibility.~ _update: it's cute but no better; prefer this approach_

For these two reasons I will mark this PR as draft for the moment.
_update_: I've made it ready to merge and am satisfied with the work if the group is.

### DIFF
--- a/aries_cloudagent/connections/models/connection_target.py
+++ b/aries_cloudagent/connections/models/connection_target.py
@@ -2,7 +2,7 @@
 
 from typing import Sequence
 
-from marshmallow import fields
+from marshmallow import EXCLUDE, fields
 
 from ...messaging.models.base import BaseModel, BaseModelSchema
 from ...messaging.valid import INDY_DID, INDY_RAW_PUBLIC_KEY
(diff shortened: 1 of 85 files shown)


sklump:
I've decided to abandon the other approach I was on, here: https://github.com/sklump/aries-cloudagent-python/tree/exclude-extra-schema-fields-take-1, with https://github.com/sklump/aries-cloudagent-python/blob/67110f634dd8e1dca631e79449f44754bcde350e/aries_cloudagent/messaging/models/base.py#L206 being too inaccessible to troubleshoot in corner cases. Experience tells me it would become a magnet for expensive and futile attention - even when not at fault - while not really reducing the code base enough for the magic it adds.

So I'm going to try to make this PR not a draft.

nrempel:
If I understand this correctly, this allows extra ignored parameters in requests to the admin API? As far as I was aware, extra parameters in protocol messages are already ignored.

sklump:
Ha ha! Apparently so. Mostly, anyway. Inner components of messages do not: in particular, credential proposal.

There must be a less intrusive way forward here.

nrempel:
Agreed, it looks good. Just checking on what this covers!

PR #664, opened 2020-08-14
Mapping of inbound messages based on key in message body -> eliminatess need for individual path mappings

LukasSMueller:
This PR introduces the following changes:
- Wallet id is passed in `context.settings` instead of set in provider itself
- Some routes updated to use adjusted  local context
- Ledger und storage provider handle multiple intances to align with their own stored wallet instance


aries_cloudagent/protocols/connections/v1_0/manager.py
```diff
@@ -317,6 +335,14 @@ async def create_request(
...
         await connection.save(self.context, reason="Created connection request")
 
+        # MULTITENANCY: Add new key to handled keys to associate wallet to new connection.
+        ext_plugins = self.context.settings.get_value("external_plugins")
+        if ext_plugins and 'aries_cloudagent.wallet_handler' in ext_plugins:
+            wallet_handler: WalletHandler = await self.context.inject(WalletHandler)
+            wallet_id = self.context.settings.get_value("wallet.id")
+            wallet: BaseWallet = await self.context.inject(BaseWallet)
+            await wallet_handler.add_key(my_info.verkey, wallet_id)
+
```
LukasSMueller, line 345: To be able to associate inbound messages for this connection a mapping between the `verkey` of this connection and the `wallet_id` is stored.

aries_cloudagent/wallet_handler/handler.py
```diff
@@ -228,6 +248,33 @@ def get_paths(self, wallet: str) -> []:
...
+            kid = recipient['header']['kid']
+            try:
+                wallet_id = await self.get_wallet_for_key(kid)
+            except KeyNotFoundError:
+                wallet_id = None
+            wallet_ids.append(wallet_id)
+
+        return wallet_ids
+
     async def get_wallet_for_path(self, path: str) -> str:
```
LukasSMueller, line 278: To associate a wallet to the inbound message we parse the body and check the key send in `kid` with the mapping stored in the handler instead of having an inbound path specific for that connection which than is  used as mapping to reference wallet.

aries_cloudagent/wallet_handler/handler.py
```diff
@@ -255,6 +302,16 @@ async def add_connection(self, connection_id: str, wallet_id: str):
...
+    async def add_key(self, key: str, wallet_id: str):
+        """
+        Add a mapping between the given connection and wallet.
+
+        Args:
+            connection_id: Indentifier of the new connection.
+            wallet_id: Identifier of the wallet the connection belongs to.
+        """
+        self._handled_keys[key] = wallet_id
+
```
LukasSMueller, line 314: Each verkey used  for a connection needs to be stored and mapped to the wallet to which the connection belongs.

LukasSMueller:
PR now includes update for mapping of inbound messages to wallet. Agent checks key included in message and checks this against a stored mapping from keys to wallets.

aries_cloudagent/wallet_handler/handler.py
ianco: Does the other agent need to do anything differently for this to work?

ianco (approved):

aries_cloudagent/wallet_handler/handler.py
LukasSMueller: No the other agent doesn't need to know if it is a multitenancy or standard agent. It just uses the informtion defined in standard [Aries RFC 0019: Encryption Envelope](https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope)

PR #667, opened 2020-08-17
Endpoint type merge

sklump:
Merge current master into Dominic Wörner's PR#589 (which became #634) endpoint-type work en route to merging back into master, unit test coverage

### DIFF
--- a/aries_cloudagent/ledger/base.py
+++ b/aries_cloudagent/ledger/base.py
@@ -6,6 +6,8 @@ from typing import Tuple, Sequence
 
 from ..issuer.base import BaseIssuer
 
+from .util import EndpointType
+
 
 class BaseLedger(ABC, metaclass=ABCMeta):
     """Base class for ledger."""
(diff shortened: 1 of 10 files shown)


sklump:
Merges current master into https://github.com/hyperledger/aries-cloudagent-python/pull/634 in preparation for merge of this (endpoint-type) work back into aca-py master.

domwoe:
Thanks!

domwoe:
> Is there a need to allow for provisioning with different endpoint types? Since only `endpoint_type==EndpointType.ENDPOINT` affects the wallet, I am guessing that there is no such need?

This could make sense. In particular if you want to do all ledger operations during provisioning

sklump:
> 
> 
> > Is there a need to allow for provisioning with different endpoint types? Since only `endpoint_type==EndpointType.ENDPOINT` affects the wallet, I am guessing that there is no such need?
> 
> This could make sense. In particular if you want to do all ledger operations during provisioning

I think that could be separate PR though. This one introduces a significant change on its own. I can get on it once it's merged.

Issue #652, opened 2020-08-05
API to query the current role of our agent


Sarangc1203:
It would be beneficial from the controller's point of view to know what the my role (corresponding the current set public DID) is so that we can separate out application specific functionalities according to them.

swcurran:
What set of roles are you talking about?

Howlla:
I think roles refer to "Steward", "Endorser", "Trustee" etc.

swcurran:
That would mean there would be a need to include the public DID being referenced, as there is not the concept of the agent having "a DID'.  Is there an existing admin API point related to DIDs that could be used?

I would add that if a peer DID is submitted the result could be "Peer".

sklump:
There's a `register-nym()` but we have no corresponding `get-nym()` taking a public DID. I could put it on the list.

sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/671

PR #671, opened 2020-08-20
get-nym-role from ledger for public DID

sklump:


sklump:
addresses https://github.com/hyperledger/aries-cloudagent-python/issues/652

sklump:
... as it turns out, somehow I moved these changes into PR#672 and then wiped them out, so the net delta was zero. Resubmitted in #673.

Issue #670, opened 2020-08-19
TAA acceptance should not be required when running in read-only mode


esune:
When starting an agent in read-only mode, acceptance of the TAA should not be required since no writing to the ledger will ever be performed.

To start aca-py in read-only mode:
- use the `--read-only-ledger` startup argument
- do NOT set a `--seed`
- do NOT set a `--endpoint`

swcurran:
I'm surprised about not setting the `--endpoint` -- why is that?  I thought that parameter was to set the endpoint for DIDComm, and was not to do with a public DID or anything to do with the ledger.  Is that parameter for something else?

esune:
Setting the `--endpoint` parameter causes the agent to want to update it's endpoint registered on the ledger, throw an exception because the ledger is not writable and exit. I am not sure whether that is the desired/expected behaviour, or something needs tweaking.

sklump:
I can take a look later today. Recall that read-only was a bolt-on and there are likely a few gotchas left to find.

sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/672

sklump:
@esune does PR #672 fix the issue to your satisfaction?

esune:
I have not tried yet re-provisioning an agent, but based on the code changes I would say so. Thanks!

PR #672, opened 2020-08-20
Fix provisioning on read only ledger

sklump:


sklump:
addresses https://github.com/hyperledger/aries-cloudagent-python/issues/670

sklump:
Hold on, I've mashed #671 into #672: please don't merge this just yet

Issue #676, opened 2020-08-21
GET /wallet/did returns a list where "public" attribute is not detecting dids on ledger and is not boolean


jcourt562:
V0.5.3 of ACA-py (have also tries with latest from master)

The swagger model for "GET /wallet/did" indicates an attribute of "public" is returned indicating "Whether DID is on the ledger".  Which is of type "boolean".  Following is the output response from a running ACA-py for two dids.  Both have been written to the ledger which was verified by both looking at the VON ledger and by subsequent calls to "GET /ledger/did-verkey" which is how I am working around this issue currently.

{
  "results": [
    {
      "did": "5zWfemJCNAjRYZQcQYsBy4",
      "verkey": "3io2gJqGxtuZKPrfcqYHE3Mbohs4a8pe95F73W7o3bLH",
      "public": "false"
    },
    {
      "did": "FfTM7q3uZLipmQnXjghcbZ",
      "verkey": "8zXY4oDUzc73X8GnJesfu2VtKWiajdUUbDvSShjuYPfM",
      "public": "true"
    }
  ]
}

NOTE: the type of the "public" attribute is declared as "Boolean" in the swagger interface but is being returned as a "string" in the json body also.

PR #681, opened 2020-08-21
match get-did-list to response "public" schema spec and fix misleadin…

sklump:
…g description text

https://github.com/hyperledger/aries-cloudagent-python/issues/676

### DIFF
--- a/aries_cloudagent/ledger/base.py
+++ b/aries_cloudagent/ledger/base.py
@@ -81,6 +81,15 @@ class BaseLedger(ABC, metaclass=ABCMeta):
         """
 
     @abstractmethod
+    async def get_nym_role(self, did: str):
+        """
+        Return the role registered to input public DID on the ledger.
+
+        Args:
+            did: DID to register on the ledger.
+        """
+
+    @abstractmethod
     def nym_to_did(self, nym: str) -> str:
         """Format a nym with the ledger's DID prefix."""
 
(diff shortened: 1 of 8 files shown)


jcourt562:
So the scenario I am doing is that a user creates multiple private DIDs using POST /wallet/did/create.  The intention for this could be to create connections or use in create credential definitions with a different DID for various reasons (i.e avoiding correlation).  At some point they may need to write those DIDs to the ledger using POST /ledger/register-nym.  In order to show them which DIDs are on at the ledger (i.e. and so can't be written again) so that for example if they want to swap their current DID in use, I need to know which ones are written to ledger.  Since GET /wallet/did doesn't return an indication of which are written to ledger I am currently having to perform GET /ledger/did-verkey.  Is there a better way ? I assume GET /wallet/did even with filtering indicating public won't work as if it did it would make sense to return the "on ledger" status in the existing responses rather than just the indication for which did is currently being used by ACA-py ?

When you say they have been superseded, I don't see anything in the swagger ui indicating GET ledger/did-verkey is deprecated or superseded ?  If they are what is the replacement ?

Thanks

sklump:
You are following the best practice as it stands.

We plan to revisit public/local DIDs after 0.5.4 because of the awkwardness you raise here. On-ledger or off-ledger ought to be in metadata and available via the wallet API. However, I suspect this will have a lot of tricky affected-bys and so it waits until we get the current release out.

jcourt562:
Thanks for the feedback !

sklump:
Please check your rocket chat for further discussion.

PR #690, opened 2020-08-26
Started integrating WalletRecord model into multitenancy implementation

LukasSMueller:


aries_cloudagent/config/wallet.py
```diff
@@ -29,6 +31,27 @@ async def wallet_config(context: InjectionContext, provision: bool = False):
...
+    #    tag_filter["thread_id"] = request.query["thread_id"]
+    #post_filter = {
+    #    k: request.query[k]
+    #    for k in ("connection_id", "role", "state")
+    #    if request.query.get(k, "") != ""
+    #}
+    post_filter = {"wallet_name": "base"}
+    wallet_records = await WalletRecord.query(context, post_filter_positive=post_filter)
+    #wallet_records = await WalletRecord.query(context)
+
```
LukasSMueller, line 54: At startup we need to store the record  for the base wallet.

aries_cloudagent/protocols/connections/v1_0/manager.py
```diff
@@ -190,6 +193,20 @@ async def create_invitation(
...
+            wallet_id = self.context.settings.get_value("wallet.id")
+            await wallet_handler.add_connection(connection.connection_id, wallet_id)
+            await wallet_handler.add_key(connection.invitation_key, wallet_id)
+            post_filter = {"wallet_name": self.context.settings.get("wallet.name")}
+            wallet_records = await WalletRecord.query(
+                self.context)
+            wallet_record = wallet_recordes[0]
+            wallet_record.add_key(connection.invitation_key)
+            wallet_record.save(self.context)
+
```
LukasSMueller, line 209: The last 4 lines would replace the ones above. Problem is that here we are in the "specific wallet context"(!=base) and therefore we  don't find  the records I think. Wallet_records are []

PR #691, opened 2020-08-26
Did posture include ledger non public

sklump:


sklump:
aw heck, I just saw this ran afoul of the new black version: fixing and pushing update

Issue #698, opened 2020-08-27
Proof request specifying a non-revocation interval fails when providing non-revocable credentials 


esune:
When a verifier submits a proof request that contains a non-revocation interval and the holder responds using non-revocable credentials that meet the remaining restrictions, the verification of the proof fails and the verifier throws the following exception:
```
Faber      | 2020-08-26 17:23:52,896 aries_cloudagent.verifier.indy ERROR Validation of presentation on nonce=561590513150476300113269 failed with error
Faber      | Traceback (most recent call last):
Faber      | Proof = false
Faber      | Presentation: state = verified , presentation_exchange_id = 1b704966-bbba-4def-8cb9-26b3ea956ec8
Faber      |   File "/home/indy/aries_cloudagent/verifier/indy.py", line 211, in verify_presentation
Faber      |     json.dumps(rev_reg_entries),
Faber      |   File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/indy/anoncreds.py", line 1756, in verifier_verify_proof
Faber      |     verifier_verify_proof.cb)
Faber      | indy.error.CommonInvalidStructure
```

This seems to be caused by "missing" revocation information when verifying the proof.

The situation described above is the default use-case for authentication scenarios (e.g.: [vc-authn-oidc](https://github.com/bcgov/vc-authn-oidc)) where the Identity Provider will ALWAYS send a non-revocation interval in the proof request regardless of the type of credentials (revocable or not) that can be accepted, since it cannot be known ahead of time.

@sklump

sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/699 addresses.

PR #699, opened 2020-08-27
Remove superfluous non revoc intervals

sklump:
Also a ton of black tweaks arising from the new version.

Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/698


esune:
I have tested this using non-revocable credentials as well as revocable credentials and it works as expected.
Still need to test using a mix of both.

Issue #696, opened 2020-08-27
Incorrect Cred id on send_presentation on Admin API is returning a 500 Internal Server Error


nodlesh:
## Issue
Tested thought the Admin API and setup and sent through the send-presentation operation, where the cred_id is incorrect for the credential.
```
{
   "presentation": {
      "comment": "This is a comment for the send presentation.",
      "requested_attributes": {
         "address_attrs": {
            "cred_type_name": "Schema_DriversLicense",
            "revealed": true,
            "cred_id": "d413f359-9595-4ab6-94cf-1aa191d2c884"
         }
      }
   }
}
```

## Outcome
The response from the Admin API is 500 Internal Server Error. 
However, the logs in the agent is showing the errors below. 
```
2020-08-26 22:04:38,911 aries_cloudagent.core.dispatcher ERROR Handler error: presentation_exchange_send_presentation
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/holder/indy.py", line 275, in get_credential
    self.wallet.handle, credential_id
  File "/usr/local/lib/python3.7/site-packages/indy/anoncreds.py", line 923, in prover_get_credential
    prover_get_credential.cb)
indy.error.WalletItemNotFound

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/protocols/present_proof/v1_0/routes.py", line 927, in presentation_exchange_send_presentation
    comment=body.get("comment"),
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/protocols/present_proof/v1_0/manager.py", line 297, in create_presentation
    await holder.get_credential(credential_id)
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/holder/indy.py", line 281, in get_credential
    credential_id, self.wallet.name
aries_cloudagent.wallet.error.WalletNotFoundError: Credential d413f359-9595-4ab6-94cf-1aa191d2c884 not found in wallet aca-py352817
2020-08-26 22:04:38,912 aries_cloudagent.admin.server ERROR Handler error with exception: Credential d413f359-9595-4ab6-94cf-1aa191d2c884 not found in wallet aca-py352817
2020-08-26 22:04:38,912 aiohttp.server ERROR Error handling request
Traceback (most recent call last):
[...]

PR #700, opened 2020-08-27
catch and raise PR for WalletNotFoundError on wrong cred_id for creat…

sklump:
…e-proof operation


sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/696.

Issue #701, opened 2020-08-27
Issuer doesn't download tails file from server if not present in local filesystem


esune:
### Issue

When issuing a credential that supports revocation, if the tails file is NOT on the issuer agent's filesystem as expected, the issuing process is disrupted resulting in a credential permanently in the `pending` state. In the issuer logs, the following exception is observable:
```
Faber      | 2020-08-27 23:34:18,557 aries_cloudagent.core.dispatcher ERROR Handler error: credential_exchange_issue
Faber      | Traceback (most recent call last):
Faber      |   File "/home/indy/aries_cloudagent/protocols/issue_credential/v1_0/routes.py", line 1057, in credential_exchange_issue
Faber      |     ) = await credential_manager.issue_credential(cred_ex_record, comment=comment)
Faber      |   File "/home/indy/aries_cloudagent/protocols/issue_credential/v1_0/manager.py", line 551, in issue_credential
Faber      |     tails_path,
Faber      |   File "/home/indy/aries_cloudagent/issuer/indy.py", line 208, in create_credential
Faber      |     if tails_file_path is not None
Faber      |   File "/home/indy/aries_cloudagent/indy/__init__.py", line 15, in create_tails_reader
Faber      |     raise FileNotFoundError("Tails file does not exist.")
Faber      | FileNotFoundError: Tails file does not exist.
Faber      | 2020-08-27 23:34:18,560 aries_cloudagent.admin.server ERROR Handler error with exception: Tails file does not exist.
Faber      | 2020-08-27 23:34:18,561 aiohttp.server ERROR Error handling request
Faber      | Traceback (most recent call last):
Faber      |   File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 418, in start
Faber      |     resp = await task
Faber      |   File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aiohttp/web_app.py", line 458, in _handle
Faber      |     resp = await handler(request)
Faber      |   File "/home/indy/.pyenv/versions/3.6.9/lib/python3.6/site-packages/aiohttp/web_middlewares.py", line 119, in impl
Faber      |     return await handler(request)
Faber      |   File "/home/indy/aries_cloudagent/admin/server.py", line 154, in ready_middleware
Faber      |     raise e
Faber      |   File "/home/indy/aries_cloudagent/admin/server.py", line 144, in ready_middleware
[...]

PR #702, opened 2020-08-28
do not assume as issuer that local tails file is in place

sklump:


sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/701

Issue #694, opened 2020-08-27
Verify receipt and add method to change to send out message types with "https://didcomm.org" prefix


swcurran:
Work was done previously to ensure that we support the receipt of either "did:sov:" or "https://didcomm.org" as the prefix of a message type.

We need to make absolutely sure that we can receive the new format as the community has decided anyone can send out using the new format.

As well, we need to change ACA-Py to always send out the new format, likely with a startup flag that lets us use the old format.

The RFC to transition the community to the new format is about to allow that.

This is the PR that implemented the change to support the new type: #379

This is a link to the [community transition RFC](https://github.com/hyperledger/aries-rfcs/tree/master/features/0348-transition-msg-type-to-https) that is about to be in Step 2.

swcurran:
@sklump -- would you please be able to look at this in the next while?  Let me know if you need more details.  We'd like to demonstrate that the community can do one of these updates without wreaking havoc upon our world.

Transition Step 2 has not officially happened but will Real Soon Now,

sklump:
Probably next week I can start to look at it

sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/705 addresses

sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/713 backs off to emit old prefix by default until official Stage 2 adoption.

PR #705, opened 2020-09-02
Support message type transition phase 2

sklump:
Command line flag is `--emit-old-didcomm-prefix`, default to new, always accept either


sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/694

andrewwhitehead (approved):
Looks good

PR #709, opened 2020-09-04
Push tails file retries

sklump:


sklump:
I've marked this as draft while I ruminate over the state machine for revocation registries. It may be frayed since the introduction of STAGED.

aries_cloudagent/tails/indy_tails_server.py
```diff
@@ -10,13 +10,23 @@ class IndyTailsServer(BaseTailsServer):
     """Indy tails server interface."""
 
     async def upload_tails_file(
-        self, context, rev_reg_id: str, tails_file_path: str
+        self,
+        context,
+        rev_reg_id: str,
+        tails_file_path: str,
+        interval: float = 1.0,
+        backoff: float = 0.25,
```
esune, line 18: It may just be me not being familiar with this, but these interval and backoff do not appear to be used: this would cause the attempts to be sequentially kicked-off based on the time it took to complete (fail) the previous request.

aries_cloudagent/utils/http.py
```diff
@@ -96,9 +101,65 @@ async def fetch(
...
+
+async def put(
+    url: str,
+    file_data: dict,
+    extra_data: dict,
+    *,
+    retry: bool = True,
+    max_attempts: int = 5,
+    interval: float = 1.0,
+    backoff: float = 0.25,
```
esune, line 121: Same as previous comment: they don't appear to be used anywhere?

esune:
I tested this by simulating an outage on the tails-server:
1. take tails-server down
2. try generating a cred_def that supports revocation
3. bring tails-server up after a few moments

I managed to break it once by waiting ~20s before restarting the tails-server and have it working (upload successful and cred_def creation successful) by waiting a slightly shorter amount of time, about 10s.

aries_cloudagent/utils/http.py
sklump: Sure they are: see below

aries_cloudagent/utils/http.py
```diff
@@ -96,9 +101,65 @@ async def fetch(
...
+
+    """
+    (data_key, file_path) = [k for k in file_data.items()][0]
+    data = {**extra_data}
+    limit = max_attempts if retry else 1
+
+    if not session:
+        session = ClientSession(connector=connector, connector_owner=(not connector))
+    async with session:
+        async for attempt in RepeatSequence(limit, interval, backoff):
```
sklump, line 151: limit, interval, backoff. See utils/http.py. Parameters interval and backoff shape the power series heuristic for retry times.

sklump:
More troubleshooting: we're still converging toward greatness

aries_cloudagent/utils/http.py
ianco: FYI in the bc registries issuer we use the "backoff" library for this, for example:

https://github.com/bcgov/von-bc-registries-agent/blob/aries-refactoring/data-pipeline/bcreg/credssubmitter.py#L114

aries_cloudagent/utils/http.py
esune: Ha, I missed that line. Thanks.

aries_cloudagent/tails/indy_tails_server.py
esune: Based on the comment to `aries_cloudagent/utils/http.py ` it then looks like these are in fact not being used, and the default values defined for the utility method are used instead.

sklump:
PR #717 Supersedes.

PR #717, opened 2020-09-11
Troubleshoot tails file rollover

sklump:
Put tails file up from admin API.
Simplify issuer revocation registry state machine and related issue-credential logic.


esune:
I think we got it! Tested as follows:

1. create new test schema
2. create new cred_def based on new schema, supporting revocation
   - simulated outage on tails-server during creation so that tails files could not get uploaded
3. the agent created two revocation registries (that would not have a tails file on the server): marked both of them full
4. issue credentials until the first registry is full, revoke all of them
5. continue issuing credentials on the next registry, and revoke them

Everything seemed to work as expected so hopefully we're good. There is a chance of additional outages happening right when one of the "next" tails-files is being uploaded to the tails-server, but haven't pushed my luck that far with testing since it really looks like an even rarer edge case considering now we have multiple retries to upload.

esune (approved):

andrewwhitehead (approved):

PR #725, opened 2020-09-28
Outbound msg limit

krgko:
Currently, The ACA-py allows for setting `--max-message-size` for inbound transport only. I think someone who uses the webhook feature might need to apply that config for the outbound transport max message limit also due to large messages will lead to `OutboundTransportError('Unexpected response status 413, caused by: Request Entity Too Large')`

This PR might be useful for some use cases which use the webhook feature and large messages.


andrewwhitehead:
I can't be sure if this is actually fixing an issue or if it's just putting a maximum size on the outbound requests. It seems like the 413 error is being sent by the receiving agent because the payload is too large for it, not because there's a limit on the outbound payload size.

krgko:
Hello @andrewwhitehead , I've rechecked the AioHTTP document and found that `max_client_size` can config for inbound transportation. Thus, this PR will not help because the outbound max request size relies on the allowed inbound request size for destination service. 

Thank you for your comment.

Issue #721, opened 2020-09-23
Credential response json doesn't match Swagger documentation


ldej:
The expected structure of a credential from `/credential/{credential_id}` and `/credentials` is:
```json
{
  "values": {
    "additionalProp1": {
      "raw": "Alex",
      "encoded": "412821674062189604125602903860586582569826459817431467861859655321"
    },
    "additionalProp2": {
      "raw": "Alex",
      "encoded": "412821674062189604125602903860586582569826459817431467861859655321"
    },
    "additionalProp3": {
      "raw": "Alex",
      "encoded": "412821674062189604125602903860586582569826459817431467861859655321"
    }
  },
  "rev_reg": {
    "accum": "21 136D54EA439FC26F03DB4b812 21 123DE9F624B86823A00D ..."
  },
  "signature_correctness_proof": {},
  "witness": {
    "omega": "21 129EA8716C921058BB91826FD 21 8F19B91313862FE916C0 ..."
  },
  "rev_reg_id": "WgWxqztrNooG92RXvxSTWv:4:WgWxqztrNooG92RXvxSTWv:3:CL:20:tag:CL_ACCUM:0",
  "schema_id": "WgWxqztrNooG92RXvxSTWv:2:schema_name:1.0",
  "signature": {},
  "cred_def_id": "WgWxqztrNooG92RXvxSTWv:3:CL:20:tag"
}
```
However, the result I get back is:
```json
{
  "referent": "my-credential-identifier",
  "attrs": {
    "name": "b",
    "email": "a"
  },
  "schema_id": "UpFt248WuA5djSFThNjBhq:2:my-schema:1.0",
  "cred_def_id": "UpFt248WuA5djSFThNjBhq:3:CL:107:my-schema",
  "rev_reg_id": null,
  "cred_rev_id": null
}
```

The main difference is the missing `referent` in the expected response, and `attrs` vs `values`.

I can update either if I get some instructions on which is the preferred structure.

sklump:
Leave it with me and I will take care of it first thing Monday.

sklump:
https://github.com/hyperledger/aries-cloudagent-python/pull/726 addresses

ldej:
It's almost right. After #726 `attrs` is still formatted like:
```json
  "attrs": {
    "additionalProp1": "Alex",
    "additionalProp2": "Alex",
    "additionalProp3": "Alex",
  },
```
But the documentation says:
```json
      "attrs": {
[...]

PR #726, opened 2020-09-28
holder credential queries return cred briefs, not full in-wallet cred…

sklump:
…entials


sklump:
Addresses issue #721

andrewwhitehead (approved):

Issue #521, opened 2020-05-21
Improve response messages and http status codes in admin API


nrempel:
With regards to aca-py error handling I've created a couple tickets as examples of the kinds of cases we should handle better:

https://github.com/hyperledger/aries-cloudagent-python/issues/520
https://github.com/hyperledger/aries-cloudagent-python/issues/519
https://github.com/hyperledger/aries-cloudagent-python/issues/518

These are just examples though. These sorts of cases are all over the place. I don't think it's worth writing them all up as tickets. Some cases are responses to admin API requests that can return more useful information and appropriate status codes. Others might be related to how the server has been run (with or without a ledger connection, for example). It should be possible to run aca-py without a ledger connection, but we should gate access to particular functionality better.

andrewwhitehead:
Status code handling has been improved but we might still want some changes around missing ledger connections

ianco:
Looking through some of Nick's tickets it seems that the startup parameters need to be reviewed as well.

E.g. should "--wallet indy" be required?  Should "--endpoint" be required?

I don't know if it makes sense to run an aca-py agent without some of these parameters specified

ianco:
In general the API's *should* return:

- HTTP 400 (with an error message) for errors on input parameters (user can retry with different parameters)
- HTTP 404 if a record is expected and not found (generally for GET requests that fetch a single record)
- HTTP 500 if there is some other processing error (i.e. won't make any difference what parameters the user tries) with an error message

.. and should *not* return:

- HTTP 500 with a stack trace (we should handle error conditions with a 400 or 404 response, and catch errors and provide a meaningful error message)

swcurran:
I definitely think we need a better way to document what is needed.  The plan of going to a config file would really help with that -- here are the recommended settings, and a doc section on why some options would/would not be needed.

ianco:
aca-py start: error: the following arguments are required: -it/--inbound-transport, -ot/--outbound-transport

What can the agent do with this minimal set of parameters?

There are basically 4 roles we can run an agent:  Issuer, Holder/Prover, Verifier, Router/Mediator

Correct?

The first 3 (currently) require an Indy ledger (and therefore an Indy wallet), I don't think Router/Mediator does ...

ianco:
In terms of startup parameters, I suggest when running "aca-py start":

- Endpoint should be required ("-e" or "--endpoint")
- Ledger should be required ("--genesis-transactions", "--genesis-file" or "--genesis-url")
[...]

PR #739, opened 2020-10-02
Aca-py startup arg updates

ianco:
Make "--endpoint" required.  Make genesis required (url, file or txns) unless "--no-ledger" is specified.  Make some "indy" wallet parameters required (name, key, etc) if indy wallet type is specified.

Added "--arg-file <file>" to allow params to be included in a file.  (File is just a list of args, not structured with yml etc)  Had to change some related logic - couldn't flag args as "required" so moved the check to the "get_settings()" method and threw "ArgsParseError"; display usage on any ArgsParseError.  (Had to fix a couple of unit tests that expected the "old" behaviour.

Related to issue #521


swcurran (commented):
Nice!

swcurran:
Hmmm... -0.03% takes us below 99%.  That's sad...

ianco:
Note that configargparse is pointing to my fork (in requirements.txt) pending them accepting my PR

ianco:
Some discussion around environment variables - the "ConfigArgParse" library adds support for env vars, but they must be configured per parameter, like for example:

        parser.add_argument("--wallet-storage-config", env_var="ACAPY_WALLET_STORAGE_CONFIG", help='....')

If the environment variable is present the value gets used, regardless of whether there is a config file or not.

However you *cannot* substitute env vars in the config file, like the following:

        input-transport = [["http", "${HOST}", "${PORT}"]]

The list of available environment variables (from the parser.add_argument() functions) is included in the usage() ("aca-py -h"), like:

 --wallet-name <wallet-name>
                        Specifies the wallet name to be used by the agent.
                        This is useful if your deployment has multiple
                        wallets. [env var: ACAPY_WALLET_REKEY]

swcurran:
> Some discussion around environment variables - the "ConfigArgParse" library adds support for env vars, but they must be configured per parameter, like for example:
> 
> ```
>     parser.add_argument("--wallet-storage-config", env_var="ACAPY_WALLET_STORAGE_CONFIG", help='....')
> ```
> 
> If the environment variable is present the value gets used, regardless of whether there is a config file or not.
> 
> However you _cannot_ substitute env vars in the config file, like the following:
> 
> ```
>     input-transport = [["http", "${HOST}", "${PORT}"]]
> ```
> 
> The list of available environment variables (from the parser.add_argument() functions) is included in the usage() ("aca-py -h"), like:
> 
> --wallet-rekey
> Specifies .... [env var: ACAPY_WALLET_REKEY]

IMHO - this is even better.  We get DRY documentation, we tightly tie the environment variables to the arguments.  It does change how we have to environment variables, but the change looks like it moves us in the right direction.  Ideal to use this always.

Next challenge -- being able to use argparse as defined here in bash scripts :-)  -- that would be awesome...

ianco:
It's not perfect - it would be nice to have the param substitution *within* the config file as well.

For example multi-value parameters (like "input-transport", which is also a multi-instance parameter = multi instances of multi values) would be confusing to jam into a single environment variable.

swcurran:
Except there is a very clear rule on how to do it, with usage information that makes it precise. What we are doing now is mixing and matching approaches that vary from variable to variable.  Neither is perfect, but I would lean to consistency.

But my vote doesn't count much (and rightly so :-) ) in these decisions.

ianco:
Env vars configured for:

--admin-api-key:  ACAPY_ADMIN_API_KEY
--webhook-url:  ACAPY_WEBHOOK_URL
--seed:  ACAPY_WALLET_SEED
--wallet-key:  ACAPY_WALLET_KEY
--wallet-rekey:  ACAPY_WALLET_REKEY
--wallet-storage-config:  ACAPY_WALLET_STORAGE_CONFIG
--wallet-storage-creds:  ACAPY_WALLET_STORAGE_CREDS

esune:
> 
> --admin-api-key: ACAPY_ADMIN_API_KEY
> --webhook-url: ACAPY_WEBHOOK_URL
> --seed: ACAPY_WALLET_SEED
> --wallet-key: ACAPY_WALLET_KEY
> --wallet-rekey: ACAPY_WALLET_REKEY
> --wallet-storage-config: ACAPY_WALLET_STORAGE_CONFIG
> --wallet-storage-creds: ACAPY_WALLET_STORAGE_CREDS

Additional environment variables that we will likely need:
`--label`
`--tails-server-base-url`
`--wallet-type`
`--wallet-storage-type`
`--wallet-storage-creds`
`--wallet-storage-config`
`--endpoint`

If it was possible to use an environment variable to drive whether the agent starts in insecure mode or uses an api-key it would be good, but maybe out of scope now (this could also be tied to whether the admin interface is activated or not - I can't think of a use for an agent that has no admin interface).

In general I would think having environment variables for all - or at least as many as possible - of the settings would be a good idea, both for consistency and especially when working with cloud deployments of any kind, where they are the easiest way to deal with configurations.

ianco:
> > --admin-api-key: ACAPY_ADMIN_API_KEY
> > --webhook-url: ACAPY_WEBHOOK_URL
> > --seed: ACAPY_WALLET_SEED
> > --wallet-key: ACAPY_WALLET_KEY
> > --wallet-rekey: ACAPY_WALLET_REKEY
> > --wallet-storage-config: ACAPY_WALLET_STORAGE_CONFIG
> > --wallet-storage-creds: ACAPY_WALLET_STORAGE_CREDS
> 
> Additional settings that we will likely need an environment variables for:
> `--label`
> `--tails-server-base-url`
> `--wallet-type`
> `--wallet-storage-type`
> `--wallet-storage-creds`
> `--wallet-storage-config`
> `--endpoint`
> 
> If it was possible to use an environment variable to drive whether the agent starts in insecure mode or uses an api-key it would be good, but maybe out of scope now (this could also be tied to whether the admin interface is activated or not - I can't think of a use for an agent that has no admin interface).
> 
> In general I would think having environment variables for all - or at least as many as possible - of the settings would be a good idea, both for consistency and especially when working with cloud deployments of any kind, where they are the easiest way to deal with configurations. At the same time, I realize that some of the environment variables (see the api-key/admin interface example above) would potentially force us to build additional logic to check whether other environment variables are set based on their value. This is a bigger job than we maybe willing to do now, but will definitely make the agents more user-friendly and resilient.

Added

requirements.txt
```diff
@@ -11,4 +11,5 @@ prompt_toolkit~=2.0.9
 pynacl~=1.3.0
 requests~=2.23.0
 pyld==2.0.1
-py_multicodec==0.2.1
\ No newline at end of file
+py_multicodec==0.2.1
+git+https://github.com/ianco/ConfigArgParse.git#egg=ConfigArgParse
```
esune, line 15: Should we log an issue to remember updating this dependency, when you PR is merged?

esune (approved):
Looks good! Thanks @ianco

WadeBarnes (approved):
Looks good!

PR #740, opened 2020-10-02
remove leading slash from scripts in path: pwd already has it

sklump:


esune:
Works correctly under Windows/Git Bash.

WadeBarnes (approved):

PR #748, opened 2020-10-08
Convert config files to yaml and force yaml formal for config files

ianco:

Convert config files to yaml format and force yaml format

### DIFF
--- a/demo/local-indy-args.yaml
+++ b/demo/local-indy-args.yaml
@@ -32,6 +32,6 @@ wallet-name: testwallet
 #    docker run --name some-postgres -e POSTGRES_PASSWORD=mysecretpassword -d -p 5432:5432 postgres:10
 wallet-storage-type: postgres_storage
 # could be sent using env var ACAPY_WALLET_STORAGE_CONFIG
-wallet-storage-config: "{\"url\":\"localhost:5432\",\"max_connections\":5}"
+wallet-storage-config: '{"url":"localhost:5432","max_connections":5}'
 # could be sent using env var ACAPY_WALLET_STORAGE_CREDS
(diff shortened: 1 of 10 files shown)


ianco:
Note: works with the released version of ConfigArgParse and yaml format

demo/local-indy-args.yaml
```diff
@@ -1,43 +1,37 @@
...
 # run a local postgres (docker) like:
 #    docker run --name some-postgres -e POSTGRES_PASSWORD=mysecretpassword -d -p 5432:5432 postgres:10
-wallet-storage-type = postgres_storage
+wallet-storage-type: postgres_storage
 # could be sent using env var ACAPY_WALLET_STORAGE_CONFIG
-wallet-storage-config = {"url":"localhost:5432","max_connections":5}
+wallet-storage-config: "{\"url\":\"localhost:5432\",\"max_connections\":5}"
 # could be sent using env var ACAPY_WALLET_STORAGE_CREDS
-wallet-storage-creds = {"account":"postgres","password":"mysecretpassword","admin_account":"postgres","admin_password":"mysecretpassword"}
+wallet-storage-creds: "{\"account\":\"postgres\",\"password\":\"mysecretpassword\",\"admin_account\":\"postgres\",\"admin_password\":\"mysecretpassword\"}"
```
andrewwhitehead, line 37: This could be formatted with single quotes or a flow scalar, or I think we could adjust the wallet code to automatically convert the value to JSON if it receives a dict

ianco: Converted to single quotes

andrewwhitehead (approved):

PR #749, opened 2020-10-08
Updates to the mobile demo for PWD

ianco:


demo/AliceGetsAPhone.md
```diff
@@ -118,25 +120,41 @@ ngrok-tails-server_1  | t=2020-05-13T22:51:14+0000 lvl=info msg="started tunnel"
...
 The `TAILS_NETWORK` parameter lets the demo script know how to connect to the tails server (which should be running in a separate shell on the same machine).
 
-If you are running in _Play with Docker_, navigate to [The demo direcory](/demo) and run:
+#### Running in Play with Docker?
+
+If you are running in _Play with Docker_, navigate to [The demo directory](/demo) and run:
 
 ```bash
-PUBLIC_TAILS_URL=https://def456.ngrok.io LEDGER_URL=http://test.bcovrin.vonx.io ./run_demo faber --revocation --events
+PUBLIC_TAILS_URL=https://c4f7fbb85911.ngrok.io LEDGER_URL=http://test.bcovrin.vonx.io ./run_demo faber --revocation --events
```
swcurran, line 146: So they still have to run ngrok because of the https requirement, right?  I had been thinking that the public tails file would be local.  Can the run ngrok on PWD?

Also -- I assume that they need to replace the ngrok subdomain with the one they are using, right?  I don't see anything to tell them that.  Obvious, but....  :-)

ianco: When you run the tails server docker-compose it starts ngrok automatically, so it's the same running local or on PWD

I gave it a test this afternoon

andrewwhitehead (approved):

PR #750, opened 2020-10-09
Update version to 0.5.5

andrewwhitehead:


swcurran (approved):
Sorry for the delay -- I missed this. Checked and no other changes since, so still good to go.

PR #757, opened 2020-10-19
Update demo README.md

burdettadam:
Update demo readme to suggest the correct flag to enable routing during the performance demo.
Here is a link to the argument configuration, https://github.com/hyperledger/aries-cloudagent-python/blob/master/demo/runners/performance.py#L544


ianco (approved):

swcurran (approved):

swcurran (changes requested):
DCO not used so this change can't be merged.

burdettadam:
I had created this pr with GitHub webpage and overlooked DCO. I will close this one in favor of #761, which includes the needed DCO. Thanks.

PR #765, opened 2020-10-27
allow interval in holder cred rev check via ledger

sklump:


andrewwhitehead:
@sklump It looks like holder/base.py should be updated to match indy.py?

andrewwhitehead (approved):

PR #767, opened 2020-10-28
Enable to set custom web hook in demo

bcersows:
Enhance `agent.py` to read env variable `WEBHOOK_TARGET`, which will be forwarded as a `--webhook-url` to aca-py.
`run_demo` received a new option `--webhook-url`, which (together with env variable `WEBHOOK_URL`) will forward the URL to the agent.

Example usage: `WEBHOOK_URL=host.docker.internal:8888 ./run_demo faber --webhook-url`


andrewwhitehead (approved):
Looks good, thanks!

Issue #736, opened 2020-09-30
Add new RFC, design of "Sign Transaction" protocol -- perhaps Indy specific


swcurran:
A common use case on Indy is the need for a Transaction Author that is not an Endorser to prepare a transaction for the ledger, pass it to an Endorser for a signature and on receipt, execute the transaction.  This is explored in this [HackMD document](https://hackmd.io/5LzMhfsMQBevB5V2tKz4hA?view), in this presentation (https://docs.google.com/presentation/d/1bF_Su5m-kfFxcFL7xVlzPAEo3isnQrkqP4T2s3mMB2Y/edit?usp=sharing) and was discussed in this [Aries WG Call](https://wiki.hyperledger.org/pages/viewpage.action?pageId=39616513) ("Please Sign This" item).

Several questions are open about this:

- Should this be an "Indy Only" protocol that deals specifically with Indy issues?
   - If not Indy specific, perhaps the first cut should be, to deal with this specific case.
- The structure of the current (as I write this) HackMD document is not (IMHO) quite right. It should be closer to the "Issue Credential V2" approach of an attachment and a format/handler in the message for how to handle the attachment.  Suggest reworking the message formats accordingly
   - This approach would allow building an initial Indy-specific way to do this, while remaining open to other options.
- There are sometimes needs to have n of m signatures on a transactions -- e.g. send the message out to 5 (m) entities and once 3 (n) have been returned, send a "nevermind" to the other 2 (m-n) and complete the protocol
   - When handling multi-signatures, specific transactions that are being signed apply the signatures in format-specific ways. For example, Indy has a specific way of handling that.
- The Admin interface for this protocol needs to be defined.
  - How is the process initiated?  How does an Endorser/signer controller get notified of the request and respond?
- The specific transactions that need to be endorsed are handled in ACA-Py today -- ledger transactions to create DIDs, Schema, Cred Defs, RevReg and RevRegEntries.  How do we inject this endorser process into those existing flows?
- How does the agent that needs an endorser know who to send the request to?
   - How does it establish a connection with the Endorser on startup?
   - How does it know the connection to use on restart?
   - Is a startup option defined to notify the agent that it has to do these things?

This task is to come up with a sufficient design for this feature that we can implement.

swcurran:
@ianco -- the idea here is that you work on the design.  You may also get to implement, but we've had others that might be able to do the implementation -- perhaps a dev from @jadhavajay 's team at AyanWorks.

[...]

PR #774, opened 2020-11-04
Updates in Connection Record and Connection APIs for Issue #736

ankita-p17:
Updates in Connection Record and APIs to accomodate changes specific to below issue for Sign Transaction Protcol 

https://github.com/hyperledger/aries-cloudagent-python/issues/736

Reference : 
https://hackmd.io/5LzMhfsMQBevB5V2tKz4hA?view

@swcurran @andrewwhitehead @ianco please check

### DIFF
--- a/aries_cloudagent/connections/models/connection_record.py
+++ b/aries_cloudagent/connections/models/connection_record.py
@@ -65,7 +65,9 @@ class ConnectionRecord(BaseRecord):  # lgtm[py/missing-equals]
         connection_id: str = None,
         my_did: str = None,
         their_did: str = None,
-        their_label: str = None,
+        their_label: str = None,   
+        tx_my_role:list = [],
+        tx_their_role:list = [],
         their_role: str = None,
         initiator: str = None,
         invitation_key: str = None,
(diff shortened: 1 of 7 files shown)


ianco:
Hi @ankita-p17 please make sure to DCO sign your commits, see:  https://github.com/hyperledger/aries-cloudagent-python/pull/774/checks?check_run_id=1351780206

ankita-p17:
> Hi @ankita-p17 please make sure to DCO sign your commits, see: https://github.com/hyperledger/aries-cloudagent-python/pull/774/checks?check_run_id=1351780206
> 
> (Also make sure to double-check unit tests are passing, including the code formatting)

Done. can you check the changes made

ianco:
Hi @ankita-p17 the DCO check is still failing.  In the past when this has happened to me sometimes the fix they recommend doesn't work (you can see the commits in the commit history still have a red "x" next to them).  Sometimes I've had to create a new branch and move my code over to get a clean commit history.

Make sure you always `--signoff` on every commit.  I think you can setup your local git configuration so that this is the default.

andrewwhitehead:
I don't think we want to do this by extending the Connection record with information that isn't specific to the connection. We would need to add a new stored record to capture it.

ankita-p17:
> I don't think we want to do this by extending the Connection record with information that isn't specific to the connection. We would need to add a new stored record to capture it.

Non-indy roles should be captured as another record? So for every connection, there will be different records maintained if user opt for role transaction Author, right?

sklump:
A connection record is meant to represent an RFC 160 (or, soon, RFC 23) connection between agents.

ankita-p17:
> A connection record is meant to represent an RFC 160 (or, soon, RFC 23) connection between agents.

Got it

ankita-p17:
Duplicate of #799 which has more functionalities for the protocol

Issue #742, opened 2020-10-02
Add test to automate the detection/documentation of Admin API changes


swcurran:
@jcourt562 mentioned that he is detecting changes in the Admin API by using the swagger definition to generate a client for the API and then looking for changes from previous generations.  ACA-Py has been updated to make the generated swagger definition predictably ordered to make such comparisons easier.

It would be nice to have such a mechanism added to the repo for detecting expected and unexpected changes to the API.  Have a way to run the process, generate the client and then diff the result with the last known good version of the client.  Ideally, the test would be in the CI/CD pipeline (forcing developer to account for and document their changes), but it could also be just a manual process we run before each release to make sure that the release notes sufficiently document the API changes.

By having the expected result in the repo, we can also provide a URL for a GitHub tag-to-tag comparison of the generated client  code.

swcurran:
@jcourt562 -- interested in helping with this one?  :-)

jcourt562:
I would need to check with my CEO, however its likely we could be interested in helping here with some giveback.

Cheers

John



________________________________

jcourt562:
Let me look into this and see what I can do.

Cheers

John

jcourt562:
Hey Stephen, you can assign this one to me. I have a script (yeah ok its only bash but hey), it runs up the ACA-py docker, and then runs the swagger-cogegen docker image against it to create an openapi.json file, so its not language specific.  We should be able to use that with maybe a little bit of change from what we do afterwards for specific language generation.  One issue is that doing this in the CI/CD only really works (at least we have it working that way) with the "shell" executor since it needs to share the fs for input output.  I might submit it as a manual script for now until I understand your CI/CD a little better.

swcurran:
That works.  Now that we have merged all of the configuration systems for ACA-Py into a unified system (#739 ) we should also have a "usage" run as part of this that we also check for changes on a merge by merge basis.  Don't feel obligated to do that as well, but I assume we'll piggyback on what you do.

@WadeBarnes can help with any CI/CD questions, and can take what you produce to move into CI/CD.

Thanks!

sklump:
bash is way underrated as a scripting language. Minimal external dependencies!

PR #777, opened 2020-11-09
WIP : Script and first run output to show when Open API changes

jcourt562:
Signed-off-by: john court <jcourt@anonyome.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,3 +186,6 @@ $RECYCLE.BIN/
 # Docs build
 _build/
 **/*.iml
+
+# Open API build
+open-api/.build
--- a/open-api/openAPIJSON.config
+++ b/open-api/openAPIJSON.config
@@ -1,0 +1,3 @@
+{
+  "validateSpec" : "false"
+} 
--- a/open-api/openapi.json
+++ b/open-api/openapi.json
@@ -1,0 +1,4569 @@
+{
+  "swagger" : "2.0",
+  "info" : {
+    "version" : "v0.5.6",
+    "title" : "Aries Cloud Agent"
+  },
+  "tags" : [ {
+    "name" : "action-menu",
+    "description" : "Menu interaction over connection"
+  }, {
+    "name" : "basicmessage",
+    "description" : "Simple messaging",
+    "externalDocs" : {
+      "description" : "Specification",
+      "url" : "https://github.com/hyperledger/aries-rfcs/tree/527849ec3aa2a8fd47a7bb6c57f918ff8bcb5e8c/features/0095-basic-message"
+    }
+  }, {
+    "name" : "connection",
+    "description" : "Connection management",
+    "externalDocs" : {
+      "description" : "Specification",
+      "url" : "https://github.com/hyperledger/aries-rfcs/tree/9b0aaa39df7e8bd434126c4b33c097aae78d65bf/features/0160-connection-protocol"
+    }
+  }, {
+    "name" : "credential-definition",
+    "description" : "Credential definition operations",
+    "externalDocs" : {
+      "description" : "Specification",
+      "url" : "https://github.com/hyperledger/indy-node/blob/master/design/anoncreds.md#cred_def"
+    }
+  }, {
+    "name" : "credentials",
+    "description" : "Holder credential management",
+    "externalDocs" : {
+      "description" : "Overview",
+      "url" : "https://w3c.github.io/vc-data-model/#credentials"
+    }
+  }, {
+    "name" : "introduction",
+    "description" : "Introduction of known parties"
+  }, {
+    "name" : "issue-credential",
+    "description" : "Credential issue, revocation",
+    "externalDocs" : {
+      "description" : "Specification",
+      "url" : "https://github.com/hyperledger/aries-rfcs/tree/bb42a6c35e0d5543718fb36dd099551ab192f7b0/features/0036-issue-credential"
+    }
+  }, {
+    "name" : "ledger",
+    "description" : "Interaction with ledger",
+    "externalDocs" : {
+      "description" : "Overview",
+      "url" : "https://hyperledger-indy.readthedocs.io/projects/plenum/en/latest/storage.html#ledger"
+    }
+  }, {
+    "name" : "out-of-band",
+    "description" : "Out-of-band connections",
+    "externalDocs" : {
+      "description" : "Design",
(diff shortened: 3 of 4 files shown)


jcourt562:
This pull request provides a script to generate the openapi.json spec file from a running ACA-py.  The openapi.json can be checked in and used to compare in subsequent releases for unexpected changes.   Apart from explicit changes to the REST API the only changes should be the random/timestamp related "example" fields in the spec file.  This submit is in response to issue #742

WadeBarnes (commented):
I'd recommend refactoring `waitActiveWebInterface` to function along the lines of https://github.com/hyperledger/aries-agent-test-harness/blob/master/manage#L279.  That way you get a status indication in cases where the container takes a bit of time to spin up.

ianco:
How does this sort the output openapi.json file?

I tried running the script and the file I generated was in a completely different order than the checked-in file (so there were a lot of spurious differences between the files).

jcourt562:
It doesn't sort the output at all, or shouldn't.  The intent is that the output coming from ACA-py is already sorted.  This was the intent of the script to detect changes.  I have run it many times but maybe there is something specific to my installation.  Can I ask what version of ACA-py you are running it against ?  as prior to V0.5.6 you may well see issues since this is why the script was created.

ianco:
I have the latest master branch checked out, and I merged the code from this PR.

swcurran:
@ianco -- the idea for this came about because of @sklump 's change to make the openapi.json file come out sorted already. If that is not working, we need to look at where it is coming from.  Perhaps that feature was lost somewhere?

ianco:
@swcurran  what's the PR?  I'll check to make sure it's included in my local ...

swcurran:
Looks like it is #738.

ianco:
ok that's definitely included in my local

jcourt562:
Can you post the output you are seeing ?   I may be able to tell what is happening based on what the differences are and having seen the wrong behaviour from ACA-py previously.   So the only other possibility is that the swagger codegen for some reason outputs different openapi.json randomly from the same raw swagger.json from ACA-py.  I haven't seen this to date and in particular there are options to swagger to artificially sort which we intentionally aren't using because that would cause other problems around detecting ACA-py changes.

jcourt562:
So I just pulled up to the latest master and re-ran, it behaved as expected but one thing I realised is that if you are already running an ACA-py on your local machine, its possible the script will fail to start its own and instead end up pulling the swagger file from the pre-existing instance.  @ianco is this a possibility with your environment ?  I would still like to see your openapi.json but will make sure the script exits if the ACA-py start phase fails.  I will also change the admin port to be a non-standard one everyone seems to use so that there is less chance of connection to an existing instance by mistake.

ianco:
I made sure everything was shut down and re-ran the script, the output was the same.  (It's possible another aca-py was running when I tested this earlier, I'm not sure.)

Here's my output:

[openapi.json.zip](https://github.com/hyperledger/aries-cloudagent-python/files/5531070/openapi.json.zip)

ianco:
FYI the api list on the swagger page is now sorted (according to the PR that @swcurran referenced) however the generated opeapi json file is not.  It looks like the file you generate (checked into your PR) is different from mine, and neither are sorted the same as the swagger page.

jcourt562:
Yep, it looks like there is an issue with swagger not maintaining the order of paths from the raw input.  The actual issues I had filed previously were around parameter order and object model order since they caused me wrapper code breakages.  I don't understand why it has consistently created the same output for me so I didn't see the issue (in fact I didn't even realise ACA-py was also sorting paths).  It could be as simple as different CPU speeds or timezone affecting internal ordering.    The sad thing is that it doesn't look like either OpenAPI or Swagger code generators have really addressed this issue so I might have to look at a different approach to formatting the "raw" interface file from ACA-py Into something more diff friendly. Will investigate.

jcourt562:
@ianco I just need to challenge your assertion that the raw swagger output from ACA-py is sorting the paths ?  I took the raw output and feed it through an online JSON pretty printer and the result doesn't show paths sorted in any meaningful way ?
I went through the painful process of checking the first few hadn't changed order between raw and the pretty format.  Attaching pretty format file.
[jsonformatter.txt](https://github.com/hyperledger/aries-cloudagent-python/files/5533640/jsonformatter.txt)

jcourt562:
I have also just gone through #738.  That seems to be sorting swagger_dict["definitions"]  I don't see anywhere in the code where swagger_dict["paths"] gets sorted, I do see parameters within paths being sorted.

ianco:
I haven't looked into the underlying files, so possibly you're right.

In the displayed swagger page (i.e. in the browser) everything is displayed sorted by message type and then path, I don't know where all this sorting is taking place.

Screen Shot 2020-11-12 at 3 38 33 PM

jcourt562:
I suspect it must be happening in the web server part of the code not in the core.

ianco:
> I have also just gone through #738. That seems to be sorting swagger_dict["definitions"] I don't see anywhere in the code where swagger_dict["paths"] gets sorted, I do see parameters within paths being sorted.

It sorts recursively, "asciibetically", so paths will get sorted within definitions etc.

https://github.com/hyperledger/aries-cloudagent-python/pull/738/files#diff-f3648d939ba43f8f8b8a42f8786f921dd598def57267dd9e3904a3a9ff391d6dR348

jcourt562:
@ianco how do you come to that conclusion ? They are seperate dictionaries aren't they ?  I assume definitions are the model definitions : 

       # order tags alphabetically, parameters deterministically and pythonically
        swagger_dict = self.app._state["swagger_dict"]
        swagger_dict.get("tags", []).sort(key=lambda t: t["name"])
        for path in swagger_dict["paths"].values():
            for method_spec in path.values():
                method_spec["parameters"].sort(
                    key=lambda p: (p["in"], not p["required"], p["name"])
                )

        # order definitions alphabetically by dict key
        swagger_dict["definitions"] = sort_dict(swagger_dict["definitions"])

sklump:
I can also find some time to look at this today

sklump:
Do you need me to sort paths too? I see that server startup only sorts the values therein explicitly.

jcourt562:
@sklump yes I think that is why @ianco is seeing different output when he runs it to me.  I really think this must be due to simple memory, cpu differences causing order differences on "paths" as I run it with different ACA-py versions and don't see any unexpected changes.

Thanks !

sklump:
See https://github.com/hyperledger/aries-cloudagent-python/pull/785

PR #778, opened 2020-11-09
feature/Connection Invitation with routingKeys

burdettadam:
This pr contains extended admin API to allow connection invitations to be created with routing and recipient keys as well as service endpoints. This functionality is needed for establishing connections using a mediator. 

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/routes.py
+++ b/aries_cloudagent/protocols/connections/v1_0/routes.py
@@ -353,9 +353,13 @@ async def connections_create_invitation(request: web.BaseRequest):
     connection_mgr = ConnectionManager(context)
     try:
         (connection, invitation) = await connection_mgr.create_invitation(
-            auto_accept=auto_accept, public=public, multi_use=multi_use,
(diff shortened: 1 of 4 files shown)


sklump (approved):

aries_cloudagent/protocols/connections/v1_0/routes.py
```diff
@@ -318,6 +339,10 @@ async def connections_create_invitation(request: web.BaseRequest):
     alias = request.query.get("alias")
     public = json.loads(request.query.get("public", "false"))
     multi_use = json.loads(request.query.get("multi_use", "false"))
+    body = await request.json()
```
andrewwhitehead, line 342: It doesn't look like the method currently requires a JSON message body, so will this blow up if current controllers aren't adjusted? It could be modified to only parse the request body as JSON if one is provided.

sklump: It's a POST: it has a body, necessarily

sklump: Ha ha! It turns out you were right: POST doesn't need a body. All these years I never knew! I will fix it today.

PR #785, opened 2020-11-16
Openapi json order paths

sklump:
Updates PR #777


ianco:
OK I do this:

```
$ ./scripts/generate-open-api-spec
```

... and then this:

```
$ git status
```

... and I get this:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   open-api/openapi.json

no changes added to commit (use "git add" and/or "git commit -a")
```

Am I missing something?  The `openapi.json` file is supposed to be generated in a consistent format/order, but the version generated on my local is very different from the one checked into github.

andrewwhitehead:
@ianco Can you upload the generated version somewhere

ianco:
Generated openapi.json file:

[openapi.json.txt](https://github.com/hyperledger/aries-cloudagent-python/files/5592027/openapi.json.txt)

sklump:
Isn't this odd:
* openapi.json.txt top-level order: swagger, info, tags, paths, definitions
* start aca-py, go to /api/docs/swagger.json, click on "raw data" to get (admin-server-generated) order: paths, info, swagger, definitions, tags.

Puzzling. Does the openapi.json.txt copy always come out in the same order? The admin-server-generated one does.

ianco:
@sklump yes seems to be consistent.

I think the openapi.json file checked into the repo is *wrong*, i.e. the one I generate seems to be sorted properly (e.g. path entries are sorted within the top-level paths).  Therefore I'm going to change my review to "approved" and we can merge this puppy and move on.

ianco (approved):

sklump:
Hold on a moment, I've made the one from server.py come out to match the one that the script generates.

sklump:
I'm just running the unit test suite in case of flake or black items. It'll be in place within 15 or so.

sklump:
I'm just going to merge master and see if it breaks anything - I don't think it does

jcourt562:
So just to close on that loop.  The reason the checked in version of openapi.json was different is because it was the version before @sklump put the fix in to sort the paths.  Swagger always outputs the sections in the same order of swagger, info, tags, paths, definitions.

PR #796, opened 2020-11-20
multitenancy admin api

TimoGlastra:
This adds the initial multitenancy admin api (+ some more). It is based on a combination of the current multitenancy branch and the alternate implementation by nick.

It includes:
- `--multitenant` parameter to enable handling of subwallets (checks for `Authorization` header)
- `--multitenant-admin` parameter to enable the `/multitenancy` plugin with routes
- `--jwt-secret` parameter which expects a secret to be used for jwt signing / verifying
- create, get list, delete wallet endpoints. Code is structured to support `unmanaged` wallets, but currently it only works with `managed` wallets
- Separate context per  request, this was global first (set on `app` instead of `request`). I've also updated all `test_routes.py` to use the request context instead of the app context. This is why the PR became so large, but this is also why a lot of tests are failing in the current multitenancy branch.

I will add tests for the admin routes later, as the routes still need a bit of ironing out and the PR is already quite large. There are no restrictions yet (subwallets can call multitenancy admin api, base wallet can call normal operations).

---
To test it out start an ACA-Py instance with the 3 new parameters (`--multitenant --multitenant-admin --jwt-secret very_secret_secret`) and call `POST /multitenancy/wallet`. You'll get back a json object with `token`. If you now set that token as header (`Authorization: Bearer <token>`) all data will be stored in the corresponding subwallet (calling `create-invitation` without Authorization header and then calling get `connections`  with authorization header will return no connections for example. Removing the header will) .


TimoGlastra:
FYI @ianco

aries_cloudagent/admin/server.py
```diff
@@ -249,6 +251,20 @@ def is_unprotected_path(path: str):
...
+
+            # This sets the context and message_router on the request
+            # This allows for different context per request, which
+            # is needed for multitenancy.
+            request["context"] = context
+            request["outbound_message_router"] = self.responder.send
+
+            return await handler(request)
+
+        middlewares.append(set_request_context)
```
TimoGlastra, line 266: Even if multitenancy is not enabled we now set the conext on the request, this is to avoid endless if/else statements.

aries_cloudagent/admin/server.py
```diff
@@ -285,7 +301,58 @@ async def collect_stats(request, handler):
...
+                except Exception:
+                    raise web.HTTPUnauthorized()
+
+                # set the updated request context
+                request["context"] = context
+
+                return await handler(request)
+
+            middlewares.append(set_multitenant_wallet)
+
```
TimoGlastra, line 350: If multitenancy is enabled, we check for the `Authorization` header. If it is available we check for validity and update the request context settings. Whenever the wallet is injected, the subwallets config will be used from now on

aries_cloudagent/holder/routes.py
```diff
@@ -118,7 +118,7 @@ async def credentials_get(request: web.BaseRequest):
         The credential response
 
     """
-    context = request.app["request_context"]
+    context = request["context"]
```
TimoGlastra, line 121: all requests now use context from the request object. This allows each admin request to have a different context

aries_cloudagent/messaging/credential_definitions/tests/test_routes.py
```diff
@@ -44,13 +41,13 @@ def setUp(self):
...
-        self.app = {
-            "request_context": self.context,
+        self.request_dict = {
+            "context": self.context,
         }
 
     async def test_send_credential_definition(self):
         mock_request = async_mock.MagicMock(
-            app=self.app,
+            __getitem__=async_mock.Mock(side_effect=self.request_dict.__getitem__),
```
TimoGlastra, line 50: I couldn't find a better approach to `__getitem__` to set dict values on the `Request` object. Setting them with `request["context"]=self.context` doesn't work and setting them with `context=self.context` also doesn't work. MagicMock doesn't play well with items I think.

aries_cloudagent/wallet/models/wallet_record.py
```diff
@@ -0,0 +1,105 @@
...
+
+from ...messaging.models.base_record import (
+    BaseRecord,
+    BaseRecordSchema,
+)
+from ...messaging.valid import UUIDFour
+from ..base import BaseWallet
+
+
+class WalletRecord(BaseRecord):
```
TimoGlastra, line 17: `WalletRecord` stores all subwallet data

requirements.txt
```diff
@@ -14,3 +14,4 @@ pyld==2.0.1
 py_multicodec==0.2.1
 pyyaml~=5.3.1
 ConfigArgParse~=1.2.3
+pyjwt~=1.7.1
```
TimoGlastra, line 17: `pyjwt` dependency added to sign/verify JWTs

ianco (approved):

PR #798, opened 2020-11-23
Agency multitenant - add multitenancy support to alice/faber

ianco:
Add support for multitenance alice/faber demo


ianco:
@TimoGlastra this is not yet working, not sure why.

Faber fails when trying to promote a wallet did to a public did.

Alice fails when trying to accept a connection.

Possibly logic errors in the alice/faber logic, but not sure so posting a draft PR with my changes

TimoGlastra:
@ianco I'll take a look regarding the creating the wallet + did and then promoting it. That should work.

The alice part fails because I haven't implemented inbound handling of multitenant messages. The base wallet is still used for this.

I'm gonna add very simple relay functionality for now, and can later integrate with the mediator functionality.

PR #800, opened 2020-11-24
Simplify storage record search API

andrewwhitehead:
- Removes unused methods, simplifies API around storage record search
- Avoid having a constructor in an interface class (BaseStorageRecordSearch)


andrewwhitehead:
Adding a couple more adjustments here

sklump (approved):
Nothing jumps out at me - I am surprised there is so much dead code

PR #802, opened 2020-11-24
results template for present-proof list-creds op by pres req

sklump:


andrewwhitehead (approved):

jcourt562:
So the code gen still gets empty definitions for the types of IndyCredPrecisCredInfo and IndyCredPrecisInterval ? on the : `GET /present-proof​/records​/{pres_ex_id}​/credentials` response ?

In the output openapi.json these fields are IndyCredPrecis_cred_info and IndyCredPrecis_interval respectively and just have a type of "object" with no "properties" defined.  I am assuming there should be "properties" to define the Indy structure of credential information and intervals.

jcourt562:
Please disregard my previous comment.  I was running a stale version of ACA-py docker image due to a mis-config of my AWS repo.  I can see all the related definitions for this method now.

PR #807, opened 2020-11-30
Different upload-tail-file path

muisit:
Added support for having a different upload path than download path for the public tails file. This supports the case where the upload path is a restricted, internal route to a central file distribution server and the download path is a public interface for a static file download. In our case using dockers, the upload path is an internal docker-resolved path and the public path is the controller application that links to the cloud-agent application.

### DIFF
--- a/aries_cloudagent/revocation/models/issuer_rev_reg_record.py
+++ b/aries_cloudagent/revocation/models/issuer_rev_reg_record.py
@@ -149,7 +149,7 @@ class IssuerRevRegRecord(BaseRecord):
         issuer: BaseIssuer = await context.inject(BaseIssuer)
         tails_hopper_dir = indy_client_dir(join("tails", ".hopper"), create=True)
 
-        LOGGER.debug("create revocation registry with size:", self.max_cred_num)
+        LOGGER.debug("create revocation registry with size:" + str(self.max_cred_num))
 
         try:
             (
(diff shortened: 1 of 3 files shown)


aries_cloudagent/revocation/models/issuer_rev_reg_record.py
```diff
@@ -149,7 +149,7 @@ async def generate_registry(self, context: InjectionContext):
         issuer: BaseIssuer = await context.inject(BaseIssuer)
         tails_hopper_dir = indy_client_dir(join("tails", ".hopper"), create=True)
 
-        LOGGER.debug("create revocation registry with size:", self.max_cred_num)
+        LOGGER.debug("create revocation registry with size:" + str(self.max_cred_num))
```
sklump, line 152: `("create revocation registry with size: %d", self.max_cred_num)` matches the rest of the code base: Using a `+` here matches none of the code base in any case: python string concatenation works fine with `"a " "b"` and looks more natural when you have to break a line to satisfy flake; e.g.,
```
(
    "hello "
    "world"
    "again"
)
```

muisit: Yeah, you are undoubtedly right. I noticed an error in my console that this 'self.max_cred_num' value could not be stringified, so I added this old-school concoction to get rid of it. It ended up in this PR, but it shouldn't have been included I guess.

sklump: It's not the end of the world, just noting that there is a convention here. Good catch in any case.

sklump:
Follow in https://github.com/hyperledger/aries-cloudagent-python/pull/829

PR #808, opened 2020-11-30
Double REG_DEF/REG_ENTRY submissions

muisit:
Removed sending and submitting rev_reg definition after calling registry_record.generate_registry. This duplicates activity also called inside generate_registry and creates additional ledger transactions. 
Also removed the pending_registry_record, which seems to duplicate the already existing registry_record seemingly without adding any value.

### DIFF
--- a/aries_cloudagent/messaging/credential_definitions/routes.py
+++ b/aries_cloudagent/messaging/credential_definitions/routes.py
@@ -175,17 +175,6 @@ async def credential_definitions_send_credential_definition(request: web.BaseReq
             await registry_record.set_tails_file_public_uri(
                 context, f"{tails_base_url}/{registry_record.revoc_reg_id}"
             )
-            await registry_record.send_def(context)
-            await registry_record.send_entry(context)
-
-            # stage pending registry independent of whether tails server is OK
-            pending_registry_record = await revoc.init_issuer_registry(
-                registry_record.cred_def_id,
-                max_cred_num=registry_record.max_cred_num,
-            )
-            ensure_future(
-                pending_registry_record.stage_pending_registry(context, max_attempts=16)
-            )
 
             tails_server: BaseTailsServer = await context.inject(BaseTailsServer)
             (upload_success, reason) = await tails_server.upload_tails_file(


muisit:
Due to the lack of documentation with the code, I'm unsure about the actual implications: what was the reason for creating a new pending_registry_record based solely on the already created registry_record? Why were manual send_def and send_entry calls added at that spot?

I see that nrempel introduced the additional sending of def and entry on july 21 in f98327704c15dcbe78248f55cdf6cab423844eb0. And then subsequently introduced the pending_registry_record. But the commit comments are very generic, so I probably don't understand the reasoning.

muisit:
Apparently this is not doing what I first saw, closing this until I can properly fix my issue locally.

swcurran:
Hi @muisit -- not exactly sure what you are seeing, but I suspect it is the initial creation of two rev-regs. The approach we are taking to rev-regs is that we always have one more ready than we are using.  We start with two, fill up the first, switch to the second and create a third.  That way, there is no delay during an issuance for creating a new reg -- we create the "next one" in the background.

muisit:
I think you're probably spot-on, thanks. I must say this is a strange feature to implement here. The controller application can also decide by itself to create a back-up rev-reg after the initial creation. If time is of the essence at that point already, the initial rev-reg should have been larger in any case. 
Plus you'll run into the same problem once the pending rev-reg is activated. The controller app will need to know how to create a new rev-reg from a cred-def in any case.
But it matches the rest of the framework, that seems to supply ready-to-use hooks for specific application-types and less of generic-hooks for generic applications.

sklump:
Do NOT touch this code.

All the manual stuff is for administrative emergency fixes if and when the network goes out at a bad time and a super-user has to clean up the mess to get it into a workable state.

swcurran:
@muisit -- here is the background.  On the first cut of revocation, we left all of the handling to the controller -- creating the rev reg, managing the tails file, etc.  When we realized how problematic that was, we implemented the (almost) fully automated approach -- tell us you want revocation, how big to make each registry and when you want to revoke credentials, and the framework will do the rest.  Easy!

However, when we did that, we decided to leave in the original APIs so the controller COULD do the handling if necessary -- and in particular, a person running an instance could use the API to "clean up" a RevReg if it got into a bad state.  That's pretty unlikely now -- and with new capabilities we'll eliminate, but out of an abundance of caution, we decided to keep the original API "just in case".

Hope that helps.

bartekus:
This seems like a quite an unwanted functionality that perhaps should have some sort of a feature flag to disable it.

From an end user perspective, I cannot fathom the benefit of having a duplicate `REVOC_REG_ENTRY` and `REVOC_REG_DEF`, nor do I see a clear way to explain this as an intended behaviour to users that raise this as an glaring issue which they can clearly [observe](https://indyscan.io/txs/SOVRIN_BUILDERNET/domain?page=1&pageSize=50&filterTxNames=[]&sortFromRecent=true&search=XGpSYZvFaHHPT5p1LhHeRV) 

What recourse is there to alleviate this bug of a feature besides a fork-and-patch, it doesn't seem there is any or am I missing something here?

swcurran:
This is a very valuable feature and should be automatic.  We know this because we tried to use it without and it's horrible for the controller.  Here is the background:

When you create a revocable cred def, you have to create the rev registry of a give size as well -- separate transactions on the ledger.  You then issue credentials for awhile.  Eventually, you run our of space in your first rev registry, and you have to create a new rev registry before you can issue your next credential.  When we first added revocation support in ACA-Py, we just exposed the capabilities, and left it to the controller to handle all of the details.  It was horrible for the controller writer.

The first benefit of what is now in ACA-Py is that the controller does not have to track the size of the rev reg, how many credentials have been issued and execute the revreg transactions to the ledger.  The controller just says to ACA-Py -- "create a cred def with a rev reg of size X", and ACA-Py takes care of all of the details.

The benefit of pre-creating the second rev reg at start time is that there is no delay when the first rev reg runs out of space.  The controller issuing credentials doesn't even know the rev reg was switched, and there is no delay.  ACA-Py detects the filled rev reg, switches to the pre-created but unused rev reg, issues the credential and the controller can continue.  Then, in the background, ACA-Py creates another RevReg, so it is ready when the the second one runs out of credentials.  And so on.

The controller only has to track the ID if the issued credentials (a combination of RevReg ID + Credential Index). When necessary, the controller passes the ID in to have the credential revoked.  The controller also tells ACA-Py when to publish revocations -- when to write "Rev Entries" to the ledger.  ACA-Py tracks all of the RevRegs for which unpublished revocations exist, and writes the necessary transactions to the ledger.  Again -- the controller doesn't have to track all of that detail.

Hope that makes sense, and makes it easier to explain.  Let me know if you have further questions.  If this is not covered in a doc, we need to add it.

PR #811, opened 2020-12-01
ptvsd debugger for pytest

burdettadam:
This PR allows Visual Studio Code ptvsd to attach to the docker running pytest from the run_tests script. The `--debug` flag can now be used with run_tests, for example.
```bash
./scripts/run_tests --debug
```
The flag and any other arguments are then passed into the docker container directly to pytest, allowing calls like.
```bash
./scripts/run_tests --debug aries_cloudagent/protocols/connections
```

### DIFF
--- a/DevReadMe.md
+++ b/DevReadMe.md
@@ -143,6 +143,18 @@ To run the ACA-Py test suite, use the following script:
 ./scripts/run_tests
 ```
 
+To run the ACA-Py test suite with ptvsd debugger enabled:
+
+```bash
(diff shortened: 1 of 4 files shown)


andrewwhitehead:
This seems to be overloading the --debug flag, it is still passed to pytest which causes it to write a debug log. Not that it's necessarily a bad thing.

burdettadam:
I agree. I decided this to be a feature. I can change it if needed. I am open to any direction.

andrewwhitehead (approved):

PR #812, opened 2020-12-02
Add Profile, ProfileSession abstractions

andrewwhitehead:
This PR adds a `Profile` base class which represents an identity and service provider. This and the `ProfileManager` base class replace the wallet management methods of `BaseWallet`, which is reduced to only key and DID management operations. A profile has an `InjectionContext` and can inject certain base classes, but in order to inject implementations of `BaseLedger`, `BaseStorage`, or `BaseWallet` for instance, a `ProfileSession` must first be created using the `session` or `transaction` methods. This is to help enable support for the Askar backend.

The `RequestContext` passed to message handlers no longer inherits from `InjectionContext`, but has a `profile` property and convenience methods for creating a session or injecting base class implementations. Handler tests are updated to use the new `test_context` method.

Similarly, an `AdminRequestContext` class is added to provide a similar interface for admin route handlers. This is moved from `request.app["request_context"]` to `request["context"]` allowing the profile and context to be adjusted on a per-request basis (which would be relevant for multi-tenancy).

Other changes:

- `BasicStorage` and `BasicWallet` are renamed to `InMemoryStorage` and `InMemoryWallet`.
- Indy-SDK support code is moved into the `indy` package and renamed (`Base -> Indy`, `Indy -> IndySdk`).
- `Injector.inject` and `Provider.provide` are no longer async methods.
- The wallet provider, storage provider, and ledger provider classes are removed (functionality moved to profile managers).
- It is no longer possible to select the `BaseStorage` implementation independent of the wallet type.
- The `BaseLedger` implementation is managed directly by the profile implementation (`IndySdkProfile` for now) - base class to be renamed to `IndyLedger`.
- The provision command should fail if an existing wallet is found, unless `--recreate-wallet` is passed.
- The start command will not create a new wallet unless `--auto-provision` is passed.
- The `auto-remove` property of a wallet previously served double duty, to remove an existing wallet on open and to remove the current wallet on close. An `auto_recreate` property is added to determine the behaviour on open.


swcurran:
@andrewwhitehead --- I notice this is going into Master.  Do we need a notification to others in the world on this?  Does this affect external protocols for example?  Or is this all internal and hence all good.

@ianco -- I assume this may impact your multi-tenancy, but hopefully in a good way. :-). 

@TimoGlastra FYI

andrewwhitehead:
@swcurran  I don't believe it affects any external protocols, but plugins would need to be updated (like issuer registration) and it would be in a 0.6 release. Deployments may need to add `--auto-provision` if they are starting an agent without provisioning it separately.

ianco:
I did a quick test and there are a large number of conflicts with the multitenancy code (about 40 files).

However a lot of the conflicts seem to fall into the pattern replacing:

context = request["context"]

with:

TimoGlastra:
That's quite some changes, it's hard to keep up. 

But looking at the code it seems like it's gonna be a good thing for multi-tenancy :)

@burdettadam This is probably also of your interest

ianco (approved):

ianco:
Didn't do a detailed review, but ran tests & alice/faber demo and everything looked ok (including revocation).

Interested in @TimoGlastra 's review once he tries to merge into the multitenancy branch ...

aries_cloudagent/config/tests/test_wallet.py
```diff
@@ -1,25 +1,58 @@
 from asynctest import TestCase as AsyncTestCase, mock as async_mock
 
+from ...config.injector import Injector
```
sklump, line 3: Quibble: already in config: don't need to go up and down into it again.

sklump (approved):

andrewwhitehead:
I've also run the `performance` demo on both master and this branch to make sure there was no negative impact from the session abstraction, it looks to be the same.

victormartinez-work:
> But looking at the code it seems like it's gonna be a good thing for multi-tenancy :)

Can you elaborate on that @TimoGlastra  @ianco  ? :) Thanks!

TimoGlastra:
> Can you elaborate on that @TimoGlastra @ianco ? :) Thanks!

I haven't looked at in detail, but It contains some abstractions that make it easier to handle multiple contexts / sessions I believe. That's exactly what we need for multi-tenancy

ianco (approved):

ianco:
Looks good to me.

PR #820, opened 2020-12-07
Profiles follow-up

andrewwhitehead:
Update CredentialManager, InboundTransportManager, InboundSession to accept a Profile instance. Route outbound messages with a Profile rather than a context.

General refactoring: combine InjectorError and ProviderError into InjectionError.


sklump (approved):
enormous effort

PR #824, opened 2020-12-07
Update routes and tests for action menu, basicmessage, trustping

andrewwhitehead:


ianco (approved):
Ran AATH and everything passed

PR #825, opened 2020-12-07
Add __main__ to package

andrewwhitehead:
This makes `python -m aries_cloudagent` equivalent to running `aca-py`.

### DIFF
--- a/aries_cloudagent/__main__.py
+++ b/aries_cloudagent/__main__.py
@@ -4,57 +4,67 @@ import os
 import sys
 
 
-ENABLE_PTVSD = os.getenv("ENABLE_PTVSD", "").lower()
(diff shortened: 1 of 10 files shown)


andrewwhitehead:
@ianco I see this would also cause issues for the test harness because it assumes that the `aca-py` script is a Python script. I think it's more compatible not to try hard-coding the python path in the first line of the script and assume the existence of `sh` instead, though. I think this could be updated to just return `list(flatten(([cmd_path, "start"], self.get_agent_args())))` to work with either version in the docker environment though: https://github.com/hyperledger/aries-agent-test-harness/blob/dc679d144d6fe1be0b4cadec57da89b4abb0bd39/aries-backchannels/acapy/acapy_backchannel.py#L668

sklump (approved):

PR #830, opened 2020-12-08
invitation url in invitation record allows QR code inclusion

sklump:

### DIFF
--- a/aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
+++ b/aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
@@ -90,11 +90,11 @@ class InvitationMessage(AgentMessage):
 
         """
         c_json = self.to_json()
-        c_i = bytes_to_b64(c_json.encode("ascii"), urlsafe=True)
+        oob = bytes_to_b64(c_json.encode("ascii"), urlsafe=True)
         result = urljoin(
             base_url
             or (self.service_blocks[0].service_endpoint if self.service_blocks else ""),
-            "?c_i={}".format(c_i),
+            "?oob={}".format(oob),
         )
         return result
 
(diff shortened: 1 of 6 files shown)


aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
```diff
@@ -91,7 +91,11 @@ def to_url(self, base_url: str = None) -> str:
         """
         c_json = self.to_json()
         c_i = bytes_to_b64(c_json.encode("ascii"), urlsafe=True)
-        result = urljoin(base_url or "", "?c_i={}".format(c_i))
+        result = urljoin(
+            base_url
+            or (self.service_blocks[0].service_endpoint if self.service_blocks else ""),
+            "?c_i={}".format(c_i),
```
andrewwhitehead, line 97: In the OutofBand RFC it's using `?oob=` instead of `?c_i=`, I think everything else is consistent though

ianco (dismissed):
Tested with Trisnic and esatus and it crashes both apps

Test:
- on PWD http://play-with-von.vonx.io/
- clone repo
- run faber: LEDGER_URL=http://test.bcovrin.vonx.io ./run_demo faber
- scan QR code with mobile app

andrewwhitehead:
@ianco It doesn't support the OOB protocol so that's not expected to work, but it shouldn't crash either.

ianco:
> @ianco It doesn't support the OOB protocol so that's not expected to work, but it shouldn't crash either.

How are mobile apps supposed to work with aca-py then?

andrewwhitehead:
It still supports the connections protocol and connection invitations.

ianco:
I'm just testing using these instructions:  https://github.com/hyperledger/aries-cloudagent-python/blob/master/demo/AliceGetsAPhone.md

andrewwhitehead:
Yes the agents may need an optional flag to switch to OOB/did-exchange rather than the old protocol.

ianco:
> Yes the agents may need an optional flag to switch to OOB/did-exchange rather than the old protocol.

Yes and the old protocol should be the default to maintain compatibility with mobile agents.

andrewwhitehead (approved):

PR #836, opened 2020-12-10
feat(conn_record): add arbitrary metadata key-value storage

dbluhm:
After some brief discussion with @andrewwhitehead and @ianco on rocket chat about a need to be able to track info not strictly related to the connection protocol but still associated with the connection, I threw this together as a quick proof of concept. I get the feeling there might be a better optimized way to do this or other more elegant solutions but figured I'd get the conversation started with the obvious/simple choice.

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -367,8 +367,8 @@ class ConnRecord(BaseRecord):
         await self.clear_cached_key(session, cache_key)
 
     async def metadata_get(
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
I think that looks like a good solution. I was a little worried about metadata setters writing the connection record itself, potentially causing overwrites when there's no synchronization, and this avoids that issue. Maybe it should accept arbitrary JSON-compatible data though instead of a string?

andrewwhitehead (approved):

PR #840, opened 2020-12-10
[Multitenancy] Profile integration

TimoGlastra:
Took me a while, but the profile features are now fully integrated with the multitenancy features

resource management is not implemented yet. This means wallets will only get opened but not closed after a while/ if there are open too many.

@ianco


swcurran:
@burdettadam --- noticed this commit message - `FOR THE ALLIANCE`.  Is this a Star Wars reference?

ianco (approved):

PR #841, opened 2020-12-11
Fix Faber demo to work with mobile PWD demo

ianco:
See https://github.com/hyperledger/aries-cloudagent-python/blob/master/demo/AliceGetsAPhone.md

### DIFF
--- a/demo/runners/faber.py
+++ b/demo/runners/faber.py
@@ -1,5 +1,4 @@
 import asyncio
-import io
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


swcurran (dismissed):
It works, right?

ianco:
> It works, right?

It's approved so it must :-D

ianco:
Argh formatting error

andrewwhitehead (approved):

PR #849, opened 2020-12-14
Rfc 441 compliance

sklump:
Verifiers, provers choose credentials and vet presentations deterministically as per RFC 441.


aries_cloudagent/verifier/base.py
```diff
@@ -0,0 +1,39 @@
+"""Base Verifier class."""
```
andrewwhitehead, line 1: I don't think this is meant to be added?

sklump: Right: it's an artifact

andrewwhitehead (approved):

PR #861, opened 2020-12-16
[Multitenancy] general updates

TimoGlastra:
- sync with master
- allow to set label on wallet creation (@esune)
- allow to set key management mode on wallet creation
- add admin api openapi response schemas
- general cleanup / improvements
- add multitenancy authorization header to openapi / swaggerui
- add route protection for base and subwallets
  - subwallets cannot call /multitenancy endpoints
  - base wallet cannot call general workflow actions (issue-credential ....)
- add wallet id to webhooks (x-wallet-id header) 
- move relay to session class. This makes it transport agnostic and doesn't require code on the transport layer

@ianco


TimoGlastra:
Things still missing in core:
- profile caching and cache invalidation
- full demo update (with performance)
- add unit tests for new features (old ones are already passing)
- documentation
- integration tests

Additional features still missing:
- webhook urls per subwallet (currently one for all with x-wallet-id header)
- mediator integration
- update wallet settings
- import /export wallet
- rotate wallet key
- unmanaged wallet scenario

TimoGlastra:
@ianco I'll update with master and then it's ready to merge

ianco (approved):

PR #864, opened 2020-12-16
store routing keys as public keys in did. 

burdettadam:
bugfix for mediation connection protocol integration.

- `create_did_doc(...)` now stores routing keys as public keys objects instead of strings. when outbound transport consumes routing keys it requires a public key, not a string.
- code duplication is reduced by use of a decorator to check mediation parameters
- added mediation_record parameter to optimize decorator behavior
- small documentation changes

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/manager.py
+++ b/aries_cloudagent/protocols/connections/v1_0/manager.py
@@ -66,7 +66,7 @@ class ConnectionManager:
         """
         return self._session
 
-    def validate_mediation(func):
+    def validate_mediation(id_index, record_index):
         """
         Decorate method for validation of mediation input.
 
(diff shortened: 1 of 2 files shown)


dbluhm:
Spotted some issues; working with @burdettadam to address them. Specifically, using the decorator as it is now, it assumes that mediation_id and mediation_record are always passed as keyword arguments which doesn't match the method signatures decorated. Rather than modifying method signatures, I think we likely just need to simplify the validation into a plain old method.

burdettadam:
I wanted to see what it would take to get the decorator handling args and not just kwargs, so I added some tests and updated the decorator accordingly. I think the complexity of passing parameters into a decorator is not ideal. I will update this pr to reflect @dbluhm suggestions.

andrewwhitehead:
Looks pretty DRY at least!

andrewwhitehead (approved):

PR #866, opened 2020-12-18
fix: minor tweaks to run scripts

dbluhm:
These tweaks better support using docker-compatible container runtimes (i.e. podman, which is the default for RHEL based systems these days) for running tests by specifying an alternate runtime with the environment variable `CONTAINER_RUNTIME`. Additionally, I made changes as recommended by shellcheck.

I believe these changes should be pretty benign but I don't have a reliable way to test that for all environments.

### DIFF
--- a/scripts/run_docker
+++ b/scripts/run_docker
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-cd $(dirname $0)
+cd "$(dirname "$0")" || exit 1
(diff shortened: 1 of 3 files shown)


andrewwhitehead:
Looks good to me, and tests okay on OS X. Would you mind rebasing it?

andrewwhitehead (approved):

PR #869, opened 2020-12-19
Multiple Webhook targets w/ API Key in Header

shaangill025:

### DIFF
--- a/aries_cloudagent/admin/base_server.py
+++ b/aries_cloudagent/admin/base_server.py
@@ -26,13 +26,15 @@ class BaseAdminServer(ABC):
     def add_webhook_target(
         self,
         target_url: str,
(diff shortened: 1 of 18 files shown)


TimoGlastra:
Thanks for taking multitenancy into account @shaangill025!

Could you maybe give me a little context on the purpose of webhook targets? The functionalities they provide don't seem to be used. Because we can now create an unlimited amount of wallets, and not all wallets are loaded at all times, It would be easier to not have to create a webhook target, but use the webhook settings object directly. But this is where my lack of understanding of the webhook targets gets in the way.

ianco:
@Shaanjot you need to DCO signoff your commits (`--signoff`) also please check for the formatting issues.

PR #799, opened 2020-11-23
Changes related to Sign Transaction Protocol  #736

HarshMultani:
This is a WIP pull request which includes following changes related to the "Sign Transaction" Protocol to be implemented.
Below are the related details : 

https://github.com/hyperledger/aries-cloudagent-python/issues/736

https://hackmd.io/5LzMhfsMQBevB5V2tKz4hA?view

It contains changes in connection record, connection related classes and new module is added "endorse_transaction"

PR #870, opened 2020-12-21
Sign Transaction Protocol implementation - Related to#799

HarshMultani-AyanWorks:
This feature is to allow author to send the transaction request to Endorser for signature and write to ledger.
This PR is related to a previous one #799 

Below are the related details :

#736

https://hackmd.io/5LzMhfsMQBevB5V2tKz4hA?view

#836 

It contains changes in connection metadata, connection manager and routes. Also added handler and a new message type for sending role to another agent.
A new module is added "endorse_transaction"


aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py
```diff
@@ -0,0 +1,440 @@
...
+
+        ledger: BaseLedger = self.session.inject(BaseLedger, required=False)
+
+        if not ledger:
+            reason = "No indy ledger available"
+            if not self.session.settings.get_value("wallet.type"):
+                reason += ": missing wallet-type?"
+            raise web.HTTPForbidden(reason=reason)
+
+        async with ledger:
```
ianco, line 99: Copying my feedback over from the previous PR:

@HarshMultani @swcurran @ankita-p17 ... I’ve been doing some thinking about the design of the Endorser api. I think that the current approach (creating a signed transaction and dealing with TAA inside the api) is not the best. The existing api’s that write to the ledger (create schema, create cred def, etc.) already have all this logic, so adding it to the Endorser API is duplication.

I think to keep separation of concerns and to keep the code DRY:
• Update the ledger-updating api's (create schema, create cred def etc.) to either write directly to the ledger, or return a prepared (signed & TAA-accepted) transaction
• The controller can then pass this transaction as an attachment to the /transactions/request-endorsement endpoint
• (this endpoint then doesn't have to worry about anything other than packaging the attachment and sending to the Endorser for endorsement)
• Once the endorsement is received, the controller can call another endpoint (e.g. /transactions/write-to-ledger) to write the endorsed transaction to the ledger
• (This last step will involve some indy-specific work, adding the endorser signature to the indy transaction)

Since we already have code for preparing the indy transactions, I don't think we should add this to the endorser api's at all

aries_cloudagent/protocols/connections/v1_0/manager.py
```diff
@@ -1311,3 +1313,56 @@ async def update_inbound(
             if conn_info.verkey == recip_verkey:
                 connection.routing_state = routing_state
                 await connection.save(self._session)
+
+    async def set_transaction_my_role(
```
ianco, line 1317: Probably want to use a different term other than "role", and probably don't want to add these methods to the connection manager (since they are endorser protocol specific)

ianco (changes requested):

aries_cloudagent/protocols/connections/v1_0/manager.py
HarshMultani-AyanWorks: Hi @ianco, Ok. We would move the entire code which sets the transaction related roles in a connection record to the endorse_transaction folder, so that it comes with the set of endorse_transaction APIs

HarshMultani-AyanWorks: Hi @ianco, In the latest commit we have separated the code which sets the metadata and transferred it to the set of endorse-transaction APIs. Also, we have changed the name from roles to jobs wherever it is required. Can you check it?

aries_cloudagent/messaging/schemas/routes.py
```diff
@@ -115,32 +129,98 @@ async def schemas_send_schema(request: web.BaseRequest):
...
-            reason += ": missing wallet-type?"
-        raise web.HTTPForbidden(reason=reason)
-
-    issuer = context.inject(IndyIssuer)
-    async with ledger:
-        try:
-            schema_id, schema_def = await shield(
-                ledger.create_and_send_schema(
-                    issuer, schema_name, schema_version, attributes
+    auto_endorse = json.loads(request.query.get("auto_endorse", "null"))
```
ianco, line 132: Default "true"

aries_cloudagent/messaging/credential_definitions/routes.py
```diff
@@ -129,92 +143,160 @@ async def credential_definitions_send_credential_definition(request: web.BaseReq
 
     """
     context: AdminRequestContext = request["context"]
+    auto_endorse = json.loads(request.query.get("auto_endorse", "null"))
```
ianco, line 146: default "true"

aries_cloudagent/messaging/credential_definitions/tests/test_routes.py
```diff
@@ -240,6 +256,8 @@ async def test_send_credential_definition_ledger_x(self):
             }
         )
 
+        self.request.query = {"auto_endorse": "true"}
```
ianco, line 259: Needs tests with "auto_endorse": "false"

aries_cloudagent/messaging/schemas/tests/test_routes.py
```diff
@@ -83,6 +85,7 @@ async def test_send_schema_x_ledger(self):
                 "attributes": ["table", "drink", "colour"],
             }
         )
+        self.request.query = {"auto_endorse": "true"}
```
ianco, line 88: Needs tests with "auto_endorse": "false"

aries_cloudagent/protocols/endorse_transaction/v1_0/routes.py
```diff
@@ -0,0 +1,550 @@
...
+
+
+@docs(
+    tags=["endorse-transaction"],
+    summary="For author to send a transaction request",
+)
+@querystring_schema(TranIdMatchInfoSchema())
+@querystring_schema(ConnIdMatchInfoSchema())
+@response_schema(TransactionRecordSchema(), 200)
+async def transaction_create_request(request: web.BaseRequest):
```
ianco, line 138: For this endpoint there are inputs for a transaction id and connection id, but I don't know how to attach the transaction (was expecting a message attachment?)

E.g. I expect:
- call create schema with "auto_endorse" = "false"
- get a transaction object (signed and taa-endorsed)
- pass this object to this "create-request" endpoint

aries_cloudagent/messaging/credential_definitions/routes.py
```diff
@@ -129,92 +143,160 @@ async def credential_definitions_send_credential_definition(request: web.BaseReq
...
+        ledger: BaseLedger = session.inject(BaseLedger, required=False)
 
-    return web.json_response({"credential_definition_id": cred_def_id})
+        if not ledger:
+            reason = "No indy ledger available"
+            if not session.settings.get_value("wallet.type"):
+                reason += ": missing wallet-type?"
+            raise web.HTTPForbidden(reason=reason)
+
+        async with ledger:
```
ianco, line 260: This code is all duplicated from existing code in the ledger object, is there some way to reuse (e.g. call ledger to create the transaction object) rather than repeating the code?

aries_cloudagent/messaging/schemas/routes.py
```diff
@@ -115,32 +129,98 @@ async def schemas_send_schema(request: web.BaseRequest):
...
+            except (IndyIssuerError, LedgerError) as err:
+                raise web.HTTPBadRequest(reason=err.roll_up) from err
+
+        return web.json_response({"schema_id": schema_id, "schema": schema_def})
+
+    else:
+        transaction_message = await request.json()
+        session = await context.session()
+
+        wallet: BaseWallet = session.inject(BaseWallet, required=False)
```
ianco, line 165: This code is all duplicated from existing code in the ledger object, is there some way to reuse (e.g. call ledger to create the transaction object) rather than repeating the code?

ianco (changes requested):
Change look good so far!  I've added a couple of comments, but in general:

- make sure the alice/faber demo works (there are docs in the "demo" folder) - it doesn't necessarily need to include support for the new endorser functions, but make sure nothing breaks
- also make sure the changes are non-breaking/backward-compatible of possible (i.e. make the new options for creating schema and cred def optional)
- need some documentation on how to create, endorse and write a transaction (for example create a new schema) using the new api's

ianco:
One additional point for discussion (@swcurran @andrewwhitehead ) should we have a new aca-py argument to enable the new endorser functionality?  e.g. "--enable-transaction-endorser"

ianco:
Hey @HarshMultani-AyanWorks this looks pretty good!  I was able to get most of the way through the process, some documentation would be good ;-)

What I did:
- started the Alice and Faber demo's and completed the connection between the 2 agents
- add a new public DID for Alice (so she could be the endorser) (also added this DID to the ledger)
- Set the Alice connection to Endorser and Faber to Author
- Faber - created a new schema with "auto-endorse = False" (returned transaction id)
- Faber - created a request (attach this transaction request to the Alice connection)
- Alice - view and endorse the request
- Faber - confirm that the endorsed request was received

I have some more requested changes, but let's try to schedule a call and discuss these before we do more work, I'd also like to get @swcurran on the call for his perspective.

My feedback:
- I find the terminology a bit confusing around setting the "set-transaction-jobs" endpoint (I would use "roles" rather than "jobs" but I think we already decided not to use the word "roles"?)
- I think the endpoints like create schema, create cred def etc should *just* return the un-signed transaction, *not* return a full endorser structure (this binds these methods too tightly to the endorser protocol as implemented by aca-py) and should also *not* automatically save the transaction.  I think the `/transactions/create-request` endpoint should take the json structure as an attachment, so the author would have to:
    - call (for example) create schema and get the ledger transaction info
    - create a json structure to pass to the `/transactions/create-request` endpoint (with the ledger transaction as an attachment)
    - call the `/transactions/create-request` endpoint and pass this structure - this will create the transaction record in the aca-py database and send the message to the endorser
    - etc ...
- I couldn't figure out how to write the transaction to the ledger once I got it endorsed (i expected a `/ledger/write-transaction` endpoint (or something like that)

ianco (dismissed):
Overall really good!  I added a comment with some items to discuss before we finalize.

andrewwhitehead:
It's looking good to me. It seems a bit odd to have `auto_endorse` options defaulting to `true` and overridden with `false`, rather than having the option opt-in to the much less common option of requesting signatures from other parties. I'm not sure what the option would be called if reversed though. `as_transaction` maybe?

wip-abramson:
Hey, I am quite out of the loop but I had a scan over the protocol in the branch and am wondering if there is also the intention to define the more generic "please sign this" protocol as it is described in the HackMd. 

I would also be interested in any thoughts on how this could also be used to allow me to request **my agent** to please sign this. I.e. if I had a public DID. Although I think in the future the crypto API's could be tweaked to support group signatures from base credential objects which would be cool.

E.g. please sign this article I have written using the credential that proves I am an epidemiologist or whatever.

Note: I don't necessarily want to send this to anyone rather I might want to publish it on the internet for anyone to check.

ianco:
> Hey, I am quite out of the loop but I had a scan over the protocol in the branch and am wondering if there is also the intention to define the more generic "please sign this" protocol as it is described in the HackMd.
> 
> I would also be interested in any thoughts on how this could also be used to allow me to request **my agent** to please sign this. I.e. if I had a public DID. Although I think in the future the crypto API's could be tweaked to support group signatures from base credential objects which would be cool.
> 
> E.g. please sign this article I have written using the credential that proves I am an epidemiologist or whatever.
> 
> Note: I don't necessarily want to send this to anyone rather I might want to publish it on the internet for anyone to check.

Hi @wip-abramson , the underlying protocol (agent-to-agent) implements the generic "please sign this" protocol, however the aca-py admin api is specific to endorsing indy transactions to write to the ledger.  There is an RFC in progress that focuses on the generic agent-to-agent protocol.

To support the more generic use case that you suggest would require an additional admin api (or set of api's) to expose the more generic functionality.

ianco:
There is still some outstanding work, but I recommend we get this PR merged and create tickets for any outstanding tasks.  (Some tasks will require review/discussion.)

- naming of “set-transaction-jobs” endpoint and parameters
- naming of the “auto-endorse” parameter
- include endorser DID as connection meta-data?
- cleanup/re-factor the code to save the non-secrets records
- review messages vs the draft RFC
- general code review for new/updated methods

ianco (approved):

PR #779, opened 2020-11-10
[Multitenancy] functionality of webhooks, labels and image_url per wallet

baegjae:
### Description:
The webhook functionality is an essential feature, because the controller should receive messages from the aca-py server. This PR enables the functionality of webhooks, labels and image_url per wallet. Also, it contains useful other features. See updates and APIs changes below.

### Terms:
**Base wallet**: Wallet created when aca-py is launched.

### Updates:
1. `webhook_urls` works per wallet
2. `label` works per wallet
3. `image_url` works per wallet
4. on demand wallet loading (reduce boot-up time)
5. wallet mapping (`connection_id` or `key` -> `wallet_name`) is stored in the storage (preparation for scale-out aca-py)
6. When registering nym, optional `wallet_name` is used to register the did of another wallet
7. When creating local did, optional `seed` is used

### APIs changes:
- `POST /wallet` : Add new wallet (added `webhook_urls`, `label`, `image_url`) - public
- `GET /wallet` : Get information of wallets with query - base wallet only
- `PUT /wallet/{wallet_id}` : Update information of a wallet - base wallet only
- `DELETE /wallet/{wallet_id}` : Remove a wallet - base wallet only
- `GET /wallet/me` : Get information of my wallet - user wallet
- `PUT /wallet/me` : Update information of my wallet - user wallet
- `DELETE /wallet/me` : Remove my wallet - user wallet
- `/wallet/did/create` : Create a local DID (added `seed`)
- `/ledger/register-nym` : Send a NYM registration to the ledger (added `wallet_name` to register the did of another wallet)

### Demo changes:
- The `--multitenant` flag enables multitenancy support. With this, Faber and Alice will use a single Base agent. Please see the detailed flows in README-multitenant.md.

### TODO
- Test code updates (Several `send_webhook` functions included `context` as a parameter to make webhook works per wallet)

PR #872, opened 2020-12-21
Multiple Webhook targets w/ API Key in Header

shaangill025:
Signed-off-by: Shaanjot <Gill>

### DIFF
--- a/scripts/run_docker
+++ b/scripts/run_docker
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-cd $(dirname $0)
+cd "$(dirname "$0")" || exit 1
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
 
-docker build -t aries-cloudagent-run -f ../docker/Dockerfile.run .. || exit 1
+$CONTAINER_RUNTIME build -t aries-cloudagent-run -f ../docker/Dockerfile.run .. || exit 1
 
 ARGS=""
 for PORT in $PORTS; do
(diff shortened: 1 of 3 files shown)


ianco:
Missing DCO signoff!  Make sure you `--signoff` your commits

andrewwhitehead:
Okay so I think we've got two somewhat separate things going on here, the addition of the API key support and the wallet_id handling. The API key support looks good and it might be easier to put that into a smaller PR first.

For the profile/wallet_id handling, I think the way that is going to be handled is via the Responder objects. In the Admin server, the responder (AdminResponder) is created separately for each request, with enough context to load the relevant profile (aka wallet). Currently it always uses the root profile but that will change:
https://github.com/hyperledger/aries-cloudagent-python/blob/b9c59da5a6d0482d63ab60e6b3c184b864b8a6f0/aries_cloudagent/admin/server.py#L284

That's how it works for Admin (REST) requests. For agent messages, a different Responder object is created by the Dispatcher:
https://github.com/hyperledger/aries-cloudagent-python/blob/b9c59da5a6d0482d63ab60e6b3c184b864b8a6f0/aries_cloudagent/core/dispatcher.py#L161

In this case the responder is given a Context instead of a Profile, although we might want to change that for consistency (not sure if the multitenancy branch adjusts this). In both cases we have a reference to the correct Profile, so if there are additional webhook targets to be added they can be added on the responder object at the time that it is created. There's no need to register the targets globally on the admin server by adjusting add_webhook_target - those are the targets that every webhook is sent to.

andrewwhitehead:
Actually I see the multitenancy branch has already implemented support for sending the wallet ID. I don't see support for custom webhook URLs but it would be likely easier to implement once that is merged.

https://github.com/hyperledger/aries-cloudagent-python/commit/34e6aae3f8c2c8fb95da1180e7d374b50303a2bb

victormartinez-work:
> Actually I see the multitenancy branch has already implemented support for sending the wallet ID. I don't see support for custom webhook URLs but it would be likely easier to implement once that is merged.
> 
> Commit here: [34e6aae](https://github.com/hyperledger/aries-cloudagent-python/commit/34e6aae3f8c2c8fb95da1180e7d374b50303a2bb)

I think custom webhook Url per wallet was added in the old multitenancy branch. Not sure if the PR could be reused.  
https://github.com/hyperledger/aries-cloudagent-python/pull/779

shaangill025:
I Will be doing a small PR for the API key in the header feature.
I agree with @victormartinez-work, the multitenancy branch implementation will be able to handle custom webhook URL. There is a webhook_urls [list] property inside the wallet record. We can retrieve this list using the context setting. 
In multitenancy, if it is the base wallet then it will dispatch to webhooks registered with the server, and if not, then it will dispatch to webhooks associated with that wallet record. Is there a use case where we might want to dispatch to both for the multitenant wallet?

PR #873, opened 2020-12-21
Agency multitenant

ianco:
@andrewwhitehead this can be merged into master.  There is still some work in progress:

- unit tests
- documentation
- wallet caching
- integration testing
- full integration of mediator, managed wallet and web hooks
- aome alice/faber/performance demo updates

However multitenancy works and can be used, and all existing aca-py functionality is working.

FYI @TimoGlastra


swcurran (approved):
I'm good with this.  Obvs didn't review or run, but I'm happy to get this into the primary branch.

andrewwhitehead (approved):

andrewwhitehead:
A couple things I noticed to clean up later:

- Instead of injecting `IndySdkLedgerPool` everywhere it should inject a base class - `IndyLedgerPool` - so that indy-vdr can be used and the app doesn't blow up if indy-sdk isn't installed. I'll adjust that on the shared-components branch. I think the adjustments to the `CachedProvider` are only used in that one case, so it might make sense to move that logic into the pool provider itself or a mixin class instead of wrapping it.

- There were a couple admin server tests deleted that should be restored

- We need to document `pyjwt` as a new dependency. I'm not sure if that has a binary component but if it's hard to install it could become a separate flag - ie. `pip install aries_cloudagent[jwt]`

PR #883, opened 2020-12-24
Coordinate Mediation Admin API

dbluhm:
This PR adds an Admin API for working with the coordinate mediation protocol.


dbluhm:
We've got a few items to sort out. An endpoint for deleting an existing mediation request is likely an appropriate addition. Got a bug or two to squash as well.

aries_cloudagent/config/argparse.py
```diff
@@ -925,6 +925,34 @@ def add_arguments(self, parser: ArgumentParser):
...
+            " will be sent. Default: false.",
+        )
+        parser.add_argument(
+            "--auto-respond-mediation-grant",
+            action="store_true",
+            env_var="ACAPY_AUTO_RESPOND_MEDIATION_GRANT",
+            help="Automatically respond to mediation grant message with a keylist update"
+            "message containing a newly created did verkey for use as a recipient key"
+            ". Default: false.",
+        )
```
dbluhm, line 955: These changes probably don't belong in this PR. Any thoughts to the contrary, @burdettadam?

aries_cloudagent/protocols/coordinate_mediation/v1_0/models/mediation_schemas.py
```diff
@@ -0,0 +1,89 @@
+"""Admin API mediation schemas for parater validation."""
```
dbluhm, line 1: This file needs a review; I think some of these are unused now.

aries_cloudagent/protocols/coordinate_mediation/v1_0/routes.py
```diff
@@ -0,0 +1,456 @@
...
+    """Parameters and validators for create Mediation request query string."""
+
+    mediator_terms = MEDIATOR_TERMS_SCHEMA
+    recipient_terms = RECIPIENT_TERMS_SCHEMA
+
+
+class MediationCreateRequestQuerySchema(OpenAPISchema):
+    """Create mediation request query string schema."""
+
+    conn_id = CONNECTION_ID_SCHEMA
```
dbluhm, line 68: `CONNECTION_ID_SCHEMA` marks it as optional but it should be required in this case. Given that, maybe we should consider making this a path match var again.

dbluhm (commented):
Some thoughts for consideration, @burdettadam.

aries_cloudagent/config/argparse.py
burdettadam: Correct, I thought I had already removed some of them.

aries_cloudagent/protocols/coordinate_mediation/v1_0/routes.py
burdettadam: CONNECTION_ID_SCHEMA appears to be pulled in from the mediation protocol so we can change it without messing up other admin API's. We can also create a new field in place easily if required is all that's needed.
something like,
`fields.UUID( 
    description="Connection identifier (required)",
    required=True,
    example=UUIDFour.EXAMPLE,
)`
or possibly dynamically change the required key: value inside the CONNECTION_ID_SCHEMA.
I have no preference for how the API should look, path match var or body match var.

burdettadam (approved):

ianco:
Overall looks pretty good, is there any documentation?

dbluhm:
> Overall looks pretty good, is there any documentation?

At this point, just what is in the code but we plan to open another PR with a demo for mediation as well as more general documentation where appropriate

andrewwhitehead (approved):

PR #884, opened 2020-12-24
Webhooks dispatch model w/ multitenancy

shaangill025:

- Looking for feedback for this approach, this is a draft version with no test cases added yet
- Updated multitenancy Admin API to include associated webhook_urls and dispatch type
- Support different dispatch types, "default" will send only to associated webhook_urls, "base" will send to only to AdminServer's webhook_targets, and "both"
- In server.py [send_webhook function], the following approach was giving a WalletNotFound error.
```
     wallet = await WalletRecord.retrieve_by_id(session, wallet_id)
```
- Current implementation is using values from profile settings.


aries_cloudagent/multitenant/admin/routes.py
```diff
@@ -205,12 +224,20 @@ async def wallet_create(request: web.BaseRequest):
...
+    # wallet_dispatch_type
+    # default: dispatch only to webhook_urls defined
+    # base: dispatch only to webhook_targets
+    # both: dipatch to both
     settings = {
         "wallet.type": body.get("wallet_type") or "in_memory",
         "wallet.name": body.get("wallet_name"),
         "wallet.key": wallet_key,
+        "wallet.webhook_urls": body.get("wallet_webhoook_urls") or [],
+        "wallet.dispatch_type": body.get("wallet_dispatch_type") or "default",
```
TimoGlastra, line 236: Maybe we can use `admin.webhook_urls` here? This way we can keep the send webhook method "dumb".

Based on the dispatch_type we will extend the base context differently (`get_wallet_profile` method in `MultitenantManager`):
- both: `admin.webhook_urls` = combination of base webhook urls and the webhook urls stored in the wallet_record
- base: do not store webhook_urls in wallet record. When extending the webhook urls of the base settings will be used
- default: do store in webhook urls in wallet record and overwrite settings when extending context

This way we don't need to store any new settings, and don't have to do any custom logic to determine if we're in base of subwallet context.

aries_cloudagent/admin/server.py
```diff
@@ -725,24 +726,60 @@ def remove_webhook_target(self, target_url: str):
...
+        wallet_name = profile.settings.get("wallet.name")
+        wallet_dispatch_type = profile.settings.get("wallet.dispatch_type")
+        wallet_webhook_urls = profile.settings.get("wallet.webhook_urls")
+        base_wallet = wallet_id == wallet_name
+
+        # async with profile.session() as session:
+        #     wallet = await WalletRecord.retrieve_by_id(session, wallet_id)
+        #     base_wallet = wallet.wallet_id == wallet.wallet_name
+        #     wallet_dispatch_type = wallet.wallet_dispatch_type
+        #     wallet_webhook_urls = wallet.wallet_webhoook_urls
```
TimoGlastra, line 757: The wallet record cannot be found in this case because the profile is in context of the subwallet. All wallet records are stored in the base wallet. This should work if you use `self.root_profile` instead of `profile`, but I think you can get all required settings from the settings dict

aries_cloudagent/multitenant/admin/routes.py
shaangill025: @TimoGlastra I appreciate the feedback, I have made some changes to align with your suggested approach. 
- I also had to make changes to the send_webhook function [change to original logic is minimal], as the target URLs are now available in profile.settings.get("admin.webhook_urls") v/s webhook_targets dictionary which has only has the base URLs. 
- Attributes such as max_attempt and topic_filter can also be supported, if you think this approach is good then I can go about handling these.

shaangill025:
@TimoGlastra I appreciate the feedback, I have made some changes to align with your suggested approach.

- I also had to make changes to the send_webhook function [change to original logic is minimal], as the target URLs are now available in profile.settings.get("admin.webhook_urls") v/s webhook_targets dictionary which has only has the base URLs.

- Attributes such as max_attempt and topic_filter [missed by not including webhook_target] can also be supported, if you think this approach is good then I can go about handling these.

TimoGlastra:
This approach looks good to me. However it seems like the webhook urls are not stored in the wallet record now. This means it works for the `create_wallet` call, but if a profile is retrieved later on (e.g. after a restart) it won't have the webhook urls anymore. Maybe I'm missing something?

shaangill025:
@TimoGlastra Thanks for the feedback. Sorry, I misunderstood your earlier comments and thought that the webhook_urls does not need to be saved in wallet_record. I will add the webhook_url related attributes to wallet_record again.

aries_cloudagent/multitenant/admin/routes.py
```diff
@@ -58,6 +58,25 @@ class CreateWalletRequestSchema(OpenAPISchema):
...
+        description="Webhook target dispatch type for this wallet. \
+            default - Dispatch only to webhooks associated with this wallet. \
+            base - Dispatch only to webhooks associated with the base wallet. \
+            both - Dispatch to both webhook targets.",
+        example="default",
+        default="default",
+        validate=validate.OneOf(["default", "both", "base"]),
+    )
+
+    wallet_webhoook_urls = fields.List(
```
baegjae, line 71: Typo here. wallet_webhoook_urls -> wallet_webhook_urls.

aries_cloudagent/multitenant/manager.py
```diff
@@ -170,7 +185,9 @@ async def create_wallet(
             # In unmanaged mode we don't want to store the wallet key
             if key_management_mode == WalletRecord.MODE_UNMANAGED:
                 del settings["wallet.key"]
-
+            # Adding wallet_webhook_urls and wallet_dispath_type to wallet_record
+            settings["wallet.webhook_urls"] = wallet_webhook_urls
```
baegjae, line 189: This is redundant. How about `wallet_webhook_urls` and `wallet_dispatch_type` add at `settings` in wallet_create of route.py. So. we can remove these parameters in create_wallet function.

aries_cloudagent/multitenant/manager.py
```diff
@@ -124,6 +124,17 @@ async def get_wallet_profile(
                 "wallet.type": None,
             }
 
+            dispatch_type = extra_settings.get("wallet.dispatch_type")
```
baegjae, line 127: I think we don't need to pass extra_settings for `wallet.webhook_urls` and `wallet.dispatch_type` in here, because we already have these values in `wallet_record.wallet.webhook_urls` and `wallet_record.wallet.dispatch_type`.

aries_cloudagent/multitenant/manager.py
TimoGlastra: I agree! We can store `wallet.webhook_urls` and `wallet.dispatch_type` inside the settings of the wallet record. 

Only thing we need to do then is modify `admin.webhook_urls` based on `wallet.dispatch_type`

aries_cloudagent/multitenant/admin/routes.py
shaangill025: @baegjae Thanks for the feedback, I have fixed the typos and adjusted the redundant logic.

aries_cloudagent/multitenant/manager.py
shaangill025: Updated implementation based upon feedback, now storing these attributes inside the settings of the wallet record and modifying admin.webhook_urls based upon dispatch_type.

TimoGlastra (approved):
LGTM. I like this flexible approach

Only thing missing may be the topic filters, but it doesn't seem they were used anywhere?

andrewwhitehead (approved):

PR #889, opened 2021-01-04
[Multitenancy] unit tests (and image url option)

TimoGlastra:
This adds unit tests for almost all code added for multitenancy.

It also add an image url option to the agent startup. The multitenancy branch had this option, but there was no way to set it for a non-multitenant wallet. This is just for feature parity.

PR #892, opened 2021-01-05
unit test coverage improvement

sklump:

### DIFF
--- a/aries_cloudagent/wallet/models/tests/test_wallet_record.py
+++ b/aries_cloudagent/wallet/models/tests/test_wallet_record.py
@@ -1,0 +1,66 @@
+from asynctest import TestCase as AsyncTestCase
+
+from .. import wallet_record as test_module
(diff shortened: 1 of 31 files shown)


aries_cloudagent/wallet/models/tests/test_wallet_record.py
```diff
@@ -0,0 +1,66 @@
+from asynctest import TestCase as AsyncTestCase
+
+from .. import wallet_record as test_module
+from ..wallet_record import WalletRecord
+
+
+class TestIssuerCredRevRecord(AsyncTestCase):
```
andrewwhitehead, line 7: I think this needs renaming? It looks like it might duplicate some additional testing in PR #889 as well

sklump:
OK, merged and augmented with more

andrewwhitehead (approved):

PR #894, opened 2021-01-06
Add optional seed field when creating did

baegjae:
- This PR adds an optional `seed` field when creating DID, and the seed is used for did derivation.
- Especially in **multitenancy** situations, we may need to use the assigned DID. In our case, we need to use the steward DID to use registering nym.
- I know the seed is sensitive data, but the seed is not recorded in this API. So, we only need to be careful when calling the API.

@ianco @TimoGlastra

### DIFF
--- a/aries_cloudagent/wallet/routes.py
+++ b/aries_cloudagent/wallet/routes.py
@@ -56,7 +56,7 @@ class DIDCreateSchema(OpenAPISchema):
         description="Optional seed for did derivation",
         required=False,
         example="00000000000000000000000Endorser1",
-        validate=validate.Length(equal=32)
+        validate=validate.Length(equal=32),
     )
 
 
--- a/aries_cloudagent/wallet/routes.py
+++ b/aries_cloudagent/wallet/routes.py
@@ -9,7 +9,7 @@ from aiohttp_apispec import (
     response_schema,
 )
 
-from marshmallow import fields
+from marshmallow import fields, validate
 
 from ..admin.request_context import AdminRequestContext
 from ..ledger.base import BaseLedger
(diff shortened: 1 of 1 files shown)


ianco:
Build error:

```
self = <aries_cloudagent.wallet.tests.test_routes.TestWalletRoutes testMethod=test_create_did>

    async def test_create_did(self):
        with async_mock.patch.object(
            test_module.web, "json_response", async_mock.Mock()
        ) as json_response:
            self.wallet.create_local_did.return_value = DIDInfo(
                self.test_did, self.test_verkey, DIDPosture.WALLET_ONLY.metadata
            )
>           result = await test_module.wallet_create_did(self.request)

aries_cloudagent/wallet/tests/test_routes.py:86: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

request = <MagicMock id='139850041889568'>

    @docs(tags=["wallet"], summary="Create a local DID")
    @request_schema(DIDCreateSchema)
    @response_schema(DIDResultSchema, 200, description="")
    async def wallet_create_did(request: web.BaseRequest):
        """
        Request handler for creating a new local DID in the wallet.
    
        Args:
            request: aiohttp request object
    
        Returns:
            The DID info
    
        """
        session = await context.session()
        wallet = session.inject(BaseWallet, required=False)
        if not wallet:
            raise web.HTTPForbidden(reason="No wallet available")
>       body = await request.json() if request.body_exists else {}
E       TypeError: object MagicMock can't be used in 'await' expression

aries_cloudagent/wallet/routes.py:250: TypeError
```

ianco:
This is controversial, we've had discussions before about including an optional seed @andrewwhitehead @swcurran

baegjae:
@swcurran I agree your opinion. I was hesitant to post this PR due to security concerns. It's a good idea to provide it as a plugin.

@ianco Thanks. The build error log is useful when I make the next PR.

PR #895, opened 2021-01-06
[Multitenancy] fix: public dids could not be used for implicit invitations

TimoGlastra:
Public DIDs were not registered for relaying until it was used in an invitation. However a public did by itself could be used as an an implicit invitation, therefore it should be routed to the corresponding subwallet


sklump:
Looks OK to me, but there has been a lot of activity in master - could you merge it? Some unit tests look like I might have duplicated them, more or less.

TimoGlastra:
Updated with master

@sklump I'm not 100% sure which unit tests you refer to? These tests address a new addition, so I don't think they're duplicates. Or do you mean unit tests in general? If that's the case I'll take a look at the unit tests and address them in a separate PR

andrewwhitehead (approved):

PR #899, opened 2021-01-07
convert invitation to raw keys on use, do not save raw in invitation

sklump:


nodlesh (approved):
Looks good. I no longer have the failure.

andrewwhitehead (approved):

PR #900, opened 2021-01-07
make openapi tags align with rest of aca-py

sklump:

I don't believe keylist merits its own tag, since it follows the same RFC (0211) with the rest of coordinate-mediation. I note that the spec itself vacillates between "keylist" and "key list" _(I'd prefer "key list" since it's correct English, but it's not my spec after all)_.


dbluhm (approved):
LGTM. We'd have to address the "keylist" vs "key-list" in the RFC since we were just following suit in the implementation though I agree that it's awkward.

PR #901, opened 2021-01-07
fix: improperly handled storage errors from bad mediation ids in connections admin api

dbluhm:


dbluhm:
I think there's a similar situation for receive invite, too, actually... I'll take a look at that and push more commits

andrewwhitehead (approved):

PR #902, opened 2021-01-07
feature: set default mediator for use in all newly created connections

dbluhm:
It seems like a good idea to be able to set a default mediator for use in all newly created connections. This would also support multitenancy with external mediator.


dbluhm:
### Progress update

We have the majority of the needed tests written out for the default mediator functionality. Coverage is not happy about the additions to the conductor but we haven't spotted exactly where tests should go for that yet. We're leaning towards getting this section of the functionality merged and then following up with additions to the Admin API given that the API requires quite a bit of time to test thoroughly.

We're doing integration tests by hand now to make sure all the different pieces are fitting together correctly.

cc @burdettadam

dbluhm:
Everything is ready for this PR; a follow up with Admin API is already in the works. cc @burdettadam

andrewwhitehead (approved):

PR #903, opened 2021-01-08
Fix error query connections by alias

ianco:


sklump (approved):

sklump:
All statements are true of the empty set

PR #905, opened 2021-01-11
[Multitenancy] fix responder only bound to request context

### LINKED ISSUE #898: [multitenancy] wallet_id  is not passed correctly in send_webhook
https://github.com/hyperledger/aries-cloudagent-python/blob/5687d287040c2abb4d37999d64fd873ea494a9b8/aries_cloudagent/admin/server.py#L733

In this code, `wallet_id` identifies what this webhook message belongs to.

I checked that `wallet_id` is passed correctly (for newly created wallet) in cases of topic: `connections`, `revocation_registry`.

But, in the case of topic `issue_credential`, `wallet_id` was None.
Although the webhook message belongs to the newly created wallet, I checked that the `profile` passed is for the base wallet (configured when aca-py runs).

Thanks!

@TimoGlastra

TimoGlastra:
The responder that is set in multitenancy authorization middleware was bound to the AdminContext, but not the profile itself. Therefor not all places had access to the correct responder. The responder is now bound to the profile, which is used in the AdminContext.

Fixes #898

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -330,17 +330,17 @@ class AdminServer(BaseAdminServer):
                 except (jwt.InvalidTokenError, StorageNotFoundError):
                     raise web.HTTPUnauthorized()
 
-            # TODO may dynamically adjust the profile used here according to
-            # headers or other parameters
-            admin_context = AdminRequestContext(profile, context=profile.context)
-
             # Create a responder with the request specific context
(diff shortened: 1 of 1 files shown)


baegjae:
@TimoGlastra This change fixes an #898  I posted. And it looks very similar to the code I tested yesterday. I think it's a clear solution. Thanks!

ianco (approved):

PR #907, opened 2021-01-12
Subwallet webhook demonstration [Alice-Faber Demo]

shaangill025:
Webhook dispatch is demonstrated inside the terminal as follows:
image

### DIFF
--- a/demo/runners/support/agent.py
+++ b/demo/runners/support/agent.py
@@ -361,7 +361,7 @@ class DemoAgent:
         webhook_port: int = None,
     ):
         if webhook_port is not None:
-            await self.listen_webhooks(webhook_port)     
+            await self.listen_webhooks(webhook_port)
         self.log(f"Register or switch to wallet {target_wallet_name}")
         if target_wallet_name == self.agency_wallet_name:
             self.ident = self.agency_ident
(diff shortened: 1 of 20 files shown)


ianco:
Web hooks are displayed optionally by providing an `--events` parameter, prob best to stick with this (i.e. display the web hok info optionally).

demo/runners/support/agent.py
```diff
@@ -510,6 +524,11 @@ async def handle_webhook(self, topic: str, payload, headers: dict):
             wallet_id = headers.get("x-wallet-id")
             method = getattr(self, handler, None)
             if method:
+                self.log(
+                    f"Agent called controller webhook: POST {self.webhook_url}",
+                    (f"\nFor wallet: {wallet_id}" if wallet_id else ""),
+                    (f"\nPayload: {repr_json(payload)}" if payload else repr_json({})),
+                )
                 EVENT_LOGGER.debug(
```
ianco, line 532: Web hook will be logged twice if run with `--events`

ianco (changes requested):
Overall looks good!

demo/runners/support/agent.py
shaangill025: Now only relying on event_logger for demonstration.
image

ianco (approved):

andrewwhitehead (approved):

PR #908, opened 2021-01-13
Out of Band Protocol [ref: Aries RFC 0434 & 0496]

shaangill025:
Objectives

- Support 0160 Connections protocol [Done]
- Connection Reuse
1. Use an existing connection when only request~attach is present [done]
2. Use an existing connection when only handshake_protocols is present [done]
3. Both handshake_protocols and request~attach are present [done]
- Attachment [PRESENTATION_REQUEST done]
- Testing [in-progress]

https://github.com/hyperledger/aries-rfcs/tree/master/features/0496-transition-to-oob-and-did-exchange
https://github.com/hyperledger/aries-rfcs/tree/master/features/0434-outofband


shaangill025:
Pending final debugging and more test coverage for OOBManager.

andrewwhitehead:
Looks good, can you please move the base manager to `/connections` rather than the connections protocol.

aries_cloudagent/connections/base_manager.py
```diff
@@ -4,19 +4,21 @@
...
+from ..wallet.base import BaseWallet
+from ..protocols.connections.v1_0.messages.connection_invitation import (
+    ConnectionInvitation,
+)
+from ..ledger.base import BaseLedger
+from ..core.error import BaseError
+from .models.connection_target import ConnectionTarget
+from ..wallet.util import did_key_to_naked
 import logging
 from typing import Sequence, Tuple
```
andrewwhitehead, line 18: Minor import ordering issue - external package imports should be in a block before internal ones. I'm not sure why flake8 isn't complaining about that.

shaangill025: I have changed the order as requested

aries_cloudagent/connections/base_manager.py
```diff
@@ -4,6 +4,9 @@
 For Connection, DIDExchange and OutOfBand Manager.
 """
 
+import logging
+from typing import Sequence, Tuple
+
 from ..core.profile import ProfileSession
```
andrewwhitehead, line 10: I'll clean this up later but just FYI, it should be ordered by (reverse) proximity.. ie. `..core` and `..wallet` come before `.models`, and from there things should be ordered alphabetically.

andrewwhitehead (approved):

aries_cloudagent/connections/base_manager.py
shaangill025: Thanks, I will keep this in mind from now on.

PR #909, opened 2021-01-13
Outbound queue integration: Redis and pluggable interface

adamsc64:
**Background**
- Messages often stay in the internal ACA-py queue for long periods of time without being delivered. This internal queue is ephemeral and in-memory. The team wants to add an option such that outbound messages can be sent to a durable queue external to ACA-py before transport.
- This PR implements the use case of storing outbound messages in a durable queue. A new set of commandline options have been added to provide a way for users to "opt in" to use of the outbound queue. This is intended to be an optional replacement to the current ACA-py outbound transport.
- For the sake of simplicity, the first queue supported by ACA-py itself, introduced in this PR, is a Redis `LIST`. Users will need to configure [Redis persistence](https://redis.io/topics/persistence) to gain message durability benefits in Redis.
- However this PR exposes a pluggable interface extensible to other queue types. The new commandline option `--outbound-queue-class` is available for this use case. Developers can specify the dotpath to a custom queue backend which developers can define themselves, implementing the base class `BaseOutboundQueue`.
- The queue in this PR does not actually deliver the message to another agent. A delivery service ([one prototype can be found here](https://github.com/andrewwhitehead/aca-deliver)) would need to be run. This service should pick up a message from the queue and then deliver that message. What the queue in this PR does provide is simply a new persistence layer that acts as a durable buffer between ACA-py and this new delivery service. In one sense, this PR can be characterized as a form of decomposition, using a queue between two services (ACA-py on the one hand, and a new delivery service on the other).
- A delivery service consuming this queue in order to send outbound messages over transport will need to pop from the left side of the queue (i.e. the Redis `LPOP` command) to get messages in the order they were sent.
- Connections sometimes time out from the server side. For example this may be possible for a persistent connection to Redis. But the redis client provided by `aioredis` seems to be resilient in the case of serverside connection timeout. I tested this on a running redis instance with a timeout set. No errors were generated and the client silently autoconnects without ACA-py needing to handle the event in any way.
- Unit tests should pass when running Redis on a local development environment, defining the environment variable `TEST_REDIS_CONFIG`. If that variable is not defined, the unit tests that require a live Redis connection will simply be skipped. If you want to run the unit test suite with Redis, the easiest thing is to run redis with the default settings (`$ redis-server`) in one terminal screen, then run the unit tests in another with the following syntax: `$ TEST_REDIS_CONFIG=redis://host.docker.internal:6379 ./scripts/run_tests`

**Known Remaining Issues**
- Include an additional `.md` file, along the lines of `Multitenancy.md` or `Mediation.md`, that describes what's been done.

**Changes in this branch**
- This branch introduces three new commandline options, `-oq` to define an outbound queue, `-oqp` to define a prefix to use when generating the queue key, and `-oqc` to specify the location of a custom queue class.
- Changed commandline validation to require either `-ot` or `-oq`, but not both at the same time.
- The `print_banner()` method is updated, to print out information on the queue, and to hide a transport in the case there is no transport.
- Refactored a lot of `print_banner()` into new class, `Banner`, with reusable line generating functions.
- Defined new `RedisOutboundQueue` and its `BaseOutboundQueue` base class type.
- Refactored in `core/conductor.py` to make external vs internal queue save logic clear.
- Handled Redis connection management inside of `transport.start()` and `transport.stop()`.
- Used `msgpack` to serialize before storing in Redis. The message contains three toplevel keys: `headers`, `endpoint`, and `payload`.
- Added optional unit tests for Redis backend. Redis test endpoint can be specified through env var TEST_REDIS_CONFIG.
- Updated run_tests script to proxy TEST_REDIS_CONFIG environment variable to docker.
- Removed a test `test_setup_x()` which seemed to test that a KeyError raised by the outbound transport manager always propagates up the stack. But this manager is now optional.
- Added `aioredis` to requirements.

### DIFF
--- a/scripts/run_tests
+++ b/scripts/run_tests
@@ -35,6 +35,10 @@ if [ "$OSTYPE" == "msys" ]; then
   CONTAINER_RUNTIME="winpty docker"
 fi
 
+if [ ! -z "$TEST_REDIS_CONFIG" ]; then
+  DOCKER_ARGS="$DOCKER_ARGS -e TEST_REDIS_CONFIG=$TEST_REDIS_CONFIG"
+fi
+
 $CONTAINER_RUNTIME run --rm -ti --name aries-cloudagent-runner \
 	-v "$(pwd)/../test-reports:/usr/src/app/test-reports:z" \
 	$DOCKER_ARGS aries-cloudagent-test "$@"
--- a/OutboundQueue.md
+++ b/OutboundQueue.md
@@ -1,0 +1,53 @@
+# Outbound Queues in ACA-py
+
+## Background
+
+By default, messages often stay in ACA-py memory for long periods of time without being delivered. As a result, when the ACA-py Python process is terminated unexpectedly, messages are lost.
+
+But with recent changes, outbound messages can now be sent to a message queue of your choice instead of being delivered by ACA-py. This queue is external to the ACA-py process, and can be configured for the durability requirements you want. This new concept of an "outbound queue" is intended to be an optional replacement to the current ACA-py outbound transport (i.e. option `-ot`, `--outbound-transport`).
+
+If you run an outbound queue, you will also need to run a new service, a delivery agent, to actually deliver the message. See more details below.
+
+## Usage Details
+
+A new set of commandline options have been added to provide a way for users to "opt in" to use of the outbound queue. These new options are as follows:
+
+- `-oq`, `--outbound-queue`: specifies the queue connection details.
+- `-oqp`, `--outbound-queue-prefix`: defines a prefix to use when generating the topic key.
+- `-oqc`, `--outbound-queue-class`: specify the location of a custom queue class.
+
+Only the first, `--outbound-queue`, is required if you would like to opt into the outbound queue to replace `--outbound-transport`. The input for this option takes the form `[protocol]://[host]:[port]`. So for example, if the queue I want to use is Redis, on host `myredis.mydomain.com` using the default port for Redis, the string would be as follows: `redis://myredis.mydomain.com:6379`
+
+The second option, `--outbound-queue-prefix`, specifies the queue topic prefix. The queue topic is generated in the following form: `{prefix}.outbound_transport`. The default value for this commandline option is the value `acapy`, so a queue key of `acapy.outbound_transport` is generated in the case of the default settings. ACA-py will send messages to the queue using this generated key as the topic.
+
+The third option, `--outbound-queue-class`, specifies the queue backend. By default, this is `aries_cloudagent.transport.outbound.queue.redis:RedisOutboundQueue`, which specifies ACA-py's builtin Redis `LIST` backend. Users can define their own class, inheriting from `BaseOutboundQueue`, to implement a queue backend of their choice. This commandline option is the official entrypoint of ACA-py's pluggable queue interface. Developers must specify a Python dotpath to a module importable in the current `PYTHONPATH`, followed by a colon, followed by the name of their custom class.
+
+## Delivery Agent
+
+When using `--outbound-queue` instead of `--outbound-transport`, ACA-py no longer delivers the messages to destinations. Instead, a delivery service ([a prototype can be found here](https://github.com/andrewwhitehead/aca-deliver)) would need to be run. This service should pick up a message from the queue and then deliver that message. 
+
+When running `--outbound-queue`, ACA-py serializes messages to be sent to the queue by using MessagePack. MessagePack is a protocol to serialize content into a compact binary format. ACA-py generates keys in MessagePack as follows:
+- `endpoint` - specifies the endpoint for the message.
+- `headers` - specifies a set of key-value pairs representing message headers.
+- `payload` - the raw binary content of the message.
+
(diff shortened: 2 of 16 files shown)


TimoGlastra:
Awesome! I have a few questions / notes.

> This implementation serves the use case of storing outbound messages in a durable queue, as opposed to the in-memory queue in ACA-py.

I see you focussed on outbound messages. Not to get ahead of the current work, but do you think this would be easily adaptable to also queue incoming messages?

> Is the api_key required to send as part of the queue message?

I think it depends. The api-key is only used for webhook events and ACA-Py stores the api-key, however you would need to be able to retrieve the correct api key again after retrieving the message from the queue. Thinking from a multitenant perspective where context changes based on the wallet you use you would need to add the data required to identify the subwallet it came from to the queued message. 

To manage resources subwallets can be closed if they're inactive for a while. This means you would need to reopen the wallet. In an 'unmanaged' scenario, ACA-Py doesn't store the key to open the wallet, so that'd mean it cannot send the outbound messages until the wallet is reopened again. However this is very specific to how multi-tenancy works.

ianco:
My thoughts inline below.

> **Open Questions**
> 
> * We can define a "key" per queue inside Redis. Different keys would let us effectively manage separate queues. In this case, I have defined the key as a single static string `acapy.outbound_transport`. Should we have multiple keys, one for each destination point for outbound messages? I.e., should there be a single queue per "mediator" or per "wallet" or per "endpoint," etc.? I don't know ACA-py enough to have an opinion about this, but it is an option for us.

I suggest a single queue, unless there is a specific reason for multiple queues.  We can always add support for multiple queues later.

> * I assume we want to support multiple outbound transport queues at once? If so, what should the parser argument syntax be?

Not sure (?) I don't know if we've ever configured aca-py with multiple outbound queues.  It depends on what transports are supported by the agents we're communicating with, we need to match protocols in order to communicate (i.e. http/http, redis/redis, etc)

> * How do I access connection information inside `redis.py` if there is no `InjectionContext` available? Should I add it as a required argument to `BaseOutboundTransport.__init__()`? If not, how should this be managed?

Not sure, do we need to add some meta-data to the message that we post to redis?

> * Should we use `msgpack` instead of `json` to serialize and save the messages in the queue? I'm leaning towards yes but open to comments.

Not sure, @andrewwhitehead ?

> * Is the `api_key` required to send as part of the queue message?

Depends on the target of the message.  With agent-to-agent messages there is no api key.  For webhooks we need to include the api key (as configured) somehow.  Maybe as meta-data?

In general when we send an outbound message to aa queue, there should be another agent (at least one) somewhere monitoring this queue as an inbound queue, or else in the case of web hooks there will be a controller (at least one) monitoring the queue.  Correct?  So however we implement the outbound messaging it must be consistent with how we handle the eventual inbound messaging.  As long as we can successfully send messages from A to B using a redis queue I think we can be flexible on exactly how we do it.

Also re known issues - eventually this should be pluggable, we don't necessarily need to worry about this right away but need to keep this requirement in the issues list.

ianco:
DCO signoff is missing for commits, make sure you "git commit -s ..." (or "--signoff") to include the DC signoff information.

adamsc64:
Rebased onto latest master.

andrewwhitehead:
I've added a sample delivery service here: https://github.com/andrewwhitehead/aca-deliver

It's definitely not as fast as it could be (no multiplexing) and doesn't support websocket delivery yet. The messages would also need to be enqueued with an `endpoint` parameter.

adamsc64:
Requesting review from:
@TimoGlastra @TelegramSam @dbluhm

adamsc64:
@swcurran requests an additional `.md` file, along the lines of `Multitenancy.md` or `Mediation.md`, that describes what's been done.

scripts/run_tests
```diff
@@ -35,6 +35,14 @@ if [ "$OSTYPE" == "msys" ]; then
...
+  DOCKER_ARGS="$DOCKER_ARGS -e TEST_REDIS_CONFIG=$TEST_REDIS_CONFIG"
+fi
+
 $CONTAINER_RUNTIME run --rm -ti --name aries-cloudagent-runner \
 	-v "$(pwd)/../test-reports:/usr/src/app/test-reports:z" \
 	$DOCKER_ARGS aries-cloudagent-test "$@"
+
+$DOCKER run --rm -ti --name aries-cloudagent-runner \
+	-v "$(pwd)/../test-reports:/usr/src/app/test-reports" \
+	$DOCKER_ARGS aries-cloudagent-test "$@"
```
andrewwhitehead, line 48: This addition seems like a possible merge error?

andrewwhitehead:
Thank you @adamsc64, I added one comment. If you could rebase this on the latest branch I think this is ready to go!

scripts/run_tests
adamsc64: 👍

adamsc64:
Resolved the merge error that @andrewwhitehead pointed out in `script/run_tests`. Rebased onto latest `main` branch in upstream. Let me know if you need anything else!

andrewwhitehead (approved):

PR #917, opened 2021-01-19
[multi-tenancy] Documentation

TimoGlastra:
@ianco


ianco (approved):

andrewwhitehead (approved):

Multitenancy.md
```diff
@@ -0,0 +1,226 @@
...
+## Webhooks
+
+### Webhook URLs
+
+ACA-Py makes use of [webhook events](./AdminAPI.md#administration-api-webhooks) to call back to the controller. Multiple webhook targets can be specified, however in multi-tenant mode it may be desirable to specify different webhook targets per wallet.
+
+When creating a wallet `wallet_dispatch_type` be used to specify how webhooks for the wallet should be dispatched. The options are:
+
+- `default`: Dispatch only to webhooks associated with this wallet.
+- `base`:` Dispatch only to webhooks associated with the base wallet.
```
lijiachuan1982, line 142: There is one additional `after :

Multitenancy.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+GET /connections [headers="Authorization: Bearer {token}]
+```
+
+The `Authorization` header is in addition to the Admin API key. So if the `admin-api-key` is enabled (which it should be in production) both the `Authorization` and the `x-api-key` headers should be provided when making calls to a sub wallet. For calls to a base wallet only the `x-api-key` should be provided.
+
+### Getting a token
+
+A token can be obtained in two ways. The first method is the `token` parameter from the response of the create wallet (`POST /multitenancy/wallet`) endpoint. The second option is using the get wallet token endpoint (`POST /multitenancy/wallet/{wallet_id}/token`) endpoint.
```
lijiachuan1982, line 204: duplicate words of `endpoint`

Multitenancy.md
```diff
@@ -0,0 +1,226 @@
...
+
+In umanaged mode the get token endpoint also requires the `wallet_key` parameter to be included in the request body. The wallet key will be included in the so the wallet can be unlocked when making request to the admin API.
```
lijiachuan1982, line 206: `The wallet key will be included in the` ? `so the wallet can be unlocked when marking request to the admin API`

PR #919, opened 2021-01-20
Make RoutingManager and MediationManager accept a Profile instead of a session

andrewwhitehead:


dbluhm:
I'm curious how you would summarize the need for this change. I'm not seeing a huge difference between letting the managers open sessions directly and letting the caller worry about the session management so I'm sure I'm missing something from the bigger picture. I'd like to understand so I know what's going on when I'm fixing the toolbox plugin after applying these changes lol

andrewwhitehead:
@dbluhm These changes are mainly about shortening the length of the sessions for the new backends, which wasn't really considered in the original set of changes. As there is generally a limit on the number of concurrent connections to the database, keeping them open excessively (especially in an async task which may be competing with many other tasks across awaits) can cause latency to increase. I will probably soon add an option to just inject BaseStorage/BaseWallet via the profile when explicit session management isn't needed, as in the previous model.

aries_cloudagent/protocols/present_proof/v1_0/routes.py
```diff
@@ -484,11 +483,12 @@ async def presentation_exchange_list(request: web.BaseRequest):
...
 
     try:
-        records = await V10PresentationExchange.query(
-            session=session,
-            tag_filter=tag_filter,
-            post_filter_positive=post_filter,
-        )
+        async with context.session() as session:
+            records = await V10PresentationExchange.query(
+                session=session,
```
sklump, line 488: where does this session come from, since the delta doesn't create it on line 475?

aries_cloudagent/protocols/present_proof/v1_0/routes.py
```diff
@@ -558,9 +559,10 @@ async def presentation_exchange_credentials_list(request: web.BaseRequest):
     )
 
     try:
-        pres_ex_record = await V10PresentationExchange.retrieve_by_id(
-            session, presentation_exchange_id
-        )
+        async with context.session() as session:
+            pres_ex_record = await V10PresentationExchange.retrieve_by_id(
+                session, presentation_exchange_id
```
sklump, line 564: Similarly, I don't see where the code initializes session here?

sklump (approved):
Immense - I imagine my incoming PR will need some manual merging once it's in main.

PR #928, opened 2021-01-28
[multi-tenancy] add profile resource management

TimoGlastra:
This PR adds a very simple approach to multi-tenant profile management. A `ProfileCache` is added that has a max `capacity` (initialized with 100 for now) and keeps count of references using `sys.getrefcount`. It will close wallets when the capacity is exceeded and the profile is not used at the moment. The ledger implementation uses a custom refcounter using async context managers which may be nice to also use for this scenario. However this implementation allowed to build the cache on top of the profiles without modifying the profile itself.


aries_cloudagent/multitenant/cache.py
```diff
@@ -0,0 +1,94 @@
...
+            # 2 = self.profiles.items()
+            # 3 = profile above
+            # 4 = sys.getrefcount
+            if sys.getrefcount(profile) <= 4:
+                LOGGER.debug(f"closing profile with id {key}")
+                del self.profiles[key]
+                await profile.close()
+
+                if len(self.profiles) <= self.capacity:
+                    break
```
andrewwhitehead, line 40: It seems a bit odd that this will always iterate through the profiles, short-circuiting if there is one to be cleaned up and we are now below max capacity. Also `self.profiles.items()` will increase the refcount of the dict itself but not the entries it contains, as I understand it (and according to my test just now).

TimoGlastra: I'm not that happy with this approach myself. I'll see if I can rework it a bit. 

Do you have a suggestion on how to approach this? Do you think the way ledger handles refcounting will also work for here?

baegjae:
@TimoGlastra I think this feature is essential when hosting multi-tenancy enabled aca-py services. Do you have any plans to complete it?

TimoGlastra:
Yes definitely, good you bring it up. Just need to revisit the implementation and resolve the comment from Andrew

PR #931, opened 2021-01-29
Fix implicit invitations

sklump:
Split did-exchange (rfc23) auto-acceptance of out-of-band requests by those referring to peer DIDs vs. public DIDs, since responses to out-of-band implicit invitations do not refer to an invitation and so acceptance of those requests must be by configuration. Nudged connection (rfc160) protocol analagously for symmetry.

Reduced scope of invitation records as they are for returning via admin API, not for saving, given the above.


andrewwhitehead:
Sorry, created some conflicts.

sklump:
@andrewwhitehead please take a look again? Thanks

andrewwhitehead (approved):

PR #937, opened 2021-02-02
OutOfBand Test Coverage Improvement, Fixes and cleanup

shaangill025:
100% Test Coverage - Done

Following discussion with Andrew, changed pthid for reuse messages to the existing conn_record invitation_msg_id [original] instead of the received OOB invitation_msg_id [previous impl] - Done

Cleanup: reordering of import according to the convention - Done

### DIFF
--- a/aries_cloudagent/protocols/out_of_band/v1_0/manager.py
+++ b/aries_cloudagent/protocols/out_of_band/v1_0/manager.py
@@ -384,7 +384,7 @@ class OutOfBandManager(BaseConnectionManager):
                             session=self._session, key="reuse_msg_state"
                         )
                 except asyncio.TimeoutError:
-                    # If no reuse_accepted or problem_report message was recieved within
+                    # If no reuse_accepted or problem_report message was received within
                     # the 15s timeout then a new connection to be created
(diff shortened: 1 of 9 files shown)


aries_cloudagent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -384,7 +381,7 @@ async def receive_invitation(
                             session=self._session, key="reuse_msg_state"
                         )
                 except asyncio.TimeoutError:
-                    # If no reuse_accepted or problem_report message was recieved within
```
sklump, line 387: Above, near line 80, create_invitation() has two parameters[include_handshake, use_connections] for the handshake protocols. I propose an Enum here. I will create it, push to my repo, and message you.

sklump (changes requested):
As noted below: out-of-band manager.py near line 80, create_invitation() has two parameters[include_handshake, use_connections] for one item (handshake protocols list). I propose an Enum here. I will create it, push to my repo, and message you.

sklump:
Continued in https://github.com/hyperledger/aries-cloudagent-python/pull/939

PR #939, opened 2021-02-03
Out-of-band test coverage improvement fixes, cleanup + handshake

sklump:
Encapsulate handshake protocols into an enum for tighter admin API integration.

99% of this work is @shaangill025: this PR finesses https://github.com/hyperledger/aries-cloudagent-python/pull/937/files#

### DIFF
--- a/aries_cloudagent/protocols/didcomm_prefix.py
+++ b/aries_cloudagent/protocols/didcomm_prefix.py
@@ -49,7 +49,7 @@ class DIDCommPrefix(Enum):
     def qualify_current(slug: str = None) -> str:
         """Qualify input slug with prefix currently in effect and separator."""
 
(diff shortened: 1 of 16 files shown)


aries_cloudagent/protocols/didcomm_prefix.py
```diff
@@ -49,7 +49,7 @@ def qualify_all(cls, messages: dict) -> dict:
     def qualify_current(slug: str = None) -> str:
         """Qualify input slug with prefix currently in effect and separator."""
 
-        return qualify(slug, environ.get("DIDCOMM_PREFIX", DIDCommPrefix.NEW.value))
```
sklump, line 52: (this was a tiny holdover from when we thought the default was going to be the new prefix - one day it will be again)

andrewwhitehead (approved):

PR #949, opened 2021-02-08
Fixes for performance demo w/ revocation enabled

andrewwhitehead:
For compatibility with 0.6 updates

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v2_0/manager.py
+++ b/aries_cloudagent/protocols/issue_credential/v2_0/manager.py
@@ -554,67 +554,64 @@ class V20CredManager:
 
         tails_path = None
         if cred_def["value"].get("revocation"):
-            async with self._profile.session() as session:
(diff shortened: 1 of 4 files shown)


ianco (dismissed):
tested alice/faber working ok

ianco (changes requested):
Alice/faber is ok but I get the following error running `./run_demo performance --revocation`:

`Error: agent Faber has no method handle_issuer_cred_rev to handle webhook on topic issuer_cred_rev`

andrewwhitehead:
@ianco I get that error too, but it doesn't seem to affect the functionality, it's just an ignored webhook.

ianco (approved):
Reported error is expected

ianco:
> @ianco I get that error too, but it doesn't seem to affect the functionality, it's just an ignored webhook.

Yep you're right it didn't affect the results

PR #951, opened 2021-02-08
Magic invitation

sklump:
also, look for auto-accept-invites on didx-manager receiving invitation

### DIFF
--- a/aries_cloudagent/protocols/didexchange/v1_0/manager.py
+++ b/aries_cloudagent/protocols/didexchange/v1_0/manager.py
@@ -272,7 +272,7 @@ class DIDXManager(BaseConnectionManager):
             except StorageNotFoundError:
                 raise DIDXManagerError("No invitation found for pairwise connection")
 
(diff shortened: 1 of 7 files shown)


andrewwhitehead:
@sklump It seems like two 'public' invitations could be created if there's a race between different requests, are we okay with that? I don't think it would create a problem.

sklump:
I think I have got the wrong idea in any case. Although all implicit invitations are on public DIDs, maybe not all inviations on public DIDs are implicit.

Is it possible to create an explicit invitation on a public DID? Re-reading RFC434 again, perhaps it is, in which case aca-py doesn't need a 'magic' invitation to stand in for implicit.

I don't know whether we should adopt this for 0.6.0 as-is but I think I have to work on yet another approach here, for 0.6.0 or to gate to the main branch thereafter.

sklump:
See https://github.com/hyperledger/aries-cloudagent-python/pull/952 for this approach, which is overall cleaner and broader I think.

I never trust magic.

PR #954, opened 2021-02-10
Revised readme to reflect recent updates

swcurran:


victormartinez-work:
I would add a section explaining that aca-py also can be deployed as a mediator or can be "connected" to a mediator by using Aries protocols (request meditation protocol)

andrewwhitehead (approved):

swcurran:
@victormartinez-work -- good idea.  We'll add that shortly.

PR #956, opened 2021-02-11
Add option to emit the new(ish) DIDComm MIME type

andrewwhitehead:


swcurran:
@nodlesh -- once this is merged, can you please run a test where you use this new option "--emit-new-didcomm-mime-type" with ACA-Py in an AATH test run with AF.NET.  It is likely to fail, but I would like it demonstrating the failure with AATH so that we can get that fixed in the community.  I have an issue in Aries RFCs about what to do with this and I can add a note about your test to that.

Thanks!

sklump (approved):

Issue #958, opened 2021-02-12
Storage for 0.6.0 not compatible with 0.5.6


blu3beri:
Connection records in ACA-Py 0.5.6 had an `initiator` prop. This property is removed in 0.6.0 / main branch. Retrieving connections now gives an error about an unknown property `initiator`. I know 0.6.0 introduces breaking changes, but the change of the storage voids all data persisted in previous version of ACA-Py.

So the question is whether it is expected that data persisted in 0.5.6 of ACA-Py is not valid anymore in 0.6.0? 

See error below, thanks!

```
2021-02-12 16:38:23,403 aries_cloudagent.core.dispatcher ERROR Handler error: connections_list
Traceback (most recent call last):
  File "/home/beri/Git/aries-cloudagent-python/aries_cloudagent/protocols/connections/v1_0/routes.py", line 336, in connections_list
    records = await ConnRecord.query(
  File "/home/beri/Git/aries-cloudagent-python/aries_cloudagent/messaging/models/base_record.py", line 295, in query
    result.append(cls.from_storage(record.id, vals))
  File "/home/beri/Git/aries-cloudagent-python/aries_cloudagent/messaging/models/base_record.py", line 111, in from_storage
    return cls(**params)
  File "/home/beri/Git/aries-cloudagent-python/aries_cloudagent/connections/models/conn_record.py", line 187, in __init__
    super().__init__(
TypeError: __init__() got an unexpected keyword argument 'initiator'
2021-02-12 16:38:23,403 aries_cloudagent.admin.server ERROR Handler error with exception: __init__() got an unexpected keyword argument 'initiator'

```

swcurran:
@andrewwhitehead @sklump @shaangill025 -- please take a look at this ASAP.  Is this something to be adjusted before 0.6.0?

Thanks for documenting this.

sklump:
initiator was redundant vs. role, for what it's worth

swcurran:
FYI - @sklump  is looking at restoring the property.  Lots of "fun" -- particularly with marshmallow to get the API defined.

PR #959, opened 2021-02-12
Restore initiator to conn rec nominally

sklump:


sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/958

andrewwhitehead:
It looks like the reported error was going through the `from_storage` class method, which seems to bypass the schema – we probably want to adjust that later but there may be a reason it works that way. I added the `rfc23_state` parameter because it was also causing an issue once I adjusted the test.

andrewwhitehead (approved):

sklump:
I don't understand how rfc23_state could get into the record. It's neither a tag nor a value. It's a bit of a worry.

sklump:
Enforcing the schema through `from_storage()` will wreck a lot of the unit tests as many throwaway records are missing mandatory fields. Doable but not a quick fix.

PR #962, opened 2021-02-13
Adjust include path in sphinx conf.py to see if docstrings will be included on RTDs

swcurran:


swcurran:
Another attempt to get RTDs website generating the docstrings -- working locally, but not with RTD.

@andrewwhitehead

andrewwhitehead (approved):

PR #967, opened 2021-02-16
[Multi-tenancy] Fix incorrect responder profile

baegjae:
After aca-py restarts, there are two routes to load the sub wallet profile.

1. API call
2. inbound message

In case 1, the responder is updated with the loaded profile.
https://github.com/hyperledger/aries-cloudagent-python/blob/982a260ca327e0e9432a45dbb381b7b80c4c8ed2/aries_cloudagent/admin/server.py#L333

However, in case 2, since there is no such process as above, the responder is based on the profile of the base wallet.
We observed the problem that webhook message came to the webhook url of the base wallet, not the webhook url of the sub wallet.

This PR enable to update the responder with the loaded profile in case 2.

### DIFF
--- a/aries_cloudagent/transport/inbound/session.py
+++ b/aries_cloudagent/transport/inbound/session.py
@@ -171,9 +171,15 @@ class InboundSession:
                     self.profile.context, wallet
                 )
 
-                # Update a responder profile with wallet profile
-                responder: AdminResponder = profile.inject(BaseResponder)
-                responder._profile = profile
+
+                base_responder: AdminResponder = profile.inject(BaseResponder)
+            
+                # Create new responder based on base responder
+                responder = AdminResponder(
+                    profile,
+                    base_responder._send,
+                    base_responder._webhook,
+                )
                 profile.context.injector.bind_instance(BaseResponder, responder)
 
                 # overwrite session profile with wallet profile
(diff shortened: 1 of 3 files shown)


aries_cloudagent/transport/inbound/session.py
```diff
@@ -169,6 +171,11 @@ async def handle_relay_context(self, payload_enc: Union[str, bytes]):
                     self.profile.context, wallet
                 )
 
+                # Update a responder profile with wallet profile
+                responder: AdminResponder = profile.inject(BaseResponder)
+                responder._profile = profile
```
TimoGlastra, line 176: Doesn't this also change the profile for the base wallet? As you're not creating a new responder instance but modifying the old one.

I.e. something like this (not tested):

```suggestion

                base_responder: AdminResponder = profile.inject(BaseResponder)
            
                # Create new responder based on base responder
                responder = AdminResponder(
                    profile,
                    base_responder._send,
                    base_responder._webhook,
                )
                responder._profile = profile
```

baegjae:
@TimoGlastra Thanks for the suggestion. I agree it's a safer way. The suggestion applied. Also, I updated a unit test.

TimoGlastra (approved):
LGTM!

andrewwhitehead (approved):

PR #968, opened 2021-02-16
cover easy modules 100%

sklump:
@TimoGlastra is there meant to be a separation between `wallet_webhook_urls` values of `[]` (empty list) vs. `None` (multitenant/admin/routes.py lines 355-370)? Please advise.

### DIFF
--- a/aries_cloudagent/core/tests/test_in_memory.py
+++ b/aries_cloudagent/core/tests/test_in_memory.py
@@ -17,3 +17,7 @@ class TestInMemoryWallet:
 
         assert "InMemoryProfile" in str(profile)
         assert profile.created
+
+    @pytest.mark.asyncio
+    async def test_profile_clear(self):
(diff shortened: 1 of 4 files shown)


TimoGlastra:
I would say no. It influences whether we send messages to the base webhook urls or webhooks specific to that subwallet.

But maybe we are doing too much magic and we should leave it to the controller to manage the `wallet_dispatch_type` and `wallet_webhook_urls`

@shaangill025 you wrote this. How as it intended to work?

shaangill025:
In regards to (multitenant/admin/routes.py 355-370)
There is a separation between `wallet_webhook_urls` values between None and []. None means `wallet_webhook_urls` was not specified and will not be updated in the settings. [] will mean that this sub-wallet has been updated to have no associated webhook targets and the dispatch_type will be updated to "base" so that we dispatch to base_wallet webhook targets [set during agent startup].

andrewwhitehead (approved):

Issue #975, opened 2021-02-18
'their_did' not queryable in wallet after making connection


daidoji:
## Summary: 
Agent A makes a connection after accepting invitation from Agent B.  In Agent A's wallet, `my_did` for the connection_id representing Agent B's pairwise did from that connection shows up with a `null` endpoint when querying the `/wallet/get-did-endpoint` and `their_did` returns 404 from that endpoint.  An example from Alice-Faber demo after making the first connection:

## From Alice
**Note: these results are symmetrical on the Faber agent (ie its `my_did` returns a null endpoint and `their_did` gives a 404)**
`curl -X GET "http://worklaptop:8021/connections" -H "accept: application/json"`
`{
  "results": [
    {
      "connection_id": "611d0f9e-00f7-420b-a949-243d7030451b",
      "routing_state": "none",
      "my_did": "NtWa4QEvG7JKYVWut15VtU",
      "accept": "auto",
      "invitation_key": "HwWaBvdgvNjRxw2cA1e3AWhp2hzsAAkpvkXsFjT3Km2z",
      "their_label": "Alice.Agent",
      "their_did": "3nxsFWa8f9naoQTjLqmJNf",
      "created_at": "2021-02-18 14:26:07.471934Z",
      "their_role": "invitee",
      "updated_at": "2021-02-18 14:28:21.131473Z",
      "invitation_mode": "once",
      "rfc23_state": "completed",
      "state": "active"
    }
  ]
}`

`curl -X GET "http://worklaptop:8021/wallet/get-did-endpoint?did=NtWa4QEvG7JKYVWut15VtU" -H "accept: application/json"`
`{
  "did": "NtWa4QEvG7JKYVWut15VtU",
  "endpoint": null
}`

`curl -X GET "http://worklaptop:8021/wallet/get-did-endpoint?did=3nxsFWa8f9naoQTjLqmJNf" -H "accept: application/json"`
`404: Unknown DID: 3nxsFWa8f9naoQTjLqmJNf.`

## To reproduce
An easy way to reproduce is to run alice-faber demo as far as the first "accept-invitation" step and then query the wallet with the dids listed in the only connection id that exists on the agent.  

## What we'd expect to happen
[...]

PR #976, opened 2021-02-18
query endpoints by connection id

sklump:


sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/975

Note that endpoint query is through connection, not wallet. It takes the connection id and returns my, their endpoints for the connection

andrewwhitehead (approved):

Issue #977, opened 2021-02-18
http connectivity out of the company network behind the proxy


vitalrev:
reopening from issue: https://github.com/hyperledger/aries-cloudagent-python/issues/697

if AcaPy is running in a firm network behind the firm proxy, the proxy configuration is necessary for http connections.

The **aiohttp** library supports proxy configuration:
> by passing `trust_env=True` into `aiohttp.ClientSession` constructor for extracting proxy configuration from `HTTP_PROXY` or `HTTPS_PROXY` environment variables (both are case insensitive):
> ```
> async with aiohttp.ClientSession(trust_env=True) as session:
>     async with session.get("http://python.org") as resp:
>         print(resp.status)
> ```
> Proxy credentials are given from ~/.netrc file if present (see aiohttp.ClientSession for more details).

See: https://docs.aiohttp.org/en/stable/client_advanced.html?highlight=proxy#proxy-support

The problem is, that in AcaPy we can not pass the `trust_env` on to `aiohttp.ClientSession`.

Solution can be as example in transport/outbound/http.py:
```
        session_args["cookie_jar"] = DummyCookieJar()
        session_args["connector"] = self.connector
        session_args["trust_env"] = os.environ.get("TRUST_ENV", False)
        self.client_session = ClientSession(**session_args)
```

vitalrev:
implemented from andrewwhitehead so that trust_env is every time true

PR #980, opened 2021-02-18
read the environment var TRUST_ENV and pass it on to aiohttp.ClientSe…

vitalrev:
The issue is described here: https://github.com/hyperledger/aries-cloudagent-python/issues/977

it's my first PR in AcaPy...  I hope I did not forget anything

### DIFF
--- a/aries_cloudagent/transport/outbound/http.py
+++ b/aries_cloudagent/transport/outbound/http.py
@@ -1,6 +1,7 @@
 """Http outbound transport."""
 
 import logging
+import os
 from typing import Union
 
 from aiohttp import ClientSession, DummyCookieJar, TCPConnector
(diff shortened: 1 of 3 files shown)


vitalrev:
Thanks for the quick response! If you implement/set the `trust_env=True` for `ClientSession(trust_env=True)`, it's fine for me... than we can close this PR :)

vitalrev:
an another problem with proxy is the bug in aiohttp: https://github.com/aio-libs/aiohttp/pull/4445
I tested with patched aiohttp and patched AcaPy 0.5.6 version...

andrewwhitehead:
Thanks, hopefully aiohttp adds that fix into a release soon.

PR #982, opened 2021-02-19
Avoid error during import if marshmallow library is mocked

andrewwhitehead:
for docs generation


swcurran (approved):
Docs now work -- good stuff.

PR #990, opened 2021-02-23
Fix start cmd for Python versions >= 3.9

borancar:
asyncio.Task.all_tasks() was moved to asyncio.all_tasks()

### DIFF
--- a/aries_cloudagent/commands/start.py
+++ b/aries_cloudagent/commands/start.py
@@ -4,6 +4,7 @@ import asyncio
 import functools
 import logging
 import signal
+import sys
 from configargparse import ArgumentParser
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
Thanks! It looks like there is also a reference to `asyncio.Task.current_task` which was removed at the same time. There's a unit test in `aries_cloudagent/commands/tests/test_start.py` that would probably be failing with this PR if our test image wasn't using python 3.6.

borancar:
Let me have a look and check the tests on 3.9.

borancar:
Updated the tests and the code and ran the `test_start.py` locally on Python 3.9.

andrewwhitehead (approved):

Issue #988, opened 2021-02-22
AriesOpenAPIDemo Docs need to be updated


yunxi-zhang:
The doc on this link: https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/AriesOpenAPIDemo.md is outdated. It still shows faber and alice as two agents will set up 1st connection step by step whereas the latest code shows this process can be automated when attributes "--auto-accept-invites" is enabled. Please update this doc.

sklump:
I suspect that you might have omitted the `--no-auto` parameter as the doc prescribes.

Regardless, as of https://github.com/hyperledger/aries-cloudagent-python/pull/992, the error message notes that the connection record is not available in the invitation state, when the (RFC160) responder double-taps the accept-invitation step for a received invitation.

PR #992, opened 2021-02-23
finesse error message for double-tap to receive request

sklump:
Address https://github.com/hyperledger/aries-cloudagent-python/issues/988


andrewwhitehead:
Hm, maybe it would be clearer as something like "No matching pairwise connection was found in state '{ConnRecord.State.INVITATION}'. A previous connection request may have updated the connection state." ?

andrewwhitehead (approved):

PR #1000, opened 2021-02-25
Update to version 0.6.0

andrewwhitehead:


swcurran (approved):
w00t! And it's PR #1000.

Issue #1006, opened 2021-03-02
Ledger keepalive missing


ducadellaforcoletta:
The parameter ledger.keepalive is missing in the argument that can be provided in the start, but is retrieved from settings with a default (5 seconds) value when used. In order to avoid continue open/close ledger, could be usefull to allow at startup a custom value also for this parameter. Maybe --ledger-keepalive

PR #1009, opened 2021-03-04
#1006 Added ledger keepalive startup parameter

ducadellaforcoletta:
Signed-off-by: DucaDellaForcoletta <a.vignapiano@gmail.com>

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -559,6 +559,13 @@ class LedgerGroup(ArgumentGroup):
             This must be set if running in no-ledger mode.  Overrides any\
             specified ledger or genesis configurations.  Default: false.",
         )
+        parser.add_argument(
+            "--ledger-keepalive",
+            default=5,
+            type=ByteSize(min_size=5),
+            env_var="ACAPY_LEDGER_KEEP_ALIVE",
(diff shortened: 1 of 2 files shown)


aries_cloudagent/config/argparse.py
```diff
@@ -559,6 +559,13 @@ def add_arguments(self, parser: ArgumentParser):
             This must be set if running in no-ledger mode.  Overrides any\
             specified ledger or genesis configurations.  Default: false.",
         )
+        parser.add_argument(
+            "--ledger-keepalive",
+            default=5,
+            type=ByteSize(min_size=5),
```
andrewwhitehead, line 565: This looks like a good thing to add, but I don't think the type is ByteSize (just int).

ducadellaforcoletta: I agree with you about the type int. I changed it. 
Pay attention, the ByteSize is used also for "max-outbound-retry" params, I suppose for the same reason I used it: the lower and upper bound limit embedded check. Could an utility class with the same embedded check for the int type could be helpful for the parser? Maybe IntegerValue could be added.

andrewwhitehead (approved):
I agree we want to add another type for restricting the range of integers, I'll add something for that soon. Thanks!

PR #1016, opened 2021-03-09
Correction to out_of_band protocol v1_0 in manager.py

konsmosc:
In line 532 of aries_cloudagent/protocols/out_of_band/v1_0/manage.py the initialization 
of PresentationManager should take Profile as an argument and not ProfileSession.

The error bellow occurs when the receive_invitation is called:
_AttributeError: 'IndySdkProfileSession' object has no attribute 'session'_

In order to avoid the above error, this line(532) **proof_present_mgr = PresentationManager(self._session)**  
must be replaced with this **proof_present_mgr = PresentationManager(self._session.profile)**.

As a result, with this change the above error disappears.


andrewwhitehead (approved):

andrewwhitehead:
Thanks!

PR #1018, opened 2021-03-09
Initial support for VC storage and retrieval

andrewwhitehead:
- Includes in-memory and Indy-SDK storage implementations
- Askar implementation: can take advantage of multi-value tag storage, will be added later on shared-components branch

### DIFF
--- a/aries_cloudagent/holder/in_memory.py
+++ b/aries_cloudagent/holder/in_memory.py
@@ -81,6 +81,7 @@ class InMemoryVCHolder(VCHolder):
             cred: The VCRecord instance to store
         Raises:
             StorageDuplicateError: If the record_id is not unique
+
         """
         record = serialize_credential(cred)
         await self._store.add_record(record)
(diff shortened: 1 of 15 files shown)


TimoGlastra:
This is perfect. Thanks Andrew.

I don't see any tags for the credential values. Is the idea to retrieve all credentials that match according to issuer/schema/... and then to the filtering outside of storage?

aries_cloudagent/holder/in_memory.py
```diff
@@ -0,0 +1,189 @@
...
+VC_CRED_RECORD_TYPE = "vc_cred"
+
+
+def load_credential(record: StorageRecord) -> VCRecord:
+    """Convert an in-memory stored record into a VC record."""
+    tags = {}
+    contexts = []
+    types = []
+    subject_ids = []
+    schema_id = None
```
TimoGlastra, line 23: A credential can have multiple schemas. Probably not something we're going to use atm, but maybe good to allow it in the API

andrewwhitehead:
@TimoGlastra For the attribute values, I'm currently leaving it up to the caller to set those as tags (probably in the form `attr:{name} = value`.

andrewwhitehead:
While attribute tags could be handled more explicitly (probably setting an attribs dict on the record), this would require rewriting the tag query to add the required prefixes, and then there might be a need for two tag queries (one for attribs and one for custom tags). I'm happy to wait and see what changes might be needed there later.

sklump (approved):

PR #1033, opened 2021-03-17
Pluggable DID Resolver Interface

dbluhm:
This PR is a work in progress.

This PR adds interfaces to enable plugging in one or more DID resolvers. This enables ACA-Py to resolve DIDs from any VDR as long as there is a resolver plugin supporting it. By using a plugin that connects to a universal resolver instance, ACA-Py can inherit support for all methods supported by the universal resolver (with security and trust implications as a result of DIDs being resolved remotely).


TimoGlastra:
This looks really good! Nice :)

I wrote a `DIDKey` implementation for ACA-Py: https://github.com/animo/aries-cloudagent-python/blob/cwu-bbs-main/aries_cloudagent/did/did_key.py

Integrating this with the resolver should be as simple as:

```python
did = "did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"
did_doc = DIDKey.from_did(did).did_doc
```

So once the two codebases are merged we can add the resolver. For now it just returns a dict, but it may nice to utilize the `DIDDoc` class you wrote to construct the did document.

---

I thought maybe we could host both these things under a root `did` directory. So the `resolver` directory will be part of that.

```
aries_cloudagent/
  did/
    resolver/
      other_resolver_stuff.py
    did_key.py
```

OR I could host the `DIDKey` class in the resolver directory.

dbluhm:
Closing in favor of #1070

PR #1038, opened 2021-03-18
Endorse txn unit tests

sklump:
Suspect: magic values as of August 2020 for timeouts in `messaging/{schemas,credential_definitions}/routes.py` for transaction record expiry. What are these for? Do we mean to create stale records? If so, zero would be clearer here? Please advise.


ianco:
> Suspect: magic values as of August 2020 for timeouts in `messaging/{schemas,credential_definitions}/routes.py` for transaction record expiry. What are these for? Do we mean to create stale records? If so, zero would be clearer here? Please advise.

Good catch, I don't think this attribute is explicitly used anywhere, but should be a date I think:

`protocols/endorse_transaction/v1_0/models/transaction_record.py:166:        example={"expires_time": "2020-12-13T17:29:06+0000"}, required=False`

I suggest to either leave out the value, or else set to `today + 24 hours`

ianco (approved):

PR #1045, opened 2021-03-22
tentative admin plumbing for DIF cred get/list/delete

sklump:
For consideration. I think we may be able to consolidate holder/routes.py into fewer endpoints later, once revocation is in place for DIF credentials, but this allow admin API access for the current state of affairs.


aries_cloudagent/holder/routes.py
```diff
@@ -274,6 +329,114 @@ async def credentials_list(request: web.BaseRequest):
...
+        Verifiable credential record
+
+    """
+    context: AdminRequestContext = request["context"]
+    credential_id = request.match_info["credential_id"]
+
+    session = await context.session()
+    holder = session.inject(VCHolder)
+    try:
+        vc_record = await holder.retrieve_credential_by_given_id(credential_id)
```
TimoGlastra, line 355: This fetches the credential by its given id. This is an optional field, and the `id` property in the credential itself. I think `retrieve_credential_by_id` should be used here, and that `given_id` becomes a query parameter for the query credentials.

TimoGlastra (commented):
Nice @sklump! I left a few comments / questions

sklump:
OK, stay tuned, I will amend tomorrow morning early.

aries_cloudagent/holder/routes.py
```diff
@@ -292,6 +455,13 @@ async def register(app: web.Application):
             ),
             web.delete("/credential/{credential_id}", credentials_remove),
             web.get("/credentials", credentials_list, allow_head=False),
+            web.get(
+                "/credential/dif/{credential_id}",
+                dif_cred_get,
+                allow_head=False,
+            ),
+            web.delete("/credential/dif/{credential_id}", dif_cred_remove),
+            web.get("/credentials/dif", dif_creds_list, allow_head=False),
```
TimoGlastra, line 464: `TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.`

aries_cloudagent/holder/routes.py
```diff
@@ -274,6 +329,114 @@ async def credentials_list(request: web.BaseRequest):
...
+        search = holder.search_credentials(
+            contexts, types, schema_ids, issuer_id, subject_id, tag_query
+        )
+        records = await search.fetch(max_results)
+    except StorageNotFoundError as err:
+        raise web.HTTPNotFound(reason=err.roll_up) from err
+    except StorageError as err:
+        raise web.HTTPBadRequest(reason=err.roll_up) from err
+
+    return web.json_response({"results": records})
```
TimoGlastra, line 437: `TypeError: Object of type 'VCRecord' is not JSON serializable`

```suggestion
        search = holder.search_credentials(
            contexts, types, schema_ids, issuer_id, subject_id, tag_query
        )
        records = await search.fetch(max_results)
        results = [record.serialize() for record in records]
    except StorageNotFoundError as err:
        raise web.HTTPNotFound(reason=err.roll_up) from err
    except StorageError as err:
        raise web.HTTPBadRequest(reason=err.roll_up) from err

    return web.json_response({"results": results})
```

aries_cloudagent/storage/vc_holder/vc_record.py
```diff
@@ -52,6 +78,50 @@ def __eq__(self, other: object) -> bool:
...
+        description="Issuer identifier",
+        example=DecentralizedId.EXAMPLE,
+    )
+    subject_ids = fields.List(
+        fields.Str(
+            description="Subject identifier",
+            example="did:example:ebfeb1f712ebc6f1c276e12ec21",
         )
+    )
+    value_json = fields.Str(description="(JSON) credential value", **JSON_DUMP)
```
TimoGlastra, line 118: I think it would be nicer to return this as valid JSON instead of a JSON string

TimoGlastra (commented):
Perfect!

andrewwhitehead (approved):

PR #1046, opened 2021-03-22
Add GHA workflow for aca-py integration tests

ianco:
Workflow can run on PR (hyperledger repo) or manually.  Runs a specific sub-set of bdd tests since some tests cause test execution to "hang" under GHA:

`./run_bdd` - runs all the tests, can be run manually
`./run_bdd -t @GHA` - runs a sub-set of tests - should be sufficient to smoke test existing functionality

Workflow is setup to run its own indy ledger and tails server, but can be run with external ledger/tails server using environment settings.

"Actions" are re-factored the same way as they are setup under AATH.


andrewwhitehead:
Looks good, I think we may want to merge some open PRs first before rebasing this, in case the new tests trip them up

ianco:
> Looks good, I think we may want to merge some open PRs first before rebasing this, in case the new tests trip them up

There isn't any reason to wait to merge this PR.  If the GHA fails it won't (yet) block any of the PR's, and it would be good to get this merged so if the GHA *does* fail then it will be visible.

andrewwhitehead (approved):

PR #1056, opened 2021-03-26
Small AIP-2 updates

andrewwhitehead:
- Remove `--emit-new-didcomm-prefix` and `--emit-new-didcomm-mime-type` in favour of `--aip-version` (1 or 2). I believe the .NET framework will currently only accept the old MIME type.
- Change the DID document service type to `did-communication` when using the `did-exchange` protocol.
- Remove DID shortening from DIDDoc and related classes.
- Do not accept DID documents without an `id` property.
- Qualify connection DIDs with `did:peer:` (does NOT yet qualify other DIDs with `did:sov:`).
- Updates the `INDY_DID` pattern to accept `did:sov:` or `did:peer:`, which isn't technically correct. We should be using a generic DID pattern in most places.
- Should permissively accept DID documents in the current v1.0 format, or in the older format (which I'm calling v0). When using the `did-exchange` protocol, the new format (with `verificationMethod`) is used for serialization regardless of the AIP version setting. When using the `connections` protocol, the old format (with `publicKey`) is used.

These changes might be better addressed though the generic resolver PR depending on the timeline, but this is for testing compatibility with other agents.

### DIFF
--- a/aries_cloudagent/connections/models/diddoc/diddoc.py
+++ b/aries_cloudagent/connections/models/diddoc/diddoc.py
@@ -30,6 +30,11 @@ from .util import canon_ref
 LOGGER = logging.getLogger(__name__)
 
 
+def parseError(msg: str):
+    LOGGER.debug(msg)
+    raise ValueError(msg)
+
+
 class DIDDoc:
     """
     DID document, grouping a DID with verification keys and services.
@@ -38,7 +43,8 @@ class DIDDoc:
     everything else as URIs (oriented toward W3C-facing operations).
     """
 
-    CONTEXT = "https://w3id.org/did/v1"
+    CONTEXT_V0 = "https://w3id.org/did/v1"
+    CONTEXT_V1 = "https://www.w3.org/ns/did/v1"
 
     def __init__(self, did: str = None) -> None:
         """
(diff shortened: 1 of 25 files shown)


aries_cloudagent/connections/models/diddoc/diddoc.py
```diff
@@ -228,61 +252,53 @@ def deserialize(cls, did_doc: dict) -> "DIDDoc":
...
+                if key.id in rv.pubkey:
+                    _parseError(f"duplicate key id: {key.id}")
                 rv.pubkey[key.id] = key
 
+        pubkeys = did_doc.get("verificationMethod", did_doc.get("publicKey")) or {}
+        for (
+            pubkey
+        ) in pubkeys:  # include all public keys, authentication pubkeys by reference
+            pubkey_type = PublicKeyType.get(pubkey["type"])
+            key = PublicKey(  # initialization canonicalizes id
```
sklump, line 288: could lose the comment now: it is an orphan

aries_cloudagent/connections/models/diddoc/util.py
```diff
@@ -74,27 +50,13 @@ def canon_ref(did: str, ref: str, delimiter: str = None):
...
-
-    if urlparse(ref).scheme:  # e.g., https://example.com/messages/8377464
+    if (
+        ref.startswith("did:")
+        or urlparse(ref).scheme  # e.g., https://example.com/messages/8377464
+    ):
         return ref
 
-    return "did:sov:{}{}{}".format(did, delimiter if delimiter else "#", ref)  # e.g., 3
+    return "{}{}{}".format(did, delimiter if delimiter else "#", ref)
```
sklump, line 59: return f"{did}{delimiter or '#'}{ref}" is slightly briefer but it is not so important

aries_cloudagent/messaging/valid.py
```diff
@@ -194,7 +194,7 @@ class IndyDID(Regexp):
     """Validate value against indy DID."""
 
     EXAMPLE = "WgWxqztrNooG92RXvxSTWv"
-    PATTERN = rf"^(did:sov:)?[{B58}]{{21,22}}$"
+    PATTERN = rf"^(did:sov:|did:peer:)?[{B58}]{{21,22}}$"
```
sklump, line 197: I thought did:peer used multibase encoding? I could be wrong?

sklump (dismissed):

dbluhm:
Indeed, there is some overlap here with #1033. We're aiming for a more "complete" solution that makes it easy to work DIDs of various methods and DID Documents in various formats; there will be some code conflicts to resolve but I think the Document version distinction introduced in this PR maps well onto our goals. Especially if this PR is already ready to go, I think merging this PR now (or when ready) and incorporating/superseding the changes with #1033 is probably the way to go.

aries_cloudagent/messaging/valid.py
sklump, line 197: https://identity.foundation/peer-did-method-spec/index.html#matching-regex

Not sure whether indy-sdk would take these though

sklump: but note that it's bad regex; the trailing $) characters appear twice. Better:
^did:peer:[01](z)([1-9a-km-zA-HJ-NP-Z]{46,47})$

Issue #1005, opened 2021-03-02
Aries OpenAPI Demo Errors 


mobinauten:
Following the Demo Script I have following issues:

1. Issuing a credential I receive the following error - "Error: agent Faber.Agent has no method handle_issue_credential to handle webhook on topic issue_credential"
2. Trying to find the Public DID on the ledger http://dev.greenlight.bcovrin.vonx.io. returns Null 

Thanks 
Olli

mobinauten:
This is Alice side's Error: agent Alice.Agent has no method handle_issue_credential to handle webhook on topic issue_credential

sklump:
The code now uses v2 of these protocols: I will update the docs accordingly

PR #1060, opened 2021-03-31
AriesOpenAPI, Acme, Alice Phone demos: update text pics & code. Check…

sklump:
…uto-issue before issuing in demo webhooks. Also accommodate multiple detail records on same cred ex id in cred ex record deletion -- which should never happen but which controllers could erroneously produce.


sklump:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/1005

ianco (approved):

Issue #950, opened 2021-02-08
Webhooks for mediation protocol


genaris:
When running an ACA-Py instance specifying a webhook-url, I can get all the webhooks related to all steps in connection (/topic/connections) and basic messaging protocols (/topic/basicmessaging) among other things. However, that is not the case for operations related to mediation, where I need to poll through the admin API to look for updates on them.

For a controller it can be of interest to receive notifications about mediation requests, and then, depending on the business logic, use admin API to grant or deny such requests. And the same could apply for the reception of these responses from a mediator, without the need to poll for mediation request status.

If this feature is already present, is there a particular configuration to enable it?

dbluhm:
This feature is not yet present but should be a relatively simple addition. I definitely agree that this would be useful if not an outright requirement -- it was on our (Indicio/SICPA) to-do list but it got bumped down the list for the time being. I'd be happy to do a bit of coaching through a PR if you (or anyone else who could use this feature) has the drive to implement it.

genaris:
Thanks, Daniel! Sure, this is something that we'll definitely need, so I think I can manage to work on that soon.

airoasis:
@genaris Hi. Have you started to implement this feature already?

TimoGlastra:
This is somewhat related to this issue, so posting here

We want to use ACA-Py as a mediator for mobile agents, and with it we also want to notify users of undelivered messages through push notifications. The idea is to add a custom DIDComm message to notify ACA-Py of device information required for push notifications (as implemented in [.NET](https://github.com/hyperledger/aries-framework-dotnet/blob/b9824699fab00386f0e757a036ee0349ef58fae0/src/Hyperledger.Aries.Routing/AddDeviceInfoMessage.cs#L12)).

We'll need to be able to notify the controller of received `forward` messages, so the controller can in turn ping the push notification service. However I'd like to include in the `forward` webhook event, whether the message could be delivered directly (so we don't send push notifications for already processed messages).

@dbluhm Do you think this would be straightforward to add in ACA-Py? Seems like it may be hard to do with where this information is produced (in the transport layer)

We could just dismiss the notification on the mobile side, however I'm interested if we could handle it on the server side.

dbluhm:
Yeah, I think this would be tricky at the moment. We would need feedback from the responder send methods which itself is probably going to be tricky due to the number of layers that get called between the responder and actually attempting to post a message.

[...]

PR #1063, opened 2021-04-05
Feature/event bus

farskipper:
This adds an event bus and utilizes it to simplify webhooks. Events are emitted and the Admin Server subscribes to some of them to be forwarded as webhooks. Here is a diagram outlining the refactor: https://hackmd.io/IoDqYR_cS3yRAe7DI8JAiA

Events are namespaced. `acapy::*` events are anything originating from ACA-Py this way plugins can use their own events without collision.

#### Responder.send_webhook

Is deprecated, it emits a warning and uses the `acapy::webhook::{TOPIC}` event to retain backwards compatibility.

#### BaseRecord webhooks

Instead of sending a webhook, it now emits an event. As such, `WEBHOOK_TOPIC` -> `RECORD_TOPIC` and `send_webhook` -> `emit_event`

The record event topic follows this pattern: `acapy::record::{RECORD_TOPIC}::{STATE}`

#### Removed Admin Server webhook_targets

It appears that the admin server webhook_targets are no longer used. See: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/admin/server.py#L754-L755 Instead we just rely on `profile.settings.get("admin.webhook_urls")`

As such, the following are removed:

- BaseAdminServer.add_webhook_target
- BaseAdminServer.remove_webhook_target
- BaseAdminServer.send_webhook


swcurran:
Something for @shaangill025 @andrewwhitehead and @TimoGlastra  to look at it.  And anyone else working on ACA-Py, of course!

Thanks!

airoasis:
Is it resolving the issue https://github.com/hyperledger/aries-cloudagent-python/issues/950 as well?

ianco:
Are there any changes to aca-py startup parameters, or any changes to how the controller interacts with aca-py?

ianco:
I ran a quick test of the Alice/Faber demo (startup the 2 agents, establish connection, issue credential from faber -> alice and then ask for proof request), it looks like Alice doesn't receive the proof request on the last step.

dbluhm:
The goal is for the webhook listeners and controller to operate exactly as before. @farskipper we'll have to check out the issues with the demo.

dbluhm:
> Is it resolving the issue #950 as well?

@airoasis This does not impact #950. This change just adds a layer between ACA-Py and the Webhook emitter that enables plugins to process events without needing to run a whole webserver to listen for webhooks.

farskipper:
> Are there any changes to aca-py startup parameters, or any changes to how the controller interacts with aca-py?

There shouldn't be.

farskipper:
@ianco I updated the branch and made a small fix. Now I'm able to get through the Alice/Faber demo.

andrewwhitehead:
This looks great to me so far, and running the `performance` demo it seems to be performing 20-30% faster running native on my Mac (it takes longer to finish starting the credential exchanges, but finishes sooner). @ianco Are you able to test the stability on openshift?

ianco:
> This looks great to me so far, and running the `performance` demo it seems to be performing 20-30% faster running native on my Mac (it takes longer to finish starting the credential exchanges, but finishes sooner). @ianco Are you able to test the stability on openshift?

We need to build a docker image in order to deploy/test on openshift, prob worth doing once we merge this and any other big PR's (for example I'd like to get the shared-components onto openshift as well).

ianco:
Overall looks really good.  Is the EventBus pluggable?  Seems like a good fit for the persistent queue.

dbluhm:
> Overall looks really good.  Is the EventBus pluggable?  Seems like a good fit for the persistent queue.

Interesting question... The current implementation aims to be simple so that components within ACA-Py and attached plugins can receive the same events originally sent only by the admin server via web hooks. Using a plugin, you can subscribe to events and then push those to an external queue or notification system. I'm not sure I see what you mean about using it with the persistent queue yet.

ianco (approved):

aries_cloudagent/core/event_bus.py
```diff
@@ -0,0 +1,119 @@
...
+                index = self.topic_patterns_to_subscribers[pattern].index(processor)
+            except ValueError:
+                return
+            del self.topic_patterns_to_subscribers[pattern][index]
+            if not self.topic_patterns_to_subscribers[pattern]:
+                del self.topic_patterns_to_subscribers[pattern]
+            LOGGER.debug("Unsubscribed: topic %s, processor %s", pattern, processor)
+
+
+class MockEventBus(EventBus):
```
Luis-GA, line 109: It is a test class for test purposes that only is used on unit tests.
This shouldn't be declared here, isn't it?

dbluhm: This follows a pattern used elsewhere in ACA-Py so I don't see a problem with defining it here. This makes it easier to reuse in tests.

andrewwhitehead (approved):

PR #1069, opened 2021-04-06
make max client request size configurable

sklump:

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -99,54 +99,73 @@ class AdminGroup(ArgumentGroup):
             nargs=2,
             metavar=("<host>", "<port>"),
             env_var="ACAPY_ADMIN",
-            help="Specify the host and port on which to run the administrative server.\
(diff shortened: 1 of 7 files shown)


aries_cloudagent/config/argparse.py
```diff
@@ -99,54 +99,73 @@ def add_arguments(self, parser: ArgumentParser):
             nargs=2,
             metavar=("<host>", "<port>"),
             env_var="ACAPY_ADMIN",
-            help="Specify the host and port on which to run the administrative server.\
-            If not provided, no admin server is made available.",
+            help=(
```
andrewwhitehead, line 102: DId you reformat all these descriptions or did Black do it?

sklump: I did it. Black accepts them either way, but everywhere else in the code base they look like the style to which I've adapted them here.

sklump: Let me just merge main again though.

andrewwhitehead (approved):

PR #1070, opened 2021-04-07
Pluggable DID Resolver Interface (cleaned)

dbluhm:
This PR cleans up #1033. The change set was too large to really reason about so I've isolated changes to just the addition of the resolver and will follow up with further integration.

This PR adds interfaces to enable plugging in one or more DID resolvers. This enables ACA-Py to resolve DIDs from any VDR as long as there is a resolver plugin supporting it and the plugin is registered. By using a plugin that connects to a universal resolver instance, ACA-Py can inherit support for all methods supported by the universal resolver (with security and trust implications as a result of DIDs being resolved remotely).


dbluhm:
>I thought maybe we could host both these things under a root did directory. So the resolver directory will be part of that.
>
> ```
> aries_cloudagent/
>   did/
>     resolver/
>       other_resolver_stuff.py
>     did_key.py
> ```
> OR I could host the DIDKey class in the resolver directory.

To @TimoGlastra's previous comment, I think moving to the `did` package will make sense

dbluhm:
Worth noting that this PR sidesteps potential conflicts with #1056.

andrewwhitehead:
Looks good to me, but I can't update the branch.

andrewwhitehead (approved):

PR #1071, opened 2021-04-07
Prover remove superfluous timestamp

sklump:


aries_cloudagent/protocols/present_proof/v1_0/manager.py
```diff
@@ -294,6 +294,21 @@ async def create_presentation(
                     await holder.get_credential(credential_id)
                 )
 
+        for cred_id in credentials:
+            if not credentials[credential_id].get("rev_reg_id"):
```
andrewwhitehead, line 298: `credential_id` or `cred_id`?

sklump: actually it's much worse than that - I don't know how this could possibly have worked

andrewwhitehead (approved):

PR #1073, opened 2021-04-08
query config from admin API

sklump:

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -559,19 +559,7 @@ class AdminServer(BaseAdminServer):
                 config["admin.webhook_urls"][index],
             )
         
-        return web.json_response(
-            {
-                k: self.context.settings[k]
-                for k in self.context.settings
-                if k not in [
-                    "admin.admin_api_key",
-                    "wallet.key",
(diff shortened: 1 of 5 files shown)


aries_cloudagent/admin/server.py
```diff
@@ -523,6 +524,27 @@ async def plugins_handler(self, request: web.BaseRequest):
...
+
+        Returns:
+            The web response
+
+        """
+        return web.json_response(
+            {
+                k: self.context.settings[k]
+                for k in self.context.settings
+                if k != "admin.admin_api_key"
```
ianco, line 544: Other parameters that should be filtered:  wallet key, wallet auth attributes (if using postgres), web hooks (if they contain api keys), wallet seed

ianco (changes requested):

aries_cloudagent/admin/server.py
sklump: great, I'll amend accordingly

sklump: Are wallet auth attributes `wallet.storage.creds` ? Are there others?

ianco (approved):

PR #1077, opened 2021-04-09
Use DID Resolver to resolve public DIDs in invitations

dbluhm:
This PR uses the DID Resolver to resolve connection info from a public DID received in an invitation.

Credit to @Luis-GA for his awesome work on this feature.


sklump (approved):
Looks OK to me

andrewwhitehead (approved):

andrewwhitehead:
Tested the performance demo, and there doesn't seem to be any impact (it doesn't measure the time for forming a connection, anyway). I still have python 3.6.8 for testing so I had to add `--ignore-requires-python` to install pydid, but it works happily.

Issue #1078, opened 2021-04-09
Bucket ticket to collect and fix discrepancies in the swagger doc


etschelp:
There are many smaller and bigger issues in the swagger.json that prevents users from automatically building clients in their respective programing language. The idea is to collect these issues in the comments.

etschelp:
`/present-proof/records/{pres_ex_id}/credentials`

schema_id, cred_def_id, rev_reg_id, cred_rev_id are missing from cred_info instead they are one level above

example model: 
```
[
  {
    "cred_def_id": "WgWxqztrNooG92RXvxSTWv:3:CL:20:tag",
    "cred_info": {
      "attrs": {
        "additionalProp1": "24"
      },
      "referent": "3fa85f64-5717-4562-b3fc-2c963f66afa6"
    },
    "cred_rev": "12345",
    "interval": {
      "to": 1617202025
    },
    "presentation_referents": [
      "1_age_uuid"
    ],
    "rev_reg_id": "WgWxqztrNooG92RXvxSTWv:4:WgWxqztrNooG92RXvxSTWv:3:CL:20:tag:CL_ACCUM:0",
    "schema_id": "WgWxqztrNooG92RXvxSTWv:2:schema_name:1.0"
  }
]
```

actual result:

```
[
  {
    "cred_info": {
      "referent": "fcc897b0-8af7-4138-aac4-17560a05178a",
      "attrs": {
        "bic": "111",
        "iban": "111"
      },
      "schema_id": "M6Mbe3qx7vB4wpZF4sBRjt:2:bank_account:1.0",
      "cred_def_id": "M6Mbe3qx7vB4wpZF4sBRjt:3:CL:571:Bank Account V2",
      "rev_reg_id": null,
      "cred_rev_id": null
    },
    "interval": null,
    "presentation_referents": [
      "referent"
    ]
  }
]
```

sklump:
I can take a look today and into early next week.

swcurran:
@etschelp -- is this issue resolved with #1080?

etschelp:
Yes that one is fixed, so I will close the ticket. As discussed in the [channel](https://chat.hyperledger.org/channel/aries-cloudagent-python?msg=jJEW6niX454TKMbcA) there are some open issues with missing type information, but there is no easy way fixing them at the moment.

PR #1080, opened 2021-04-09
Admin API data structure vs. schema audit

sklump:
Progress: https://github.com/hyperledger/aries-cloudagent-python/issues/1078


sklump:
Stay tuned, I should be able to finish this off in a few hours.

sklump:
@ianco - note that send-schema, send-cred-def results accommodate endorsement protocol, but have an additional layer for "sent" item vs. "txn" to endorse.

andrewwhitehead (approved):

PR #1085, opened 2021-04-13
implement basic ECDH-1PU crypto functionality according to RFC 587

Windforcer:
This PR is a WIP implementing RFC 587 with the goal of updating ACA-Py Encryption Envelope to achieve Aries agent interop with AFGo. Specifically, this PR implements the ECDH-1PU algorithm, built using the AFGo codebase as a reference (https://github.com/hyperledger/aries-framework-go/blob/main/pkg/crypto/tinkcrypto/wrap_support.go#L71) and using the python-ecdsa library.

Further work includes adding more test cases and implementing the OOB header to indicate which version of the encryption envelope is being used.

Any feedback from the ACA-Py team is appreciated!


swcurran:
@ianco @andrewwhitehead -- heads up on this.

@Windforcer -- some of the commits you have provided have not be signed per DCO. Please see the details in the checks above.

andrewwhitehead:
Thanks, this looks like a good start. I'm also adding EDCH-1PU and ECDH-ES to [aries-askar](https://github.com/hyperledger/aries-askar) which will provide the new wallet back-end. Something like this will likely be needed for the in-memory wallet, which is used to test the protocols without depending on Indy-SDK (or Askar).

The 'otherinfo' used in the ConcatKDFHash is not correct. It would be something like `len_u32(alg) + alg + len_u32(apu) + apu + len_u32(apv) + apv + len_u32(256)` where `len_u32(b: bytes) = len(b).to_bytes(4, "big")`.

Windforcer:
Thanks for the feedback :)

DCO is now passing.

@andrewwhitehead Just want to confirm, we're not duplicating work right? ECDH-1PU is needed in both aries-askar and ACA-Py? 
Also thank you for the correction in ConcatKDF, I will fix soon.

andrewwhitehead:
I don't think there's any code duplication, it's similar to what we have with the current 'pack' algorithm which is implemented [here](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/wallet/crypto.py) for the in-memory wallet, but also in Indy-SDK and in Askar.

Windforcer:
ECDH-1PU functionality is implemented. Please see `Test_Appendix_Example()` in `test_1pu.py` for an example key generation workflow. The test can be compared with [this](https://tools.ietf.org/id/draft-madden-jose-ecdh-1pu-03.html#rfc.appendix.A) as proof the implementation is correct and conforms to the ECDH-1PU spec.

Additional feedback is always appreciated. @andrewwhitehead @swcurran 
Special thanks to Andrew Whitehead for continued guidance throughout the development process.

andrewwhitehead:
Thanks! Are you able to fix the conflict on requirements.txt, maybe rebasing the branch on `main`?

I added one comment but I can also clean that up after this is merged if you're out of time.

Windforcer:
I can add these fixes now, no problem

Windforcer:
@andrewwhitehead 

Both fixes are added :)

andrewwhitehead (approved):

PR #1087, opened 2021-04-14
Relax formats rfcs 453 454

sklump:
Ignore unrecognized formats rather than rejecting their messages


TimoGlastra (commented):
Nice!

Small note, I think the `attachment()` methods (e.g. in cred_request.py) should also be modified as the `V20CredFormat.Format.get` could return None

sklump:
I should get to it later today. I will include it here and ping you again @TimoGlastra before I merge it

sklump:
@TimoGlastra how about now? I think this is better overall - in any list of formats, getting an attachment without specifying a format picks the first recognized rather than trying what's at index zero.

TimoGlastra (commented):
Perfect!

andrewwhitehead (approved):

PR #1088, opened 2021-04-14
demo uses v2 protocols, not v1

sklump:


sklump:
@ianco any idea why the integration tests would fail? Is there some magic hook on the other side that needs 0036, 0037 instead of 0453, 0454?

sklump:
Oh - I see there is more. I will get to it today.

sklump:
OK @ianco, good to go now. Thanks for your patience.

ianco:
> OK @ianco, good to go now. Thanks for your patience.

Glad I could help :-D

ianco (approved):

PR #1092, opened 2021-04-15
Py did vs str did

sklump:
Note that throughout aca-py, `did` is typically a string. This gave rise to some ambiguity in passing public (PyDID) DID objects into what expected a string.

Also refactored the DIDInfo duplication away.

### DIFF
--- a/aries_cloudagent/resolver/base.py
+++ b/aries_cloudagent/resolver/base.py
@@ -61,19 +61,17 @@ class BaseDIDResolver(ABC):
 
     async def resolve(self, profile: Profile, did: Union[str, DID]) -> DIDDocument:
         """Resolve a DID using this resolver."""
-        if isinstance(did, str):
-            did = DID(did)
+        py_did = DID(did) if isinstance(did, str) else did
 
-        if not self.supports(did.method):
+        if not self.supports(py_did.method):
             raise DIDMethodNotSupported(
-                f"{self.__class__.__name__} does not support DID method {did.method}"
+                f"{self.__class__.__name__} does not support DID method {py_did.method}"
             )
 
-        did = str(did)
-        did_document = await self._resolve(profile, did)
+        did_document = await self._resolve(profile, str(py_did))
         result = DIDDocument.deserialize(did_document)
         return result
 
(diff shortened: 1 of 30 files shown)


aries_cloudagent/resolver/base.py
```diff
@@ -61,19 +61,17 @@ def supports(self, method: str) -> bool:
...
 
-        did = str(did)
-        did_document = await self._resolve(profile, did)
+        did_document = await self._resolve(profile, str(py_did))
         result = DIDDocument.deserialize(did_document)
         return result
 
     @abstractmethod
-    async def _resolve(self, profile: Profile, did: DID) -> dict:
+    async def _resolve(self, profile: Profile, py_did: DID) -> dict:
```
dbluhm, line 76: This line is actually a mistake; the type for `did` should be `str` with the intention being that resolver plugins need not concern themselves with PyDID if they don't want to. Sorry for the mix-up.

aries_cloudagent/resolver/default/indy.py
```diff
@@ -37,21 +37,20 @@ def supported_methods(self) -> Sequence[str]:
         """Return supported methods of Indy DID Resolver."""
         return ["sov"]
 
-    async def _resolve(self, profile: Profile, did: str) -> dict:
+    async def _resolve(self, profile: Profile, py_did: DID) -> dict:
```
dbluhm, line 40: See previous comment for this method.

dbluhm (changes requested):
It seems we've triggered some confusion on the `_resolve` method parameter types -- sorry about that. I actually had a fix locally that I just hadn't opened a PR for yet.

aries_cloudagent/resolver/base.py
sklump: It's no trouble, let's just converge on what's right

aries_cloudagent/resolver/default/indy.py
sklump: How about now? Better?

dbluhm (approved):

andrewwhitehead (approved):

PR #1096, opened 2021-04-18
Loosen type checks in responder

dbluhm:
This one might be a bit of a stretch but here's my reasoning. I think I'm one of the most active ACA-Py plugin developers at the moment and one of the things I've found is that it can be somewhat burdensome to follow the same pattern ACA-Py does for defining messages and handlers in all cases. I've experimented with using a simplified base class other than the `AgentMessage` in ACA-Py and have had favorable results. There _is_ a loss of functionality but, with these tweaks, the core functions work. From what I've gathered so far, the lost functionality is limited primarily to timing and tracing features, which might be a reasonable trade-off in a plugin context if a different root message class integrates better with the rest of your code.

This PR uses `typing.Protocol` to define the expected methods on messages to send, loosening the original explicit type check for `AgentMessage` to a "shape" check. This uses the `typing-extensions` package to backport `typing.Protocol` for python 3.6.

Rather than defining just what the responder needs to send a message, it might make sense to define something somewhere else that outlines the minimum functional agent message attributes. To get a message from decrypted object to being handled, a minimum of `Handler`, and `deserialize` is required. To get a message from creation (in a handler, for instance) to being sent as an encrypted message, a minimum of `to_json` and `_thread_id` is required.

Interested to hear thoughts :slightly_smiling_face:

### DIFF
--- a/aries_cloudagent/core/dispatcher.py
+++ b/aries_cloudagent/core/dispatcher.py
@@ -17,7 +17,7 @@ from ..messaging.agent_message import AgentMessage
 from ..messaging.error import MessageParseError
 from ..messaging.models.base import BaseModelError
 from ..messaging.request_context import RequestContext
-from ..messaging.responder import BaseResponder
+from ..messaging.responder import BaseResponder, Message
 from ..messaging.util import datetime_now
 from ..protocols.connections.v1_0.manager import ConnectionManager
 from ..protocols.problem_report.v1_0.message import ProblemReport
@@ -267,7 +267,7 @@ class DispatcherResponder(BaseResponder):
         self._webhook = send_webhook
 
     async def create_outbound(
-        self, message: Union[AgentMessage, str, bytes], **kwargs
+        self, message: Union[Message, AgentMessage, str, bytes], **kwargs
     ) -> OutboundMessage:
         """
         Create an OutboundMessage from a message body.
--- a/aries_cloudagent/messaging/responder.py
+++ b/aries_cloudagent/messaging/responder.py
@@ -6,7 +6,7 @@ in response to the message being handled.
 """
 
 from abc import ABC, abstractmethod
-from typing import Sequence, Union
+from typing import Optional, Sequence, Union
 from typing_extensions import Protocol, runtime_checkable
 
 from ..core.error import BaseError
@@ -26,7 +26,7 @@ class Message(Protocol):
         """Serialize the message to json."""
 
     @property
-    def _thread_id(self) -> str:
+    def _thread_id(self) -> Optional[str]:
         """Get the message's thread id."""
 
 
--- a/aries_cloudagent/messaging/responder.py
+++ b/aries_cloudagent/messaging/responder.py
@@ -7,16 +7,27 @@ in response to the message being handled.
 
 from abc import ABC, abstractmethod
 from typing import Sequence, Union
+from typing_extensions import Protocol, runtime_checkable
 
 from ..core.error import BaseError
 from ..connections.models.connection_target import ConnectionTarget
 from ..transport.outbound.message import OutboundMessage
 
-from .agent_message import AgentMessage
-
 
 class ResponderError(BaseError):
     """Responder error."""
+
+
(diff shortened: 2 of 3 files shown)


dbluhm:
While I'm on the subject of my experiments lol, I've also tried out using class decorators to remove a lot of the boilerplate currently required for defining `AgentMessage`s with good success. However, I think I prefer the alternate base class approach because the decorator approach introduces a lot more magic.

burdettadam (approved):

andrewwhitehead:
That's not a bad option. I think if we want to create a protocol for outbound messages, then `to_json` might not be the ideal way to do that, though. With DIDComm v2 the structure of the messages will change, and at a minimum the serializer will need the message contents separated from the decorators. A generic way to do that might be to pass in an instance of a 'MessageSerializer' type interface and have the message implementation call methods on that to produce the result. An easier way would be for the serializer to just pull out the decorators and move them (but it's less than ideal if it has to parse the generated JSON to do that).

Another option here would be to create an abstract base class for serializable messages with a method (not `to_json`) used to produce the expected result, then implement it for `AgentMessage` by just delegating to `to_json`, at least for now.

dbluhm:
Ah, interesting. My changes were very much coming at this from the minimal changes required approach lol. Parsing through your comment, I think I'm missing something still. In what way is a `to_json` method insufficient in anticipation of DIDComm v2? Is it a backwards compatibility issue where we want messages to be able to produce JSON conforming to either the original format or DIDComm? Or producing output in forms other than JSON? Or am I off the mark entirely?

andrewwhitehead:
@dbluhm  My concern is just that we don't want each message implementation to be worrying about what DIDComm version is currently being used. Or rather that they do have to worry about that, but in the AgentMessage case it can all be handled in one place, and otherwise there's no standard way to pass in that information. If we want to formalize a protocol for converting the message to its external representation then it should probably include some context in the parameters at least, and we may want to do the same when deserializing.

dbluhm:
That makes more sense to me now, thanks for the clarification. After a fair amount of thought, I came up with the following:

```python
class DIDCommVersion(Enum):
    """Message formats."""

    v1 = auto()
    v2 = auto()


class BaseMessage(ABC):
    """Abstract base class for messages."""

    @abstractproperty
    def type(self) -> str:
        """Return message type."""

    @abstractproperty
    def id(self) -> str:
        """Return message id."""

    @abstractmethod
    def serialize(
        self, msg_format: DIDCommVersion = DIDCommVersion.v1
    ) -> dict:
        """Return serialized message in format specified."""

    @abstractclassmethod
    def deserialize(
        cls,
        value: dict,
        msg_format: DIDCommVersion = DIDCommVersion.v1
    ):
        """Return message object deserialized from value in format specifified."""

    @abstractproperty
    def Handler(self) -> Type:
        """Return reference to handler class."""
```

This formally defines a "minimum viable message" and provides an unopinionated class for plugins to extend in whatever way makes sense in the context of the plugin. I considered continuing with usage of `typing.Protocol` and splitting up the different pieces (`SerializableMessage`, `DeserializableMessage`, `HandleableMessage`, etc.) but did not find much value in that approach. On top of that, I think explicitly inheriting from a base class for messages communicates our intent clearer than doing the structural sub-typing and checking shape before use.

I spent a good while thinking about what providing a `MessageSerializer` instance would look like. I like the idea -- I think it fits nicely with python's `json.dumps(cls=AlternateEncoder)` -- but given how we tend to use the `to_json`/`serialize` methods, providing a serializer that knows how to handle plugged in message classes and outputs the message in the desired format seems unworkable, unless we define another intermediate format for the serializer to consume. In the end, just providing the desired output format as an enum was the most flexible method I thought up.

If this approach seems agreeable, I'll probably just close out this PR and open a new one with these classes added and using `BaseMessage` in key places like the dispatcher and responder in place of `AgentMessage`.

andrewwhitehead:
Okay, I think that's a good approach.

dbluhm:
Closing in favor of #1133

PR #1101, opened 2021-04-20
Fix introductions protocol

sklump:
Fix introductions protocol. Integrate did-exchange problem reports with problem-report protocol (leave connections as per its RFC 160, predating problem-report protocol).


aries_cloudagent/storage/base.py
```diff
@@ -91,6 +91,7 @@ async def find_record(
         if not results:
             raise StorageNotFoundError("Record not found")
         if len(results) > 1:
+            print(f"\n!! !! TOO Many results {results}")
```
andrewwhitehead, line 94: Is this debugging code?

sklump: Yes - oops

andrewwhitehead (approved):

PR #1111, opened 2021-04-22
secured api_key against timing attacks

wadeking98:
Signed-off-by: wadeking98 <wkingnumber2@gmail.com>
@andrewwhitehead 
comparing user input to secrets and then responding makes this application vulnerable to timing attacks. See resource here: https://blog.sqreen.com/developer-security-best-practices-protecting-against-timing-attacks/  
  
While an attacker isn’t likely to be able to exploit this from somewhere over the internet due to network jitter, they might be able to pull it off if the agent ever supports TOR, HTTP/2 (timeless timing attack), or if they get access to a pod inside the same openshift cluster that can send requests to the agent (ie: the api pod)

### DIFF
--- a/aries_cloudagent/transport/outbound/tests/test_http_transport.py
+++ b/aries_cloudagent/transport/outbound/tests/test_http_transport.py
@@ -1,6 +1,7 @@
 import asyncio
 import pytest
 
+from hmac import compare_digest
 from aiohttp.test_utils import AioHTTPTestCase, unittest_run_loop
 from aiohttp import web
 from asynctest import mock as async_mock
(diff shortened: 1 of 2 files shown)


aries_cloudagent/transport/outbound/tests/test_http_transport.py
```diff
@@ -66,7 +67,7 @@ async def send_message(transport, payload, endpoint, api_key):
             send_message(transport, "{}", endpoint=server_addr, api_key=api_key), 5.0
         )
         assert self.message_results == [{}]
-        assert self.headers.get("x-api-key") == api_key
+        assert api_key and compare_digest(self.headers.get("x-api-key"), api_key)
```
andrewwhitehead, line 70: I don't think there's any need to use this comparison method in the test

wadeking98: okay, I'll remove it

andrewwhitehead (approved):

Issue #1114, opened 2021-04-23
Send Offer Failing for Issue Credential


nodlesh:
There seems to be an issue in Acapy. The problem started yesterday by a merge that happened 1 or 2 days ago. The test runsets in the daily test executions are failing on issuing a credential. This impacts over half of the tests for runsets that include acapy. I am fairly certain it isn’t anything I’ve submitted in AATH because the same AATH code works on older Acapy from the beginning of the week. This issue is on the admin call issue-credential/send-offer. AATH receives a 500 Internal server error and the agent throws this.
```
2021-04-23 20:08:28,936 aries_cloudagent.core.dispatcher ERROR Handler error: credential_exchange_send_bound_offer     
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/protocols/issue_credential/v1_0/routes.py", line 870, in credential_exchange_send_bound_offer
    body = await request.json()
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_request.py", line 614, in json
    return loads(body)
  File "/usr/local/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
2021-04-23 20:08:28,938 aries_cloudagent.admin.server ERROR Handler error with exception: Expecting value: line 1 column 1 (char 0)

=================
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/admin/server.py", line 183, in ready_middleware
    return await handler(request)
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/admin/server.py", line 220, in debug_middleware
    return await handler(request)
  File "/usr/local/lib/python3.7/site-packages/aiohttp_apispec/middlewares.py", line 45, in validation_middleware
    return await handler(request)
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/admin/server.py", line 389, in setup_context
    return await task
[...]

PR #1117, opened 2021-04-26
not having a body is OK to send bound offer

sklump:
Fix https://github.com/hyperledger/aries-cloudagent-python/issues/1114


sklump:
weird: will resubmit

Issue #1123, opened 2021-04-28
Can't emit webhooks for stateless records anymore


TimoGlastra:
The event bus PR that was merged yesterday made a change that now won't allow you to emit events for stateless records anymore. Is this intended?

PR: #1063
Blame: https://github.com/hyperledger/aries-cloudagent-python/pull/1063/files#diff-fe738d195b46caab22aaa9e802de272264dc1f986e589aad2a2bd0965f98ac14R382

@farskipper @dbluhm

dbluhm:
Not intentional -- I'll take a look

TimoGlastra:
Verified that #1124 fixed the issue

PR #1124, opened 2021-04-28
Fix: stateless record webhooks

dbluhm:
This should hopefully address #1123.

### Changes
**fix: stateless records emit events**

Error on the side of caution for BaseRecord.emit_event, using topic
without state when none is given, and inserting the serialized record as
the payload if no payload is given.

Events are only not emitted by records when no record topic is set.

**fix: relax admin server record event pattern**

Should now accept stateless record event topics

**fix: add event namespace to base records to allow plugins to override**

**test: record topic parsing for webhook**

### DIFF
--- a/aries_cloudagent/admin/tests/test_admin_server.py
+++ b/aries_cloudagent/admin/tests/test_admin_server.py
@@ -1,13 +1,14 @@
-from aries_cloudagent.core.event_bus import Event
 import json
 
 from aiohttp import ClientSession, DummyCookieJar, TCPConnector, web
 from aiohttp.test_utils import unused_port
 import pytest
 
+from ...core.event_bus import Event
 from asynctest import TestCase as AsyncTestCase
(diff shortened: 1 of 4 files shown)


aries_cloudagent/admin/tests/test_admin_server.py
```diff
@@ -1,7 +1,9 @@
+from aries_cloudagent.core.event_bus import Event
```
TimoGlastra, line 1: nit: absolute import

TimoGlastra (approved):
Looks good! Thanks for the quick fix :)

aries_cloudagent/admin/tests/test_admin_server.py
dbluhm: Auto-import failed me :slightly_smiling_face: fixed

ianco (approved):

PR #1130, opened 2021-04-30
feat: feedback from responder methods about message delivery

dbluhm:
As a result of some discussion in #950, I did a quick investigation of what it would look like to provide some feedback to the responder send methods about message delivery so that from the `forward` message handler, for example, we could determine whether the message was sent directly or not. This doesn't provide a ton of information but I think it provides just enough to know whether a push notification should be sent to a mobile device or not from the forward message handler, as @TimoGlastra described in the comments of that issue.

I think there are a lot of ways to end up at similar functionality though this one ended up being simpler than I anticipated. Interested to hear thoughts.

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -1,5 +1,6 @@
 """Admin server classes."""
 
+from aries_cloudagent.transport.outbound.status import OutboundSendStatus
 import asyncio
 from hmac import compare_digest
 import logging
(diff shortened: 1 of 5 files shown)


dbluhm:
Taking the next step to emitting an event in the forward message handler should be pretty simple from here.

aries_cloudagent/admin/server.py
```diff
@@ -1,5 +1,6 @@
 """Admin server classes."""
 
+from aries_cloudagent.transport.outbound.status import OutboundSendStatus
```
dbluhm, line 3: Auto-import defaulting to absolute imports... I'll fix this

aries_cloudagent/core/dispatcher.py
```diff
@@ -5,6 +5,7 @@
 lifecycle hook callbacks storing state for message threads, etc.
 """
 
+from aries_cloudagent.transport.outbound.status import OutboundSendStatus
```
dbluhm, line 8: Same

aries_cloudagent/messaging/responder.py
```diff
@@ -6,6 +6,7 @@
 """
 
 from abc import ABC, abstractmethod
+from aries_cloudagent.transport.outbound.status import OutboundSendStatus
```
dbluhm, line 9: Same

TimoGlastra:
This is awesome! Thans for looking into this @dbluhm! Thought this was going to be a lot more work. Glad I asked :)

andrewwhitehead (approved):

PR #1133, opened 2021-05-01
Add "minimum viable message" base class for use in plugins

dbluhm:
As discussed in #1096, when writing plugins, having the flexibility to implement messages differently than the pattern followed within ACA-Py can be really significant. This PR defines a minimized base message class that is unopinionated and trivial to extend.

This PR also addresses concerns around message serialization in anticipation of changes coming with DIDComm v2 by defining `serialize` and `deserialize` methods that specify the output/input format. At present, the `AgentMessage` class only implements DIDComm v1 message structure.

### DIFF
--- a/aries_cloudagent/core/dispatcher.py
+++ b/aries_cloudagent/core/dispatcher.py
@@ -15,6 +15,7 @@ from aiohttp.web import HTTPException
 
 from ..core.profile import Profile
 from ..messaging.agent_message import AgentMessage
+from ..messaging.base_message import BaseMessage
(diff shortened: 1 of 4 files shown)


andrewwhitehead:
Small conflict created by another PR

andrewwhitehead (approved):

dbluhm:
Side note @ianco, integration tests failed the first time around. I should have noted the scenario (not sure if those logs are stored somewhere) but I believe it had "Issuer sends credential offer but then revokes." Not sure if that's something we should look into.

PR #1135, opened 2021-05-03
fix: add missing connection_id to issue v2 schema

TimoGlastra:
Noticed while writing documentation that the connection_id was missing.


sklump (changes requested):
Not so: this schema is for the out-of-band creation method. It does not take a connection id.

PR #1148, opened 2021-05-05
feat: send webhooks for forward messages 

TimoGlastra:
This adds an option to ACA-Py to send webhook events on received forward messages (`--monitor-forward`, taken from `monitor-ping`).

I extended the `OutboundSendStatus` work from Daniel and added a few more status options. These are then included in the webhook event, so the controller can react to the event. 

If `--monitor-forward` is not enabled, the event is still dispatched in the event bus (just not as webhook), so other parts of the codebase/plugins can hook into this if they want to manage it in ACA-Py instead of the controller

### DIFF
--- a/AdminAPI.md
+++ b/AdminAPI.md
@@ -54,6 +54,19 @@ When a webhook is dispatched, the record `topic` is appended as a path component
  * `content`: the contents of the agent message
  * `state`: `received`
 
+#### Forward Message Received  (`/forward`)
+
+Enable using `--monitor-forward`.
+
+ * `connection_id`: the identifier of the connection associated with the recipient key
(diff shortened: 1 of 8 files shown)


TimoGlastra:
@dbluhm

dbluhm (dismissed):
Nice! Looks good!

dbluhm:
Failing tests are my fault, fix in #1151

TimoGlastra:
Updated

andrewwhitehead (dismissed):

andrewwhitehead:
The branch needs updating/rebasing

andrewwhitehead (approved):

PR #1154, opened 2021-05-05
Fix/type in ld proof

domwoe:
* Fix issue described in [Rocket chat ](https://chat.hyperledger.org/channel/aries-cloudagent-python?msg=CDike9uCQciDRohRu)
* Allow `type` property in proof validation schema
* Set default proof type if property is not provided

Currently only `Ed25519Signature2018` is allowed anyways...


aries_cloudagent/messaging/jsonld/routes.py
```diff
@@ -37,6 +37,7 @@ class SignRequestSchema(OpenAPISchema):
                             "creator": fields.Str(required=False),
                             "verificationMethod": fields.Str(required=False),
                             "proofPurpose": fields.Str(required=False),
+                            "type": fields.Str(required=False),
```
dbluhm, line 40: I ended up looking at this issue as well before seeing you'd gotten here first but in my fix I made `verificationMethod` and `proofPurpose` required. Based on my reading of the VC data model and LD Proof spec, these are required values that, in the context of these routes at least, we don't know how to populate automatically. Shouldn't we make these attributes required?

domwoe: Yes, good point. I'll update the PR later today.

aries_cloudagent/messaging/jsonld/routes.py
```diff
@@ -94,8 +95,9 @@ class Meta:
...
             {
                 "creator": fields.Str(required=False),
-                "verificationMethod": fields.Str(required=False),
-                "proofPurpose": fields.Str(required=False),
+                "verificationMethod": fields.Str(required=True),
+                "proofPurpose": fields.Str(required=True),
+                "type": fields.Str(required=False),
             }
         )
     )
```
domwoe, line 103: This schema seems to be not used anymore. Should we remove it? @dbluhm @burdettadam

etschelp: Also be aware that there are checks further down the need to be in sync with the schema see [create_verify_data](https://github.com/hyperledger/aries-cloudagent-python/blob/b0f0d1ed4b844c2e4aa0d3822655b7881e1168ad/aries_cloudagent/messaging/jsonld/create_verify_data.py#L52). Also there is/was a _creator_ that becomes the _verificationMethod_ again.

domwoe: You're correct. This is in consistent. Either we allow both `creator` and `verificationMethod` but none is required in the schema and we raise an error later if none is provided or we remove the `creator` property. I'd vote for removing the property.

dbluhm: Hm... The `DocSchema` should be used in the `VerifyRequestSchema`. I'm not sure why it's not -- merge artifact perhaps?

domwoe: I think I'm done. You want to have a look? :)

dbluhm (approved):
LGTM!

PR #1160, opened 2021-05-07
allow other service/endpoint types in indy did docs

domwoe:
Ability for the native indy resolver to get additional services / endpoint types from an indy ledger.

### DIFF
--- a/aries_cloudagent/resolver/default/indy.py
+++ b/aries_cloudagent/resolver/default/indy.py
@@ -2,13 +2,14 @@
 
 Resolution is performed using the IndyLedger class.
 """
+
 from typing import Sequence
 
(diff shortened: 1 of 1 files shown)


domwoe:
@dbluhm  I had to set a priority for the did-communication service, otherwise I get the following error from `pydid`
```
File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/pydid/doc/builder.py", line 200, in add_didcomm
    priority = priority or self._determine_next_priority()
  File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/pydid/doc/builder.py", line 161, in _determine_next_priority
    if self.services
ValueError: max() arg is an empty sequence
```

andrewwhitehead (approved):

PR #1174, opened 2021-05-13
DID Exchange - Create and send requests against implicit invitations

### LINKED ISSUE #1137: DID Exchange - Issue with implicit invitations
I'm experimenting with the DID Exchange protocol in order to create a connection based on a public DIDs implicit invitation.
A public DID allows to retrieve the necessary keys and service information from a given VDR, i.e. all information that is needed to pack and send the respective request message.

ACA-Py currently provides the following APIs:

image

`create-request` allows to create a connection request and returns it, but there seems to be no exposed functionality to actually send the request. 

I tried a simple hack adding
```
 # Send request
        responder = self._session.inject(BaseResponder, required=False)
        if responder:
            await responder.send(request, connection_id=conn_rec.connection_id)
```
for testing, but `fetch_connection_targets` fails because there's no invitation attached to the connection.

In addition, I don't understand the use case for `receive-request`, since I'd say requests should use the DIDcomm channel. Everything else should use the Out of Band API.

shaangill025:
closes #1137

### DIFF
--- a/aries_cloudagent/connections/base_manager.py
+++ b/aries_cloudagent/connections/base_manager.py
@@ -291,7 +291,9 @@ class BaseConnectionManager:
         ):
             if connection.invitation_msg_id or connection.invitation_key:
                 invitation = await connection.retrieve_invitation(self._session)
-                if isinstance(invitation, ConnectionInvitation):  # conn protocol invitation
+                if isinstance(
+                    invitation, ConnectionInvitation
+                ):  # conn protocol invitation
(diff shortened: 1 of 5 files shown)


shaangill025:
@domwoe Will this work for your use case?

domwoe:
@shaangill025 looks good! However, I went the path to add an additional route because I think it makes sense to return the connection record instead of the request when sending the request over DIDcomm.

sklump (approved):

PR #1178, opened 2021-05-14
fix: don't load indy resolver if no indy ledger

dbluhm:


aries_cloudagent/resolver/__init__.py
```diff
@@ -1,5 +1,6 @@
 """Interfaces and base classes for DID Resolution."""
 
+from aries_cloudagent.ledger.base import BaseLedger
```
andrewwhitehead, line 3: This should be a relative import with the others

dbluhm: Ah, woops. Auto-import... thanks for the catch.

andrewwhitehead (approved):

PR #1182, opened 2021-05-18
DIDX create-request [return ConnRecord]

shaangill025:
Work left from #1174 and #1137.

- /didexchange/create-request now return ConnRecord
- Based upon feedback from @domwoe and  @andrewwhitehead, this will make it more consistent


domwoe:
I think if do this we should also always send the request, i.e. don't use a send flag, since the request itself is not returned or stored. Using the API now with send=false (default) does not make sense.

sklump (approved):

PR #1186, opened 2021-05-20
Tags on connection metadata

Luis-GA:
With this PR, It is possible to add tags to connections. The tags are saved as a metadata key-value on the connections record.
Once the connections have been tagged, it is possible to group them on tags.

The tags are added on `/connections/create-invitation` & `/connections/receive-invitation` via a query parameter called `tags` in the request.  The input is a string, it is possible to add multiple tags if they are separated by commas. 

The tags could be filtered on `/connections` via a query parameter called `tags` in the request. The input is a string, it is possible to filter by multiple tags if they are separated by commas.

**Examples**:

- Create invitation

> Request
`curl -X POST "http://localhost:8021/connections/create-invitation?tags=tag1%2Ctag3" -H "accept: application/json" -H "Content-Type: application/json" -d "{ }"`

> Response
`{
  "connection_id": "8e1d9605-5843-4844-99d1-96141329e006",
  "invitation": {
    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/invitation",
    "@id": "1b62e664-a8d9-4c97-bb7a-2ba3b5329ebb",
    "recipientKeys": [
      "Gsc9q2FskuKYH2bAMnd3UkQadEkJ4QFwwA8RUdyGqsTo"
    ],
    "label": "faber.agent",
    "serviceEndpoint": "https://0639b9bd3d2a.ngrok.io"
  },
  "invitation_url": "https://0639b9bd3d2a.ngrok.io?c_i=eyJAdHlwZSI6ICJkaWQ6c292OkJ6Q2JzTlloTXJqSGlxWkRUVUFTSGc7c3BlYy9jb25uZWN0aW9ucy8xLjAvaW52aXRhdGlvbiIsICJAaWQiOiAiMWI2MmU2NjQtYThkOS00Yzk3LWJiN2EtMmJhM2I1MzI5ZWJiIiwgInJlY2lwaWVudEtleXMiOiBbIkdzYzlxMkZza3VLWUgyYkFNbmQzVWtRYWRFa0o0UUZ3d0E4UlVkeUdxc1RvIl0sICJsYWJlbCI6ICJmYWJlci5hZ2VudCIsICJzZXJ2aWNlRW5kcG9pbnQiOiAiaHR0cHM6Ly8wNjM5YjliZDNkMmEubmdyb2suaW8ifQ=="
}`

- Filter connections

> Request
`curl -X GET "http://localhost:8021/connections?tags=tag1" -H "accept: application/json"`

> Response
`{
  "results": [
    {
      "updated_at": "2021-05-20 12:06:41.865730Z",
      "created_at": "2021-05-20 12:06:41.865730Z",
      "accept": "manual",
      "rfc23_state": "invitation-sent",
      "their_role": "invitee",
      "routing_state": "none",
      "state": "invitation",
      "invitation_mode": "once",
      "connection_id": "8e1d9605-5843-4844-99d1-96141329e006",
      "invitation_key": "Gsc9q2FskuKYH2bAMnd3UkQadEkJ4QFwwA8RUdyGqsTo",
      "metadata": {
        "tags": [
          "tag1",
          "tag3"
        ]
      }
    }
  ]
}`

Based on @burdettadam idea, https://github.com/sicpa-dlab/aries-cloudagent-python/issues/49.

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -180,6 +180,7 @@ class ConnRecord(BaseRecord):
         accept: str = None,
         invitation_mode: str = None,
         alias: str = None,
+        metadata: dict = None,
         their_public_did: str = None,
         rfc23_state: str = None,  # from state: formalism for base_record.from_storage()
         initiator: str = None,  # for backward compatibility with old ConnectionRecord
@@ -210,6 +211,7 @@ class ConnRecord(BaseRecord):
         self.accept = accept or self.ACCEPT_MANUAL
         self.invitation_mode = invitation_mode or self.INVITATION_MODE_ONCE
         self.alias = alias
+        self.metadata = metadata
         self.their_public_did = their_public_did
 
     @property
@@ -235,6 +237,7 @@ class ConnRecord(BaseRecord):
                 "invitation_mode",
                 "invitation_msg_id",
                 "alias",
+                "metadata",
                 "error_msg",
                 "their_label",
                 "state",
@@ -598,6 +601,13 @@ class ConnRecordSchema(BaseRecordSchema):
         description="Optional alias to apply to connection for later use",
         example="Bob, providing quotes",
     )
+
+    metadata = fields.Dict(
+        description="Optional metadata to tag connections",
+        required=False,
+        example={"tags": ["tag1", "tag2"]},
+    )
+
     their_public_did = fields.Str(
         required=False,
         description="Other agent's public DID for connection",
--- a/aries_cloudagent/messaging/models/base_record.py
+++ b/aries_cloudagent/messaging/models/base_record.py
@@ -45,26 +45,41 @@ def match_post_filter(
         alt: set to match any (positive=True) value or miss all (positive=False)
             values in post_filter
     """
+
+    def filter_keys(record, k, filter_tags):
+        if k == "tags":
+            metadata = record.get("metadata")
+            if not metadata:
+                return False
+            tags = metadata.get(k, [])
+            intersect = list(set(filter_tags) & set(tags))
+            return tags and len(intersect) == len(filter_tags)
+
+        return record.get(k) and record.get(k) in filter_tags
+
     if not post_filter:
         return True
 
     if alt:
         return (
             positive
-            and all(
-                record.get(k) and record.get(k) in alts
-                for k, alts in post_filter.items()
-            )
+            and all(filter_keys(record, k, alts) for k, alts in post_filter.items())
         ) or (
             (not positive)
-            and all(
-                record.get(k) and record.get(k) not in alts
-                for k, alts in post_filter.items()
-            )
+            and all(not filter_keys(record, k, alts) for k, alts in post_filter.items())
         )
 
     for k, v in post_filter.items():
-        if record.get(k) != v:
+        if record.get(k) == v:
+            continue
+        elif k == "tags":
+            record_tags = record.get("metadata").get(k)
+            filter_tags = v
+            intersect = list(set(filter_tags) & set(record_tags))
+            if len(intersect) != len(filter_tags):
+                return not positive
+
+        else:
             return not positive
 
     return positive
@@ -500,6 +515,7 @@ class BaseRecord(BaseModel):
                     ret[k] = cls.prefix_tag_filter(v)
                 else:
                     ret[tag_map.get(k, k)] = v
+
         return ret
 
     def __eq__(self, other: Any) -> bool:
--- a/aries_cloudagent/messaging/models/base_record.py
+++ b/aries_cloudagent/messaging/models/base_record.py
@@ -46,16 +46,21 @@ def match_post_filter(
             values in post_filter
     """
 
-    def filter_keys(record, k, filter_tags):
-        if k == "tags":
-            metadata = record.get("metadata")
-            if not metadata:
-                return False
-            tags = metadata.get(k, [])
-            intersect = list(set(filter_tags) & set(tags))
-            return tags and len(intersect) == len(filter_tags)
+    def filter_tags(record, k, filter_tags):
 
-        return record.get(k) and record.get(k) in filter_tags
+        metadata = record.get("metadata")
+        if not metadata:
+            return False
+        tags = metadata.get(k, [])
+        intersect = list(set(filter_tags) & set(tags))
+        return tags and len(intersect) == len(filter_tags)
+
+    def filter_keys(record, k, alts):
+        if k == "tags":
+
+            return filter_tags(record, k, alts)
+
+        return record.get(k) and record.get(k) in alts
 
     if not post_filter:
         return True
@@ -73,10 +78,7 @@ def match_post_filter(
         if record.get(k) == v:
             continue
         elif k == "tags":
-            record_tags = record.get("metadata").get(k)
-            filter_tags = v
-            intersect = list(set(filter_tags) & set(record_tags))
-            if len(intersect) != len(filter_tags):
+            if not filter_tags(record, k, v):
                 return not positive
 
         else:
--- a/aries_cloudagent/messaging/models/tests/test_base_record.py
+++ b/aries_cloudagent/messaging/models/tests/test_base_record.py
@@ -15,7 +15,7 @@ from ....storage.base import (
 
 from ...util import time_now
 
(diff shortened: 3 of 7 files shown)


sklump:
Seems fraught to me: name collisions with non-secrets tags, existing metadata operations
https://github.com/hyperledger/aries-cloudagent-python/blob/a322a8650a94b0488bd288a14fc59a2792f90fd3/aries_cloudagent/connections/models/conn_record.py#L412

I don't know what the answers ought to be though.

Luis-GA:
@sklump Thank you for your suggestions. Yes, I agree with you that `tags` is a used name for other purposes, however, I could not found a better word to explain the grouping (I thought also in labels, but they are used also). Finally, I opted to name them as tags because I think the variables are in different contexts. 

About the second comment, the filtering by tags logic is managed on `aries_cloudagent/messaging/models/base_record.py` by the function  `match_post_filter` in the same way that currently filtering by `alias` works. 
https://github.com/hyperledger/aries-cloudagent-python/blob/a322a8650a94b0488bd288a14fc59a2792f90fd3/aries_cloudagent/messaging/models/base_record.py#L32
In this function, the connection is a dictionary object instead of a ConnRecord instance. So I think we can not use the `ConnRecord.metadata_get` for this PR.

Let me know if you consider any other name to replace "tags" or if there is something in the code that could be done in a better way.

ianco:
Hi @Luis-GA how does this functionality relate to the existing endpoints GET and POST `/connections/{conn_id}/metadata`?

Luis-GA:
Hi @ianco the tags are just a metadata key-value to filter connections. Those endpoints could retrieve the connection metadata or overwrite the metadata (including its tags). 
Do you think we should have a different approach for that?

Thank you for your question!

ianco:
Hi @Luis-GA I'm not sure about the approach for your particular use case, I just think we shouldn't have 2 separate ways to add meta-data to a connection

burdettadam:
This feature is an attempt to reduce the need to search a wallet multiple times. Currently, the pattern is to retrieve a filtered connection record list by WQL, then for each connection record retrieve a meta-data record by WQL and filter connections again on the desired meta-data. This can be seen in retrieving a default mediator or looking up a connection that can be used to resolve a did in the resolver plugin. While this works well in small single-user wallets, it may not work as well in large wallets with multiple users. I believe the approach to support 2 separate ways was to maintain backward compatibility. Do you suggest we deprecate meta-data records? This may be premature optimization, should we wait till this feature is needed?

Luis-GA:
Hi @ianco. I think it is not really two ways to add metadata. It only tags the metadata at creation (create & receive invitations). 

Later on, the only endpoints for CRUD operations are GET and POST `/connections/{conn_id}/metadata`. 
They are the only way to add metadata or tags in those metadata after a connection creation.

ianco:
Ah ok thanks for the clarification @Luis-GA !

ianco:
I think the new tags is using a different method for creating/storing metadata than the existing GET/POST `/connections/{conn_id}/metadata` endpoints.  I created a new invitation with tag `tag1` and then accepted it on the other agent with tag `tag2`.  These tags show up when you query the connection but aren't retrieved when you call the `/metadata` endpoint.  Likewise when you add an element via the `/metadata` endpoint it is not returned when you query the connection.  This means that you create the tags when you create or accept an invitation, but you can never update the tags.  You can create metadata once the connection is established, but you can't search on metadata (the same way you can with tags).

Wouldn't it be better to have one set of metadata for a connection, and to support searching with the elements added through the `/metadata` endpoints?

Luis-GA:
Hi @ianco,  I have done some tests:

> I create a connection
> `curl -X POST "http://localhost:8021/connections/create-invitation?tags=tag-example" -H "accept: application/json" -H "Content-Type: application/json" -d "{}"`
> The response was:
> `{
>   "connection_id": "3ce7b9e0-23a3-4f5f-90bd-351e199cf088",
>   "invitation": {
>     "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/invitation",
>     "@id": "030eabdb-80e1-4349-918f-7ceac9b07c41",
>     "serviceEndpoint": "http://efef47449e07.ngrok.io",
>     "label": "faber.agent",
>     "recipientKeys": [
>       "2HS1kiMGEWa6g5ZWUYP3SwzCxdLwGatJcJ5iPXpwKkQF"
>     ]
>   },
>   "invitation_url": "http://efef47449e07.ngrok.io?c_i=eyJAdHlwZSI6ICJkaWQ6c292OkJ6Q2JzTlloTXJqSGlxWkRUVUFTSGc7c3BlYy9jb25uZWN0aW9ucy8xLjAvaW52aXRhdGlvbiIsICJAaWQiOiAiMDMwZWFiZGItODBlMS00MzQ5LTkxOGYtN2NlYWM5YjA3YzQxIiwgInNlcnZpY2VFbmRwb2ludCI6ICJodHRwOi8vZWZlZjQ3NDQ5ZTA3Lm5ncm9rLmlvIiwgImxhYmVsIjogImZhYmVyLmFnZW50IiwgInJlY2lwaWVudEtleXMiOiBbIjJIUzFraU1HRVdhNmc1WldVWVAzU3d6Q3hkTHdHYXRKY0o1aVBYcHdLa1FGIl19"
> }`


> Then I get the metadata:
> `curl -X GET "http://localhost:8021/connections/3ce7b9e0-23a3-4f5f-90bd-351e199cf088/metadata" -H "accept: application/json"`
> The response was:
> `{
>   "results": {
>     "tags": [
>       "tag-example"
>     ]
>   }
> }`

> Later I overwrite the metadata:
> `curl -X POST "http://localhost:8021/connections/3ce7b9e0-23a3-4f5f-90bd-351e199cf088/metadata" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"metadata\": {\"tags\":[\"tag-example2\"]}}"`
> The response was:
> `{
>   "results": {
>     "tags": [
>       "tag-example2"
>     ]
>   }
> }`

> I retrieve again the metadata:
> `curl -X GET "http://localhost:8021/connections/3ce7b9e0-23a3-4f5f-90bd-351e199cf088/metadata" -H "accept: application/json"`
> The response was:
> `{
>   "results": {
>     "tags": [
>       "tag-example2"
>     ]
>   }
> }`

> Then I get the connection itself:
> `curl -X GET "http://localhost:8021/connections/3ce7b9e0-23a3-4f5f-90bd-351e199cf088" -H "accept: application/json"`
> The response was:
> `{
>   "invitation_key": "2HS1kiMGEWa6g5ZWUYP3SwzCxdLwGatJcJ5iPXpwKkQF",
>   "routing_state": "none",
>   "state": "invitation",
>   "created_at": "2021-05-27 08:05:45.331895Z",
>   "invitation_mode": "once",
>   "connection_id": "3ce7b9e0-23a3-4f5f-90bd-351e199cf088",
>   "metadata": {
>     "tags": [
>       "tag-example"
>     ]
>   },
>   "accept": "manual",
>   "updated_at": "2021-05-27 08:05:45.331895Z",
>   "rfc23_state": "invitation-sent",
>   "their_role": "invitee"
> }`

**My conclusions:**

The metadata operations work properly, however, the connections calls do not retrieve the last version of the connection. I have checked this is a previous bug that should be fixed in another PR to not lose the scope of this PR. 

What do you think?

Luis-GA:
Error fixed on https://github.com/hyperledger/aries-cloudagent-python/pull/1210

andrewwhitehead:
> @burdettadam Currently, the pattern is to retrieve a filtered connection record list by WQL, then for each connection record retrieve a meta-data record by WQL and filter connections again on the desired meta-data. This can be seen in retrieving a default mediator or looking up a connection that can be used to resolve a did in the resolver plugin.

At the moment I don't really see how this change improves that situation, as it has to fetch all the connection records and post-filter by their metadata contents. It's not taking advantage of tag indexing in the database. The linked issue seems to suggest adding tags to metadata records (in addition to `key` and `connection_id`), which would allow filtering those metadata records with WQL and then fetching the associated connection record by ID. If that were implemented, we would probably want to auto-prefix the associated tags in order to avoid collisions.

dbluhm:
I see the reasoning for tagging connection metadata but I think this set of changes is unfortunately missing the mark. @Luis-GA @burdettadam, let's discuss this offline and see what solutions we can come up with.

PR #1188, opened 2021-05-20
AIP-2 base64url consistency

### LINKED ISSUE #1108: raw base64url consistency?
ACA-Py seems to be permissive, in a certain sense, with the format that a base64 attachment payload can take: [`b64_payload = unpad(set_urlsafe_b64(self.base64, True))`](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/messaging/decorators/attach_decorator.py#L426). However, this still enforces that the message sender must convert their base64 payload to rawURL format.

For interop, we'll need to decide how we standardize, and ensure it gets into AIP2.0. [aries-rfcs/#561](https://github.com/hyperledger/aries-rfcs/pull/561) (and also specifying whether to include/exclude padding), for total consistency? If that's not possible, do we push to standardize attachment signatures on the way you're converting a base64 payload to unpadded base64URL?

shaangill025:
- In regards to decoding, we handle base64 permissively, passing `urlsafe` as `True`.
- If `emit_new_didcomm_mime_type` and `emit_new_didcomm_prefix` arguments are set, send everything as base64url [updates `data_base64` accordingly] else leave eveything as it is.
closes #1108

### DIFF
--- a/aries_cloudagent/messaging/decorators/attach_decorator.py
+++ b/aries_cloudagent/messaging/decorators/attach_decorator.py
@@ -571,6 +571,7 @@ class AttachDecorator(BaseModel):
         cls,
         mapping: Mapping,
         *,
+        flag_aip2: bool = False,
         ident: str = None,
         description: str = None,
         filename: str = None,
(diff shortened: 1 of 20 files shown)


swcurran:
The concern I have is the use of the `emit_new_didcomm_mime_type` and `emit_new_didcomm_prefix` arguments as the trigger.  Should we have a separate `aip2` argument as a trigger for a number of behaviours in ACA-Py? Alternative is a new one specifically for this, but aren't we trying to make a number of breaking changes at once?

shaangill025:
This PR should be ready, it needs minor change [uncommenting `# if self._session.profile.settings.get_value("aip_version", 1) >= 2]` once #1056 is merged into `main`.

dbluhm:
@shaangill025 Is this PR still relevant?

PR #1190, opened 2021-05-20
Fix bug with problem report 2.0 handler and message family

esune:
Fixes a bug caused by inconsistent naming that would result in problem reports not being handled correctly by the receiving agent for `issue-credential-2.0`.

I am surprised all the unit tests passed before: is there a testing pattern I should be looking into to add one, or is this something that will/should be caught by the test harness?


sklump (approved):

sklump:
Note that the unit tests use a mock responder: they check that messages are sent, not what another agent does on receipt. Good catch.

esune:
> 
> 
> Note that the unit tests use a mock responder: they check that messages are sent, not what another agent does on receipt. Good catch.

Thanks for the explanation! That makes sense, I guess this should be covered in the integration tests from AATH - c.c. @nodlesh

PR #1193, opened 2021-05-21
Fix: OOB Invitation against Public DIDs

shaangill025:
- Now able to create and receive OOB invitations against public DIDs.
- Fixes not being able to load `IndyDIDResolver`.
- Implement `supported_did_regex` in `KeyDIDResolver` and `IndyDIDResolver`.

### DIFF
--- a/aries_cloudagent/resolver/__init__.py
+++ b/aries_cloudagent/resolver/__init__.py
@@ -4,7 +4,8 @@ import logging
 
 from ..config.injection_context import InjectionContext
 from ..config.provider import ClassProvider
-from ..ledger.base import BaseLedger
+# from ..ledger.base import BaseLedger
+
 from .did_resolver_registry import DIDResolverRegistry
 
 LOGGER = logging.getLogger(__name__)
@@ -23,7 +24,8 @@ async def setup(context: InjectionContext):
     await key_resolver.setup(context)
     registry.register(key_resolver)
 
-    if context.inject(BaseLedger, required=False):
+    # ledger = context.inject(BaseLedger, required=False)
+    if not context.settings.get("ledger.disabled"):
         indy_resolver = ClassProvider(
             "aries_cloudagent.resolver.default.indy.IndyDIDResolver"
         ).provide(context.settings, context.injector)
--- a/aries_cloudagent/resolver/__init__.py
+++ b/aries_cloudagent/resolver/__init__.py
@@ -5,8 +5,6 @@ import logging
 from ..config.injection_context import InjectionContext
 from ..config.provider import ClassProvider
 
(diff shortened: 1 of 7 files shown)


andrewwhitehead:
We might need to look at where the DID resolver is provided. A BaseLedger instance can't generally be provided globally, because (for indy-sdk at least) it's tied to the wallet. So the DID resolver might need to be provided at the Profile level in that case.  @dbluhm

dbluhm:
I see the issue with the indy ledger loading "fix" I added... @andrewwhitehead the IndyDIDResolver will use the profile passed to the resolve method to get the ledger instance so that is handled correctly, I think. My fix was short-sighted -- the issue we were experiencing was that the ledger disabled setting can be false even while no indy wallet is loaded so the IndyDIDResolver should not be loaded.

aries_cloudagent/resolver/__init__.py
```diff
@@ -4,7 +4,9 @@
 
 from ..config.injection_context import InjectionContext
 from ..config.provider import ClassProvider
-from ..ledger.base import BaseLedger
+
+# from ..ledger.base import BaseLedger
```
andrewwhitehead, line 8: Please remove commented code

aries_cloudagent/resolver/default/indy.py
```diff
@@ -35,9 +39,18 @@ async def setup(self, context: InjectionContext):
...
+        Return supported methods of Indy DID Resolver.
+
+        DEPRECATED: Use supported_did_regex instead.
+        """
         return ["sov"]
 
+    @property
+    def supported_did_regex(self) -> Pattern:
+        """Return supported_did_regex of Indy DID Resolver."""
+        return re.compile(IndyDID.PATTERN)
```
andrewwhitehead, line 52: I think it's worth creating this globally and returning a pre-compiled `Pattern` instance.

aries_cloudagent/resolver/default/key.py
```diff
@@ -22,9 +26,18 @@ async def setup(self, context: InjectionContext):
...
+        Return supported methods of Key DID Resolver.
+
+        DEPRECATED: Use supported_did_regex instead.
+        """
         return ["key"]
 
+    @property
+    def supported_did_regex(self) -> Pattern:
+        """Return supported_did_regex of Key DID Resolver."""
+        return re.compile(DIDKeyType.PATTERN)
```
andrewwhitehead, line 39: Same, can be pre-compiled

shaangill025: I think all requested changes are in.

andrewwhitehead (approved):

PR #1195, opened 2021-05-21
Mitigate rich schema thirst

sklump:
fix protocols/problem_report/v1_0/__init__.py internal_error() en passant: hadn't been creating compliant RFC 35 ProblemReport objects, resulting in cascading failure


sklump:
OK I see I have a few things to work out still

andrewwhitehead (approved):

PR #1202, opened 2021-05-24
Updates to OOB InvitationMessage & Endpoint example - Consistent with RFC0434

### LINKED ISSUE #1138: OOB InvitationReceiveRequest model
The model specifies `service_dids` and `services`, but the example value in the Open API definition only specifies `service_dids`.
I guess `service_dids` is deprecated. I can't find anything about it in the OOB RFC and ACA-Py will return an error when `services` is not specified. `Services` should either accept a (fully-qualified) DID or a DIDcomm service object.

shaangill025:
- resolve #1138

### DIFF
--- a/aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
+++ b/aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
@@ -2,7 +2,7 @@
 
 from collections import namedtuple
 from enum import Enum
-from re import sub, compile
+from re import sub
 from typing import Sequence, Text, Union
 from urllib.parse import parse_qs, urljoin, urlparse
 
@@ -97,9 +97,7 @@ class ServiceOrDIDField(fields.Field):
         if isinstance(value, dict):
             return Service.deserialize(value)
         elif isinstance(value, str):
-            # To adjust after #1193 merge
-            # if bool(IndyDID.PATTERN.match(value)):
-            if bool(compile(IndyDID.PATTERN).match(value)):
+            if bool(IndyDID.PATTERN.match(value)):
                 return value
             else:
                 raise ValidationError(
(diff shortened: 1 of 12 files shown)


aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
```diff
@@ -86,6 +85,26 @@ def aka(self) -> int:
...
+    def _serialize(self, value, attr, obj, **kwargs):
+        if isinstance(value, Service):
+            return value.serialize()
+        return value
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, dict):
+            return Service.deserialize(value)
+        elif isinstance(value, str):
+            if bool(IndyDID.PATTERN.match(value)):
```
andrewwhitehead, line 100: I don't think this is the right place to be restricting the implementation to only Indy DIDs. Even now we should be able to accept an invitation with multiple DID services as long as one of them is supported.

open-api/openapi.json
```diff
@@ -6241,22 +6241,29 @@
             "$ref" : "#/definitions/AttachDecorator"
           }
         },
-        "service_blocks" : {
+        "services" : {
```
andrewwhitehead, line 6244: This one trailing space seems slightly odd to me

andrewwhitehead:
Looks good but maybe we can be less restrictive about what DID formats are supported at the message validation level.

aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
shaangill025: I have created `DIDValidation` pattern [less retrictive] in `messaging/valid.py`.

andrewwhitehead (approved):

PR #1207, opened 2021-05-27
Fix message types of problem report 1.0

baegjae:
Similar fix with https://github.com/hyperledger/aries-cloudagent-python/pull/1190
Thanks!

@sklump 


sklump:
Also covered in https://github.com/hyperledger/aries-cloudagent-python/pull/1212

PR #1210, opened 2021-05-27
BUG FIX: set connections metadata

Luis-GA:
Due https://github.com/hyperledger/aries-cloudagent-python/pull/1186#issuecomment-849452360 

There was a bug that does not save the metadata into the records once they were set.
This PR fixes that bug, updating the connections metadata everywhere.

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/routes.py
+++ b/aries_cloudagent/protocols/connections/v1_0/routes.py
@@ -451,6 +451,8 @@ async def connections_metadata_set(request: web.BaseRequest):
         for key, value in body.get("metadata", {}).items():
             await record.metadata_set(session, key, value)
         result = await record.metadata_get_all(session)
+        record.metadata = result
+        await record.save(session)
     except StorageNotFoundError as err:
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
This seems excessive, although it is a real problem. I think all that is needed is to clear the cache key: https://github.com/hyperledger/aries-cloudagent-python/blob/2b542949495313904ad9f5493d3769f58eb43af8/aries_cloudagent/connections/models/conn_record.py#L409

andrewwhitehead:
Actually no, that won't do it. At the moment I can't see why the connections endpoint is returning metadata at all. It certainly shouldn't be stored on the record.

Luis-GA:
Hi @andrewwhitehead the connections request returns metadata because I added it into the `ConnRecord` for the following PR, https://github.com/hyperledger/aries-cloudagent-python/pull/1186

dbluhm:
Metadata is not part of the connection record. Metadata records are separate but are tagged with a connection record ID in order to recall metadata about a specific connection record. Not saving directly to the connection record on metadata_set endpoint is the correct behavior since there is not actually any new data to save to the connection record.

PR #1213, opened 2021-05-27
Support indy 1.16 predicate restrictions

sklump:
Still to work out:
* general WQL (+ `"$and"`) for restrictions in proof request requested attributes/predicates
* not necessarily relying on these being a list: they can be a (WQL statement) dict or a list of (WQL statement) dicts, each (WQL statement) dict possibly being its own compound WQL statement; e.g., this is a valid restriction:
```
{
  "$and": [
    {
      "$or": [
        {
          "schema_name": "membership"
        },
        {
          "schema_name": "members"
        }
      ]
    },
    {
      "$neq": {
        "schema_version": "0.1"
      }
    },
    {
      "issuer_did": "abc123"
    },
    {
      "$not": {
        "$in": {
          "attr::name::value": [
            "joe",
            "dolly",
            "bill"
          ]
        }
      }
    }
  ]
}
```

### DIFF
--- a/aries_cloudagent/indy/sdk/models/proof_request.py
+++ b/aries_cloudagent/indy/sdk/models/proof_request.py
@@ -15,38 +15,11 @@ from ....messaging.models.base import BaseModel, BaseModelSchema
 from ....messaging.models.openapi import OpenAPISchema
 from ....messaging.valid import (
     INDY_CRED_DEF_ID,
-    INDY_DID,
     INDY_PREDICATE,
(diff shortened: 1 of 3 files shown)


sklump:
Reference: https://hyperledger-indy.readthedocs.io/projects/sdk/en/latest/docs/design/011-wallet-query-language/README.html

But "$and" is in the code where it's implicit in the standard.

ianco (approved):

PR #1143, opened 2021-05-04
Feature/didweb

domwoe:
Work in progress PR to add the did web capabilities to ACA-PY

* Native did:web resolver
* Serve did document under `/.well-known/did.json`
* Create a did document from wallet content (uses public DID and endpoint) + additional keys (e.g.) BBS+ keys can be referenced
* Connection/DID Exchange based on implicit invitation with public did:web

Multitenancy is currently not supported.
Needs cleaning and tests.
I'm not an experienced Python programmer so there might be stupid things I've done. Would be cool if someone could take a look.

Example usage:
Assumption: ACA-Py has configured a public DID, admin interface with HTTPS (e.g. via ngrok) and a BBS+ did:key 
```
POST:  /didweb​/create-from-wallet​/{did}

{
  "verification_methods": [{
"did":"did:key:zUC71v7BaQAEpNCN9wVetcqMWtWPygSuQ2t4MJ1H9654Aio3DypS8wCd253CZ29C1CiLSMmC8MrepFYvKrvdMHBatyEoQa5pffr8HMqvRR98Vb7NtEBkpN9Ld73jyeyAqYxg8Fy", "
verification_relationships": ["assertion_method"]
}]}
```

will produce:
```
{
  "@context": "https://www.w3.org/ns/did/v1",
  "id": "did:web:478eb39f6464.ngrok.io",
  "verificationMethod": [
  {
    "id": "did:web:478eb39f6464.ngrok.io#key-1",
    "type": "Ed25519VerificationKey2018",
    "controller": "did:web:478eb39f6464.ngrok.io",
    "publicKeyBase58": "DAwrZwgMwkTVHUQ8ZYAmuvzwprDmX8vFNXzFioxrWpCA"
  },
  {
     "id": "did:web:478eb39f6464.ngrok.io#key-2",
     "type": "Bls12381G2Key2020",
     "controller": "did:web:478eb39f6464.ngrok.io",
     "publicKeyBase58": 
"n5ZJWiW1TkL9jzpoig99Db9UjQ8hN4L8UuRTfEcFRqSEpNroSGoUDd5XQ2nwUuAhJ9MK5wzqSMzxNzCWC1qs51i5cEBii2ie1i9XXCWG1dyWXKr9jRyETJdmWUEHFzoodef"
}
],
"authentication": [
"did:web:478eb39f6464.ngrok.io#key-1"
],
"assertionMethod": [
"did:web:478eb39f6464.ngrok.io#key-1",
[...]

PR #1218, opened 2021-06-02
Feature/native did web resolver

domwoe:
Add native [did:web](https://w3c-ccg.github.io/did-method-web/) resolver. Carved out and touched up from #1143 

- Uses DID and DID Doc validation from pyDID
- ~Does not support encoding of ports in DIDs yet (e.g. `did:web:localhost%3A8443 -> https://localhost:443/.well-known/did.json` yet, because pyDID does not allow '%' char in DID (See: https://github.com/dbluhm/pydid/issues/34)~

### DIFF
--- a/aries_cloudagent/resolver/default/web.py
+++ b/aries_cloudagent/resolver/default/web.py
@@ -1,19 +1,23 @@
 """Web DID Resolver."""
 
-from typing import Sequence, Pattern
-import aiohttp
 import urllib.parse
+
+from typing import Sequence, Pattern
+
+import aiohttp
+
+from pydid import DID, DIDDocument
 
 from ...config.injection_context import InjectionContext
 from ...core.profile import Profile
+from ...messaging.valid import DIDWeb
+
 from ..base import (
     BaseDIDResolver,
     DIDNotFound,
     ResolverError,
     ResolverType,
 )
-from ...messaging.valid import DIDWeb
-from pydid import DID, DIDDocument
 
 
 class WebDIDResolver(BaseDIDResolver):
--- a/aries_cloudagent/resolver/default/web.py
+++ b/aries_cloudagent/resolver/default/web.py
@@ -1,0 +1,80 @@
+"""Web DID Resolver."""
+
+import json
+from typing import Sequence, Pattern
+import aiohttp
+import urllib.parse
+
+from ...config.injection_context import InjectionContext
+from ...core.profile import Profile
+from ..base import (
+    BaseDIDResolver,
+    DIDNotFound,
+    ResolverError,
+    ResolverType,
+)
+from ...messaging.valid import DIDWeb
+from pydid import DID, DIDDocument, DIDDocumentError
+
+
+class WebDIDResolver(BaseDIDResolver):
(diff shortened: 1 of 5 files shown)


aries_cloudagent/resolver/default/web.py
```diff
@@ -0,0 +1,79 @@
...
+from ...config.injection_context import InjectionContext
+from ...core.profile import Profile
+from ..base import (
+    BaseDIDResolver,
+    DIDNotFound,
+    ResolverError,
+    ResolverType,
+)
+from ...messaging.valid import DIDWeb
+from pydid import DID, DIDDocument
```
andrewwhitehead, line 16: Please reorder imports:

```
import urllib.parse

from typing import Sequence, Pattern

import aiohttp

from pydid import DID, DIDDocument

from ...config.injection_context import InjectionContext
from ...core.profile import Profile
from ...messaging.valid import DIDWeb

from ..base import (
    BaseDIDResolver,
    DIDNotFound,
    ResolverError,
    ResolverType,
)
```

aries_cloudagent/resolver/default/web.py
```diff
@@ -0,0 +1,83 @@
...
+        """Perform required setup for Web DID resolution."""
+
+    @property
+    def supported_did_regex(self) -> Pattern:
+        """Return supported_did_regex of Web DID Resolver."""
+        return DIDWeb.PATTERN
+
+    def supported_methods(self) -> Sequence[str]:
+        """Return list of supported methods."""
+        return ["web"]
```
dbluhm, line 40: This method should no longer be required, technically.

dbluhm (approved):

swcurran:
I was trying this out and have found that while `did:web:did.actor:alice` resolves properly (e.g. matches more or less what the [DIF Resolver](https://resolver.identity.foundation/) returns. However, the other actors DIDs (e.g. mike -- `did:web:did.actor:mike`, URL: https://did.actor/mike/did.json ) does not resolve successfully.  For Mike, the error is:

```bash
Error: Response was incorrectly formatted. Failed to deserialize DIDDocumentRoot:. 1 validation error for ParsingModel[DIDDocumentRoot]. __root__ -> @context -> 1. str type expected (type=type_error.str). [ErrorWrapper(exc=ValidationError(model='DIDDocumentRoot', errors=[{'loc': ('@context', 1), 'msg': 'str type expected', 'type': 'type_error.str'}]), loc=('__root__',))].
```

I'm guessing this is because of the JSON-LD contexts in the Mike document (and Carol). Is that the issue?

Thanks

dbluhm:
PyDID is complaining about the embedded object in the `@context` which isn't strictly spec conforming. At this point, I'm seriously questioning whether we'll ever be able to expect DID Documents to be even mostly spec compliant lol... I thought I had loosened expectations around parsed DID Documents in use by ACA-Py though. Was this error raised while just querying the plugged in resolvers through `/resolver/resolve/{did}` or another operation?

swcurran:
Yes -- I was using `/resolver/resolve/{did}`

PR #1220, opened 2021-06-02
Feature/didx request with public did

domwoe:
Introduces an optional flag `use_public_did` in `/didexchange/create-request`

Enabling the flag will create a DIDX request with the public DID of the requester.
The approach is not ideal since the did doc will still be attached to the request instead of just setting the `did` property as suggested in the [RFC](https://github.com/hyperledger/aries-rfcs/blob/master/features/0023-did-exchange/README.md#request-message-attributes) (but not required :) ).

Setting only the `did` property leads to issues because of inconsistencies in ACA-Py's DID Doc class and the DID Doc class returned by the resolver.


aries_cloudagent/protocols/didexchange/v1_0/tests/test_manager.py
```diff
@@ -236,6 +233,30 @@ async def test_create_request_implicit(self):
 
             assert conn_rec
 
+    # async def test_create_request_implicit_use_public_did(self):
```
domwoe, line 236: I'd need some help with this test. I want to setup a wallet with a public DID and assert that the did in the connection record is equal to the public DID, but I'm not sure how to setup/mock a wallet with a public DID

domwoe:
This would be useful for BPA 2 BPA connections @ianco

aries_cloudagent/protocols/didexchange/v1_0/tests/test_manager.py
ianco: Hi @domwoe sorry for the delay in responding to this.

Take a look at the tests for the in memory wallet, there is an example of creating a public did:

https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/wallet/tests/test_in_memory_wallet.py#L339

ianco (changes requested):

aries_cloudagent/protocols/didexchange/v1_0/tests/test_manager.py
domwoe: Thx! This was actually pretty simple. Don't know why I couldn't get it working back then...

ianco (approved):

PR #1231, opened 2021-06-08
fix regex for did:web pattern

### LINKED ISSUE #1228: did:web resolver: DIDWeb.PATTERN too permissive
The regular expression used for `DIDWeb.PATTERN` is currently too permissive causing other DIDs (such as `did:github:dbluhm`) to be pushed to the `did:web` resolver for handling.

The current value used is: `^(did:web:)?([a-zA-Z0-9%._-]*:)*[a-zA-Z0-9%._-]+$`

I'm not sure how this is used elsewhere but at least for resolver selection, I think the `?` on `(did:web:)` is the part causing us trouble. If the `did:web` part is intended to be optional for other uses, we might need to use a different regex for resolver selection. The resolver selection regex is really used only to sort out whether it should handle a given DID and does not need to be exhaustive. Further validation can be performed by the resolver itself after selection. `^did:web:.*$` would likely be sufficient.

@domwoe

domwoe:
- removes the accidental `?`
- resolves #1228 

### DIFF
--- a/aries_cloudagent/messaging/valid.py
+++ b/aries_cloudagent/messaging/valid.py
@@ -228,7 +228,7 @@ class DIDWeb(Regexp):
     """Validate value against did:web specification."""
 
     EXAMPLE = "did:web:example.com"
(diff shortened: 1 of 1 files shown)


dbluhm (approved):
That did the trick. Thanks for the fix!

PR #1234, opened 2021-06-09
Include record_id in the returned list of credentials

ianco:

Not sure if this is the best approach, but need to get the record_id of each credential to be able to add these to the presentation.


shaangill025 (approved):

TimoGlastra (approved):
LGTM

PR #1235, opened 2021-06-09
Update Alice/Faber demo to include json-ld credentials and proofs

ianco:
Still work in progress - working on updates to the documentation, and adding BDD tests.  Logic in the demo is pretty straightforward and will be enhanced.

The holder `credentialSubject.id` (in the credential) must be specified in order for the presentation to validate.  I have added this as an optional parameter to the `/create-request` endpoint, if the holder supplied a did then it will be inserted into the credential request as the `credentialSubject.id`.  (Not sure if this is the best approach @TimoGlastra @andrewwhitehead @swcurran )


demo/AliceWantsAJsonCredential.md
```diff
@@ -0,0 +1,560 @@
...
+                }
+              }
+            ]
+          }
+        }
+      ]
+    },
+    "record_ids": [
+      "1496316f972e40cf9b46b35971182337"
+    ]
```
TimoGlastra, line 508: This should be a dict right?

ianco: Good catch thanks!

aries_cloudagent/protocols/issue_credential/v2_0/formats/ld_proof/handler.py
```diff
@@ -353,14 +353,20 @@ def _get_proof_purpose(
...
+    ) -> LDProofVCDetail:
         # Add BBS context if not present yet
         if (
             detail.options.proof_type == BbsBlsSignature2020.signature_type
             and SECURITY_CONTEXT_BBS_URL not in detail.credential.context_urls
         ):
             detail.credential.add_context(SECURITY_CONTEXT_BBS_URL)
 
+        # add holder_did as credentialSubject.id (if provided)
+        if holder_did and holder_did.startswith("did:key"):
```
andrewwhitehead, line 367: I don't understand why this would be limited to did:key, shouldn't any DID be treated the same here?

andrewwhitehead (approved):

PR #1247, opened 2021-06-16
DIF PE Updates and present_proof_v1_0 OOB Attachment Fix

### LINKED ISSUE #1240: DIF PE handler should not determine whether "schema.uri" is type or context based on "#"
As I already mentioned in https://github.com/hyperledger/aries-cloudagent-python/pull/1125#discussion_r630778201 it is wrong to assume the `schema.uri` in PE is an `@type` if it contains a `#`.

Take a look at this playground: https://tinyurl.com/ye9whcca 

the expanded types result in:
```json
    "@type": [
  	  "https://www.w3.org/2018/credentials#VerifiableCredential",
      "http://schema.org/Person"
    ]
```

https://github.com/hyperledger/aries-cloudagent-python/blob/319a2ff30d9ddd4a9fdb4b3ab69d1946c110c21b/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py#L201-L204

@shaangill025

### LINKED ISSUE #1242: DIF PE handler throws error when the only "schema.uri" is VC
Is there a specific reason the DIF PE handler throws an error if the `schema.uri` only contains `https://www.w3.org/2018/credentials#VerifiableCredential`? 

Currently getting the following error in AATH:

```
aries_cloudagent.protocols.present_proof.v2_0.formats.handler.V20PresFormatHandlerError: Only expanded type extracted from CREDENTIALS_CONTEXT_V1_URL and VERIFIABLE_CREDENTIAL_TYPE included as the schema.uri
```

I think we should promote more explicit PE definitions, but it don't see the need to restrict it.

@shaangill025

### LINKED ISSUE #1244: Error deriving proof when non BBS credential is available
I run into an error in the following scenario:

1. Issue non-ZKP Ed25519 credential
2. Issue exactly the same credential, but with BBS signature
3. Create a proof request with `limit_disclosure: required`
4. ACA-Py tries to derive a credential with the first non-ZKP credential and errors because that's not possible.

I think if the signature suite does not support derivation/limit disclosure, it should not try to derive the credential but check beforehand if derivation is possible

Also in this case there is a valid credential, but the other, non-ZKP, credential is loaded first which breaks the whole flow.

@shaangill025

shaangill025:
- resolve #1244 
- resolve #1242 
- resolve #1240 
- present_proof_v1 and OOB attachment fix

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
+++ b/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
@@ -17,7 +17,6 @@ from ......vc.ld_proofs import (
     BbsBlsSignatureProof2020,
     WalletKeyPair,
 )
-from ......vc.ld_proofs.constants import EXPANDED_TYPE_CREDENTIALS_CONTEXT_V1_VC_TYPE
 from ......vc.vc_ld.verify import verify_presentation
 from ......wallet.base import BaseWallet
 from ......wallet.key_type import KeyType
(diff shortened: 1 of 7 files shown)


aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
```diff
@@ -186,47 +185,111 @@ async def create_pres(
...
+                )
                 for schema in input_descriptor.schemas:
+                    tag_query_or_list = []
                     uri = schema.uri
-                    required = schema.required or True
+                    if schema.required is None:
+                        required = True
+                    else:
+                        required = schema.required
                     if required:
```
andrewwhitehead, line 207: As discussed, I think this tag filtering belongs in VCHolder so that it can be specialized for the specific backends, but it's fine to merge this as-is and I can revisit it later with the shared-components work. Most important is that it works for now.

andrewwhitehead (approved):

PR #1249, opened 2021-06-17
Update get/set_did_public to use a storage record pointer

andrewwhitehead:
This should be much faster when there are many DIDs stored in the wallet.

The 'public' DID metadata property is now generally ignored except when auto-upgrading an existing wallet. The `get_posted_dids` method is updated to NOT filter the public DID, as it did previously.

Includes a couple fixes for warnings when running tests.

### DIFF
--- a/aries_cloudagent/wallet/indy.py
+++ b/aries_cloudagent/wallet/indy.py
@@ -1,5 +1,6 @@
 """Indy implementation of BaseWallet interface."""
 
+from aries_cloudagent.storage.record import StorageRecord
 import json
 
 from typing import List, Sequence, Tuple, Union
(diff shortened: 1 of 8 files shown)


aries_cloudagent/wallet/indy.py
```diff
@@ -1,5 +1,6 @@
 """Indy implementation of BaseWallet interface."""
 
+from aries_cloudagent.storage.record import StorageRecord
```
TimoGlastra, line 3: absolute import

andrewwhitehead: thanks!

swcurran (approved):
LGTM (but I don't know...) and I ran the performance demo and all good.  Not sure that is sufficient.

PR #1250, opened 2021-06-19
DIF PE - VCHolder backend-specific tag_query building and handle no given_id credentials

### LINKED ISSUE #1253: DIF PE Presentation Submission and optional id property in W3C credentials
[W3C data model specification](https://www.w3.org/TR/vc-data-model/#identifiers)
`id` property is optional, this is stored as `given_id` in `VCRecord`.
DIF PE handler uses this `given_id` in generating the `descriptor_map` for `presentation_submission`. `given_id` is also used to check if the number of applicable credentials meets the requirement specification [`submission_requirement`]
- If `id` is not included, there will be issues in DIF PE handler as currently implemented. I think as a fix, the `given_id` of the credential could either be set to a generated `UUID4` or to the `record_id`. @TimoGlastra Any feedback?
- If `id` is included, then according to spec, `the value of the id property MUST be a single URI. It is RECOMMENDED that the URI in the id be one which, if dereferenced, results in a document containing machine-readable information about the id`. As it will be unique for each credential so there shouldn't be any issues.

shaangill025:
- Addresses the following from #1247, @andrewwhitehead will this meet the requirement? 
> As discussed, I think this tag filtering belongs in VCHolder so that it can be specialized for the specific backends, but it's fine to merge this as-is and I can revisit it later with the shared-components work. Most important is that it works for now.
- resolve #1253
- Verification mechanism to check if the received presentation matches against the `constraints` in `presentation_definition`

### DIFF
--- a/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
+++ b/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
@@ -1032,14 +1032,9 @@ class DIFPresExchHandler:
             for descriptor_id in result.keys():
                 credential_list = result.get(descriptor_id)
                 for credential in credential_list:
-                    if credential.given_id:
-                        if credential.given_id not in cred_uid_descriptors:
-                            cred_uid_descriptors[credential.given_id] = {}
-                        cred_uid_descriptors[credential.given_id][descriptor_id] = {}
-                    else:
-                        if credential.record_id not in cred_uid_descriptors:
-                            cred_uid_descriptors[credential.record_id] = {}
-                        cred_uid_descriptors[credential.record_id][descriptor_id] = {}
+                    cred_id = credential.given_id or credential.record_id
+                    if cred_id:
+                        cred_uid_descriptors.setdefault(cred_id, {})[descriptor_id] = {}
 
             if len(result.keys()) != 0:
                 nested_result.append(result)
@@ -1094,27 +1089,17 @@ class DIFPresExchHandler:
 
                 if key in result:
                     for credential in result[key]:
-                        if credential.given_id and credential.given_id not in uid_dict:
+                        cred_id = credential.given_id or credential.record_id
+                        if cred_id and cred_id not in uid_dict:
                             merged_credentials.append(credential)
-                            uid_dict[credential.given_id] = {}
-                        elif (
-                            not credential.given_id
-                            and credential.record_id not in uid_dict
-                        ):
-                            merged_credentials.append(credential)
-                            uid_dict[credential.record_id] = {}
+                            uid_dict[cred_id] = {}
 
                 for credential in credentials:
-                    if credential.given_id and credential.given_id not in uid_dict:
-                        if (key + (credential.given_id)) not in exclude:
+                    cred_id = credential.given_id or credential.record_id
+                    if cred_id and cred_id not in uid_dict:
+                        if (key + cred_id) not in exclude:
                             merged_credentials.append(credential)
-                            uid_dict[credential.given_id] = {}
-                    elif (
-                        not credential.given_id and credential.record_id not in uid_dict
-                    ):
-                        if (key + (credential.record_id)) not in exclude:
-                            merged_credentials.append(credential)
-                            uid_dict[credential.record_id] = {}
+                            uid_dict[cred_id] = {}
                 result[key] = merged_credentials
         return result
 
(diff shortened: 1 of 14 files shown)


shaangill025:
This PR is ready for review.

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
```diff
@@ -1187,12 +1206,18 @@ async def create_vp(
...
-            ).startswith("urn:"):
+            if len(cred.subject_ids) > 0 and not self.check_if_cred_id_derived(
+                next(iter(cred.subject_ids))
+            ):
                 return True
         return False
 
+    def check_if_cred_id_derived(self, id: str) -> bool:
+        """Check if credential or credentialSubjet id is derived."""
+        if id.startswith("urn:") and "c14" in id:
```
TimoGlastra, line 1217: I think you can be more explicit:

```suggestion
        if id.startswith("urn:bnid:_:c14n")
```

shaangill025:
@TimoGlastra As requested, I have now implemented the mechanism to verify that the received presentation matches against the `presentation_request`.

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
```diff
@@ -1030,22 +1032,27 @@ async def apply_requirements(
...
-                        given_id_descriptors[credential.given_id] = {}
-                    given_id_descriptors[credential.given_id][descriptor_id] = {}
+                    if credential.given_id:
+                        if credential.given_id not in cred_uid_descriptors:
+                            cred_uid_descriptors[credential.given_id] = {}
+                        cred_uid_descriptors[credential.given_id][descriptor_id] = {}
+                    else:
+                        if credential.record_id not in cred_uid_descriptors:
+                            cred_uid_descriptors[credential.record_id] = {}
+                        cred_uid_descriptors[credential.record_id][descriptor_id] = {}
```
andrewwhitehead, line 1042: This can be simplified a bit, maybe something like:

```py
cred_id = credential.given_id or credential.record_id
if cred_id:
  cred_uid_descriptors.setdefault(cred_id, {})[descriptor_id] = {}
```

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
```diff
@@ -1187,12 +1206,18 @@ async def create_vp(
...
-            ).startswith("urn:"):
+            if len(cred.subject_ids) > 0 and not self.check_if_cred_id_derived(
+                next(iter(cred.subject_ids))
+            ):
                 return True
         return False
 
+    def check_if_cred_id_derived(self, id: str) -> bool:
+        """Check if credential or credentialSubjet id is derived."""
+        if id.startswith("urn:bnid:_:c14n"):
```
andrewwhitehead, line 1217: This seems too specific to me, wouldn't a different blank node identifier be generated depending on the ordering of the normalized statements?

aries_cloudagent/storage/vc_holder/in_memory.py
```diff
@@ -19,6 +19,16 @@ def __init__(self, profile: InMemoryProfile):
...
 
+    def build_type_or_schema_query(self, uri_list: Sequence[str]) -> dict:
+        """Build and return in-memory backend specific type_or_schema_query."""
+        type_or_schema_query = {"$and": []}
+        for uri in uri_list:
+            tag_or_list = []
+            tag_or_list.append({f"type:{uri}": "1"})
+            tag_or_list.append({f"schm:{uri}": "1"})
+            type_or_schema_query["$and"].append({"$or": tag_or_list})
+        return type_or_schema_query
```
andrewwhitehead, line 30: Thanks! This should make things much easier to adjust for the Askar backend.

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
shaangill025: When deriving, I think all blank node identifiers are converted into `urn:bnid:_:c14nX` node ids, @TimoGlastra any suggestions?

andrewwhitehead: So the [BBS+ Signatures 2020 draft](https://w3c-ccg.github.io/ldp-bbs2020/) doesn't seem to be explicit about the format of the blank node identifiers. I think it's trying to say that it's covered by the canonicalization algorithm, but the link is broken. Assuming this is the the correct link, it doesn't really specify the format of the identifiers either although it gives some guidelines:
https://json-ld.github.io/rdf-dataset-canonicalization/spec/#blank-node-identifier-issuer-state

In any case, I think the prefix you're using is probably consistent with the current implementations.

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
shaangill025: Simplified using the above template in a few places.

andrewwhitehead (approved):

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
TimoGlastra: This is indeed how current implementations handle it: https://github.com/w3c-ccg/ldp-bbs2020/issues/10#issuecomment-616216278

I'm working on a PR to add this to the spec

PR #1266, opened 2021-06-24
Fix performance demo, clean up some warnings

andrewwhitehead:


swcurran (commented):
Looks good.  It's still not that fast -- 0.28 per second on my system.  Is that expected?﻿

swcurran (approved):

andrewwhitehead:
@swcurran That doesn't seem right, I generally get over 5/s on my system. Maybe docker is limiting the resources.

PR #1267, opened 2021-06-25
Add support for shared components

andrewwhitehead:
- Add `indy-vdr` ledger client
- Add `indy-credx` credential handling
- Add `aries-askar` storage backend

The `performance`, `alice`, and `faber` demo agents should all support the `--wallet-type askar` argument for testing with the new backend components, as does the agent itself when run with `aca-py`.


swcurran (dismissed):
Tested via a number of mechanisms (BDD tests, AATH tests, Alice/Faber, etc.).  When this branch is run in Indy mode, all is good. When run in Askar mode things are mostly (but not completely) good.  As such, we'll merge this now as an optional, alpha feature that is behind a feature flag ("--wallet-type askar"), and evolve the implementation from there.

Outstanding work, @andrewwhitehead !!!  w00t!!  
FYI - @jljordan42

swcurran (approved):

ianco (approved):

PR #1273, opened 2021-06-28
Fix manual event handling

ianco:

Fixes the OpenAPI example


demo/runners/alice.py
```diff
@@ -68,13 +68,14 @@ async def handle_connections(self, message):
         if (
             message["connection_id"] == self.connection_id
             and message["rfc23_state"] == "completed"
-            and not self._connection_ready.done()
+            and (self._connection_ready and not self._connection_ready.done())
```
shaangill025, line 71: This is not necessary/required with changes to `input_invitation`.

shaangill025 (approved):

PR #1274, opened 2021-06-28
Add `wait_for_event` method on event bus

dbluhm:
We've found this to be generally useful in some of our coding efforts and after a brief discussion with @esune noting that it would be of use in some of his ongoing work, we figured it would be good to pop this out of our ongoing work so it can be used elsewhere.


esune:
Thanks @dbluhm! Changes look good to me, not sure why the integration tests are failing though...

dbluhm:
Integration test failure was:
```
Failing scenarios:
  features/0453-issue-credential.feature:56  Issue a credential with revocation, with the Issuer beginning with an offer, and then revoking the credential -- @1.2
```
I've run into this intermittently before so seeing if re-running the integration tests has a different result.

ianco (approved):

PR #1275, opened 2021-06-28
Update bdd tests to support askar; fix ledger method signatures

ianco:
Ledger methods are updated to provide endorser support for revocation updates

bdd updates are to allow tests to be run using askar wallet


swcurran (approved):
LGTM.  I haven't run both tests with and without and compared.  Did you?  If so, what was the result?

ianco:
> LGTM. I haven't run both tests with and without and compared. Did you? If so, what was the result?

I can run the bdd tests with `--wallet-type askar` (with a couple of small tweaks) and the tests run fine if I run individually; but when I run the full suite with `askar` I get weird behaviour with some tests failing and/or hanging

swcurran:
Hmmm...weird. So a test will fail in the full suite, but run fine individually?  That sucks...

swcurran:
In test T004-RFC0453 I'm seeing this: `aries_cloudagent.transport.error.RecipientKeysError: ('Error trying to extract recipient keys from JWE', Error('Incorrect padding',))`

swcurran:
I guess I should be putting stuff like that into issues, not into this PR :-) ... Doh.

PR #1276, opened 2021-06-29
DIF PE - giving Holder more control on how to specify presentation_request

### LINKED ISSUE #1262: Giving the Controller control over what to use to satisfy a presentation request when using DIF PE
When processing a proof request, the general flow within ACA-Py is:

- Search the wallet for the credentials that could be used to satisfy the presentation
- Construct the presentation using the identified credentials

As well, at some point, the Controller needs to be contacted to provide consent in responding to the presentation request.

In some cases, the search of the wallet will return multiple credentials that could be used in the presentation, and a selection is made to choose the one(s) that will be used in the presentation sent to the verifier.  Some examples:

- The request might be for a single type of credential, but the holder has more than one of those in their wallet.
- In DIF PE, there are places where the user can specify "ORs" about what is to be presented (e.g. "present your passport or driver's license credential") and the prover holds credentials that satisfy several of the "OR" conditions.
- The request might allow the presentation of a credential from one of multiple issuers, and the prover's wallet contains credentials from more than one of the permitted issuers.

The desired flow is that the prover code in ACA-Py should default to picking one (e.g. the most recent one issued in the first case), and then the Controller shown the request, the selected credential(s) to be used, and the list of other credentials that could be used.  The Controller can then respond with the credentials that they want to use for the presentation.  This "selection" process should be included as part of the process of getting consent to send the presentation from the Controller.

If the "auto" flag is set for sending a proof request, then the default set of credentials should be used to send the proof.

This capability is in place (at least I think it is...) in the Indy processing, and it should also be in place for DIF PE.

Let's discuss the complexity of this and make sure it is a reasonable request.  I think it is, but...

shaangill025:

- resolve #1262 
[includes detail about the approach and discussion around the use cases]


aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
```diff
@@ -292,6 +293,11 @@ async def create_pres(
                 # For now, setting to 1000
                 max_results = 1000
                 records = await search.fetch(max_results)
+                # sort records by issuanceDate in reverse_order
+                records.sort(
+                    key=lambda v: dateutil_parser(v.cred_value.get("issuanceDate")),
+                    reverse=True,
+                )
```
andrewwhitehead, line 300: I think this needs to be a little more tolerant of errors. It looks like `dateutil_parser` will throw an error if the issuanceDate is undefined, and it probably would if the date's not in a supported format.

aries_cloudagent/protocols/present_proof/v2_0/routes.py
```diff
@@ -602,6 +602,11 @@ async def present_proof_credentials_list(request: web.BaseRequest):
                     pd_uri_list=uri_list,
                 )
                 records = await search.fetch(count)
+                # sort records by issuanceDate in reverse_order
+                records.sort(
+                    key=lambda v: dateutil_parser(v.cred_value.get("issuanceDate")),
+                    reverse=True,
+                )
```
andrewwhitehead, line 609: Same as above, maybe this needs to be split into a utility method.

aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
shaangill025: The `issuanceDate` is required and must be in `RFC3339` format. This is enforced in `./issue_credential/v2_0/formats/ld_proof/handler.py` when receiving and storing credential with `./vc/vc_ld/models/credential.py -> VerifiableCredential`. I can validate against `RFC3339DateTime` before sorting but I think it will be a redundant check.
https://www.w3.org/TR/vc-data-model/#issuance-date

andrewwhitehead: Okay, I was just thinking to catch exceptions if any occur.

shaangill025: I get your point regarding catching exceptions, I will add that in.

shaangill025: Exception handling added.

andrewwhitehead (approved):

PR #1284, opened 2021-06-30
PR#1276 Rework - Moved sorting of VCRecords into VCHolder

shaangill025:
- Following discussion with Andrew, decided to move the sorting logic into `./storage/vc_holder` for all 3 backends. 
- If there is a `dateutil` exception triggered when sorting then return the `records` list as it is
- Reverted changes made to `./present_proof/v2_0/formats/dif/handler.py`, `./present_proof/v2_0/routes.py` and corresponding tests.


aries_cloudagent/storage/vc_holder/tests/test_indy_vc_holder.py
```diff
@@ -61,11 +61,211 @@ def test_record() -> VCRecord:
...
+            },
+        },
     )
 
 
 @pytest.mark.indy
 class TestIndySdkVCHolder(in_memory.TestInMemoryVCHolder):
     # run same test suite with different holder fixture
-    pass
+
```
andrewwhitehead, line 94: If these two tests are the same as the ones in `TestInMemoryVCHolder`, then they don't need to be duplicated here

shaangill025: Now removed.

shaangill025:
This should be ready now.

andrewwhitehead (approved):

PR #1286, opened 2021-07-01
Add outbound message transport events

dbluhm:
This PR introduces events relating to outbound messages. These changes are looking forward to more separation between the core ACA-Py message processing and generation components and message delivery mechanisms, which is fraught with nuance and largely dictated by use case. In such a state where there is a separation between message generation and delivery, we propose that these events (and/or similar events) act as the primary interface between the two. For now, however, these events are simply emitted and do not replace any of the previous functionality with one significant exception: considering how `OutboundSendStatus` was returned up the call stack from `outbound_message_router` and beyond, an `OutboundStatusEvent` seemed preferable. This change is mostly self-contained except for the recent `forward` message handler additions made to help with emitting push notifications to clients of mediators.

@TimoGlastra we were hoping to get your opinion on this change. As implemented, do you think the new `OutboundStatusEvent` would provide enough context to a push notifier or do we want to continue with a specific forward message event encapsulating the send status of the forwarded message?


andrewwhitehead:
Hi @dbluhm, there's one test failing at the moment in TestForwardHandler:

```py
        # emit event that a forward message is received (may trigger webhook event)
        await context.profile.notify(
            "acapy::forward::received",
            {
                "connection_id": recipient.connection_id,
>               "status": send_status.value,
                "recipient_key": context.message.to,
            },
        )
E       AttributeError: 'NoneType' object has no attribute 'value'
```

dbluhm:
Indeed, this was actually a semi-intentional failure. This is the subject of my question for @TimoGlastra as his thoughts will influence updates to the forward message handler and its tests. Given that, it's likely appropriate to mark this as a draft. I'll go ahead and do that.

aries_cloudagent/core/conductor.py
```diff
@@ -8,6 +8,10 @@
 
 """
 
+from aries_cloudagent.core.transport_events import (
+    OutboundStatusEvent,
+    OutboundMessageEvent,
+)
```
TimoGlastra, line 14: Absolute import

TimoGlastra (approved):
LGTM.

Seems like a more general approach instead of solely focused on forward messages. How wil this impact the forward webhook API call? Can we base that on the new event?

dbluhm:
My thinking on this feature has changed slightly. I like the modification of turning the outbound send status to an emitted event rather than a value that gets returned up a deep call stack but I'm not as confident about the benefit of some of the other pieces in this PR. I will close this and open a new one when appropriate.

PR #1294, opened 2021-07-06
DIF PE - is_holder property implementation - compliant with DIF spec

### LINKED ISSUE #1261: RFC 0510 - ACA-Py DIF PE is_holder support
-[ RFC 0510](https://github.com/hyperledger/aries-rfcs/tree/master/features/0510-dif-pres-exch-attach) in regards to `is_holder` states that `preferred or required as defined in the spec.Note that this feature allows the Holder to present credentials with a different subject identifier than the DID used to establish the DIDComm connection with the Verifier.`
- This means when `is_holder` is required we should only present and sign those credentials where `credentialSubject.id` is the same as the ConnRecord my_did.
- Since, current connection DID will be `did:sov:`, it will leave out all of BbsBls credentials and not support limit_disclosure. So is_holder cannot be used alongside limit_disclosure

Is this as expected and should be implemented in ACA-Py like this? To support this:
- When fetching W3C credentials, check if `is_holder`
- add `subject_ids = [ConnRecord.my_did]` as a parameter and call `search_credentials()`

shaangill025:
- resolve #1261
- I think this meets the following attributes from the DIF spec ~~but does not use the `is_holder` property to decide whether the `VP` should be signed or not. But in case, this is needed then this can be done with minimal changes/extension~~.
1. `Verify that the Holder or Subject of the Claim still controls the identifier [DID].` [Link](https://identity.foundation/presentation-exchange/#proof-of-identifier-control)
2. `Holder of  Claim is the same as the Subject of the attribute.` [Link](https://identity.foundation/presentation-exchange/#note-2)

### DIFF
--- a/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
+++ b/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
@@ -193,7 +193,7 @@ class DIFPresExchHandler:
         else:
             reqd_key_type = KeyType.ED25519
         for cred in applicable_creds:
-            if len(cred.subject_ids) > 0:
+            if cred.subject_ids and len(cred.subject_ids) > 0:
                 if not issuer_id:
                     for cred_subject_id in cred.subject_ids:
                         if not cred_subject_id.startswith("urn:"):
@@ -1188,7 +1188,7 @@ class DIFPresExchHandler:
         submission_property = PresentationSubmission(
             id=str(uuid4()), definition_id=pd.id, descriptor_maps=descriptor_maps
         )
-        if self.is_holder and self.check_sign_pres(applicable_creds):
+        if self.is_holder:
             (
                 issuer_id,
                 filtered_creds_list,
@@ -1240,17 +1240,6 @@ class DIFPresExchHandler:
                     return signed_vp
             else:
                 return vp
-
-    def check_sign_pres(self, creds: Sequence[VCRecord]) -> bool:
-        """Check if applicable creds have CredentialSubject.id set."""
-        for cred in creds:
-            if (
-                cred.subject_ids
-                and len(cred.subject_ids) > 0
-                and not self.check_if_cred_id_derived(next(iter(cred.subject_ids)))
-            ):
-                return True
-        return False
 
     def check_if_cred_id_derived(self, id: str) -> bool:
         """Check if credential or credentialSubjet id is derived."""
--- a/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
+++ b/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
@@ -38,6 +38,7 @@ from ....vc.ld_proofs.constants import (
 )
 from ....vc.vc_ld.prove import sign_presentation, create_presentation, derive_credential
 from ....wallet.base import BaseWallet, DIDInfo
+from ....wallet.did_method import DIDMethod
 from ....wallet.key_type import KeyType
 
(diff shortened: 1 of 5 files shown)


TimoGlastra:
> does not use the is_holder property to decide whether the VP should be signed or not

I'm not dure I fully understand. Could you elaborate a bit?

shaangill025:
> > does not use the is_holder property to decide whether the VP should be signed or not
> 
> I'm not dure I fully understand. Could you elaborate a bit?

According to the DIF spec, `is_holder` is associated with `fields[*].id`, so I extended `filter_constraints()` and this is how it is handled:
- we first apply a `field` to a credential to check if it is applicable
- If it is applicable and the `field.id` is specified inside `is_holder` with `required directive`, then call `process_constraint_holders()` which checks if the `credentialSubject.id` is one of the `DIDs` for the holder. If so,  then the credential is added for the `VP`, otherwise, that credential is not added.

For signing the VP, the following logic is being used:
- If all 3 credentials have `did:...:test` as `credentialSubject.id`, VP is signed with suite having `public_key_base58` as `did:...:test's did_info.verkey`
- If one or more of the 3 credentials have `did:...:test` as `credentialSubject.id` and the remainder have no credentialSubject.id, VP is signed with suite having `public_key_base58` as `did:...:test's did_info.verkey`
- If one of 3 credentials have `did:...:other_test` and the other 2 have `did:...:test` as `credentialSubject.id`, an exception is raised stating Applicable credentials have different `credentialSubject.id`, multiple proofs are not supported currently

andrewwhitehead:
> If all 3 credentials have `did:...:test` as `credentialSubject.id`, VP is signed with suite having `public_key_base58` as `did:...:test's did_info.verkey`

I'm a little confused by this. It sounds like the VP is being signed using Ed25519Signature2020 with the subject's private key. But that only makes sense to me if the credential is self-issued, is that the scenario? Or is the signature doing something else here, maybe validating the presentation for a specific VP request or establishing ownership of the key.

aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
```diff
@@ -400,6 +416,41 @@ async def filter_constraints(
...
+
+    async def process_constraint_holders(
+        self,
+        subject_ids: Sequence[str],
+    ) -> bool:
+        """Check if holder or subject of claim still controls the identifier."""
+        if not self.local_dids:
+            await self.local_dids_list()
+        for subject_id in subject_ids:
+            if subject_id not in self.local_dids:
```
andrewwhitehead, line 450: It may be much more efficient in general to look up the relevant subject IDs in the wallet individually rather than loading every DID and then going through them.

shaangill025:
> > If all 3 credentials have `did:...:test` as `credentialSubject.id`, VP is signed with suite having `public_key_base58` as `did:...:test's did_info.verkey`
> 
> I'm a little confused by this. It sounds like the VP is being signed using Ed25519Signature2020 with the subject's private key. But that only makes sense to me if the credential is self-issued, is that the scenario? Or is the signature doing something else here, maybe validating the presentation for a specific VP request or establishing ownership of the key.

`VP` is being signed with the subject's private key. It makes sense as long as the `holder` is the same as the `subject` [not when the credential is self-issued]. I remember Timo mentioning that if we sign the `VP` with `verkey` of some other holder DID then the VP cannot be verified. Thinking about this,  `is_holder` property should be used to decide whether to sign the `VP` or not.

andrewwhitehead (approved):

PR #1296, opened 2021-07-06
Support generic public DIDs in DIDX request

domwoe:
Be more permissive in validation of target public DID in `didexchange/create-request`.
Public DIDs are resolved using the resolver interface which supports additional DID methods besides `did:sov` (e.g. `did:web`).

Only question is if we should still allow unqualified DIDs.


andrewwhitehead (approved):
Thanks!

PR #1298, opened 2021-07-06
check if did is already qualified

domwoe:
PR #1296 introduced a problem that  when the `pthid` got set to the public DID in a DIDX request with an already qualified DID it would still get prefixed with `did:sov`.


domwoe:
Sorry, missed this... :(

andrewwhitehead (approved):

PR #1301, opened 2021-07-07
fix: failure to verify jsonld on non-conformant doc but vaild vmethod

dbluhm:
Minor fix for jsonld verify route.

### DIFF
--- a/aries_cloudagent/messaging/jsonld/routes.py
+++ b/aries_cloudagent/messaging/jsonld/routes.py
@@ -152,9 +152,15 @@ async def verify(request: web.BaseRequest):
                     )
 
                 if not isinstance(ver_meth_expanded, VerificationMethod):
-                    raise InvalidVerificationMethod(
-                        "verificationMethod does not identify a valid verification method"
-                    )
+                    try:
+                        ver_meth_expanded = VerificationMethod.deserialize(
+                            ver_meth_expanded.serialize()
+                        )
+                    except ValueError:
+                        raise InvalidVerificationMethod(
+                            "verificationMethod does not identify a valid "
+                            "verification method"
+                        )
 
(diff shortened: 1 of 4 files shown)


dbluhm:
Tests to hit the error path pending, hoping to get this ready for merge ASAP

aries_cloudagent/messaging/jsonld/routes.py
```diff
@@ -152,9 +152,15 @@ async def verify(request: web.BaseRequest):
                     )
 
                 if not isinstance(ver_meth_expanded, VerificationMethod):
-                    raise InvalidVerificationMethod(
-                        "verificationMethod does not identify a valid verification method"
-                    )
+                    try:
+                        ver_meth_expanded = VerificationMethod.deserialize(
+                            ver_meth_expanded.serialize()
```
andrewwhitehead, line 157: What is it if not a VerificationMethod?

dbluhm: The dereference method can also be used to dereference a Service. The method itself has a return value of `Resource` which is the base `VerificationMethod` and `Service` build from. It's a little on the complicated side due to attempts to work with documents that don't strictly conform to the schema outlined in the DID spec. In that case, a plain `Resource` object is returned since the more feature rich parsing failed. We give that parsing another try for just the identified resource here.

Alternatively, I could see this being simpler: we default to dereferencing resources from a DID document as if the doc itself was non-conforming but modify the `dereference` method to also take a parameter asserting that the dereferenced resource is a `VerificationMethod` or `Service` and "casts" the resource to one of these objects (or whatever type that was passed in) or fails.

andrewwhitehead (approved):

Issue #1256, opened 2021-06-23
posture inconsistency issue when assigning public did


baegjae:
https://github.com/hyperledger/aries-cloudagent-python/pull/1249
I like this approach. I would like to report some issues after merging this PR.

### posture inconsistency issue
I requested POST `/wallet/did/public?did=k1ZNyPEzq4yi3kSdw4AYA`, and I got below return value.
```
{
    "result": {
        "did": "k1ZNyPEzq4yi3kSdw4AYA",
        "verkey": "QScuSjT8NmnJAfAChd4v2CyxtygSDmgKsSRxrkbJ2YU",
        "posture": "wallet_only",
        "key_type": "ed25519",
        "method": "sov"
    }
}
```
Before merging the PR, the posture was `public`.
I can get the did by requesting GET `/wallet/did/public`, but I still think the `public` posture is right.

### connection error when using public invitation
In the situation above, Faber encountered an error when receiving a connection request from Alice.
`receipt.recipient_did_public` was `None`, and it incurs an error here.

https://github.com/hyperledger/aries-cloudagent-python/blob/e94229289a6319d194360089b173da7d19f761f6/aries_cloudagent/protocols/connections/v1_0/manager.py#L490

Thanks!
@andrewwhitehead

PR #1305, opened 2021-07-09
Correct public did checking method in public invitation connection

baegjae:
Previously, `posted` metadata was used to check a public did.
However, multiple dids could have `posted` metadata, even though these dids are not a public did.
This PR uses the `get_public_did()` method to get and check a public did.

Also, This PR resolves https://github.com/hyperledger/aries-cloudagent-python/issues/1256

Thanks!

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/manager.py
+++ b/aries_cloudagent/protocols/connections/v1_0/manager.py
@@ -1063,7 +1063,8 @@ class ConnectionManager(BaseConnectionManager):
                     receipt.recipient_verkey
                 )
                 receipt.recipient_did = my_info.did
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
Essentially the 'posted' flag means that a DID has been published somewhere, ie. it is a public DID. Previously the DID might have a 'public' flag which meant that it was the default public DID used by the system, and the only DID that was acceptable for DID-based invitations. Along with the move to using a separate record to point to the public DID, we changed the connection manager to allow any posted DID to be used for this purpose. If necessary, we could add another metadata property to disallow invitations via a particular DID.

baegjae:
Oh I understand. You are changing the concept of public DID. I agree that this concept is more flexible. Thanks for the explanation.

PR #1316, opened 2021-07-14
OOB Protocol - CredentialOffer Attachment Support

shaangill025:
- Extends work done by @DaevMithran in #1216 
- Adds a `create-offer` endpoint in `issue-credential-v1` and `issue-credential-v2` to create an offer independent of connection.
- Support both `issue-credential-v1` and `issue-credential-v2` credential offer workflow 
[Faber] with auto flags, offer_sent -> request_received -> credential_issued
[Alice] with auto flags, offer_received -> request_sent -> credential_received
- Test coverage


aries_cloudagent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -172,7 +181,9 @@ async def create_invitation(
                         a_id,
                     )
                     message_attachments.append(
-                        InvitationMessage.wrap_message(cred_ex_rec.cred_offer.offer())
+                        InvitationMessage.wrap_message(
+                            cred_ex_rec.cred_offer.serialize()
```
andrewwhitehead, line 185: I see here it's using `cred_ex_rec.cred_offer` but at line 175 it's using `cred_ex_rec.credential_offer_dict`, is that correct?

aries_cloudagent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -637,7 +689,7 @@ async def _process_pres_request_v1(
                     holder=self._session.inject(IndyHolder),
                 )
             except ValueError as err:
-                self._logger.warning(f"{err}")
+                LOGGER.warning(f"{err}")
```
andrewwhitehead, line 692: Maybe add some context so it's clear where the logged error is coming from

andrewwhitehead: Actually I think you can use `LOGGER.exception("...")` and avoid capturing `err` manually.

aries_cloudagent/protocols/out_of_band/v1_0/manager.py
shaangill025: `cred_offer` in `V20CredExRecord` is same as `credential_offer_dict` in `V10CredentialExchange`. I think this is correct.

andrewwhitehead (approved):

PR #1331, opened 2021-07-23
Initial DIDComm v2 envelope support via Askar backend

andrewwhitehead:
Also cleans up and relocates the (incomplete) in-memory ECDH-1PU support.

This does not make the DIDComm v2 envelope format(s) available for communication yet, it only establishes some of the basic functionality.


swcurran:
What about adding a "DIDComm2Envelopes.md" that talks about how to use this?  How do we get this tested?

@ianco @shaangill025 -- can you please take a look at the code, at least?

Thanks!  This is another very cool addition.

swcurran:
Also -- is there any feedback to add to RFC 587 after the implementation?  I know you are commenting in the DIDComm2 Working Group, but wanted to make sure what's needed is getting back to the Aries community.  Thanks!

andrewwhitehead:
@swcurran There isn't really a way to use it at the moment, at least for agent-to-agent communication. We still need to implement a standard key lookup interface for the agent's own DID document keys as well as other agents' keys via the DID resolver (and cache). This is mainly a proof of concept to see what the complexity is and what it left to do.

swcurran:
OK -- perhaps @troyronda and @Baha-sk might be interested in taking a look?

PR #1347, opened 2021-08-11
feat: set arbitrary plugin config value at cli

dbluhm:
Doing a bit of spit balling.

We previously addressed the need for supplying configuration to plugins with the `--plugin-config` flag to load a yaml file into `settings` for use by plugins. We've found this feature to be quite helpful, especially in the case of loading multiple plugins (i.e. resolvers) that each require their own configuration. However, we have also found that it would be handy to be able to set or modify these values from the command line. So instead of having to modify the `plugin_config.yml` you're using to load the universal resolver plugin to point it to a different resolver instance, we could have a command line argument like the following:
```
--plugin-config-value http_uniresolver.endpoint=http://localhost:3000
```

Or perhaps a short form of:
```
-o http_uniresolver.endpoint=http://localhost:3000
```

With such a command line argument, we propose that dotted key values act as a shorthand for nesting values. For example, `a.b.c.d=value` becomes `{"a": {"b": {"c": {"d": "value"}}}}`. We also propose that the value be loaded as yaml to avoid parsing ourselves while still allowing us to express values more complicated than just strings and numbers.

Open to thoughts :slightly_smiling_face:

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -1,6 +1,7 @@
 """Command line option parsing."""
 
 import abc
+from functools import reduce
 from os import environ
 
 import yaml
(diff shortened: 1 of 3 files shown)


burdettadam (approved):

andrewwhitehead:
Seems reasonable to me, can you write something like `--plugin-config-value "a.b.c=[one,two]"` ?

dbluhm:
Yep, `[one, two]`, `{inline: yaml_dict}`, `true`, etc. all get parsed into the expected values

andrewwhitehead (approved):

PR #1348, opened 2021-08-11
Refactor outbound queue interface

dbluhm:
This PR adjusts the outbound queue interface to be more easily generalized. As we've worked to implement outbound queuing via Kafka, we've found that the configuration just did not map onto what was needed in Kafka. Rather than expecting a connection string and parsing protocol etc. etc., this PR just passes all settings to the outbound queue implementation and allows it to read plugin configuration from the settings. This affords much greater flexibility in the structure of the outbound queue implementation. As a result, I removed some outbound queue command line arguments that were no longer used.

Additionally, I updated aioredis to 2.0.0. From the aioredis maintainers speaking on version 1.3.1 of the library:

> As of December, 2019, this library had been abandoned. The code had grown stale, issues were piling up, and we were rapidly becoming out-of-date with Redis’s own feature-set. In light of this, the decision was made to ensure that the barrier to support for potential maintainers or contributors would be minimized.

I found that using aioredis 2.0.0 significantly simplified the code as it stood.

Some other minor consistency fixes were made to bring the outbound queue package in line with the rest of the code base.

I haven't had an opportunity to integration test the redis changes just yet. I noticed there were some tests for this but they did not appear to be run anywhere in the tests run by the github actions. Is this something we would want run as part of PR checks or is the occasional one-off sufficient?


dbluhm:
Alternatively, I wonder if the redis implementation should live a little bit further distant from the ACA-Py core. Have we considered including a separate package in the repo for these sorts of plugins?

burdettadam (approved):

andrewwhitehead (approved):

Issue #1349, opened 2021-08-11
Multitenancy with a single database and multiple profiles


MaticDiba:
We are trying to use ACA-Py with multitenancy enabled (using Askar with Postgres for storage). The current implementation of multitenancy is limiting for our use case because it creates a new database for each tenant agent. This is problematic with a growing number of possible tenants. 

We need a solution where each tenant wallet could be stored in the same database. One approach would be to enable storing tenant wallets to single multi-tenant sub wallet (separate database from main base wallet) as separate Profiles. With this approach, mapping to each tenant would be done similarly to the current approach - with the base wallet. But instead of creating a new database for each tenant, we would just use one "tenant sub wallet" with multiple profiles. And then profile id can be used to map to the correct tenant.

Askar library already supports creating multiple profiles in one database. Is there an item on the road map or proposed solution to enable this approach in Aca-Py as well?

Any suggestion/discussion would be much appreciated.

swcurran:
@andrewwhitehead , @TimoGlastra  -- any comments on this?  I know this has been discussed. Did we get to the start of a design for this?  Andrew -- pretty sure we talked about this when we were in Utah at the last Connect-a-thon :-).

Did Kiva do something like this and could that design be leveraged?

bartekus:
My recommendation here would be to have a look at https://postgrest.org and also https://aws.amazon.com/blogs/database/multi-tenant-data-isolation-with-postgresql-row-level-security/ for the specifics about achieving multi-tenancy that has scalability as well as security as their main objectives. 

By externalizing the data layer to a easily manageable API interface and relying on Postgres row-level access control, it's entirely possible to facilitate a proper access for a administrative organization to each wallet that it manages but also enable each individual holder ability to access their data with minimal overhead and development cost.

Has anyone thus far considered this approach? If yes, what downsides have been determined, if no, how could I be of service in terms of presenting this as viable option for others to consider/use?

PR #1368, opened 2021-08-23
#1349 Added option to multitenant mode to handle multiple wallets inside the same subwallet with many Askar profiles

dejsenlitro:
This mode can be run by specifying askar-profile as the wallet_type in the CLI argument --multitenancy-config
Added AskarProfileMultitenantManager class which handles the new logic to get the wallet profile. The old logic was moved to MultitenantManager. The both extend BaseMultitenantManager where most of the previous code remained
The MultitenantManagerProvider chooses the right manager to use depending on the CLI argument


acuderman:
@TimoGlastra @andrewwhitehead @swcurran Hey, this PR is awaiting a review.  Thanks in advance.

aries_cloudagent/multitenant/manager_provider.py
```diff
@@ -0,0 +1,49 @@
...
+class MultitenantManagerProvider(BaseProvider):
+    """
+    Multitenant manager provider.
+
+    Decides which manager to use based on the settings.
+    """
+
+    MANAGER_TYPES = {
+        "basic": MultitenantManager,
+        "askar-profile": AskarProfileMultitenantManager,
```
andrewwhitehead, line 24: I think this needs to use DeferLoad from utils.classloader in order to avoid importing the Askar classes until needed - as that is not currently a required dependency.

aries_cloudagent/askar/profile.py
```diff
@@ -54,6 +54,15 @@ def store(self) -> Store:
         """Accessor for the opened Store instance."""
         return self.opened.store
 
+    def opened(self) -> AskarOpenStore:
```
andrewwhitehead, line 57: Doesn't this method clash with the `opened` instance property?

aries_cloudagent/multitenant/manager_provider.py
acuderman: Agree changed, thanks

aries_cloudagent/askar/profile.py
acuderman: Yes, It is not needed.  Removed

aries_cloudagent/config/argparse.py
```diff
@@ -1439,4 +1465,18 @@ def get_settings(self, args: Namespace):
 
             if args.multitenant_admin:
                 settings["multitenant.admin_enabled"] = True
+
+            if args.multitenancy_config:
+                multitenancyConfig = json.loads(args.multitenancy_config)
```
andrewwhitehead, line 1470: There may be a more consistent way to support these options between the YAML argument handling in other places and the generic plugin configuration mechanism, but I'm okay merging this as-is for now. @dbluhm

aries_cloudagent/multitenant/admin/routes.py
```diff
@@ -1,5 +1,6 @@
 """Multitenant admin routes."""
 
+from aries_cloudagent.multitenant.base import BaseMultitenantManager
```
andrewwhitehead, line 3: This, and other instances should be a relative import and sorted consistently with the other imports. I'll clean it up after merging (since I can't amend this PR).

andrewwhitehead (approved):

PR #1366, opened 2021-08-22
Documentation and unit test cleanups

andrewwhitehead:

PR #1371, opened 2021-08-23
Updates to eliminate warnings on RTD doc generation

swcurran:

I added a mock for `aries_cloudagent.vc`, which eliminated all of the errors. Of course I suspect that it also removes the doc generation for that module, but am not certain.  But I feel better that all the warnings are gone...


andrewwhitehead:
My other PR should fix most of the warnings I think, if you want to test docs generation there.

swcurran:
Pretty sure I did, but I can try again.

swcurran:
I'm guessing you are talking about #1366?  I ran the generation on top that and got many of the same errors.  Once I mocked the one ACA-Py module (vc) on top of that PR, all but one warning (below) went away.   That said, I need to update this PR after the PR is merged, as the RST generator needs to run on the latest code.  It would be nice to have that run on every build and added (if needed) to each PR.  Not sure how to do that...

The remaining error is this.  Is this a code error?

```bash
WARNING: error while formatting arguments for aries_cloudagent.protocols.issue_credential.v1_0.models.credential_exchange.V10CredentialExchange: unhashable type: 'list'

```

andrewwhitehead:
I need more context to figure out what's causing that error, and I'm not set up to test it right now. If the errors are just warnings then I'm not sure we need to disable documentation for that module.

swcurran:
I've updated this branch with your merged PR and all is good now.

We're still generating the docs for that module (vc) AFAIK -- we're just not allowing it to be referenced from where it is called.  Not sure why we have to do that. FYI -- there had been two other modules for which we used to do that, but I just tested them and that is no longer needed.

The one error is left, but I can't figure out why it is happening.  I checked various things to do with Sphinx, but can't trace it.  I'm a bit worried it's a code issue, but that's beyond me.

This is now ready to merge.

swcurran:
@ianco -- is this integration error anything to do with my changes (seems very unlikely...) or an intermittent error?

Do I fake a commit to retrigger?

ianco (approved):

ianco:
> @ianco -- is this integration error anything to do with my changes (seems very unlikely...) or an intermittent error?
> 
> Do I fake a commit to retrigger?

This is an intermittent error.  You can re-run the rests from the workflow page (click on the details link and then there should be a button to rerun the workflow)

PR #1376, opened 2021-08-24
feat: add inject_or

dbluhm:
This PR removes the `required` parameter from `Injector.inject` and introduces `inject_or` that mimics the behavior of `inject(..., required=False)`. The reasoning for this change is that type checkers take issue with using the result of `inject(..., required=True)` when the signature says that it could still be `None`. For example:

```python
storage = context.inject(BaseStorage)
records = await storage.find_all_records(...)
```
Yields the error `"find_all_records" is not a known member of "None"` when using pyright, even though we know that the call will have resulted in an error if it was actually `None`.

This PR is mostly a proposal and does not yet fix the 109 or so instances of `inject(..., required=False)` that should be switched to `inject_or`. Open to suggestions or alternatives.

### DIFF
--- a/aries_cloudagent/admin/request_context.py
+++ b/aries_cloudagent/admin/request_context.py
@@ -114,15 +114,15 @@ class AdminRequestContext:
     def _test_session(self) -> ProfileSession:
         session = self.profile.session(self._context)
 
-        def _inject(base_cls, required=True):
+        def _inject(base_cls):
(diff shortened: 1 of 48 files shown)


dbluhm:
Given no expressed opposition lol, I went ahead and finished the implementation on this to be able to consider these changes holistically.

andrewwhitehead (approved):
Looks good, thanks!

PR #1377, opened 2021-08-25
fix: send proposal schema

TimoGlastra:
send-proposal doesn't have `connection_id` in the schema, which is inconvenient for generated clients


swcurran:
@ianco @shaangill025 -- could one or both of you please take a look? Timo's asked this be added sooner than later.

Thanks

shaangill025 (approved):

swcurran (approved):
LGTS (Looks Good to Shaanjot).  And me.

Issue #1365, opened 2021-08-20
Getting 400 Response on connectionless presentation on calling /verifiy-presentation


swcurran:
Issue created from this question on Hyperledger Chat: https://chat.hyperledger.org/channel/aries-cloudagent-python?msg=7qeKFQtQcwnHimKR8

We are trying to verify a connectionless proof which state has been set to “presentation_received”. We are using the release 0.7.0 of aca-py (also tried with 0.6.0) to spin up an agent, accompanied by the tails server for revocation functionality.

We use the following startup params:

```
--endpoint http://<IP_ADDRESS>:8020
--label Hyper42MandateService
--auto-ping-connection
--auto-respond-messages
--auto-accept-invites
--inbound-transport http 0.0.0.0 8020
--outbound-transport http
--admin aries-cloudagent-runner 8021
--admin-insecure-mode
--auto-provision
--wallet-type indy
--wallet-name <somenamehere>
--wallet-key <somekeyhere>
--seed <someseedhere>
--preserve-exchange-records
--genesis-url http://test.bcovrin.vonx.io/genesis
--webhook-url http://mandate-webhook-service:8022/webhooks
--tails-server-base-url http://docker_tails-server_1:6543
--trace-target log
--trace-tag acapy.events
--trace-label localAriesCloudAgent.trace
```

However, we are receiving a 400: Record ID not provided when we try to verify a connectionless proof request (when calling `POST /present-proof/records/{pres_ex_id}/verify-presentation`).

Logging which we found in the agent has been added as an attachment (agentlogging.txt).

We looked through Aries agent code on how it handles presentation verification (present-proof v1.0, file routes.py, method presentation_exchange_verify_presentation) and it seems that it tries to fetch Presentation Exchange by given pres_ex_id, which should be fine, and it doesn't actually fail right there. But on line 820 it wants a connection ID, and then it tries to fetch a connection record, and it fails, which seems to result in a Bad Request error. Present-proof v2.0 seems not to differ too much in this regard, but we have not tested that yet.

So our question: is this the expected behaviour? Are connectionless proof requests not able to be verified at this time or ever, or could it be something else?

After some review of vc-authn-oidc's implmentation, I see one big difference between our implementation and vc-authn-oidc.
For vc-authn-oidc aca-py is started with the option auto-verify-presentation. We do not use that option.
[...]

PR #1383, opened 2021-09-01
Remove connection check on proof verify

ianco:

Not required in general (if we are verifying a proof it means we have already received the proof) and causes an error on connectionless proof requests.


ianco:
Fixes:  https://github.com/hyperledger/aries-cloudagent-python/issues/1365

andrewwhitehead (approved):

Issue #1291, opened 2021-07-05
Update Alice/Faber Credential and Proof Request to use Aries RFC "Date Int" format


swcurran:
A fun one for @ianco when he has some time.

Please update the Alice/Faber demo to change the "age" attribute to be "DateOfBirth" and use the "DateInt" recommended format as outlined in this recently merged [PR to Aries RFCs](https://github.com/hyperledger/aries-rfcs/pull/687).  Once done, change the proof request to use that style instead of the current ">18" check.  

It would be good to demo that in practice.

ianco:
Fixed in PR https://github.com/hyperledger/aries-cloudagent-python/pull/1395

PR #1395, opened 2021-09-03
Connectionless proof demo

ianco:


ianco:
Also fixes issue https://github.com/hyperledger/aries-cloudagent-python/issues/1291

swcurran (approved):
A couple of things that would be nice to change, but not holding it up because of them:

1. It would be nice if Alice accepted the invitation and processed the connectionless proof, either be making the existing “new invitation” fancier or by adding a new option.

2. It would be nice that instead of hard-coding the predicate dateIn value, that you showed the calculation to get the date — e.g. convert_to_dateInt(now() -years(21)).

ianco:
> A couple of things that would be nice to change, but not holding it up because of them:
> 
> 1. It would be nice if Alice accepted the invitation and processed the connectionless proof, either be making the existing “new invitation” fancier or by adding a new option.
> 2. It would be nice that instead of hard-coding the predicate dateIn value, that you showed the calculation to get the date — e.g. convert_to_dateInt(now() -years(21)).

Re #2, I'll update in my next PR.

Re #1, I'm not sure what this means, since it's a connectionless proof ...

swcurran:
Re: 1 — A connection-less proof is just an invitation (plain-text) given to an any agent to which they respond (if they want, of course) with a proof.  Alice could be that agent — separate from her existing connection with Faber.  So her agent just gets the invitation with the proof request and responds to it using the data in the invitation, without really knowing who is asking.

Issue #1333, opened 2021-07-25
Review the demo for the defaults set when AIP 20 is used


swcurran:
When `run_demo --aip 20` is used some options seem to be missing, including at least the following.  Please review all the options and adjust them as needed.

- `use_did_exchange` (aka `--did-exchange`) should be `true`
- the new style message type (using `https://didcomm.org`) should be used, not the old `did:sov:...`

Those I found to be missing.  Other things that should be set:

- use the V2 Credential Exchange protocols (RFC 0453/0454)

I think that is all, but could be wrong.

It would also be nice to see the agent usage scripts. For now, the only way to see them is to pass in a bogus option, e.g.

`./run_demo alice --foo`

I'm probably going to adjust the `run_demo` script so that `./run_demo --help` works, so maybe `./run_demo alice --help` could show the `./run_demo` usage and then the Alice usage.

swcurran:
Likewise, if the user selects `--aip 10`, we should set the AIP 1.0 options:

- `use_did_exchange` should be `false`
- we should not use the new style message type prefix (although that is arguable)
- use the V1 Credential Exchange protocols (RFC 0036/0037)

There might be others, but good enough for now.

It would be nice in the usage to show the allowed (or at least the format) options for AIP -- e.g. "10" and "20".

PR #1406, opened 2021-09-09
Fix aip 20 behaviour and other cleanup

ianco:
Fixes issue https://github.com/hyperledger/aries-cloudagent-python/issues/1333

### DIFF
--- a/demo/AriesOpenAPIDemo.md
+++ b/demo/AriesOpenAPIDemo.md
@@ -507,6 +507,8 @@ Finally, we need put into the JSON the data values for the `credential_preview` 
       }
 ```
 
+(Note that the birthdate above is used to present later on to pass an "age proof".)
+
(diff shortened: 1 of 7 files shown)


demo/AriesOpenAPIDemo.md
```diff
@@ -655,6 +657,8 @@ From the Faber browser tab, get ready to execute the **`POST /present-proof/send
 }
 ```
 
+(Note that the birthdate requested above is used as an "age proof", the calculation is something like `now() - years(18)`, and the presented birthdate must be on or before thsi date.  You can see the calculation in action in the `faber.py` demo code.)
```
swcurran, line 660: Typo "thsi"

swcurran (changes requested):
Just the typo

demo/AriesOpenAPIDemo.md
ianco: Fixed

swcurran (approved):

Issue #1402, opened 2021-09-09
Plain-text DIDComm messages are sent to the external outbound queue 


acuderman:
When using an external outbound queue. messages are not packed before they are pushed to the queue. That causes problems (apart from the security) when the recipient is a multitenant Aca-py agent since the relay can not figure out who is the recipient of the message.

[Messages should be packed in the similar way they are being when using the internal queue](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/transport/outbound/manager.py#L394)

dbluhm:
Addressed by #1408

PR #1408, opened 2021-09-10
Encode DIDComm messages before sent to the queue

acuderman:
## What is this PR for?

This PR fixes issue #1402


dbluhm (approved):
Tested and working with one of our queue implementations we're working on. Thanks for fixing this one.

andrewwhitehead (approved):
Verified locally that this doesn't seem to have any significant effects on performance

PR #1409, opened 2021-09-10
Log warning when unsupported problem report code is received

### LINKED ISSUE #1397: marshmallow.exceptions.ValidationError: {'_schema': ['Value for description.code must be issuance-abandoned']}
Issue created based on https://chat.hyperledger.org/channel/aries-cloudagent-python?msg=xZ7RwjjJNwdxpKzsF

Our ACA-PY agent (version 0.7.0) is connected to the BCOVRIN TEST (--genesis-url http://test.bcovrin.vonx.io/genesis).
When sending a credential to a user that uses connect.me we receive an error. Which is to be expected, as connect.me does not use the bcovrin test network.
But the the problem-report is not handled smoothly.

Error received from connect.me:
aries_cloudagent.transport.pack_format DEBUG Expanded message: {'@id': '9d661ea0-2336-41fe-b3af-db12b3c2438f', '@type': 'did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/issue-credential/1.0/problem-report', 'comment': 'error occurred: VcxError { inner: \n\nCould not find Credential Definition on the connected Ledger networks\n\nCredential Definition not found on the Ledger }', 'description': {'code': 'invalid-credential-offer', 'en': "Couldn't create credential-request for received credential-offer."}, '~thread': {'received_orders': {'did:sov:Cst5jmVuTpyQGgs7hg3Ybt': 0}, 'sender_order': 0, 'thid': 'd8655eb0-cbf5-43fb-be86-fea8c10b23f9'}}

This results in the following stacktrace in ACA-PY:
2021-09-01 12:47:16,618 aries_cloudagent.messaging.models.base ERROR CredentialProblemReport message validation error:
File "/home/indy/aries_cloudagent/messaging/models/base.py", line 137, in deserialize
return schema.loads(obj) if isinstance(obj, str) else schema.load(obj)
File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/marshmallow/schema.py", line 723, in load
data, many=many, partial=partial, unknown=unknown, postprocess=True
File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/marshmallow/schema.py", line 904, in doload
raise exc
marshmallow.exceptions.ValidationError: {'_schema': ['Value for description.code must be issuance-abandoned']}

shaangill025:
- resolve #1397

### DIFF
--- a/aries_cloudagent/protocols/coordinate_mediation/v1_0/messages/problem_report.py
+++ b/aries_cloudagent/protocols/coordinate_mediation/v1_0/messages/problem_report.py
@@ -52,22 +52,15 @@ class CMProblemReportSchema(ProblemReportSchema):
     def validate_fields(self, data, **kwargs):
         """Validate schema fields."""
 
-        if data.get("description", {}).get("code", "") not in [
+        if not data.get("description", {}).get("code", ""):
+            raise ValidationError("Value for description.code must be present")
+        elif data.get("description", {}).get("code", "") not in [
             prr.value for prr in ProblemReportReason
         ]:
-            if (
-                data.get("description")
-                and data.get("description").get("code")
-                and len(data.get("description").keys()) >= 2
-            ):
-                locales = list(data.get("description").keys())
-                locales.remove("code")
-                LOGGER.warning(
-                    "Unexpected error code received.\n"
-                    f"Code: {data.get('description').get('code')}, "
-                    f"Description: {data.get('description').get(locales[0])}"
-                )
-            raise ValidationError(
-                "Value for description.code must be one of "
-                f"{[prr.value for prr in ProblemReportReason]}"
+            locales = list(data.get("description").keys())
+            locales.remove("code")
+            LOGGER.warning(
+                "Unexpected error code received.\n"
+                f"Code: {data.get('description').get('code')}, "
+                f"Description: {data.get('description').get(locales[0])}"
             )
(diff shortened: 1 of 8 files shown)


andrewwhitehead:
Hm, I don't think this is quite what I had in mind. It's still raising a ValidationError, so the failure would remain. It would probably make sense for the problem report message classes to not validate the code at all (apart from it being a string), and to leave that up to the message handler - just logging a warning when it's not recognized.

shaangill025:
> Hm, I don't think this is quite what I had in mind. It's still raising a ValidationError, so the failure would remain. It would probably make sense for the problem report message classes to not validate the code at all (apart from it being a string), and to leave that up to the message handler - just logging a warning when it's not recognized.

I guess something like this:
```
@validates_schema
    def validate_fields(self, data, **kwargs):
        if not data.get("description", {}).get("code", ""):
            raise ValidationError("Value for description.code must be present")
        elif data.get("description", {}).get("code", "") not in [
            prr.value for prr in ProblemReportReason
        ]:
            if len(data.get("description").keys()) >= 2:
                locales = list(data.get("description").keys())
                locales.remove("code")
                LOGGER.warning(
                    "Unexpected error code received.\n"
                    f"Code: {data.get('description').get('code')}, "
                    f"Description: {data.get('description').get(locales[0])}"
                )
```

andrewwhitehead:
Yes, I think that would work

andrewwhitehead (approved):

PR #1413, opened 2021-09-14
 Make mediation invitation parameter idempotent

chumbert:
To ease Aca-py's deployment in scenarios where command line arguments
cannot be changed easily, make `--mediator-invitation` and
`--mediator-connections-invite` available to the `provision` command.

`provision` stores the mediator invite to be used by the agent (effectively "provisioning" it).
`start` uses the stored invite. Once used, it does not establish connection again.
If the invitation changes, the  stored value is replaced by the new one, the default mediator is cleared and the new invitation
is used to establish connection.

Boy-scouting: I split the mediation invite tests apart from the other ones for the sake of easier comprehension/code management.


dbluhm (dismissed):
LGTM!

aries_cloudagent/core/conductor.py
```diff
@@ -317,26 +319,45 @@ async def start(self) -> None:
...
+                            session, MediationManager.SEND_REQ_AFTER_CONNECTION, True
+                        )
+                        await conn_record.metadata_set(
+                            session, MediationManager.SET_TO_DEFAULT_ON_GRANTED, True
+                        )
+
+                        print("Attempting to connect to mediator...")
+                        del mgr
+            except Exception as e:
+                print(e)
```
andrewwhitehead, line 360: Extra print statement

andrewwhitehead:
I don't really understand what the goal is here. It seems like it's meant to store the mediation invitation if you add the argument to the provision command, but the conductor startup process is still looking for the `mediation.invite` setting instead of the stored invitation. It seems like the argument would need to be specified again in order to be accepted at startup?

chumbert:
> I don't really understand what the goal is here. It seems like it's meant to store the mediation invitation if you add the argument to the provision command, but the conductor startup process is still looking for the `mediation.invite` setting instead of the stored invitation. It seems like the argument would need to be specified again in order to be accepted at startup?

Our issue is that in an automated scalable environment, new aca-py instances are spinned up with the same arguments as the original instance, thus retaining the `mediation.invite` setting. We want to avoid these new instances creating new connections to the mediator if one is already established, hence the need for idempotency.

However, I agree with your point that the conductor should not necessarily require the `mediation.invite` to be specified if one was provisioned and is already stored. I'll make sure this can be either/or.

aries_cloudagent/protocols/coordinate_mediation/mediation_invite_store.py
```diff
@@ -0,0 +1,165 @@
...
+            raise NoDefaultMediationInviteException(
+                "No default mediation invite: cannot mark it as used."
+            )
+
+        updated_record = MediationInviteRecord(record.invite, used=True)
+        await self.store(updated_record)
+
+        return updated_record
+
+    async def get_mediation_invite_record(
```
andrewwhitehead, line 143: It seems like this method should store and return the provided invitation, if any. At the moment it does not return the provided invitation if there is no stored invitation.

chumbert: Indeed, I missed a case. Good catch

andrewwhitehead (approved):

Issue #1238, opened 2021-06-10
[Endorser Protocol] Add auto flags and default endorsement settings


esune:
When running an instance of ACA-Py with `AUTHOR` only privileges (no permission to write to the ledger), the agent needs to explicitly request endorsement to an authorized agent when performing any action that requires writing a new transaction to the ledger. Some of these transactions are:
- creating a schema
- creating a credential definition
- creating a revocation registry
- publishing a revocation registry entry

To facilitate the endorsement pattern, it would be nice to have new global settings and `auto` flags that will:
- allow the `endorser` to automatically endorse all received endorsement requests endorsement
- allow the `author` to automatically submit endorsement requests for all relevant objects
- allow both agents to automatically setup the required metadata on the connection records

As part of implementing the above functionality, it may be required to add:
- an `--agent-role` setting/startup parameter to determine what the agent capabilities are when interacting with the ledger, and what role is held when completing a transaction
- an `--auto-endorse-transactions` setting/startup parameter to tell the endorser to automatically process incoming endorsement requests
- a setting to identify which connection corresponds to the endorser agent on the `author` (maybe by label or public did?)

c.c.: @ianco

esune:
A couple of other thoughts about the above changes:
- the scope of refactoring will likely bleed into `revocation`, as it is another fairly complicated set tasks to orchestrate
- ideally, both endorsement and revocation are "transparent" to the user: this means that there should be no need to keep track of the sequence of steps required to endorse a transaction, or to keep a revocation registry active. Everything should be taken care of by aca-py, transparently to the user.
- some of the endorsemement/revocation endpoints might need to be turned to "read only mode" when appropriate (i.e.: auto flags set) in order to prevent erroneous calls to interfere with the state of the objects in aca-py (mostly thinking about revocation registries here)

esune:
High level design document with proposed changes/approach can be found [here](https://hackmd.io/@esune/Hk1wCXz3d).

swcurran:
Reviewed and added one comment.  Looks great to me!  Let's go for it.

esune:
Perfect, I'll setup a meeting with @ianco next week to plan the work then!

ianco:
Updated the high-level design to include a section describing how the event bus will be used:

https://hackmd.io/51flx8CVS1ypyOjEqVl1tg?view#Implementation-Details

swcurran:
Closing this as I think the feature is done. If not, please add new issues left to be completed.

PR #1422, opened 2021-09-23
Endorser protocol configuration, automation and demo integration

ianco:
See issue https://github.com/hyperledger/aries-cloudagent-python/issues/1238

This PR adds automation support for requesting, endorsing and writing transactions, and for executing the workflow of creating a cred def that supports revocation.

The next stage of work will re-factor the code to use the event bus and eliminate the code duplication (see https://hackmd.io/51flx8CVS1ypyOjEqVl1tg?view#Implementation-Details)

### DIFF
--- a/Endorser.md
+++ b/Endorser.md
@@ -1,0 +1,61 @@
+# Transaction Endorser Support
+
+Note that the ACA-Py transaciton support is in the process of code refactor and cleanup.  The following documents the current state, but is subject to change.
+
+ACA-Py supports an [Endorser Protocol](https://github.com/hyperledger/aries-rfcs/pull/586), that allows an un-privieged agent (an "Author") to request another agent (the "Endorser") to sign their transactions so they can write these transactions to the ledger.
+
+Transaction Endorsement is built into the protocols for Schema, Credential Definition and Revocation, and endorsements can be explicitely requested, or ACA-Py can be configured to automate the endorsement workflow.
+
+## Setting up Connections between Authors and Endorsers
+
+Since endorsement involves message exchange between two agents, these agents must establish and configure a connection before any endorsements can be provided or requested.
+
+Once the connection is established and `active`, the "role" (either Author or Endorser) is attached to the connection using the `/transactions/{conn_id}/set-endorser-role` endpoint.  For Authors, they must additionally configure the DID of the Endorser as this is required when the Author signs the transaction (prior to sending to the Endorser for endorsement) - this is done using the `/transactions/{conn_id}/set-endorser-info` endpoint.
+
+## Requesting Transaction Endorsement
+
+Transaction Endorsement is built into the protocols for Schema, Credential Definition and Revocation.  When executing one of the endpoints that will trigger a ledger write, an endorsement protocol can be explicitely requested by specifying the `connection_id` (of the Endorser connection) and `create_transaction_for_endorser`.
+
+(Note that endorsement requests can be automated, see the secion on "Configuring ACA-Py" below.)
+
+If transaction endorsement is requested, then ACA-Py will create a transaction record (this will be returned by the endpoint, rather than the Schema, Cred Def, etc) and the following endpoints must be invoked:
+
+| Protocol Step       | Author                          | Endorser                          |
+| -------------       | ------                          | --------                          |
+| Request Endorsement | `/transactions/create-request`  |                                   |
+| Endorse Transaction |                                 | `/transactions/{tran_id}/endorse` |
+| Write Transaction   | `/transactions/{tran_id}/write` |                                   |
+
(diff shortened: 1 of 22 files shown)


swcurran:
I didn't find anything in testing this using alice/faber, but I couldn't figure out how to activate it in the demo.  Do you have instructions?

We also need some tweaks to the run_demo help handling.  I'm going to put in some suggestions in a separate issue.

ianco:
> I didn't find anything in testing this using alice/faber, but I couldn't figure out how to activate it in the demo. Do you have instructions?

The demo is in an "interim" state right now, endorser support is added but not fully added.  The following is how it works now (sets up the endorser roles to allow manual testing using the swagger page) but not ultimately how I see the demo demonstrating endorsement ...

Run the following in 2 separate shells (make sure you are running von-network and the tails server first):

```
./run_demo faber --endorser-role endorser --revocation
./run_demo alice --endorser-role author --revocation
```

Copy the invitation from faber to alice to complete the connection.

Then in the alice shell, select option "D" and copy faber's DID to alice (it is the DID displayed on faber agent startup).

This starts up the aca-py agents with the endorser role set (via the new command-line args) and sets up the connection between the 2 agents with appropriate configuration.

Then, in the alice swagger page (http://localhost:8031) you can create a schema and cred def, and all the endorser steps will happen automatically.  You don't need to specify a connection id or explicitly request endorsement (aca-py does it all automatically based on the startup args).

swcurran:
Nice -- that worked.

I'm thinking we need to have a "von-network"-like Endorser so that we can start it when needed for testing.  Perhaps it should be built into von-network?   We should brainstorm a bit how to do that.  We also need to have one that we can put into production, ahead of a BPA-based Endorser.

I think an "Endorser.md" file is needed for this to document how to use it both in real usage, and a section on how to use it in the demo.  Seems like a fun thing to do on a Friday, no?  :-). I'm thinking that should be done before merging.

I'm hoping you can demo this on Tuesday morning at the ACA-Pug meeting.  Doable?

ianco:
> I think an "Endorser.md" file is needed for this to document how to use it both in real usage, and a section on how to use it in the demo. Seems like a fun thing to do on a Friday, no? :-). I'm thinking that should be done before merging.

I added some docs both for ACA-Py as well as the demo.  These docs will need to be updated as the Endorser code is updated & re-factored.

Endorser.md
```diff
@@ -0,0 +1,61 @@
+# Transaction Endorser Support
+
+Note that the ACA-Py transaciton support is in the process of code refactor and cleanup.  The following documents the current state, but is subject to change.
```
swcurran, line 3: Do you anticipate the API (e.g. Admin API and Startup Parameters) changing, or are those more or less set?

ianco: I don't anticipate any API changes but can't guarantee we won't need any

Endorser.md
```diff
@@ -0,0 +1,61 @@
...
+
+ACA-Py supports an [Endorser Protocol](https://github.com/hyperledger/aries-rfcs/pull/586), that allows an un-privieged agent (an "Author") to request another agent (the "Endorser") to sign their transactions so they can write these transactions to the ledger.
```
swcurran, line 5: Typo "un-privileged"

Endorser.md
swcurran: Typo "transaction"

Endorser.md
swcurran: Should have a reference to this being an Indy feature/requirement.

Endorser.md
```diff
@@ -0,0 +1,61 @@
...
+
+Transaction Endorsement is built into the protocols for Schema, Credential Definition and Revocation, and endorsements can be explicitely requested, or ACA-Py can be configured to automate the endorsement workflow.
+
+## Setting up Connections between Authors and Endorsers
+
+Since endorsement involves message exchange between two agents, these agents must establish and configure a connection before any endorsements can be provided or requested.
+
+Once the connection is established and `active`, the "role" (either Author or Endorser) is attached to the connection using the `/transactions/{conn_id}/set-endorser-role` endpoint.  For Authors, they must additionally configure the DID of the Endorser as this is required when the Author signs the transaction (prior to sending to the Endorser for endorsement) - this is done using the `/transactions/{conn_id}/set-endorser-info` endpoint.
```
swcurran, line 13: I think it should be possible to have the DID specified on the command line.

swcurran (approved):
Approved -- but you might want to clean up the typos first. Or just merge.

Endorser.md
ianco: Agree this is a command line parameter but not currently used (the auto-connection-setup support is still TBD)

ianco:
> Approved -- but you might want to clean up the typos first. Or just merge.

I'll include the doc updates/fixes in the next PR

Issue #1417, opened 2021-09-16
Askar wallet sessions result in a deadlock  


acuderman:
When using Askar wallet type combined with Postgres storage and calling some endpoints at the same time, deadlocks happen. The issue is that the session (DB connection) is acquired at the start of the request and is not dropped until the end. That would be ok (not performance-wise, since this limits the number of concurrent requests to the number of max DB connections) but there are some cases where the second session is created while the first one is still active.  
  
The issue arises if X concurrent requests acquire the first connection to the database but are unable to get the second one. In this case, all requests will be stuck waiting for an available connection (where X is at least the number of max database connections).  

### Example  

[create_request](https://github.com/hyperledger/aries-cloudagent-python/blob/64c5cb5c4a68ccf984881c78dc1b788704016c64/aries_cloudagent/protocols/didexchange/v1_0/manager.py#L217) handler is an example. At the initialization, a session is created and another session is created inside [get_default_mediator function](https://github.com/hyperledger/aries-cloudagent-python/blob/64c5cb5c4a68ccf984881c78dc1b788704016c64/aries_cloudagent/connections/util.py#L27), while the first session is still active.  
 
Another example is [receive_invitation](https://github.com/hyperledger/aries-cloudagent-python/blob/64c5cb5c4a68ccf984881c78dc1b788704016c64/aries_cloudagent/protocols/didexchange/v1_0/manager.py#L66) handler where new session is created inside the [send_reply](https://github.com/hyperledger/aries-cloudagent-python/blob/64c5cb5c4a68ccf984881c78dc1b788704016c64/aries_cloudagent/protocols/didexchange/v1_0/manager.py#L147) function (when auto_accept is enabled).  

### Proposed solution  

Due to the performance and deadlock issues, the proposed solution is to create sessions only on demand in the lower scope. The session should also be closed right after the wallet query. The mentioned approach is already used by some managers, for example in the [Mediation manager](https://github.com/hyperledger/aries-cloudagent-python/blob/64c5cb5c4a68ccf984881c78dc1b788704016c64/aries_cloudagent/protocols/coordinate_mediation/v1_0/manager.py#L353)

PR #1424, opened 2021-09-24
Open askar sessions only on demand - Connections

acuderman:
This PR fixes the connections part of the #1417 issue.

### DIFF
--- a/aries_cloudagent/core/in_memory.py
+++ b/aries_cloudagent/core/in_memory.py
@@ -49,6 +49,11 @@ class InMemoryProfile(Profile):
             ),
         )
 
+    def _init_context(self):
+        """Initialize the session context."""
+        self._context.injector.bind_instance(BaseStorage, STORAGE_CLASS(self))
+        self._context.injector.bind_instance(BaseWallet, WALLET_CLASS(self))
+
     def session(self, context: InjectionContext = None) -> "ProfileSession":
         """Start a new interactive session with no transaction support requested."""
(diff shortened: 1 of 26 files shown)


aries_cloudagent/core/in_memory.py
```diff
@@ -49,6 +49,11 @@ def bind_providers(self):
             ),
         )
 
+    def _init_context(self):
```
andrewwhitehead, line 52: I don't see why this is necessary, it's already performed for the session when that is created?

andrewwhitehead:
Thanks! Looks very thorough in general

aries_cloudagent/core/in_memory.py
acuderman: This is performed on the session only when `test_session ` function is called. But in this case we call `test_profile`.

Another option would be to call init_context every time we create a new session (in constructor) .

andrewwhitehead: The session must be awaited or used as an async context manager. When this happens the `_setup` method is called, which calls `_init_context` on the session. If the session has not been initialized this way then it is not technically 'open' (the in-memory backend is more forgiving about this currently). I think that adding this workaround is just masking the issue.

acuderman: Thanks for the explanation. We removed `_init_context` on the profile and created sessions with async context.

andrewwhitehead (approved):

PR #1425, opened 2021-09-27
Multiple Indy Ledger support and State Proof verification

shaangill025:
- https://hackmd.io/BVye073CTiqgNEMXxDeVdQ?view
- Detailed document describing how multiple ledger support works can be found at `./Multiledger.md` (this PR).


ianco (changes requested):
Hi @shaangill025 overall this looks pretty good.  I haven't done a deep dive yet though.

Once thing that would be nice is a document describing how the multi-ledger support works, or how to use it, for example:

- how to I define multiple ledgers?  (I know there is a  hackmd but should prob be a doc in the github repo)
- what do I need to do for read vs write?  (I assume the verifier has to be able to auto-select the ledger based on received proofs, but what about the issuer?)
- can you add/remove ledgers at runtime or only on startup?
- what happens if I restart an agent with a different set of ledgers than previously?  (note that data in the wallet is ledger-specific, for example data associated with credential definitions)
- any impact on other functions, such as the endorser protocol (may need a different endorser for different ledgers)

See the docs for functions like multitenancy, mediation, endorser, for example.

shaangill025:
> Hi @shaangill025 overall this looks pretty good. I haven't done a deep dive yet though.
> 
> Once thing that would be nice is a document describing how the multi-ledger support works, or how to use it, for example:
> 
> * how to I define multiple ledgers?  (I know there is a  hackmd but should prob be a doc in the github repo)
> * what do I need to do for read vs write?  (I assume the verifier has to be able to auto-select the ledger based on received proofs, but what about the issuer?)
> * can you add/remove ledgers at runtime or only on startup?
> * what happens if I restart an agent with a different set of ledgers than previously?  (note that data in the wallet is ledger-specific, for example data associated with credential definitions)
> * any impact on other functions, such as the endorser protocol (may need a different endorser for different ledgers)
> 
> See the docs for functions like multitenancy, mediation, endorser, for example.

@ianco It is ready for review. I have removed the ability to update the `ledger configuration` or `write_ledger` at runtime as I observed some issues in testing (mostly around cached `DID to ledger_id`).

demo/multi_ledger_config.yml
```diff
@@ -0,0 +1,6 @@
+- id: localVON
+  is_production: true
+  genesis_url: 'http://localhost:9000/genesis'
```
ianco, line 3: This fails when running `./run_demo faber --multi-ledger` as the default von-network runs in docker, as does the faber demo (`ConnectionRefusedError: [Errno 111] Connect call failed ('127.0.0.1', 9000)`)

This needs to use the docker IP or host name to connect.

aries_cloudagent/config/argparse.py
```diff
@@ -761,11 +773,19 @@ def get_settings(self, args: Namespace) -> dict:
...
+                    settings["ledger.ledger_config_list"] = ledger_config_list
             else:
                 raise ArgsParseError(
-                    "One of --genesis-url --genesis-file or --genesis-transactions "
-                    "must be specified (unless --no-ledger is specified to "
-                    "explicitly configure aca-py to run with no ledger)."
+                    "One of --genesis-url --genesis-file, --genesis-transactions "
+                    "or --genesis-transactions-list must be specified (unless "
+                    "--no-ledger is specified to explicitly configure aca-py to"
+                    " run with no ledger)."
```
ianco, line 788: One suggestion - allow the existing `--genesis-url`, `-genesis-file` etc and *in addition* allow the multi-ledger yaml file to be specified.  This supports any existing configurations but allows additional ledgers to be added.

If `--genesis-url`, `--genesis-file` etc is specified then this would be the `write_ledger`, otherwise would have to be specified in the multi-ledger yaml file

ianco (changes requested):

ianco (approved):
LGTM

swcurran:
w00t!  Awesome stuff, @shaangill025 !

PR #1429, opened 2021-09-30
Event bus Metadata

burdettadam:
This pr adds metadata to the event passed to an event handler. While writing a queue plugin that subscribes to events on the event bus a limitation of knowing what regular expression triggered the handler being called was identified. The event bus will call multiple handlers that match the subscribed regular expression. Currently, inside the handler, we must discover what regular expression triggered the event handler. The addition of metadata with the matched pattern in an event simplifies required logic in handlers that is error-prone.

Changes:
- event_bus.py 
  - added new EventMetaData and EventWithMetaData classes to add needed context to events


ianco (approved):
Looks good to me

PR #1433, opened 2021-10-04
Profile for base outbound transport class

burdettadam:
When developing plugins for different transports it becomes difficult to do things without having access to the context profile provides. This pr adds root_profile to the base class of outbound transport which is passed in from the outbound transport manager. 

changes:
 - aries_cloudagent/transport/outbound/base.py
     - added class attribute to store profile called root_profile with supporting code
 - aries_cloudagent/transport/outbound/manager.py
     - updated the class creation to take the manager profile
     
### DIFF
--- a/aries_cloudagent/transport/outbound/base.py
+++ b/aries_cloudagent/transport/outbound/base.py
@@ -7,6 +7,7 @@ from typing import Union
 from ...core.profile import Profile
 from ...utils.stats import Collector
 
+from ...core.profile import Profile
 from ..error import TransportError
 from ..wire_format import BaseWireFormat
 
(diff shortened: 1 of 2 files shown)


ianco:
> @ianco -- can you please look at the ACA-Py Integration tests for this? Looks like it stopped after 6 hours. Something wrong with the tests?

I get this error:

```
Aries      | 2021-10-06 18:36:57,434 aries_cloudagent.utils.task_queue ERROR Error running task OutboundTransportManager.start_transport
Aries      | Traceback (most recent call last):
Aries      |   File "/home/indy/aries_cloudagent/transport/outbound/manager.py", line 175, in start_transport
Aries      |     root_profile=self.root_profile
Aries      | TypeError: __init__() got an unexpected keyword argument 'root_profile'
```

If you try running the alice/faber demo the connection never connects.

swcurran:
So is that something broken in the PR, or something else?  Just want to be clear on who needs to fix this.

ianco:
> So is that something broken in the PR, or something else? Just want to be clear on who needs to fix this.

It's broken in this PR.  I ran all the tests in the main branch and everything passed, but with these updates the connection "hangs" and I get an error running the integration tests

ianco:
Hi @burdettadam thanks for the PR!  Can you please catch up your branch with the latest updates in `main` and then we can get it reviewed and merged?

PR #1438, opened 2021-10-11
fix: incorrect return type

TimoGlastra:
Small fix in the type hinting of the store_credential method

In general I think it may be good to look into static type checking tools for ACA-Py. E.g. tools like [pyright](https://github.com/microsoft/pyright) can help find errors in typing (and thus implementation)

Signed-off-by: Timo Glastra <timo@animo.id>


swcurran:
@ianco -- do you want to look into the discussion above about using [pyright](https://github.com/microsoft/pyright)?

PR #1446, opened 2021-10-15
Add credential validation to offer issuance step

MaticDiba:
This pull request adds an additional credential instance validation for any missing attributes. With this check eliminate errors at the issuance stage where in case of missing attributes, Aca-Py returns 500 error right now. And exchange record goes into a fault state.


ianco:
Hi @MaticDiba thanks for the PR!  Can you please catch up your branch with the latest updates in `main` and then we can get it merged?

PR #1461, opened 2021-10-28
Auto connect from author to endoposer on startup

ianco:

If the appropriate startup arguments are provided, auto-initiate the connection between the author and endorser on agent startup

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -1536,14 +1536,24 @@ class EndorsementGroup(ArgumentGroup):
             ),
         )
         parser.add_argument(
+            "--endorser-invitation",
+            type=str,
+            metavar="<endorser-invitation>",
+            env_var="ACAPY_ENDORSER_INVITATION",
+            help=(
+                "For transaction Authors, specify the the invitation used to "
+                "connect to the Endorser agent who will be endorsing transactions. "
+                "Note this is a multi-use invitation created by the Endorser agent."
+            ),
+        )
+        parser.add_argument(
             "--endorser-public-did",
             type=str,
             metavar="<endorser-public-did>",
             env_var="ACAPY_ENDORSER_PUBLIC_DID",
             help=(
                 "For transaction Authors, specify the the public DID of the Endorser "
-                "agent who will be endorsing transactions.  Note this requires that "
-                "the connection be made using the Endorser's public DID."
+                "agent who will be endorsing transactions."
             ),
         )
         parser.add_argument(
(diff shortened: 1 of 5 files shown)


aries_cloudagent/config/argparse.py
```diff
@@ -1605,6 +1615,15 @@ def get_settings(self, args: Namespace):
             elif args.endorser_protocol_role == ENDORSER_ENDORSER:
                 settings["endorser.endorser"] = True
 
+        if args.endorser_invitation:
```
amanji, line 1618: Is the `endorser_invitation` meant to supplant `endorser_public_did`?

esune: It looks like you can use either: when using the invitation you would not have to manually create the connection between the two agents, though, as it would be done autmatically.

demo/runners/support/agent.py
```diff
@@ -1286,6 +1322,22 @@ async def start_endorser_agent(start_port, genesis):
     log_msg("Endorser Endpoint URL is at:", endorser_agent.endpoint)
     log_msg("Endorser webhooks listening on:", start_port + 2)
 
+    # get a reusable invitation to connect to this endorser
+    log_msg("Generate endorser multi-use invite ...")
+    endorser_agent.endorser_connection_id = None
+    endorser_agent.endorser_public_did = None
+    endorser_connection = await endorser_agent.admin_POST(
+        "/connections/create-invitation?alias=EndorserMultiuse&auto_accept=true&multi_use=true"
+    )
```
esune, line 1331: Would it make sense to use OOB invitations (also on the receiver end) rather than connection invitations? My understanding was that at some point it would replace connection invitation completely and therefore should be preferred - @shaangill025 probably has better/fresher information on this.

aries_cloudagent/config/argparse.py
```diff
@@ -1535,15 +1535,25 @@ def add_arguments(self, parser: ArgumentParser):
                 "directly."
             ),
         )
+        parser.add_argument(
+            "--endorser-invitation",
+            type=str,
+            metavar="<endorser-invitation>",
+            env_var="ACAPY_ENDORSER_INVITATION",
+            help=(
+                "For transaction Authors, specify the the invitation used to "
```
esune, line 1544: Nit.: `the the` (repeated) 😅

aries_cloudagent/config/argparse.py
amanji: In that case the `on_startup_event` function would fail if the invitation is not set but the Endorser's public DID is.

aries_cloudagent/config/argparse.py
ianco: Oops lol

demo/runners/support/agent.py
ianco: Yes good observation, I'll make the change

aries_cloudagent/config/argparse.py
ianco: The invitation is used to auto-establish the connection between the author and endorser.

The endorser DID is used to set the meta-data on the endpoint (the author needs to know the endorser DID - this has to be part of the transaction that the author signs before sending it to be endorsed).

The on_startup method will establish the connection if there is an invitation, and then set the connection meta-data if it can.  I don't think this will *fail* if the DID isn't provided, however the endorsement protocol can't be used until this DID is set.

I think I'll set the argparse validation to enforce that the endorser DID and alias *must* be provided if the invitation is provided.

demo/runners/support/agent.py
ianco: It now checks the invitation type and supports either OOB or "old-style" connections

shaangill025 (approved):

Issue #1455, opened 2021-10-21
Proof request with none_revoked restriction returns valid=true even though the credential has been revoked


etschelp:
I tested this with the latest Lissi wallet and aca-py 0.7.2-rc0. 

Steps to reproduce.

1. aca-py is started with a configured tails server, a cred def with revocation registry is created
2. Connect aca-py with the Lissi wallet
3. Issue a credential
4. Revoke the credential
5. Send a proof request with none_revoked restriction and a timestamp that is after the revocation from step 4
6. Lissi wallet says credential has been revoked, but we send it anyway
7. aca-py says proof is valid, but the verified field should be set to false in this case

In the aca-py logs I see the following

```
2021-10-21 13:57:37,745 aries_cloudagent.indy.verifier INFO Amended presentation request (nonce=1056466064839133135704935): removed non-revocation interval at requested_attributes referent Mz2w979BSbBb8SGihvYbYy:2:Bank Account:1.0; corresponding credential in proof is irrevocable
2021-10-21 13:57:37,745 aries_cloudagent.indy.verifier WARNING Amended presentation request (nonce=1056466064839133135704935); removed global non-revocation interval; no revocable credentials in proof
```
This is most likely a version incompatibility issue between agents. When I look into the [verifier.py](https://github.com/hyperledger/aries-cloudagent-python/blob/2355b5ec8f36b4507ba091587d6095f7ad61f5f5/aries_cloudagent/indy/verifier.py#L50) I see that it checks for the existence of a _timestamp_ field in the _identifiers_ section of the presentation request that is retrieved via the _sub_proof_index_. If this field is missing from the response of the other agent the none revocation check is skipped. 

Response from Lissi that is not working

```json
    "presentation": {
        "requested_proof": {
            "revealed_attrs": {},
            "revealed_attr_groups": {
                "Mz2w979BSbBb8SGihvYbYy:2:Bank Account:1.0": {
                    "sub_proof_index": 0,
                    "values": {
                        "Name": {
                            "raw": "revocationtest01",
                            "encoded": "54732868580499211735974961366354315864941016766578192688462230293521237042341"
                        }
                    }
                }
            },
            "self_attested_attrs": {},
            "unrevealed_attrs": {},
            "predicates": {}
        },
        "identifiers": [
            {
                "schema_id": "Mz2w979BSbBb8SGihvYbYy:2:Bank Account:1.0",
[...]

PR #1465, opened 2021-10-30
Fix error removing proof req entries by timestamp

ianco:

Addresses issue https://github.com/hyperledger/aries-cloudagent-python/issues/1455

### DIFF
--- a/aries_cloudagent/indy/sdk/tests/test_verifier.py
+++ b/aries_cloudagent/indy/sdk/tests/test_verifier.py
@@ -326,7 +326,7 @@ class TestIndySdkVerifier(AsyncTestCase):
                 INDY_PROOF_REQ_PRED_NAMES,
                 INDY_PROOF_PRED_NAMES,
                 "schemas",
-                "credential_definitions",
+                {"LjgpST2rjsoxYegQDRm7EL:3:CL:18:tag": {"value": {"revocation": {}}}},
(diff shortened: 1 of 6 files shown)


aries_cloudagent/indy/sdk/tests/test_verifier.py
```diff
@@ -322,20 +322,31 @@ async def test_verify_presentation(self, mock_verify):
         ) as mock_pre_verify, async_mock.patch.object(
             self.verifier, "non_revoc_intervals", async_mock.MagicMock()
         ) as mock_non_revox:
+            INDY_PROOF_REQ_X = deepcopy(INDY_PROOF_REQ_PRED_NAMES)
             verified = await self.verifier.verify_presentation(
-                INDY_PROOF_REQ_PRED_NAMES,
+                deepcopy(INDY_PROOF_REQ_X),
```
esune, line 327: Might not need this second `deepcopy` statement.

esune (approved):
Only comment I had was about something in the unit test, looks good otherwise as far as I can tell.

aries_cloudagent/indy/sdk/tests/test_verifier.py
ianco: Fixed (oops)

shaangill025 (approved):

Issue #1441, opened 2021-10-13
ACA-PY answers requests with wrong response


seriousManual:
We are currently investigating an issue that has occured when we were testing our setup together with a standard wallet app from the appstore. 

The Setup:
We are using the ACA-PY 0.7.1. The attached controller is setup in a way so that when a connection has been created, the controller is automatically issuing a credential to the requesting party.
 
The  Issue:
We were cooperating with the app's developer to investigate in detail what has been happening during the handshake between the parties. The wallet sends a connection request via HTTP against the agent (Initiated via a scanned QRCode).
This request is *not* terminating but is kept open.
 
In the meantime:
* The agent has been notifying our backend via a webhook that the connection has been requested.
* Our backend accepts the connection request, the state of the connection changes to `response`, another webhook notification reaches our backend (`state: response`).
* Our backend then sends a credential offer to the requesting party.
* The wallet receives the credential offer.
 
When this credential offer is accepted in the wallet, an error occurs in the app. Because suddenly the HTTP request, that has initiated the whole handshake, concludes successfully and returns the response that is related *to the credential accept request*.
That means though the HTTP request has initially been sent to request a connection it is now answered with the response to the credential offer.
The app then raises an error because response and request do not match.
 
A couple things we are unsure about:
Is this a known issue?
Is this maybe not an issue at all but intended behaviour?
Can you maybe give any insights on how to handle this properly?

An additional detail provided to us by the manufacturer of the wallet app: The wallet sends the `~transport` decorator, specifying the `return_route` (https://github.com/hyperledger/aries-rfcs/blob/main/features/0092-transport-return-route/README.md) along with the connection request.

Issue #1372, opened 2021-08-23
HTTP request hangs with return_route header


smithbk:
In testing the issuance flow (AIP v1) where ACAPy is the issuer and the IBM agent is the holder, the HTTP request which the holder makes to ACAPy in order to deliver the final ACK hangs.  In other words, when the IBM agent sends the HTTP reque3st containing the ACK to complete the issuance flow, the ACAPy agent never sends an HTTP response.  However, when I change my IBM agent to NOT send the return_route header in the Aries ACK message, then ACAPy immediately sends the HTTP response and all is good.

swcurran:
@ianco can you look at this one?  Or at least given an assessment of how quickly a fix is needed?

Thanks!

TimoGlastra:
Wouldn't this be somehow expected?

From the return route RFC:

> For HTTP transports, the presence of this message decorator indicates that the receiving agent MAY hold onto the connection and use it to return messages as designated.

Because there is no message to be returned, the connection is kept alive (which is a desirable feature when you want to do e.g. long polling). Another way to solve would be to close the HTTP connection on the IBM holder side, or by not including the return routing when no response is expected.

This will have impact on how mobile agents interact with ACA-Py

smithbk:
I think only messagepickup/1.0/noop should be long polled.

The reason I say this is because otherwise it puts more of a burden on the sender to know when a return_route should be added or not.  In some cases, the sender may want to send a message which may or may not terminate the thread.  Only the receiver knows for certain if the thread is complete or if/when it will send another message on the thread.  Also, I do not want to async send messages because if the message fails to be delivered via HTTP, I want in some cases to return that to the admin controller.

ianco:
![acapy-receive-inbound](https://user-images.githubusercontent.com/1852309/131887713-c4583551-4254-4580-8f5d-388dcc8ddc5e.jpg)

Not sure how legible this is, but it's the basic aca-py flow for receiving and processing a message

To close an inbound return route I suggest one of 2 options:

- add a method to the Responder class to close a return route (i.e. return 200 status)
- add something in the Conductor's `dispatch_complete` method to close the return route if necessary

I'm not sure of the implications on the Mediator, looking for some input from @TimoGlastra

swcurran:
For fun I made a plantuml version of the drawing above — @ianco please review:

[...]

PR #1468, opened 2021-11-01
Conn req reply fix

ianco:
See issues: https://github.com/hyperledger/aries-cloudagent-python/issues/1441, https://github.com/hyperledger/aries-cloudagent-python/issues/1372

This is a bit of a shot in the dark, the connection auto-replies aren't following the same convention as other protocols.


ianco:
Oops included code from another PR, will close and re-open :-(

PR #1473, opened 2021-11-02
WIP: Updating the RTDs code for Release 0.7.2

swcurran:
Update the the Read The Docs material in preparation for the next release of ACA-Py.  We'll leave this as a WIP until 0.7.2 is ready to be released, to catch the latest updates.

Currently getting generation errors from the [coordinate_mediation](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/coordinate_mediation/mediation_invite_store.py) module. Help on these would be appreciated -- it's something to do with the formatting of the documentation for the listed methods:

``` bash
WARNING: error while formatting arguments for aries_cloudagent.protocols.coordinate_mediation.mediation_invite_store.MediationInviteRecord.from_json: Handler <function update_annotations_using_type_comments at 0x7f18ab6a5ae8> for event 'autodoc-before-process-signature' threw an exception (exception: wrapper loop when unwrapping dataclasses.dataclass.from_json)
WARNING: error while formatting arguments for aries_cloudagent.protocols.coordinate_mediation.mediation_invite_store.MediationInviteRecord.to_json: Handler <function update_annotations_using_type_comments at 0x7f18ab6a5ae8> for event 'autodoc-before-process-signature' threw an exception (exception: wrapper loop when unwrapping dataclasses.dataclass.to_json)
WARNING: error while formatting arguments for aries_cloudagent.protocols.coordinate_mediation.mediation_invite_store.MediationInviteRecord.unused: Handler <function update_annotations_using_type_comments at 0x7f18ab6a5ae8> for event 'autodoc-before-process-signature' threw an exception (exception: wrapper loop when unwrapping dataclasses.dataclass.unused)
/home/swcur/repos/aries-cloudagent-python/aries_cloudagent/protocols/coordinate_mediation/mediation_invite_store.py:docstring of aries_cloudagent.protocols.coordinate_mediation.mediation_invite_store.MediationInviteStore.get_mediation_invite_record:10: WARNING: Field list ends without a blank line; unexpected unindent.
```

PR #1476, opened 2021-11-03
fix: use named tuple instead of dataclass in mediation invite store

dbluhm:
Dataclass wasn't added until python 3.7 and we're on 3.6. Not sure how this wasn't failing tests :thinking: we must have a backport package installed in the base image or similar but at any rate, I'm hopeful this will resolve issues @swcurran raised in #1473 for this module's doc generation.


dbluhm:
@swcurran I wasn't able to reproduce the exact problems you described in #1473 (I suspect due to python version differences) but curious if you can check if this fixes the issue or not.

ianco (approved):

PR #1494, opened 2021-11-16
Fix TypeError

gorgsenegger:
When calling /schemas/{schema_id}/write_record, a server error 500 occurrs. The log output says

    TypeError: inject() got an unexpected keyword argument 'required'

Other code in this file seems to indicate that the proposed change is what is needed to prevent the error from being thrown.

### DIFF
--- a/aries_cloudagent/messaging/schemas/routes.py
+++ b/aries_cloudagent/messaging/schemas/routes.py
@@ -368,7 +368,7 @@ async def schemas_fix_schema_wallet_record(request: web.BaseRequest):
 
     schema_id = request.match_info["schema_id"]
 
-    ledger = context.inject(BaseLedger, required=False)
+    ledger = context.inject_or(BaseLedger)
     if not ledger:
         reason = "No ledger available"
         if not context.settings.get_value("wallet.type"):


ianco:
Hi @gorgsenegger thanks for the PR!  Note that all commits to hyperledger repo's must be DCO signed (`--signoff` or `-s`)

See:  https://github.com/hyperledger/aries-cloudagent-python/pull/1494/checks?check_run_id=4224754831

gorgsenegger:
Hi @ianco, I'm new to this, did the signoff work? Thank you!

ianco:
Yep it worked, thanks!

swcurran:
Yup - DCO looks good.

This seems like a Big Deal.  Does this mean we are going to need an immediate 0.7.3 release?

ianco:
> This seems like a Big Deal. Does this mean we are going to need an immediate 0.7.3 release?

The error is in the `schemas_fix_schema_wallet_record()` method, which is not a critical method (imho)

ianco (approved):

shaangill025:
Same change might be needed here, https://github.com/hyperledger/aries-cloudagent-python/blob/f9506df755e46c5be93b228c8811276b743a1adc/aries_cloudagent/messaging/credential_definitions/routes.py#L391.

PR #1501, opened 2021-11-18
Outbound Queue - more usability improvements

dbluhm:
This PR improves the usability of outbound queues by:
- Fixing a bug where a queue instance was created twice when loading the class
- Add explicit startup and shutdown methods that are called in startup and shutdown of the conductor. This makes it so we can hold connections to external queues and give the outbound queue implementation the flexibility to determine how long those connections should live.
- Add `open` and `close` method to be used as `start` and `stop` were previously used. These methods are called by `aenter` and `aexit` on the base class and are therefore used each time a message is enqueued.
- All the `start`, `stop`, `open`, and `close` methods are now optional. If no special behavior is required, the inheriting class need not define an empty method.
- Outbound queue accepts a `Profile` instance for configuration instead of `Settings`. This allows the queue to subscribe to an publish events through the root profile.

This represents a BREAKING CHANGE in the outbound queue interface as the semantics of `start` and `stop` have been altered. Additionally, queue implementation init methods will need to be reconfigured to accept a profile instead of a settings object.

I think these changes, while breaking changes, go a long way towards improving the outbound queue interface and will better support a wider variety of external queue implementations.


ianco:
Tests are failing with error `AttributeError: 'Settings' object has no attribute 'settings'`

swcurran:
@shaangill025 -- when you get started on your new issue (#1513 ), this should be part of your assessment.

swcurran:
@andrewwhitehead @ianco -- can you please review this PR?  We'd like to get it into a 0.7.3 release if it makes sense -- and we'd like to get that release in progress today.  Thanks!

shaangill025 (approved):

swcurran:
To confirm -- the note above about BREAKING CHANGE (quoted below).  This means that this has to be the 0.8.0 release, correct?  Do we all agree on that (@dbluhm @shaangill025 @ianco @andrewwhitehead )?  I had been going down a "0.7.3" path, but I now realize this has to be the 0.8.0 one.

> This represents a BREAKING CHANGE in the outbound queue interface as the semantics of start and stop have been altered. Additionally, queue implementation init methods will need to be reconfigured to accept a profile instead of a settings object.
> 
> I think these changes, while breaking changes, go a long way towards improving the outbound queue interface and will better support a wider variety of external queue implementations.

ianco:
What is the breaking change exactly?  Do we need to change configuration etc on existing aca-py deployments?

andrewwhitehead:
It's just a change to the outbound queue base class I believe, so the only people affected are ones with custom outbound queue implementations.

dbluhm:
Yes, what @andrewwhitehead said. I believe this impacts very few people -- I of course don't know all the implementations out there but the only ones I'm aware of are the redis implementation bundled with ACA-Py and the Kafka implementation we're working on with SICPA. I would not consider this to be a breaking change to ACA-Py overall, necessarily, but it does get into a bit of a semantics discussion of what is "the" API of ACA-Py.

swcurran:
So, we use "0.7.3" and highlight the impact on those with an external queue?   Does that seem OK to folks?

ianco:
Yes I think `0.7.3` is the correct version number to use

PR #1509, opened 2021-11-23
Fix warnings when generating ReadTheDocs

### LINKED ISSUE #1506: Oddity in the "vc" module when generating ReadTheDocs docs -- a bug?
A number of documentation warnings are generated in the ACA-Py read-the-docs documentation when the ACA-Py module "vc" is not mocked out.  Sometimes such warnings are an indication of a problem in the code, so I think this is worth a review.   To replicate the problem:

1. Follow the steps in the [ReadMe](https://github.com/hyperledger/aries-cloudagent-python/blob/main/docs/README.md) in the docs folder of ACA-Py to locally generate the Sphinx documentation.  The second of the three steps can be ignored (although it won't hurt) -- just do the sphinx install (if needed) and then run the `sphinx build...` command.  The result is the docs should generate with one warning that you can ignore.
2. Edit the `conf.py` file in the /docs folder and remove from the list of `autodoc_mock_imports` the `aries_cloudagent.vc` entry - the last one.
3. Regenerate and you should see many warnings.
4. Investigate, resolve if possible, or at least confirm that the issues are not important.

The impact of the issue is that the documentation for that module (`vc`) is not included in the docs. And as noted above, it might indicate a larger issue.

Worst case, please document what you find and note why it is not an issue.  Best case -- submit a PR to address the issue.

For bonus points -- if you can address the one unrelated warning (the one there even with `vc` mock'd), that would be great.

Thanks!

shaangill025:
Signed-off-by: shaangill025 <gill.shaanjots@gmail.com>

- resolves #1506
- For duplicated object description, renamed references in `__init__.py`:
  - `vc.ld_proofs.crypto`, `vc.ld_proofs.purposes`, `vc.ld_proofs.suites` by adding a `_` when importing modules. Modified `vc.ld_proofs` to accomodate these changes
  -  `vc.vc_ld.models` by adding a `_` when importing modules. Modified `vc.vc_ld` accordingly
- For cross-reference warnings, adopted solution from [here](https://github.com/sphinx-doc/sphinx/issues/3866#issuecomment-768167824).
- Regarding`list not hashable` issue in `aries_cloudagent/protocols/issue_credential/v1_0/models/credential_exchange.py`, I think there was a minor bug/typo, we had `credential_request: [Mapping, IndyCredRequest] = None` instead of `credential_request: Union[Mapping, IndyCredRequest] = None`.

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v1_0/models/credential_exchange.py
+++ b/aries_cloudagent/protocols/issue_credential/v1_0/models/credential_exchange.py
@@ -68,7 +68,7 @@ class V10CredentialExchange(BaseExchangeRecord):
         ] = None,  # aries message: ..._dict for historic compat on all aries msgs
         credential_offer_dict: Union[Mapping, CredentialOffer] = None,  # aries message
         credential_offer: Union[Mapping, IndyCredAbstract] = None,  # indy artifact
-        credential_request: [Mapping, IndyCredRequest] = None,  # indy artifact
+        credential_request: Union[Mapping, IndyCredRequest] = None,  # indy artifact
         credential_request_metadata: Mapping = None,
         credential_id: str = None,
         raw_credential: Union[Mapping, IndyCredential] = None,  # indy cred as received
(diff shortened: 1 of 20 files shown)


shaangill025:
@swcurran This should fix all warnings, can you please check on your end?

swcurran (approved):
Nice!  Thanks!

PR #1519, opened 2021-11-25
added logging documentation

chriamue:
Added markdown documentation of logging in Acapy.

- [x] Command Line Arguments
- [x] Environment Variables
- [x] Acapy Config
- [x] Logging Config File
- [x] Link in dev readme

### DIFF
--- a/DevReadMe.md
+++ b/DevReadMe.md
@@ -151,6 +151,10 @@ PORTS="5000:5000 8000:8000 1000:1000" ./scripts/run_docker start --inbound-trans
 
 Refer to [the previous section](#Running) for instructions on how to run the software.
 
+### Logging
+
+You can find more details about logging and log levels [here](Logging.md).
+
 ### Running Tests
 
 To run the ACA-Py test suite, use the following script:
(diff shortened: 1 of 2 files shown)


chriamue:
Whats the best location for Logging.md file?

swcurran:
I think the root is probably fine for now, as others are there.  We should probably put them into docs, but I think we need to do them all together.

The question is where we want to put the link to this file. I think it should go into the DevReadme.md file at some useful point.

@ianco -- please take a read through this for accuracy and to approve.

Thanks for the contribution -- good stuff.

ianco:
> The question is where we want to put the link to this file. I think it should go into the DevReadme.md file at some useful point.

Yes looks good thanks for the PR.  I suggest to link from `docs/logging/docs/GettingStartedAriesDev` somewhere?

swcurran:
I would think that this section of the Developer's ReadMe would be best -- https://github.com/hyperledger/aries-cloudagent-python/blob/main/DevReadMe.md#developing

swcurran (approved):
LGTM -- Thanks!

PR #1526, opened 2021-12-01
Display QR code when generating/displaying invites on startup

ianco:


dbluhm (approved):
Tested and working :slightly_smiling_face: I considered suggesting adding a flag for whether the QR code was printed or not but I think the instances where the `--invite` flags are used are generally limited to debugging and testing or else are not going to be negatively impacted by a QR code being printed. So not immediately springing for yet another flag seems favorable.

ianco:
> I considered suggesting adding a flag for whether the QR code was printed or not ...

Yes I had the same thought.  Thanks for the review!

PR #1530, opened 2021-12-03
Enable WS Pings for WS Inbound Transport

dbluhm:
When connecting over WS, it is impossible for ACA-Py to detect when the other end has unexpectedly closed (no WS close or error message sent) without using a heartbeat and timeout. With these changes, if a mobile agent connected over WS suddenly loses connectivity (wifi or data signal drops, for instance), the window of time where a message can be lost due to ACA-Py not realizing the WS has entered a bad state is shrunk to at most `timeout_interval` seconds. An image with a version of these changes has been in testing in a live mediator deployment for a couple of months now with much success.


TimoGlastra (approved):
Nice

swcurran:
@dbluhm -- can you please update this branch and we can merge it.  Thanks.

PR #1543, opened 2021-12-08
ConnRecord tags - `their_public_did` and `invitation_msg_id`

### LINKED ISSUE #1542: didexchange fetches all connection records to determine whether a connection can be reused
As I understand the didexchange implementation at the moment, it fetches all connection records to be able to determine if a connection can be reused. It needs `their_public_did` which is not a tag, thus requiring a `post_filter`.

When dealing with a lot of connections this will probably making reusing connections quite slow.

Same happens with `invitation_msg_id`

shaangill025:
- resolve #1542

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -3,7 +3,7 @@
 import json
 
 from enum import Enum
-from typing import Any, Union
+from typing import Any, Optional, Union
 
 from marshmallow import fields, validate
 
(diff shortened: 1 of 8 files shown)


TimoGlastra (approved):
LGTM. May be good to mention in the release notes of the next release that connections made before this change (and that haven't been updated since) won't be usable for connection reuse as they don't have the tags yet

swcurran:
Would there be a way to add a process (perhaps via a startup parameter?) to run through the existing connection and add this tag?  Asking hypothetically -- perhaps not specifically for this change.  Eventually, we're going to run into a situation that this is needed by some that have deployed ACA-Py.  It would be nice to know of any approach for this in case someone does need it.

Issue #1541, opened 2021-12-08
cannot create connection with self when using oob with didexchange protocol


TimoGlastra:
I can't seem to make a connection with myself (the same agent) when using the OOB + didexchange protocol. It works fine when using OOB + connections protocol. May not be the most common use case to create connections with yourself, but it can useful for testing.

Create an OOB using the following body:

```
{
  "alias": "Larry",
  "label": "Larry on the other side",
  "handshake_protocols": [
    "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0"
  ],
  "use_public_did": false
}
```

then receive the invitation (with auto accept enabled on both sides) and an error will show in the acapy logs. If you change `didexchange` to `connections` in the `handshake_protocols` everything will works as exptected.

Stack trace:

```
2021-12-08 12:35:03,394 aries_cloudagent.core.conductor ERROR Exception in message handler:
Traceback (most recent call last):
  File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/core/dispatcher.py", line 198, in handle_message
    await handler(context, responder)
  File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/protocols/didexchange/v1_0/handlers/response_handler.py", line 34, in handle
    context.message, context.message_receipt
  File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/protocols/didexchange/v1_0/manager.py", line 682, in accept_response
    self._session, response._thread_id
  File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/connections/models/conn_record.py", line 336, in retrieve_by_request_id
    return await cls.retrieve_by_tag_filter(session, tag_filter)
  File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/messaging/models/base_record.py", line 270, in retrieve_by_tag_filter
    f", {post_filter}" if post_filter else "",
aries_cloudagent.storage.error.StorageDuplicateError: Multiple ConnRecord records located for {'request_id': '80a1aead-8ca4-4044-b914-e8845f1109de'}
[...]

PR #1544, opened 2021-12-08
Remove request_id from inviter connection record

ianco:

Addresses issue https://github.com/hyperledger/aries-cloudagent-python/issues/1541

Tested in alice/faber, there are some error messages on the faber side however the connection (didexchange) is established and can be used for issuing credentials and requesting proofs.

Not sure what is the issue of removing `request_id` from the inviter connection record, or why it's required for didexchange but not connections protocol.


swcurran:
@ianco -- what do we do with this one?  Thanks

ianco:
Replaced by PR #1670

Issue #1539, opened 2021-12-08
Credential record state not updated after revocation


DaevMithran:
There should be a credential_revoked state for credential records, currently we have to use a seperate api to know the revocation status and the state of credential records remain in credential_issued even after revocation. The update should also reflect in the credential records.

DaevMithran:
Agreed. That's a valid point. As we are supporting revocation in two ways either using cred_ex_id or rev_reg_id. The first approach will work only if the credential record is in the ACA-Py secure storage, So we can change the state of credential record to credential_revoked only when the credential record is in the ACA-Py secure storage. 
I'll create a PR for this. Thank you for the response.

swcurran:
Addressed by #1545

PR #1545, opened 2021-12-09
Add credential_revoked state

DaevMithran:
Pull request for [Issue: 1539](https://github.com/hyperledger/aries-cloudagent-python/issues/1539)
@swcurran  Added the credential_revoked state. 
For now I have considered only issue-credential 1.0, I should check for the record in V20CredentialExchange too.
Will do it soon. If you have any suggestions on my approach please comment below.

Questions:
- Is it required raise an error if the credential record doesn't exist? 
- The credential state should be changed only after publishing revocation.
   When publish is set to pending, what is the best way to implement this?. I'm planning to create retrieve credential records using rev_reg_id and cred_rev_id

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v1_0/manager.py
+++ b/aries_cloudagent/protocols/issue_credential/v1_0/manager.py
@@ -910,8 +910,11 @@ class CredentialManager:
         cred_rev_ids: Sequence[str]
     ) -> None:
         """
-        Update credential state on revocation using
-        revocation registry ID and credential revocation IDs
+        Update credentials state to credential_revoked
+
+        Args:
+            rev_reg_id: revocation registry ID
+            cred_rev_ids: list of credential revocation IDs
 
         Returns:
             None
--- a/aries_cloudagent/protocols/issue_credential/v1_0/manager.py
+++ b/aries_cloudagent/protocols/issue_credential/v1_0/manager.py
@@ -4,11 +4,11 @@ import asyncio
 import json
 import logging
 
-from typing import Mapping, Tuple
+from typing import Mapping, Tuple, Sequence
 
 from ....cache.base import BaseCache
 from ....core.error import BaseError
-from ....core.profile import Profile
+from ....core.profile import Profile, ProfileSession
 from ....indy.holder import IndyHolder, IndyHolderError
 from ....indy.issuer import IndyIssuer, IndyIssuerRevocationRegistryFullError
 from ....ledger.multiple_ledger.ledger_requests_executor import (
@@ -38,7 +38,9 @@ from .messages.credential_problem_report import (
 from .messages.credential_proposal import CredentialProposal
 from .messages.credential_request import CredentialRequest
 from .messages.inner.credential_preview import CredentialPreview
-from .models.credential_exchange import V10CredentialExchange
+from .models.credential_exchange import (
+    V10CredentialExchange,
+)
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


ianco:
Typically the credential exchange records will get deleted after the credential is issued, so in most use cases there will be no credential exchange record when the credential is revoked.

I'm not sure if it makes sense to add a status or tag on the credential itself with the status (issued vs revoked)?

swcurran:
> Typically the credential exchange records will get deleted after the credential is issued, so in most use cases there will be no credential exchange record when the credential is revoked.
> 
> I'm not sure if it makes sense to add a status or tag on the credential itself with the status (issued vs revoked)?

While "typically" they will get deleted, not always, and the intention of this PR is to updated the non-deleted protocol state objects if they are not deleted.

ianco:
> * Is it required raise an error if the credential record doesn't exist?

It shouldn't raise an error.  The credential exchange record can be deleted, and could even be deleted in between the revocation and publishing events.

> * The credential state should be changed only after publishing revocation.
>   When publish is set to pending, what is the best way to implement this?. I'm planning to create retrieve credential records using rev_reg_id and cred_rev_id

Not completely sure of the intent, but you may want to change the status to `revocation_pending` if publishing the revocation status will be delayed, otherwise it looks like the credential is still in `issued` status.  Then, when you publish revocations, you can just flip the status from `revocation_pending` to `revoked`.

Also note that there are some tests failing on this PR (formatting and unit tests).

DaevMithran:
Hi @swcurran  @ianco 
For the current approach I added a revoke_credentials method in Credential Manager which retrieves credential records using revoc_reg_id and revocation_id and changes their state to credential_revoked.

This approach supports credentials records issued after this merge. As the previous credential records did not have tags to be retrieved using revoc_reg_id and revocation_id

aries_cloudagent/protocols/issue_credential/v1_0/manager.py
```diff
@@ -893,6 +895,33 @@ async def receive_credential_ack(
 
         return cred_ex_record
 
+    async def revoke_credentials(
```
ianco, line 898: I suggest re-naming this method to something like "set_credential_revocation_state" to be more clear on what it's doing (i.e. it's not revoking the credentials, just updating the state).

ianco (changes requested):
One (fairly minor) comment, plus this functionality should also be provided for the V20 cred exchange.

aries_cloudagent/protocols/issue_credential/v1_0/manager.py
DaevMithran: Agree. Changed this

ianco (approved):

shaangill025:
> * This approach supports credentials records issued after this merge. As the previous credential records did not have tags to be retrieved using revoc_reg_id and revocation_id

I don't see any changes to `TAG_NAMES` [here](https://github.com/hyperledger/aries-cloudagent-python/blob/f94184dfc85729b0e0ecce5aa84910feaab36986/aries_cloudagent/protocols/issue_credential/v1_0/models/credential_exchange.py#L37) and [here](https://github.com/hyperledger/aries-cloudagent-python/blob/f94184dfc85729b0e0ecce5aa84910feaab36986/aries_cloudagent/protocols/issue_credential/v2_0/models/cred_ex_record.py#L37)

DaevMithran:
> > * This approach supports credentials records issued after this merge. As the previous credential records did not have tags to be retrieved using revoc_reg_id and revocation_id
> 
> I don't see any changes to `TAG_NAMES` [here](https://github.com/hyperledger/aries-cloudagent-python/blob/f94184dfc85729b0e0ecce5aa84910feaab36986/aries_cloudagent/protocols/issue_credential/v1_0/models/credential_exchange.py#L37) and [here](https://github.com/hyperledger/aries-cloudagent-python/blob/f94184dfc85729b0e0ecce5aa84910feaab36986/aries_cloudagent/protocols/issue_credential/v2_0/models/cred_ex_record.py#L37).  I am just confused by the above comment regarding `cred_ex_record` tags. Just to confirm in regards to `cred_ex_record` backward compatibility, there should be no breaking change, and only that `cred_ex_record` state will be updated as `credential_revoked` upon revocation after the merge?

Hi @shaangill025 initially I tried the tags approach but V20Cred_ex_record doesn't include values of rev_reg_id and cred_rev_id so went with a different approach. 

- First I retrieve IssuerCredRevRecord using rev_reg_id and cred_rev_id
- extract the cred_ex_id from the record 
- retrieve credential record using cred_ex_id
- update state of credential record

PR #1556, opened 2021-12-13
Added event queue support as an alternative to webhook notifications

acuderman:
## What is this PR for?

- Added external queue support as an alternative to HTTP webhook events


swcurran:
Please take a look at this PR, especially @shaangill025 @dbluhm @andrewwhitehead, as all of you have been/are doing queue related changes.

Thanks!

dbluhm:
We have implemented something with a similar function before. In our case, however, the event queuing system was completely decoupled from core ACA-Py and relied on the event bus, capturing the startup and shutdown events for queue initialization and teardown and then subscribing to all events with a handler that pushed the event data onto our external queue.

I'm not against this explicit external queuing system built into ACA-Py but I think I would recommend using the event bus rather than adding another responsibility to the conductor. I am of the opinion that a significant amount of the code currently in the conductor could/should be moved to listeners of the startup and shutdown events. That's just my opinion though. I would be interested to hear what others think.

swcurran:
@acuderman are you familiar with the event bus mentioned by @dbluhm?  There are several PRs that have been merged related to the Event Bus ([list](https://github.com/hyperledger/aries-cloudagent-python/pulls?q=is%3Apr+is%3Aclosed+bus)), with #1063 being the initial implementation.

@andrewwhitehead @shaangill025 -- please consider @dbluhm 's comment above and weigh in on the issue.

Thanks!

acuderman:
@dbluhm @swcurran Thank you for the comments and shared resources. Will take a look at the event bus later this week and update the PR accordingly / close It if everything can be done without changing the core functionality.

acuderman:
Following the [example](https://github.com/hyperledger/aries-cloudagent-python/pull/1448/files#diff-e43ab078f0713f413ca336221ba84d7b97cb4aa829e79295bc5361d75732262eR399), It is possible to subscribe to required events, without changing the core functionality. Closing the issue. Thank you for the input.

PR #1557, opened 2021-12-13
ACA-Py Upgrade command implementation

shaangill025:
- `upgrade` command
  **Example**
  ```
  ./scripts/run_docker upgrade --upgrade-config ./aries_cloudagent/acapy_upgrade_config.yml \
  --from-version v0.7.2 \
  --wallet-type indy \
  --wallet-name issuer \
  --wallet-key mykey \
  --wallet-storage-type postgres_storage \
  --wallet-storage-config '{"url":"host.docker.internal:5432","max_connections":5}' \
  --wallet-storage-creds '{"account":"postgres","password":"pwd","admin_account":"postgres","admin_password":"pwd"}'
  ```
    - `--upgrade-config` - path to YAML config file
      **Example YAML config**
      ```
      v0.6.0:
      ...
      v0.7.1:
      ...
      v0.7.2:
        resave_records:
          base_record_path:
            - "aries_cloudagent.connections.models.conn_record.ConnRecord"
          base_exch_record_path:
            - "aries_cloudagent.protocols.issue_credential.v1_0.models.credential_exchange.V10CredentialExchange"
        update_existing_records: false
      ```
      The above will re-save `ConnRecord` and `V10CredentialExchange`. `update_existing_records` can be used to handle 
      changes where existing records need to be updated, for instance, if a new required field has been added to Marshmallow 
      schema. It is handled here: [link1](https://github.com/hyperledger/aries-cloudagent-python/blob/c97c1ba35168e9ab445ec9328e19350a7ece0d47/aries_cloudagent/commands/upgrade.py#L39) and [link2](https://github.com/hyperledger/aries-cloudagent-python/blob/c97c1ba35168e9ab445ec9328e19350a7ece0d47/aries_cloudagent/commands/upgrade.py#L122) and [link3](https://github.com/hyperledger/aries-cloudagent-python/blob/c97c1ba35168e9ab445ec9328e19350a7ece0d47/aries_cloudagent/commands/upgrade.py#L183), this will have to be managed every release.
    - `--from-version` is only required when version information is not there in storage. It will be ignored if version is found in storage
    - Also accepts `WalletGroup` arguments

Update: now follows this [approach](https://github.com/hyperledger/aries-cloudagent-python/pull/1557#issuecomment-996387027).

Thanks @ianco for suggesting the command approach. This is an initial implementation, for now, it can only handle re-save and/or update records but it can be expanded upon. @andrewwhitehead, @swcurran, and @ianco I will appreciate any feedback.

### DIFF
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -175,9 +175,12 @@ async def upgrade(settings: dict):
                                 session,
                                 reason="re-saving record during ACA-Py upgrade process",
                             )
-                        print(
-                            f"All records of {str(record_type)} successfully re-saved."
-                        )
+                        if len(all_records) == 0:
+                            print(f"No records of {str(record_type)} found")
+                        else:
+                            print(
+                                f"All records of {str(record_type)} successfully re-saved"
+                            )
             # Step 2 Update existing records, if required
             if (
                 "update_existing_records" in upgrade_config
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -104,9 +104,7 @@ async def upgrade(settings: dict):
                     record_type = ClassLoader.load_class(record_path)
                 except ClassNotFoundError as err:
                     raise UpgradeError(f"Unknown Record type {record_path}") from err
-                if not issubclass(record_type, BaseRecord) and not issubclass(
-                    record_type, BaseExchangeRecord
-                ):
+                if not issubclass(record_type, BaseRecord):
                     raise UpgradeError(
                         "Only BaseRecord and BaseExchangeRecord can be resaved"
                         f", found: {str(record_type)}"
@@ -115,7 +113,8 @@ async def upgrade(settings: dict):
                     all_records = await record_type.query(session)
                     for record in all_records:
                         await record.save(
-                            session, reason="re-saving record during ACA-Py upgradation"
+                            session,
+                            reason="re-saving record during ACA-Py upgrade process",
                         )
                     print(f"All records of {str(record_type)} successfully re-saved.")
         # Step 2 Update existing records, if required
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -4,6 +4,7 @@ import asyncio
 from configargparse import ArgumentParser
 from typing import Sequence
 
+from ..core.profile import Profile
 from ..config import argparse as arg
 from ..config.default_context import DefaultContextBuilder
 from ..config.base import BaseError
@@ -31,13 +32,13 @@ async def upgrade(settings: dict):
     try:
         root_profile, public_did = await wallet_config(context)
         # Step 1 re-saving all BaseRecord and BaseExchangeRecord
-        resave_record_paths = settings.get("upgrade_resave_records")
+        resave_record_paths = settings.get("upgrade.resave_records")
         for record_path in resave_record_paths:
             try:
                 record_type = ClassLoader.load_class(record_path)
             except ClassNotFoundError as err:
-                raise UpgradeError(f"Unknown Record type: {record_path}") from err
-            if not isinstance(record_type, BaseRecord) and not isinstance(
+                raise UpgradeError(f"Unknown Record type {record_path}") from err
+            if not issubclass(record_type, BaseRecord) and not issubclass(
                 record_type, BaseExchangeRecord
             ):
                 raise UpgradeError(
@@ -50,9 +51,24 @@ async def upgrade(settings: dict):
                     await record.save(
                         session, reason="re-saving record during ACA-Py upgradation"
                     )
+                print(f"All records of {str(record_type)} successfully re-saved.")
+        # Step 2 Update existing records, if required
+        if settings.get("upgrade.update_existing_records"):
+            await update_existing_records(root_profile)
         await root_profile.close()
     except BaseError as e:
-        raise UpgradeError("Error during upgrade: ") from e
+        raise UpgradeError(f"Error during upgrade: {e}")
+
+
+async def update_existing_records(profile: Profile):
+    """
+    Handle addition of required attributes to Marshmallow schema.
+
+    Args:
+        profile: Root profile
+
+    """
+    pass
 
 
 def execute(argv: Sequence[str] = None):
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -106,8 +106,7 @@ async def upgrade(settings: dict):
                     raise UpgradeError(f"Unknown Record type {record_path}") from err
                 if not issubclass(record_type, BaseRecord):
                     raise UpgradeError(
-                        "Only BaseRecord and BaseExchangeRecord can be resaved"
-                        f", found: {str(record_type)}"
+                        "Only BaseRecord can be resaved" f", found: {str(record_type)}"
                     )
                 async with root_profile.session() as session:
                     all_records = await record_type.query(session)
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -137,13 +137,13 @@ async def upgrade(settings: dict):
             if "update_existing_records" in upgrade_config:
                 update_existing_recs_callable = (
                     version_upgrade_config_inst.get_update_existing_func(
-                        upgrade_from_version
+                        from_version_config
                     )
                 )
                 if not update_existing_recs_callable:
                     raise UpgradeError(
                         "No update_existing_records function "
-                        f"specified for {upgrade_from_version}"
+                        f"specified for {from_version_config}"
                     )
                 await update_existing_recs_callable(root_profile)
         # Update storage version
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -10,7 +10,7 @@ from ..config.default_context import DefaultContextBuilder
 from ..config.base import BaseError
 from ..config.util import common_config
 from ..config.wallet import wallet_config
-from ..messaging.models.base_record import BaseRecord, BaseExchangeRecord
+from ..messaging.models.base_record import BaseRecord
 from ..storage.base import BaseStorage
 from ..storage.error import StorageNotFoundError
 from ..storage.record import StorageRecord
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -11,7 +11,11 @@ from ..config.base import BaseError
 from ..config.util import common_config
 from ..config.wallet import wallet_config
 from ..messaging.models.base_record import BaseRecord, BaseExchangeRecord
+from ..storage.base import BaseStorage
+from ..storage.error import StorageNotFoundError
+from ..storage.record import StorageRecord
 from ..utils.classloader import ClassLoader, ClassNotFoundError
+from ..version import __version__, RECORD_TYPE_ACAPY_VERSION
 
 from . import PROG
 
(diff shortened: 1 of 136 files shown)


ianco (dismissed):
This looks good!  I wonder if we should store the agent version in the wallet, that way we know what version we are upgrading from, and we could even check on startup if an "upgrade" was required.  (Obviously existing deployments wouldn't have this so would require the "--from-version" to be provided, but if the version number were in the wallet all they would need to do is "aca-py upgrade")

andrewwhitehead:
I agree, the version should likely be stored to allow for automatic upgrades. This would probably mean auto-populating the version if it is not detected during startup, since we don't store one at the moment.

ianco:
I wouldn't recommend auto-populating the version number in the database on startup.  I suggest we insert the version number when we create the database (on `aca-py provision` or `... --auto-provision`).  On startup we query the version from the database and compare to the version of the application, and auto-shutdown if we detect a version incompatibility (which means we need to `aca-pu upgrade`).  Version incompatibility can include "no version in database".

For `aca-py upgrade` we run all the necessary updates *from* the version currently in the database (or the version the user specifies) *to* the version of the software they are trying to run.  Once the upgrade is done we insert/update the version # in the database.

I also suggest we have a VERY OBVIOUS MESSAGE somewhere that they backup their database before they run an `aca-py upgrade`.

shaangill025:
> I wouldn't recommend auto-populating the version number in the database on startup. I suggest we insert the version number when we create the database (on `aca-py provision` or `... --auto-provision`). On startup we query the version from the database and compare to the version of the application, and auto-shutdown if we detect a version incompatibility (which means we need to `aca-pu upgrade`). Version incompatibility can include "no version in database".

Working on this.

ianco:
> > I wouldn't recommend auto-populating the version number in the database on startup. I suggest we insert the version number when we create the database (on `aca-py provision` or `... --auto-provision`). On startup we query the version from the database and compare to the version of the application, and auto-shutdown if we detect a version incompatibility (which means we need to `aca-pu upgrade`). Version incompatibility can include "no version in database".
> 
> > I also suggest we have a VERY OBVIOUS MESSAGE somewhere that they backup their database before they run an aca-py upgrade.
> 
> Working on this.

OK as long as you agree with it :-D

shaangill025:
@ianco @andrewwhitehead This is ready.

ianco (approved):
Looks good to me

@andrewwhitehead do you want to take a look before we merge?

aries_cloudagent/commands/upgrade.py
```diff
@@ -0,0 +1,190 @@
...
+        # Step 1 re-saving all BaseRecord and BaseExchangeRecord
+        if "resave_records" in upgrade_config:
+            resave_record_paths = upgrade_config.get("resave_records")
+            for record_path in resave_record_paths:
+                try:
+                    record_type = ClassLoader.load_class(record_path)
+                except ClassNotFoundError as err:
+                    raise UpgradeError(f"Unknown Record type {record_path}") from err
+                if not issubclass(record_type, BaseRecord) and not issubclass(
+                    record_type, BaseExchangeRecord
```
andrewwhitehead, line 108: BaseExchangeRecord is a subclass of BaseRecord

aries_cloudagent/commands/upgrade.py
```diff
@@ -0,0 +1,190 @@
...
+                ):
+                    raise UpgradeError(
+                        "Only BaseRecord and BaseExchangeRecord can be resaved"
+                        f", found: {str(record_type)}"
+                    )
+                async with root_profile.session() as session:
+                    all_records = await record_type.query(session)
+                    for record in all_records:
+                        await record.save(
+                            session, reason="re-saving record during ACA-Py upgradation"
```
andrewwhitehead, line 118: "upgrade" or "upgrade process" not upgradation

ianco (dismissed):

swcurran:
@andrewwhitehead -- go to go on this one?  I'd like to get it into 0.73.

andrewwhitehead:
@swcurran  I think this is good to go, my only concern is that upgraded agents will refuse to start until the `upgrade` command is run. I believe the user also needs to specify an exact (and supported) version that they are upgrading from, which may not be immediately obvious, so we would need to document that well.

swcurran:
Can you propose specific changes to be made?  @shaangill025 -- can you take a look?  I'd really like to get this out today, so collaborating on this would be helpful. 

The big issue is refusing to start without the upgrade. Ideas on how best to deal with that?

Next the issue of indicating a version -- what do you propose we document to help them?  Can they look at the "version.py" from their previous version to tell them?

swcurran:
Alternatively:

- slow down on this upgrade and leave it out of the release?
- slow down on the release until we get this right?

Opinions welcome!  What would developers want?

andrewwhitehead:
I'm not sure. I think the upgrade could be made optional, as the only thing that's affected is continuation of existing did-exchange connections (AFAIK).

To get the current version of their aca-py install the easiest way is probably to run `aca-py -v`

shaangill025:
> @swcurran I think this is good to go, my only concern is that upgraded agents will refuse to start until the `upgrade` command is run. I believe the user also needs to specify an exact (and supported) version that they are upgrading from, which may not be immediately obvious, so we would need to document that well.

@andrewwhitehead I see there is a flaw, how about I raise an exception [here](https://github.com/hyperledger/aries-cloudagent-python/blob/1e92e224d685fe49f079eb1da8692c1135626e74/aries_cloudagent/config/wallet.py#L155) and ask to execute the `upgrade command with --from-version` instead of adding the version record here when they execute the `start --auto-provision` or `provision` command. 

`--from-version` will be required only for the initial release and subsequent upgrades will rely on the stored version record. I think we mention this in the release notes that they record the current version of ACA-Py running and use that to execute the `upgrade command`.

The YAML config for previous versions will be the same for this release. So even if somebody puts in the wrong `--from-version`, it shouldn't cause an issue.

```
v0.6.0:
  resave_records:
    base_record_path:
      - "aries_cloudagent.connections.models.conn_record.ConnRecord"
  update_existing_records: false
v0.7.0:
  resave_records:
    base_record_path:
      - "aries_cloudagent.connections.models.conn_record.ConnRecord"
  update_existing_records: false 
v0.7.1:
  resave_records:
    base_record_path:
      - "aries_cloudagent.connections.models.conn_record.ConnRecord"
  update_existing_records: false
v0.7.2:
  resave_records:
    base_record_path:
      - "aries_cloudagent.connections.models.conn_record.ConnRecord"
  update_existing_records: false
```

andrewwhitehead:
Hm, it doesn't make sense to me to repeat the upgrade commands for each version, they should be done in order. v0.6.0-v0.7.1 doesn't require any upgrade for instance. It seems like if the --from-version isn't specified and no version is stored then we should just apply all the defined upgrades in order. In any case, I think the upgrade command should be optional for this release (only if a user wants to use existing did-exchange connections for instance).

shaangill025:
> Hm, it doesn't make sense to me to repeat the upgrade commands for each version, they should all be done from the --from-version onward in order. v0.6.0-v0.7.1 doesn't require any upgrade for instance. It seems like if the --from-version isn't specified and no version is stored then we should just start from the first defined upgrade. In any case, I think the upgrade command should be optional for this release (only if a user wants to use existing did-exchange connections for instance).

Right now, YAML config for each version is defined separately and not incrementally from the specified version as you have proposed above. I can improve/update this later. Regarding making `upgrade` optional, maybe I only raise an exception [here](https://github.com/hyperledger/aries-cloudagent-python/blob/1e92e224d685fe49f079eb1da8692c1135626e74/aries_cloudagent/core/conductor.py#L331) if `--auto-provision` is set otherwise store the acapy_version record and continue?


> I see there is a flaw here when introducing this into ACA-Py, how about I raise an exception [here](https://github.com/hyperledger/aries-cloudagent-python/blob/1e92e224d685fe49f079eb1da8692c1135626e74/aries_cloudagent/config/wallet.py#L155) and ask to execute the `upgrade command with --from-version` instead of adding the version record when they execute the `start --auto-provision` or `provision` command

I think this change still makes sense.

andrewwhitehead:
> Regarding making `upgrade` optional, maybe I only raise an exception [here](https://github.com/hyperledger/aries-cloudagent-python/blob/1e92e224d685fe49f079eb1da8692c1135626e74/aries_cloudagent/core/conductor.py#L331) if `--auto-provision` is set otherwise add the acapy_version record and continue?

I don't think we ever want to auto-add the version record unless it's a new wallet. For an existing wallet that should only happen when `upgrade` is run. However, for 0.7.3 the application could happily run without having a version record stored, and then in 0.8.0 we could make the upgrade step mandatory just to know there's some consistency in the stored connection records. It should still be possible on 0.7.3 to manually run the `upgrade` command, but not required.

shaangill025:
> I don't think we ever want to auto-add the version record unless it's a new wallet. For an existing wallet that should only happen when `upgrade` is run. However, for 0.7.3 the application could happily run without having a version record stored, and then in 0.8.0 we could make the upgrade step mandatory just to know there's some consistency in the stored connection records. It should still be possible on 0.7.3 to manually run the `upgrade` command, but not required.

Sounds good, working on the above changes and I have added the incremental/onward config support. This should be ready in the next 2-3 hours.

andrewwhitehead:
The logic there looks a little too complicated to verify right now. Let's just leave the version check out of the wallet config.

shaangill025:
@andrewwhitehead @swcurran With the new incremental/onwards version approach, I have implemented a sorting logic that assumes major versions like `v7.2.0 .. v7.3.0` and cannot handle `release candidates`, is that a reasonable assumption?

The requested changes are in and ready for review.

andrewwhitehead:
I hadn't realized that the `--upgrade-config` path was a required parameter in order to perform the upgrade. It should be possible to find the default config file in the aries_cloudagent package, see for example how it's done in the logging configuration: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/config/logging.py#L14:L36

We can't really expect users to know the path to the configuration file, or even where the aca-py module is installed on their system.

andrewwhitehead:
Running the `upgrade` command with no wallet configuration also produces the message `aries_cloudagent.commands.upgrade.UpgradeError: ACA-Py version not found in storage and no --from-version specified.`. I think this should be failing earlier, because how would it find the version in storage without the wallet config? But also, I think it should be selecting 0.7.2 as the from-version (based on the upgrade configuration) because that is the first one to offer any impact from an upgrade.

andrewwhitehead:
Testing with the latest, if I run `python3 -m aries_cloudagent upgrade --upgrade-config ./aries_cloudagent/config/tests/test-acapy-upgrade-config.yaml` I get the following:

```
No ACA-Py version found in wallet storage and no --from-version specified. Selecting v0.7.2 as --from-version from the config.
Running upgrade process for v0.7.2
All records of <class 'aries_cloudagent.connections.models.conn_record.ConnRecord'> successfully re-saved.
All records of <class 'aries_cloudagent.protocols.issue_credential.v1_0.models.credential_exchange.V10CredentialExchange'> successfully re-saved.
```

But that's without any wallet configuration, so what records is it re-saving? The same output is produced if run a second time.

andrewwhitehead:
Oddly if I run without the `--upgrade-config` parameter it only re-saves the connection records:

```
No ACA-Py version found in wallet storage and no --from-version specified. Selecting v0.7.2 as --from-version from the config.
Running upgrade process for v0.7.2
All records of <class 'aries_cloudagent.connections.models.conn_record.ConnRecord'> successfully re-saved.
```

shaangill025:
> Testing with the latest, if I run `python3 -m aries_cloudagent upgrade --upgrade-config ./aries_cloudagent/config/tests/test-acapy-upgrade-config.yaml` I get the following:
> 
> ```
> No ACA-Py version found in wallet storage and no --from-version specified. Selecting v0.7.2 as --from-version from the config.
> Running upgrade process for v0.7.2
> All records of <class 'aries_cloudagent.connections.models.conn_record.ConnRecord'> successfully re-saved.
> All records of <class 'aries_cloudagent.protocols.issue_credential.v1_0.models.credential_exchange.V10CredentialExchange'> successfully re-saved.
> ```
> 
> But that's without any wallet configuration, so what records is it re-saving? The same output is produced if run a second time.

With no wallet configuration, we get the `InMemoryProfile(backend=in_memory, name=default)` profile, this has no `ConnRecord` or `V10CredentialExchange` records so they cannot be not re-saved. I have added a conditional around the print statement to address the confusion.

andrewwhitehead (approved):

PR #1558, opened 2021-12-13
Await asyncio.sleeps

andrewwhitehead:
Fixes a few warnings during tests.


swcurran (approved):
Tested using demo. And integration tests.

PR #1565, opened 2021-12-17
Update docker scripts to use new & improved docker IP detection

ianco:

Required to support newer versions of docker


swcurran (approved):
LGTM - tested `run_demo`

PR #1570, opened 2021-12-21
fix: use invitation key for connection query

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>

Invitation key was not taken into account for get connections endpoint.

If possible, would like to get this in the upcoming 0.7.3 release :)

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/tests/test_routes.py
+++ b/aries_cloudagent/protocols/connections/v1_0/tests/test_routes.py
@@ -1,5 +1,6 @@
 import json
 
+from unittest.mock import ANY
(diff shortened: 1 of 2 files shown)


shaangill025 (approved):
`/connections` endpoint's `invitation_key` filter now working as intended.

aries_cloudagent/protocols/connections/v1_0/tests/test_routes.py
```diff
@@ -1,5 +1,6 @@
 import json
 
+from unittest.mock import ANY
```
andrewwhitehead, line 3: Neat, I learned a new trick

Issue #1571, opened 2021-12-21
No connection re-use with OpenAPI Demo via Docker Faber-Alice Out of Band Invitations (OOB)


watkinspd:
Two potential issues:

**1: Unable to re-use connections with OOB in Faber-Alice OpenAPI demo.**
I am excercising the Aries OpenAPI Demo using the following guide:
[https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/AriesOpenAPIDemo.md](https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/AriesOpenAPIDemo.md)

I am running the demo via Docker Desktop MacOS. I am trying to use Faber-Alice to test Out of Band Invitations (OOB). I am generating OOB invitations on Faber and receiving them on Alice.

I am able to make and use connections if I specify 
`"use_public_did": false`
 in 
`/out-of-band/create-invitation` 
I verified this by sending messages over the resulting connections from both Faber and Alice.

However I am unable to generate or accept OOB invitations that result in the re-use of established connections. Each attempted re-use of an invitation, or use of a new invitation results a in a new connection.

If I change my approach to specify 
`"use_public_did": true` 
in 
`/out-of-band/create-invitation` 
I get 
`400: Public invitations are not enabled`

I reviewed

[https://github.com/hyperledger/aries-rfcs/blob/main/features/0434-outofband/README.md](https://github.com/hyperledger/aries-rfcs/blob/main/features/0434-outofband/README.md)

The tutorial does not explicitly describe the requirements for connections to be re-used so I am unsure why my efforts are not working. My hope would be that connection re-use can be done using the OpenAPI Faber-Alice demo via docker.

**2: Faber-Alice OpenAPI demo - always setting "their_label": "alice.agent" on Faber side.**
I have noticed that the connections on the Faber side always have
 `"their_label": "alice.agent"` 
when using
 `/out-of-band/receive-invitation` 
regardless of the alias or label that is used in the OOB invitation on the Alice side.

swcurran:
@ianco -- can you please take a look? Perhaps we need some docs on how to reuse connections in Alice/Faber?

ianco:
To get the reuse I believe (off the top of my head) you need to startup aca-py with the `--public-did` parameter (or something like that), which we don't do for the demo agents.

Would you like me to add this parameter?  Or add a new param to alice/faber for example `./run_demo faber --allow-public-did`?

swcurran:
[...]

PR #1574, opened 2021-12-21
Allow public invites for alice/faber demo

ianco:

Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/1571

Note - added a `--reuse-connections` parameter to faber - this will trigger faber to use a public did in the didexchange connection (and alice will always "reuse connection" if possible) and the connection indeed gets reused, however right now there are no webhooks for the "reuse" messages so the alice/faber controllers are getting hung ...

### DIFF
--- a/demo/runners/agent_container.py
+++ b/demo/runners/agent_container.py
@@ -96,7 +96,7 @@ class AriesAgent(DemoAgent):
         return self._connection_ready.done() and self._connection_ready.result()
 
     async def handle_oob_invitation(self, message):
-        print("handle_oob_invitation():", json.dumps(message))
+        print("handle_oob_invitation()")
         pass
 
     async def handle_connection_reuse(self, message):
@@ -624,7 +624,6 @@ class AgentContainer:
         self.multitenant = multitenant
         self.mediation = mediation
         self.use_did_exchange = use_did_exchange
-        print("Setting use_did_exchange:", self.use_did_exchange)
         self.wallet_type = wallet_type
         self.public_did = public_did
         self.seed = seed
@@ -827,13 +826,10 @@ class AgentContainer:
         for cred_attr in cred_attrs:
             if cred_attr["name"] in wallet_attrs:
                 if wallet_attrs[cred_attr["name"]] != cred_attr["value"]:
-                    print("Value doesn't match for:", cred_attr["name"])
                     matched = False
             else:
-                print("Attribute not found for:", cred_attr["name"])
                 matched = False
 
-        print("Matching credential received")
         return matched
 
     async def request_proof(self, proof_request):
@@ -885,12 +881,10 @@ class AgentContainer:
 
         if self.cred_type == CRED_FORMAT_INDY:
             # return verified status
-            print("Received proof:", self.agent.last_proof_received["verified"])
             return self.agent.last_proof_received["verified"]
 
         elif self.cred_type == CRED_FORMAT_JSON_LD:
             # return verified status
-            print("Received proof:", self.agent.last_proof_received["verified"])
             return self.agent.last_proof_received["verified"]
 
         else:
--- a/demo/runners/agent_container.py
+++ b/demo/runners/agent_container.py
@@ -99,6 +99,16 @@ class AriesAgent(DemoAgent):
         print("handle_oob_invitation():", json.dumps(message))
         pass
 
+    async def handle_connection_reuse(self, message):
+        print("handle_connection_reuse():", json.dumps(message))
+        # TODO we are reusing an existing connection, set our status to the existing connection
+        pass
+
+    async def handle_connection_reuse_accepted(self, message):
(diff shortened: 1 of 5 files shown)


swcurran:
Hmm...I ran this:

```
./run_demo faber --public-invites
Preparing agent image...
sha256:94c44484e255f32387e52e2b69f8f7e767625c3a36235cd1239b5481dbb67cc9
Trying to detect ngrok service endpoint
ngrok not detected for agent endpoint
172.17.0.1
Starting [faber] agent with args [--port 8020 --public-invites]
usage: faber.py [-h] [--no-auto] [-p <port>] [--did-exchange] [--revocation]
                [--tails-server-base-url <tails-server-base-url>]
                [--cred-type <cred-type>] [--aip <api>] [--timing]
                [--multitenant] [--mediation] [--multi-ledger]
                [--wallet-type <wallet-type>]
                [--endorser-role <endorser-role>] [--arg-file <arg-file>]
faber.py: error: unrecognized arguments: --public-invites
```

ianco:
> Hmm...I ran this:
> 
> ```
> ./run_demo faber --public-invites
...
> faber.py: error: unrecognized arguments: --public-invites
> ```

You don't need to add a new parameter to faber, the `--public-invites` param is added to the aca-py agent that faber (and alice) start up under the hood

swcurran:
So it works, but each time I do a new invitation, I get a new connection on both sides.  That means that reuse is not being used, which is the goal.  Alice should be responding with a "reuse" message, and her "new" connection should be removed.  Likewise on Faber -- once she responds with the reuse, the hanging connection should be removed.  This (pretty likely :-) ) works because there was a follow up tweak to DID Exchange to do the clean up after someone tried this out.

Thoughts?

swcurran:
I think the reuse _should_ be automatic, handled by ACA-Py, but am not certain how it works.

swcurran:
Also, I noticed that the "their_public_did" item on the connection object was not set, which is a bit odd.  Not sure why it wouldn't be, but I think that is a pre-requisite for "reuse" on a connection.

ianco:
I believe the auto-reuse only works when you use public DIDs for connections, which the demo is not doing

ianco:
> I believe the auto-reuse only works when you use public DIDs for connections, which the demo is not doing

I guess it depends on the intent, I though we were just trying to unblock Peter (issue https://github.com/hyperledger/aries-cloudagent-python/issues/1571 where he is using the API).

If we want to configure alice/faber to use connection reuse for the demo then there is more work required.  Let me know ...

swcurran:
Let's leave it for now.  Peter's goal is to enable connection reuse, so that's really what we are after. 

I think it would be nice to have the demo optional support connection reuse, so it would have an optional parameter, making it easy to see what the controller has to do to support reuse.  How about we leave this as a draft PR and when you get a chance, you can update it?  I'm hoping it is a pretty small change.

ianco:
> Let's leave it for now. Peter's goal is to enable connection reuse, so that's really what we are after.
> 
> I think it would be nice to have the demo optional support connection reuse, so it would have an optional parameter, making it easy to see what the controller has to do to support reuse. How about we leave this as a draft PR and when you get a chance, you can update it? I'm hoping it is a pretty small change.

Sure that works, I've converted the PR to draft.

ianco:
Fixed demo - run with `./run_demo faber --reuse-connections` - this will use faber's public DID for the connection and allow connection reuse.

(Alice will always try to reuse connections if possible.)

demo/runners/agent_container.py
```diff
@@ -96,8 +96,23 @@ def connection_ready(self):
         return self._connection_ready.done() and self._connection_ready.result()
 
     async def handle_oob_invitation(self, message):
+        print("handle_oob_invitation()")
         pass
 
+    async def handle_connection_reuse(self, message):
```
shaangill025, line 102: Is this function being used?

demo/runners/agent_container.py
```diff
@@ -96,8 +96,23 @@ def connection_ready(self):
...
+        # we are reusing an existing connection, set our status to the existing connection
+        if not self._connection_ready.done():
+            self.connection_id = message["connection_id"]
+            self.log("Connected")
+            self._connection_ready.set_result(True)
+
+    async def handle_connection_reuse_accepted(self, message):
```
shaangill025, line 109: Same question as above.

demo/runners/agent_container.py
ianco: This allows the controller (faber) to tag the active connection (handles the webhook)

demo/runners/agent_container.py
ianco: This allows the controller (alice) to tag the active connection (handles the webhook)

shaangill025 (approved):
`Alice` reuses the existing connection as intended.

Issue #1567, opened 2021-12-18
Exception when resolving a DID when multi-ledger support is active


swcurran:
Resolving a did from a ledger that is not the main (production and write set to true) ledger raises the exception below, and a recent PyDID change (#1562) didn't resolve the issue. From the exception, the guess is that the value that is handed over to pydid is empty.

My next bet is that it's related to the ledger resolution in indy.py https://github.com/hyperledger/aries-cloudagent-python/blob/3fe6715a2bb6ac30ecddf52ec5ad5e03493354df/aries_cloudagent/resolver/default/indy.py#L48

Exception:

```
2021-12-16 15:28:09,214 indy.libindy.native.indy.services.ledger INFO 	src/services/ledger/mod.rs:137 | build_get_attrib_request() => Ok("{\"reqId\":1639668489214330900,\"identifier\":\"F6dB7dMVHUQSC64qemnBi7\",\"operation\":{\"type\":\"104\",\"dest\":\"NBUXM1kVv4yNVc6ezpaa94\",\"raw\":\"endpoint\"},\"protocolVersion\":2}")
 =================
 2021-12-16 15:28:09,240 aries_cloudagent.core.dispatcher ERROR Handler error: resolve_did
 Traceback (most recent call last):
   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/resolver/routes.py", line 96, in resolve_did
     context.profile, did
   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/resolver/did_resolver.py", line 73, in resolve_with_metadata
     resolver, doc = await self._resolve(profile, did)
   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/resolver/did_resolver.py", line 54, in _resolve
     did,
   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/resolver/base.py", line 146, in resolve
     return await self._resolve(profile, did)
   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/resolver/default/indy.py", line 67, in _resolve
     Ed25519VerificationKey2018, ident="key-1", public_key_base58=recipient_key
   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/pydid/doc/builder.py", line 49, in add
[...]

PR #1583, opened 2022-01-04
Multiple Indy Ledger - Askar Fixes and Cleanup

shaangill025:

- Fixes issues with multiple ledger support when using `askar` wallet-type. Issue (#1567) reported [here](https://github.com/hyperledger/aries-cloudagent-python/issues/1567#issuecomment-1004153961).

### DIFF
--- a/aries_cloudagent/ledger/merkel_validation/domain_txn_handler.py
+++ b/aries_cloudagent/ledger/merkel_validation/domain_txn_handler.py
@@ -144,15 +144,15 @@ def prepare_get_attr_for_state(reply):
 
 def prepare_attr_for_state(txn, path_only=False):
     """Return key, value pair for state from ATTR."""
-    txn_data = txn.get(TXN).get(DATA)
+    txn_data = txn.get(TXN, {}).get(DATA)
     nym = txn_data.get(DEST)
     attr_type, attr_key, value = parse_attr_txn(txn_data)
     path = make_state_path_for_attr(nym, attr_key, attr_type == HASH)
     if path_only:
         return path
     hashed_value = hash_of(value) if value else ""
-    seq_no = txn.get(TXN_METADATA).get(SEQ_NO)
-    txn_time = txn.get(TXN_METADATA).get(TXN_TIME)
+    seq_no = txn[TXN_METADATA].get(SEQ_NO)
+    txn_time = txn[TXN_METADATA].get(TXN_TIME)
     value_bytes = encode_state_value(hashed_value, seq_no, txn_time)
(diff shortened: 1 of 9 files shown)


shaangill025:
This is ready, @etschelp has tested this and it fixes `askar` related issues.

aries_cloudagent/ledger/merkel_validation/domain_txn_handler.py
```diff
@@ -145,16 +144,15 @@ def prepare_get_attr_for_state(reply):
...
+    txn_data = txn.get(TXN).get(DATA)
     nym = txn_data.get(DEST)
     attr_type, attr_key, value = parse_attr_txn(txn_data)
     path = make_state_path_for_attr(nym, attr_key, attr_type == HASH)
     if path_only:
         return path
     hashed_value = hash_of(value) if value else ""
-    seq_no = result.get(TXN_METADATA).get(SEQ_NO)
-    txn_time = result.get(TXN_METADATA).get(TXN_TIME)
+    seq_no = txn.get(TXN_METADATA).get(SEQ_NO)
```
andrewwhitehead, line 154: This pattern is used several times, but it's not as foolproof as it looks. If the first `get` fails then it will return None which doesn't have a `get` method.

shaangill025: I think that `ATTRIB` write request reply will be structured as:
- `indy_sdk`
```
{
    'op': 'REPLY', 
    'result': {
        ...,
        "txn": {
           ...
        },
        "txnMetadata": {
            "seqNo": 10,  
            ...
        },
        ...
    }
}
```
- `indy_vdr`
```
{
    ...,
    "txn": {
       ...
    },
    "txnMetadata": {
        "seqNo": 10,  
        ...
    },
    ...
}
```
`prepare_attr_for_state` is not intended to be called directly and instead through `prepare_for_state_write` which should be able to handle this:
```
if RESULT in reply.keys():
   reply = reply.get(RESULT)
```
The same workflow happens with `prepare_for_state_read`.

andrewwhitehead: I'm just talking about the `.get(X).get(Y)`. It should be something like `.get(X, {}).get(Y)` or `[X].get(Y)` if the key is known to exist.

shaangill025: Sounds good, I understand it now.

andrewwhitehead (approved):

Issue #1590, opened 2022-01-07
RevocationError reason kwarg


dbluhm:
While working with revocation on 0.7.3-rc0, I ended up in a situation where there was an issue writing to the tails server (this might be as mundane as just a connectivity issue) but, in the process, when that error was reported another error occurred:

```
agent_1          | 2022-01-07 15:59:26,153 aries_cloudagent.core.event_bus ERROR Error occurred while processing event
agent_1          | Traceback (most recent call last):
agent_1          |   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/core/event_bus.py", line 120, in notify
agent_1          |     await processor()
agent_1          |   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/revocation/routes.py", line 1265, in on_revocation_tails_file_event
agent_1          |     f"Tails file for rev reg {revoc_reg_id} " f"failed to upload: {reason}"
agent_1          |   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aries_cloudagent/core/error.py", line 11, in __init__
agent_1          |     super().__init__(*args, **kwargs)
agent_1          | TypeError: RevocationError does not take keyword arguments
```

This appears to be a result of converting some code from being in an HTTP handler to being in an event handler with some residual `reason="..."` in the Error instantiation.

PR #1591, opened 2022-01-08
Fixes for revocable credential issuance

andrewwhitehead:
- Fix RevocationError instantiations
- Ensure transaction is not left open when IndyIssuerRevocationRegistryFullError exception is raised


swcurran:
@dbluhm  - can you please test this?

dbluhm (approved):
Confirmed that this fixes #1590

PR #1596, opened 2022-01-11
Fix tails_local_path format

DaevMithran:
Signed-off-by: DaevMithran <daevmithran1999@gmail.com>

Issue is not fixed yet. Tried replacing the white spaces with %20 in tails_local_path. I'm not receiving error in swagger now. I'm receiving error in the logs. And it creates only one rev_reg instead of two now. Can anyone guide me on this?

### DIFF
--- a/aries_cloudagent/indy/sdk/tests/test_util.py
+++ b/aries_cloudagent/indy/sdk/tests/test_util.py
@@ -57,7 +57,7 @@ class TestIndyUtils(AsyncTestCase):
         tails_local_path = tails_path("rev-reg-id")
         assert tails_local_path is None
 
-        tails_rr_dir = indy_client_dir(join("tails", "rev-reg-id"), create=True)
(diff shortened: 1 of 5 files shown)


andrewwhitehead:
You don't want to use %20 in the path everywhere, only when the tails file is actually pushed to the tails server (with `put_file` I believe). In this case it would be best to urlencode the parameter rather than replacing spaces specifically.

DaevMithran:
> You don't want to use %20 in the path everywhere, only when the tails file is actually pushed to the tails server (with `put_file` I believe). In this case it would be best to urlencode the parameter rather than replacing spaces specifically.

Even this did not work. I believe the issue is in indy-tails-server. changing this https://github.com/bcgov/indy-tails-server/blob/main/tails_server/web.py#L135 from os.path.join(storage_path, revocation_reg_id) to os.path.join(storage_path, revocation_reg_id).replace(" ", "\\") might work

swcurran:
Given that it appears the error is in indy-tails-server, should we close this PR and related issues?

swcurran:
Closing this PR.

PR #1601, opened 2022-01-18
DID updates for endorser

ianco:

Adds endorser support to the ledger register_nym function, allowing Authors to self-register a public DID.

### DIFF
--- a/aries_cloudagent/ledger/routes.py
+++ b/aries_cloudagent/ledger/routes.py
@@ -1,10 +1,14 @@
 """Ledger admin routes."""
 
+import json
+
 from aiohttp import web
 from aiohttp_apispec import docs, querystring_schema, request_schema, response_schema
 from marshmallow import fields, validate
 
 from ..admin.request_context import AdminRequestContext
+from ..connections.models.conn_record import ConnRecord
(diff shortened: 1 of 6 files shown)


aries_cloudagent/ledger/routes.py
```diff
@@ -181,6 +224,7 @@ async def register_ledger_nym(request: web.BaseRequest):
         request: aiohttp request object
     """
     context: AdminRequestContext = request["context"]
+    outbound_handler = request["outbound_message_router"]
     async with context.profile.session() as session:
         ledger = session.inject_or(BaseLedger)
```
andrewwhitehead, line 229: Hm, it's not updated by this PR, but BaseLedger should be injected via the profile not the session. Generally anything injected by the session shouldn't be touched after the async with block as it might refer to a closed session. This is also true for the IndyLedgerRequestExecutor instances which seem to be set up by the conductor.

andrewwhitehead (approved):

aries_cloudagent/ledger/routes.py
ianco: OK thanks for the tip!

PR #1607, opened 2022-01-21
Auto-promote author did to public after endorsing

ianco:
Adds a couple of options:

Author can specify to auto-promote DID to their public DID when they receive the endorsed transaction (and it is saved)

Endorser can use a different DID for endorsing than they use for their public DID (for connections) (can be specified globally by parameter, or can be specified when they call the endorse endpoint).

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -1820,7 +1820,6 @@ class EndorsementGroup(ArgumentGroup):
             if settings["endorser.author"]:
                 settings["endorser.auto_request"] = True
             else:
-                pass
                 raise ArgsParseError(
                     "Parameter --auto-request-endorsement should only be set for "
                     "transaction Authors"
@@ -1830,7 +1829,6 @@ class EndorsementGroup(ArgumentGroup):
             if settings["endorser.endorser"]:
                 settings["endorser.auto_endorse"] = True
             else:
-                pass
                 raise ArgsParseError(
                     "Parameter --auto-endorser-transactions should only be set for "
                     "transaction Endorsers"
@@ -1840,7 +1838,6 @@ class EndorsementGroup(ArgumentGroup):
             if settings["endorser.author"]:
                 settings["endorser.auto_write"] = True
             else:
-                pass
                 raise ArgsParseError(
                     "Parameter --auto-write-transactions should only be set for "
                     "transaction Authors"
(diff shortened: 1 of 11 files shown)


aries_cloudagent/config/argparse.py
```diff
@@ -1826,6 +1856,16 @@ def get_settings(self, args: Namespace):
                     "for transaction Authors"
                 )
 
+        if args.auto_promote_author_did:
+            if settings["endorser.author"]:
+                settings["endorser.auto_promote_author_did"] = True
+            else:
+                pass
```
andrewwhitehead, line 1863: Are we passing or raising? Or calling?

aries_cloudagent/wallet/routes.py
```diff
@@ -544,6 +568,31 @@ async def wallet_rotate_did_keypair(request: web.BaseRequest):
...
+        "endorser.auto_promote_author_did"
+    ):
+        did = event.payload["did"]
+        try:
+            await promote_wallet_public_did(
+                profile, profile.context, profile.session, did
+            )
+        except Exception as e:
+            # log the error, but continue
+            LOGGER.exception(
```
andrewwhitehead, line 590: `exception` will produce a stack trace with the error being handled, it shouldn't need to be added to the message

aries_cloudagent/config/argparse.py
ianco: Raising ... argh, code smell :-(

aries_cloudagent/wallet/routes.py
ianco: This is just writing the exception details to the logger, right?  It shouldn't affect any of the messaging.

andrewwhitehead: I mean the message argument that is passed to `exception`. There's a traceback added automatically so that should just be descriptive of the context. For example:

```
>>> import logging
>>> 
>>> try:
...     raise ValueError("nope")
... except:
...     logging.getLogger().exception("uh oh")
... 
uh oh
  File "<stdin>", line 2, in <module>
ValueError: nope
```

andrewwhitehead: I found we have a bunch of cases of this at the moment which is counterproductive:

```
>>> import logging
>>> 
>>> try:
...     raise ValueError("nope")
... except Exception as err:
...     logging.getLogger().exception(err)
... 
nope
  File "<stdin>", line 2, in <module>
ValueError: nope
```

ianco: Ah ok I get it thanks

ianco: Fixed

shaangill025 (approved):

PR #1614, opened 2022-01-31
feat: enable webhook events for mediation records

### LINKED ISSUE #950: Webhooks for mediation protocol

TimoGlastra:
Adds mediation webhook topic so webhook events are sent for mediation events.

Fixes #950

@dbluhm anything I'm missing here? The change seems so trivial that I may be forgetting something

### DIFF
--- a/aries_cloudagent/protocols/coordinate_mediation/v1_0/models/mediation_record.py
+++ b/aries_cloudagent/protocols/coordinate_mediation/v1_0/models/mediation_record.py
@@ -19,6 +19,7 @@ class MediationRecord(BaseRecord):
         schema_class = "MediationRecordSchema"
 
     RECORD_TYPE = "mediation_requests"
+    RECORD_TOPIC = "mediation"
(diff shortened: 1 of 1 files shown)


TimoGlastra:
Needs #1615

dbluhm (approved):
I've been away from this code for a bit and I'm surprised it really was this simple but I think this is good; should be all that is needed to get state changes for mediation pushed out as events/webhooks

PR #1616, opened 2022-01-31
PR for endorser support to update endpoints on ledger.

HarshMultani-AyanWorks:
This is a PR that contains logic which would enable the endpoint to get updated on ledger after the endorser has signed the transaction.

The endpoints that have undergone changes are : 
1.) /wallet/set-did-endpoint
2.) /wallet/did/public

These are in the /aries_cloudagent/wallet/routes.py file.

After the endpoint is updated on the ledger, it also gets updated in the wallet. The call to update it in the wallet is not included in this code, but would be implemented soon.


aries_cloudagent/ledger/indy.py
```diff
@@ -886,7 +886,12 @@ async def get_endpoint_for_did(
         return address
```
ianco, line 886: You'll need to make these same changes in the indy_vdr.py version of the ledger

(this is enabled when you use `--wallet-type askar`)

aries_cloudagent/wallet/indy.py
```diff
@@ -705,6 +705,8 @@ async def set_did_endpoint(
         endpoint: str,
```
ianco, line 705: You'll need to make these same changes in the askar.py version of the ledger

(this is enabled when you use `--wallet-type askar`)

ianco (changes requested):

ianco:
Replaced bt PR https://github.com/hyperledger/aries-cloudagent-python/pull/1696

PR #1619, opened 2022-02-05
Fixes and cleanups for issue-credential 1.0

andrewwhitehead:
- Improve logging
- Use transactions for better state management under Askar
- Add more checks around the credential exchange state in case of repeated messages


ianco (approved):
Confirm that this improves the stability of issuing credentials (tested using the performance demo).  I still get some errors but not as frequent.

ianco:
(had a successful run with `./run_demo performance --arg-file demo/postgres-indy-args.yml -c 10000 -b 100 --wallet-type askar`)

PR #1620, opened 2022-02-07
Breaking change: ConnRecord their_role and state inconsistency between 0160 and 0023

shaangill025:

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -220,19 +220,11 @@ class ConnRecord(BaseRecord):
         """Initialize a new ConnRecord."""
         super().__init__(
             connection_id,
-            state=(ConnRecord.State.get(state) or ConnRecord.State.INIT).rfc160,
             **kwargs,
         )
         self.my_did = my_did
         self.their_did = their_did
         self.their_label = their_label
-        self.their_role = (
-            ConnRecord.Role.get(their_role).rfc160
-            if isinstance(their_role, str)
-            else None
-            if their_role is None
-            else their_role.rfc160
-        )
         self.invitation_key = invitation_key
         self.invitation_msg_id = invitation_msg_id
         self.request_id = request_id
@@ -250,6 +242,31 @@ class ConnRecord(BaseRecord):
             if connection_protocol is None
             else connection_protocol.aries_protocol
         )
+        self.set_state(ConnRecord.State.get(state) or ConnRecord.State.INIT)
+        if their_role:
+            self.set_their_role(their_role)
+        else:
+            self.their_role = None
+
+    def set_state(self, given_state):
+        """Set ConnRecord state based upon connection_protocol."""
+        if self.connection_protocol == ConnRecord.Protocol.RFC_0023.aries_protocol:
+            self.state = given_state.rfc23
+        elif self.connection_protocol == ConnRecord.Protocol.RFC_0160.aries_protocol:
+            self.state = given_state.rfc160
+        else:
+            self.state = given_state.rfc23
+
+    def set_their_role(self, given_role):
+        """Set ConnRecord their_role upon connection_protocol."""
+        if isinstance(given_role, str):
+            given_role = ConnRecord.Role.get(given_role)
+        if self.connection_protocol == ConnRecord.Protocol.RFC_0023.aries_protocol:
+            self.their_role = given_role.rfc23
+        elif self.connection_protocol == ConnRecord.Protocol.RFC_0160.aries_protocol:
+            self.their_role = given_role.rfc160
+        else:
+            self.their_role = given_role.rfc23
 
     @property
     def connection_id(self) -> str:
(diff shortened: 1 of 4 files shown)


shaangill025:
Resolves issue reported [here](https://chat.hyperledger.org/channel/aries-cloudagent-python?msg=EZHgKFGgWK2aHrsRQ)

I think `rfc23_state` ([link](https://github.com/hyperledger/aries-cloudagent-python/blob/343679ada2bbf4850913805e105e5fad7dacdb6e/aries_cloudagent/connections/models/conn_record.py#L260)) becomes redundant with these changes. If this is acceptable then I can remove it and update the demo code accordingly.

I think `ATAH` might also need updates as now the `state` and `role` can be different depending upon whether the connection is `0160` or `0023`

ianco:
> I think `rfc23_state` ([link](https://github.com/hyperledger/aries-cloudagent-python/blob/343679ada2bbf4850913805e105e5fad7dacdb6e/aries_cloudagent/connections/models/conn_record.py#L260)) becomes redundant with these changes. If this is acceptable then I can remove it and update the demo code accordingly.
> 
> I think `ATAH` might also need updates as now the `state` and `role` can be different depending upon whether the connection is `0160` or `0023`

We'll need to review all the BCGov apps, I don't know offhand what state we're using in the webhooks.

PS  "AATH" I think ...

shaangill025:
Earlier for all `ConnRecord` (0160 or 0023 `connection_protocol`), the states and roles were always 0160 nomenclature:
- state
  - init
  - invitation
  - request
  - response
  - active
  - error
- role
  - invitee
  - inviter
  
But now, for `ConnRecord` (0023 `connection_protocol`), the states and roles will now be:
- state
  - start
  - invitation
  - request
  - response
  - completed
  - abandoned
- role
  - requester
  - responder

etschelp:
After testing this PR locally I think the rfc23 state has to stay because for all connections that use the connection protocol it makes sense.

So we have two use cases here:

1. connection_protocol is connections/1.0

state: active
rfc23_state: completed

2. connection_protocol is didexchange/1.0

state: completed
rfc23_state: completed

The events in this PR work exactly as before, what has changed is that the stored connection record that is returned from the rest endpoint now also has the state set to completed for the did exchange protocol whereas in the main branch it would be active. So I have to revise my comment from the meeting earlier this is indeed a storage change and not a breaking controller change.

swcurran:
This is why we need to lead with code and then nail down the RFC.  The right thing to do here would have been to keep the states the same where possible, since the same object gets created from both protocols. We pretty much started from scratch with DID Exchange/OOB vs. retaining the history.

I can't really help with the call, so encourage those with deep code knowledge to push through this. The key is determining how the controllers will deal with this.  The following are my thoughts on helping how to think about this (but could be way off)

I think we basically have three controllers approaches that we have to consider:

- those using `connections/` endpoints.
- those using 'didexchange/` endpoints.
- those transitioning from `connections/` to `didexchange/` endpoints.

I think what we want in the next release is to announce the deprecation of the use of connections and to push everyone to use didexchange.  Would it be feasible to document what happens when people use connections/ (what data gets set as we go through the states and messages), same for didexchange, and then provide guidance for what has to change in the controller when transitioning to didxchange?   Of course, that gets fun when the implementation has `--auto-accept-invitations` and `--auto-accept-requests` or not.

TimoGlastra:
Looking at this as an ACA-Py user I'd vote against making the `state` property contain didexchange state for didexchange and connection state for the connection protocol. We would like to keep supporting 0160 for a while and having to check for multiple states that semantically mean the same is not ideal. I'd rather have ACA-Py use a consistent state value across protocols (RFC 0160 right now) that you can map to your desired protocol state (rfc0023 state property)

swcurran:
Do we have a decision on what's going to happen with this one?  Are we making the change or not?  

I'll add this to the agenda of the next ACA-Pug meeting.

etschelp:
I think this task started as a fix for a minor inconsistency between documentation, connection record and connection event and then mutated into a whole discussion about the deprecation of the connection protocol. So I would propose to either just fix the inconsistency, or wait until there is a decision on how to proceed. Basically there are three open questions. 1. how to handle the connection states and 2. how to handle the deprecation. 3. how to migrate stored data.

dbluhm:
I am in agreement with @TimoGlastra on this one. Controllers coded to expect RFC 160 states would require updates to work with connections formed through DID Exchange protocol rather than connections. Since there's a 1:1 mapping from RFC 160 to RFC 23 states, there's no value added in making such a change, either.

I would prefer that (for the time being, at least), `state` always references the RFC 160 state correctly (`state` property of serialized connection records in webhooks should always correspond to RFC 160 states) and `rfc23_state` always references the RFC 23 state (not sure off the top of my head whether this value is included in webhooks or not; if it isn't, I think it ought to be).

swcurran:
I agree.  I believe that means we close this PR and ideally add a documentation change to clarify what a controller should expect in the  `state` (0160 states) and `rfc23_state` items, but we keep them the functionality the same.

shaangill025:
> I agree. I believe that means we close this PR and ideally add a documentation change to clarify what a controller should expect in the `state` (0160 states) and `rfc23_state` items, but we keep them the functionality the same.


| state | rfc23_state     |
| ------------- |:-------------:|
| init      | start |
| invitation      | invitation |
| request | request  |
| response | response |
| active | completed |
| error | abandoned |

PR #1628, opened 2022-02-11
Fix for non-revoc proof with no timestamp

ianco:


TimoGlastra:
Tested and works! Nice @ianco

TimoGlastra (approved):

PR #1642, opened 2022-02-24
Pin markupsafe at version 2.0.1

andrewwhitehead:
Temporary fix due to breaking changes in the dependency.

The 2.2 branch of aiohttp-apispec requires Jinja2 < 3, while its 3.0 (beta) branch supports newer Jinja2 versions which would likely also bypass this issue. We can either test out the current beta or wait for a new release there.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,7 @@ base58~=2.1.0
 deepmerge~=0.3.0
 ecdsa~=0.16.1
 Markdown~=3.1.1
+markupsafe==2.0.1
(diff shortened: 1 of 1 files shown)


swcurran:
Shouldn't I be able to replicate the AATH issue with `./run_docker --version` in the scripts folder of the current `main`?  I couldn't.  Happy to accept this, but just wondering about that.

swcurran (approved):

swcurran:
Confirmed that this fixed the AATH. Issue -- all good.

PR #1643, opened 2022-02-24
feat: accept taa using config

### LINKED ISSUE #1532: Allow to accept TAA using startup parameters
Currently it is not possible to automatically accept the TAA using startup parameters. It is possible if you have an interactive tty, but this is not always the case in server environments. If you know which ledger you're going to interact with, it would be possible to set the aml mechanism in the acapy startup parameters to be able to automate the complete startup process. 

With the new multi-ledger functionality we could even allow to set this per ledger, so you can handle it for all configured ledgers.

Maybe there are good reasons not to do this, so interested to hear you input

TimoGlastra:
Adds option to accept the TAA using config parameters in addition to the Admin API and using TTY.

The TAA can be accepted by providing a valid acceptance mechanism, and specifying the version. This is to make sure newer versions of the TAA aren't automatically accepted. If the version does not match, or the acceptance mechanism doesn't exist in the aml, an error is thrown and the process is terminated (this seems fine to me as it is run on startup and can be fatal to using the agent).

The configuration is as follows:

```
--accept-taa on_file 1.0
```

or in the yml config:

```yml
accept-taa: [on_file, 1.0]
```

Fixes #1532

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -800,6 +800,17 @@ class LedgerGroup(ArgumentGroup):
                 " HyperLedger Indy ledgers."
             ),
         )
+        parser.add_argument(
+            "--accept-taa",
+            type=str,
+            nargs=2,
+            metavar=("<acceptance-mechanism>", "<taa-version>"),
+            env_var="ACAPY_ACCEPT_TAA",
+            help=(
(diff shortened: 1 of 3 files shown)


andrewwhitehead (approved):
Thanks! I like that it requires passing the current TAA version.

PR #1647, opened 2022-03-01
Performance demo updates

ianco:


demo/README.md
```diff
@@ -396,6 +396,22 @@ The script starts both agents, runs the performance test, spits out performance
...
 
+You can also run the demo against a postgres database using the following:
+
+```bash
+./run_demo performance --arg-file demo/postgres-indy-args.yml
+```
+
+(Obvs you need to be running a postgres database - the command to start postgres is in the yml file provided above.)
+
+You can tweak the number of credentials issued using the `--cound` and --batch` prameters, and you can run against an askar database using the `--askar` option.
```
andrewwhitehead, line 407:
```suggestion
You can tweak the number of credentials issued using the `--count` and `--batch` parameters, and you can run against an Askar database using the `--wallet-type askar` option.
```

ianco: Nice catch thanks

andrewwhitehead (approved):

PR #1655, opened 2022-03-08
Add cred_def_id to metadata when using an Endorser

frostyfrog:
When publishing a new credential definition to the ledger (with the Author/Endorser workflow), the `cred_def_id` is nowhere to be found in the transaction response or the webhooks. I have moved the metadata definition lines for `cred_def_id` and `issuer_id` to the actual definition point for the metadata and have removed the redundant lines that were in an if block.


frostyfrog:
That was just a rebase

ianco (approved):

Issue #1657, opened 2022-03-09
Multitenancy and public DIDs


matgnt:
Hi all,

as I'm looking into acapy multitenancy a bit, I had some trouble registering public DIDs for sub-wallets.
It looks like I'm not the only one as there have been some issues here:
https://github.com/hyperledger/aries-cloudagent-python/issues/1460
and there
https://github.com/hyperledger/aries-cloudagent-python/issues/1358

I'm trying to summarize what I've learned so far:
Most important to understand:
- `/wallet/did/create` and `/wallet/did/public` endpoints are NOT enough.
- You need to use `/ledger/register-nym` IN BETWEEN those endpoints!

So far so good. The issue is, that:
Using this endpoint is not possible when using multitenancy. I think there are Options here:
1. Having a separate agent running that does the endorsement for you
2. Having a sub-wallet with endorser rights doing this for you
3. Maybe let the base-wallet do this for you

Currently I only see option 1.
Option 2 does not work because I would need to be able to pre-seed / fill the DID I create in a sub-wallet - which seems not possible right now.
`/wallet/did/create` does not allow to add a seed or existing (on the ledger) DID
`/wallet/did/public` does not allow to add seed, so I can NOT add an existing (on the ledger) DID

Option 3 would have been my first guess, but with multitenancy enabled, it requires authentication, and this I get only for a sub-wallet.

Can someone confirm my observations or give some hints where I might have some misunderstandings?

Thanks a lot!
Matthias

etschelp:
Last time I checked all the base wallet could do was create sub wallets. So I would say yes you need another agent, sub wallet or standalone, that does the endorsement for you.

As a remark there is also an open PR that endorses an ATTRIB transaction to set the endpoint on the ledger: https://github.com/hyperledger/aries-cloudagent-python/pull/1616 For holder wallets this is the best option imho. 

In the end it depends on where to place the orchestration of the api calls. If I remember correctly the consent was to keep such logic in the controller (not aca-py). But what seems odd right now is that one part of the wallet initialisation logic lies in the sub wallet and one part in the base wallet. This means that there is a difference on what a freshly initialised sub- wallet controller must do vs a standalone controller. So maybe it would make sense to have the option to provide a did/seed when calling _POST /multitenancy/wallet_ and as a second step have the option in the base wallet to write a NYM or ATTRIB transaction to the ledger (either directly or via an endorser). Like this all api calls would be in one place.

matgnt:
[...]

PR #1658, opened 2022-03-09
[RFC] Do NOT merge yet - Allow access to /ledger/register-nym with multitenancy

matgnt:
The change shows what might be necessary to support "Option 3" described in #1657 

Still needs discussion whether this is the desired way to go.

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -325,6 +325,9 @@ class AdminServer(BaseAdminServer):
                     )
                 )
 
+                # to allow base wallet to register sub-wallet DIDs on the ledger
+                is_register_nym_path = path.startswith("/ledger/register-nym")
+
                 # base wallet is not allowed to perform ssi related actions.
                 # Only multitenancy and general server actions
                 if (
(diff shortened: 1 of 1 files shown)


ianco:
You can register a nym on the ledger through an endorser.  With PR 1616 you'll also be able to update attrib through an endorser as well

chumbert:
We also considered using the root wallet as endorser but decided against it.

Apart from these endorsement considerations, I'm not in favor of opening any new endpoints. This endorsement design decision, that can make sense for some but not for others, would have an impact on all aca-py instances due to the endpoint whitelist being hardcoded. This was one of the cons when making our decision.

If we want to accommodate for this, I would suggest having the endpoints whitelist be configurable instead of hardcoded. This way, having the root wallet as an endorser can be an option, but the downsides of opening new endpoints  won't impact all multi-tenant aca-py instances.

dbluhm:
With #1836, this type of change should no longer require code changes.

matgnt:
Closing this PR. Thanks @dbluhm for your PR #1836 that should do the trick in a more configurable way.

PR #1659, opened 2022-03-09
Redis PQ Cleanup

shaangill025:


shaangill025:
Removing redundant code as this feature will be supported by the upcoming `aries-acapy-plugin-redis-events` plugin.

andrewwhitehead:
It looks like the `get_outbound_queue` function is still looking at the `transport.outbound_queue` setting, but since that isn't configurable by the command line arguments I'm not sure how it gets initialized?

aries_cloudagent/config/logging.py
```diff
@@ -124,6 +122,18 @@ def print_banner(
         )
         banner.print_spacer()
 
+        external_in_transports = set().union(
```
andrewwhitehead, line 125: You can just write `set(transport for transport ...)`

aries_cloudagent/config/argparse.py
```diff
@@ -1262,21 +1246,8 @@ def get_settings(self, args: Namespace):
         settings = {}
         if args.inbound_transports:
             settings["transport.inbound_configs"] = args.inbound_transports
-        else:
-            raise ArgsParseError("-it/--inbound-transport is required")
```
andrewwhitehead, line 1266: Is it correct that -it and -ot are becoming completely optional?

andrewwhitehead: Oh I see, that validation is moving to the conductor.

aries_cloudagent/core/conductor.py
```diff
@@ -108,6 +106,52 @@ async def setup(self):
         """Initialize the global request context."""
 
         context = await self.context_builder.build_context()
+        ext_plugin = context.settings.get("external_plugins")
+        ext_plugin_config = context.settings.get("plugin_config")
+        in_transport_config_exception = False
+        if not context.settings.get("transport.inbound_configs"):
```
andrewwhitehead, line 112: This does rather defeat the purpose of supporting external queue plugins. Maybe we can add a way for the plugin to indicate that it handles inbound or outbound transport messages once it's loaded? The plugin registry object could look for a method in the plugin metadata, for example.

shaangill025: > How about within the plugin code we add `transport.inbound_plugin_handle` and `transport.outbound_plugin_handler` [bool] to `profile.settings`. Then modify the validation in conductor as:
> 
> ```
> if not context.settings.get("transport.inbound_configs") and not context.settings.get("transport.inbound_plugin_handle"):
>         raise ArgsParseError(
>             "No --inbound-transport/-it and external transport config specified"
>         )
> if not context.settings.get("transport.outbound_configs") and not context.settings.get("transport.outbound_plugin_handle"):
>         raise ArgsParseError(
>             "No --outbound-transport/-ot and external transport config specified"
>         )
> ```

I am leaving `-it` and `-ot` validation unchanged, with the plugin it will be something like `-it redis_queue.inbound redis 0 -oq redis_queue.outbound redis 0` or `-it kafka_queue.inbound kafka 0 -oq kafka_queue.outbound kafka 0`

ianco:
Where is the documentation on how to use the external queues?  Is that in another repo?

shaangill025:
> Where is the documentation on how to use the external queues? Is that in another repo?
> 
> (Here I guess? https://github.com/bcgov/aries-acapy-plugin-redis-events)

Yes, there for redis plugin and https://github.com/sicpa-dlab/aries-acapy-plugin-kafka-events for kafka plugin.

ianco:
> Yes, there for redis plugin and https://github.com/sicpa-dlab/aries-acapy-plugin-kafka-events for kafka plugin.

I get a `404` on the SICPA repo, is that a secure repository?

shaangill025:
> I get a `404` on the SICPA repo, is that a secure repository?

I think so, I believe they gave me access a while back.

swcurran:
@victormartinez-work - what is the status of the SICPA repo referenced above?  Is it going to open source?  Moved to Hyperledger?  Thanks!

ianco:
I think we need a 10k foot overview of what's involved in deploying aca-py with persistent queues, a "persistent queues" README that should go in the aca-py repo with links to the redis and kafka plug-in repo's, i.e.:

- startup the queue (redis or kafka)
- startup aca-py with the appropriate plug-in
- run the listener(s)/relay(s) (I seem to recall there is a separate process that listens on the outbound queue and sends out messages via http or whatever)

The documentation in the redis plugin repo is good but jumps right into the details.

I seem to recall that SICPA had a nice picture.

ianco (approved):
LGTM, everything still works

PR #1660, opened 2022-03-09
Revert change to send_credential_ack return value

andrewwhitehead:
Related to #1619


swcurran:
FYI -- this fixes the AATH tests.  Does reverting mean we don't get the benefits of why the changes were made in the first place (of which I'm not sure...).  What's the story of this one?  :)

ianco (approved):
LGTM

andrewwhitehead:
@swcurran  No the benefits are still there, I had just changed a method signature so it wasn't returning a copy of the exchange record anymore but that turned out to introduce an issue (this PR also adds a test for the new exchange record state).

PR #1670, opened 2022-03-17
Qualify did exch connection lookup by role

ianco:

Fixes issue #1541

### DIFF
--- a/aries_cloudagent/protocols/didexchange/v1_0/manager.py
+++ b/aries_cloudagent/protocols/didexchange/v1_0/manager.py
@@ -702,13 +702,22 @@ class DIDXManager(BaseConnectionManager):
         conn_rec = None
         if response._thread:
             # identify the request by the thread ID
-            try:
-                async with self.profile.session() as session:
+            async with self.profile.session() as session:
+                try:
                     conn_rec = await ConnRecord.retrieve_by_request_id(
-                        session, response._thread_id, their_role="inviter"
+                        session, response._thread_id,
+                        their_role=ConnRecord.Role.RESPONDER.rfc23,
                     )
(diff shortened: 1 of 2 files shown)


aries_cloudagent/protocols/didexchange/v1_0/manager.py
```diff
@@ -702,13 +702,22 @@ async def accept_response(
...
                     )
-            except StorageNotFoundError:
-                pass
+                except StorageNotFoundError:
+                    pass
+                if not conn_rec:
+                    try:
+                        conn_rec = await ConnRecord.retrieve_by_request_id(
+                            session, response._thread_id,
+                            their_role=ConnRecord.Role.RESPONDER.rfc160,
```
ianco, line 717: It looks like the rfc23 roles should be used, but the created connection records are using the rfc160 roles ... so I included checks for both

TimoGlastra (approved):
Great! Thanks @ianco

Adding `their_role` to the `TAG_NAMES` means this will prevent old exchanges from finishing after PR has been released. This is not an issue for completed or new exchanges so I guest it's not really an issue? 

Good to keep track somehow of later added tags, as now it may look you can query record by the `their_role` property, but this is only the case for records saved _after_ this PR has been merged.

swcurran:
We do have the "migrate" mechanism to update stored data on deployment of a new version, but given the tight scope of this change, I don't think we need to use it.  Agreed?

We used the mechanism with 0.73 to update connection records.

ianco:
I don't think we need to "migrate" for this function because it's only used during the process of establishing a connection.

Once the connection is established we don't use the "request_id" to lookup the connection.

Issue #1651, opened 2022-03-03
Invalid proof presentation when requesting revocable and non-revocable credentials


mmoramartinez:
When requesting a proof presentation that contains two attribute groups, one from a revocable credential definition and one from a non-revocable definition, if the check for revocation is requested for all attributes, the proof is always invalid, even if the revocable credential is still active. 
```
"presentation_request": {
        "nonce": "1104330525903017641791888",
        "name": "twoCredCheck",
        "version": "1.0",
        "requested_attributes": {
          "attrGrp_0": {
            "names": [
              "one",
              "two"
            ],
            "restrictions": [
              {
                "cred_def_id": "5vT6qtd26Rc9huD4tL2cXm:3:CL:8055:rev"
              }
            ]
          },
          "attrGrp_1": {
            "names": [
              "one",
              "two"
            ],
            "restrictions": [
              {
                "cred_def_id": "5vT6qtd26Rc9huD4tL2cXm:3:CL:8055:non"
              }
            ]
          }
        },
        "requested_predicates": {},
        "non_revoked": {
          "to": 1646301471
        }
      },
      "state": "request_sent"
```
During verification, “non_revoc_intervals” is not empty, but there is no timestamp for the non revocable attributes (indy/verifier.py). 
`if (timestamp is not None) ^ bool(non_revoc_intervals.get(uuid))
`
Error:
`aries_cloudagent.indy.sdk.verifier ERROR Presentation on nonce=79516549657105118470684 cannot be validated: Timestamp on sub-proof #1 is missing vs. requested attribute group attrGrp_1`
```
"results": [
    {
      "updated_at": "2022-03-03T10:00:04.820234Z",
      "connection_id": "e144d8a6-7996-446c-b61a-e80a0b440de5",
      "auto_present": false,
      "thread_id": "e3ac25cf-59e0-4e0d-90e3-0d85253143f5",
      "created_at": "2022-03-03T09:57:51.268109Z",
      "presentation_request": {
        "nonce": "1104330525903017641791888",
        "name": "twoCredCheck",
        "version": "1.0",
        "requested_attributes": {
          "attrGrp_0": {
            "names": [
              "one",
              "two"
            ],
            "restrictions": [
              {
                "cred_def_id": "5vT6qtd26Rc9huD4tL2cXm:3:CL:8055:rev"
              }
            ]
          },
          "attrGrp_1": {
            "names": [
              "one",
              "two"
            ],
            "restrictions": [
              {
[...]

PR #1672, opened 2022-03-17
Add an integration test for mixed proof with a revocable cred and a n…

ianco:
…on-revocable cred

### DIFF
--- a/demo/features/0454-present-proof.feature
+++ b/demo/features/0454-present-proof.feature
@@ -40,7 +40,7 @@ Feature: RFC 0454 Aries agent present proof
 
    @T001.2-RFC0454 @GHA
    Scenario Outline: Present Proof json-ld where the prover does not propose a presentation of the proof and is acknowledged
-      Given we have "2" agents
+      Given we have "3" agents
          | name  | role     | capabilities        |
          | Acme  | issuer   | <Acme_capabilities> |
          | Faber | verifier | <Acme_capabilities> |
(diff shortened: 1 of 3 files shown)


swcurran:
Isn't that good news?  We can fix it.  We have a failed test -- the right place to start :-)

At least I hope we can fix it.

ianco:
> Isn't that good news? We can fix it. We have a failed test -- the right place to start :-)
> 
> At least I hope we can fix it.

If Indy/Askar is rejecting the proof then I'm not sure it will be so easy to fix :-(

ianco:
I double-checked indy-sdk and I didn't find any tests with mixed proofs (including revocable and non-revocable creds).  It may be worth adding a test to Indy (and/or Askar) to verify whether this is supported or not.
From the error messages it looks like the proof is being rejected by the underlying library.

swcurran:
I'm wondering if we can do some shenanigans in ACA-Py based on knowledge of the credential being revokable or not.  What we want is for the business layer (above ACA-Py) to care if the credential is revokable or not as the verifier may not know when constructing the proof request. For example, a request for a university credential where two universities use the same schema, but one has a revokable credential, the other doesn't.  To me, the verifier should always assume a revokable credential and still have things work if the holder presents a proof from a non-revokable credential.

ianco:
I was digging into this a bit more and Indy is definitely rejecting the proof.  I found this comment in the code so I suspect that this is a non-Indy supported use case:

```
        Irrevocable credentials constitute proof of non-revocation, but
        indy rejects proof requests with non-revocation intervals lining up
        with non-revocable credentials in proof: seek and remove.
```

swcurran:
Arrggh...

@andrewwhitehead -- thoughts on this?  Can/should CredX change that?

I don't like this...

ianco:
Updated the integration test per my comment on Issue #1651 - when the non-revoked interval is specified per attribute (revealed or predicate) then the proof can contain mixed revocable and non-revocable credentials.  (I think aca-py is cleverly "fixing" the proof before asking indy-sdk to validate).

If the non-revoked interval is specified globally (for the entire proof request) then all presented credentials must be revocable and non-revoked.

shaangill025 (approved):

PR #1690, opened 2022-03-25
Fix: DIF proof proposal when creating bound presentation request [Issue#1687]

### LINKED ISSUE #1687: present-proof v2 puts records to the storage it can't read back any more
I ran into this issue and I think it might be possible at the moment that someone sends data to running 3rd party acapy instances to break their usage of reading their own received presentation exchanges from acapy - and it wouldn't be straight forward to recover from it (deleting such records via the API).

Here are the steps to reproduce the situation:
I'm trying to send a W3C presentation. I'm using
`/present-proof-2.0/send-proposal` for this and I'm probably using some wrong format I guess, but this is not the problem here.
Here is the (probably wrong) data that I send to the endpoint:

```

{
    "auto_present": true,
    "comment": "test",
    "connection_id": "7a5d96b0-98fa-4208-9cec-8779ebb55c21",
    "presentation_proposal": {
      "dif": {

        "input_descriptors": [
            {
                "id": "XXX",
                "name": "CO2 Confirmation",
                "schema": [
                    {
                    }
                ],
                "constraints": {
                    "fields": [
                        {
                            "path": [
                                "$.type"
                            ],
                            "filter": {
                                "type": "string",
                                "pattern": "Co2ConfirmationCredential"
                            }
                        }
                    ]
                }
            }
        ]
      }
    },
    "trace": false
  }
```

Above data is just to make the problem reproducible.

I'm using sub-wallets with askar profile: `"--multitenancy-config", "{\"wallet_type\":\"askar-profile\"}"`

Now, what happens on the other side (receiver of the proposal, the acapy behind the given connection_id) is (a lot of auto-flags ON... but probably not relevant here) acapy stores a record that it can't read back any more.
When trying to read via `/present-proof-2.0/records`
```
marshmallow.exceptions.ValidationError: {'presentation_definition': ['Missing data for required field.'], 'input_descriptors': ['Unknown field.']}
```
It breaks in a way that the endpoint doesn't deliver any other item from this endpoint any more.
Debugging the pres_ex_id and trying to delete from the according endpoint `DELETE /present-proof-2.0/records/{pres_ex_id}` also doesn't work because it tries to read the item from storage before it removes it.

So my question - if I should try to fix this - what would be the expected behavior in such cases?

1. we should never get such wrong data items into the storage in the first place and it's ok to not being able to read this AND other data back from storage in such rare cases (current situation)
2. `DELETE` should still work and command line should print out the problematic `pres_ex_id`
3. Listing records should still work for non-problematic records
4. Listing records should still work and add a "dummy" record that just contains the `pres_ex_id` for the problematic record (being able to use the DELETE endpoint with it...) 
5. any other options?

shaangill025:
- resolve #1687 
- So the validation for `/present-proof-v2.0/send-proposal` is working but the problem was that with `auto_respond_presentation_proposal` flag on it would create an invalid bound request from the proposal [`input_descriptors` was not enclosed in `presentation_definition`]. I am surprised that I didn't catch this earlier.
- @swcurran According to the [RFC](https://github.com/hyperledger/aries-rfcs/blob/main/features/0510-dif-pres-exch-attach/README.md#examples-propose-presentation), the `propose-presentation` message only expects `input_descriptors` and no `options` [for `challenge` and/or `domain`]. With auto flag, there will be no opportunity to add a domain to the request, so I have updated the schema to allow `options` in the `propose-presentation` message. `challenge` is always required for verifying presentation, if `options` is not specified then it is randomly assigned.

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
+++ b/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
@@ -160,6 +160,8 @@ class DIFPresFormatHandler(V20PresFormatHandler):
         else:
             dif_proof_request["options"] = pres_proposal_dict["options"]
             del pres_proposal_dict["options"]
+            if "challenge" not in dif_proof_request.get("options"):
+                dif_proof_request["options"]["challenge"] = str(uuid4())
         dif_proof_request["presentation_definition"] = pres_proposal_dict
 
         return self.get_format_data(PRES_20_REQUEST, dif_proof_request)
--- a/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
+++ b/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
@@ -151,9 +151,16 @@ class DIFPresFormatHandler(V20PresFormatHandler):
             A tuple (updated presentation exchange record, presentation request message)
 
         """
-        dif_proof_request = pres_ex_record.pres_proposal.attachment(
+        dif_proof_request = {}
+        pres_proposal_dict = pres_ex_record.pres_proposal.attachment(
             DIFPresFormatHandler.format
         )
+        if "options" not in pres_proposal_dict:
+            dif_proof_request["options"] = {"challenge": str(uuid4())}
+        else:
+            dif_proof_request["options"] = pres_proposal_dict["options"]
+            del pres_proposal_dict["options"]
+        dif_proof_request["presentation_definition"] = pres_proposal_dict
 
         return self.get_format_data(PRES_20_REQUEST, dif_proof_request)
 
--- a/aries_cloudagent/protocols/present_proof/dif/pres_proposal_schema.py
+++ b/aries_cloudagent/protocols/present_proof/dif/pres_proposal_schema.py
@@ -3,7 +3,7 @@ from marshmallow import fields
 
 from ....messaging.models.openapi import OpenAPISchema
 
-from .pres_exch import InputDescriptorsSchema
+from .pres_exch import InputDescriptorsSchema, DIFOptionsSchema
 
 
 class DIFProofProposalSchema(OpenAPISchema):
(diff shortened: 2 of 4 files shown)


aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py
```diff
@@ -151,9 +151,16 @@ async def create_bound_request(
...
         """
-        dif_proof_request = pres_ex_record.pres_proposal.attachment(
+        dif_proof_request = {}
+        pres_proposal_dict = pres_ex_record.pres_proposal.attachment(
             DIFPresFormatHandler.format
         )
+        if "options" not in pres_proposal_dict:
+            dif_proof_request["options"] = {"challenge": str(uuid4())}
+        else:
+            dif_proof_request["options"] = pres_proposal_dict["options"]
```
andrewwhitehead, line 161: Should we add a challenge value if there isn't one?

shaangill025: Without a `challenge` we cannot verify the presentation.

andrewwhitehead: Right, but at the moment we seem to generate a challenge value only if `options` isn't provided?

shaangill025: Resolved [here](https://github.com/hyperledger/aries-cloudagent-python/blob/b3fe34fc75fbfb3c3896464bc5b7d832c231487c/aries_cloudagent/protocols/present_proof/v2_0/formats/dif/handler.py#L163)

swcurran:
@andrewwhitehead -- please review and (hopefully) approve the PR.

@shaangill025 do we need a reminder about when to change (or at least check) the locked down Ursa dependency?  Maybe I put something in my calendar for (when?) to check this?

andrewwhitehead (approved):

PR #1692, opened 2022-03-25
Multi-tenancy stale wallet clean up

dbluhm:
This PR is intended to supersede #928. Due to significant changes since the original PR was opened, I cherry-picked a commit but had to modify it. History has not been well preserved, unfortunately. Credit to @TimoGlastra for the original work on the `ProfileCache` especially and for assistance in finding the solution implemented here.

This PR adds a LRU cache for profiles opened by `MultitenantManager` (corresponding to wallet types `indy` and `askar`, but not `askar-profile`). `weakref.finalize` is used to ensure the profiles are closed when they fall out of scope. An ordered dictionary of profiles holds a (strong) reference to the profile until it is evicted. If that profile happens to still be in use at the time it is evicted, it will not be closed until other (strong) references to it expire, i.e. after processing of a message or admin request has finished. This exact scenario is unlikely but possible.

In addition to this profile caching mechanism, I also updated a few aspects of the `BaseMultitenantManager` and its subclasses to make them more consistent with conventions used in ACA-Py.

Also, after being very confused by handling of askar profiles (not to be confused with `AskarProfile`s) within the `AskarProfileMultitenantManager` and `AskarProfile`, I did some light updates to (I hope) improve clarity.

A brief listing of known limitations:
- As implemented, the capacity of the LRU cache is statically defined as 100.
- A temporally based system for evicting profiles from the cache is likely better than a max capacity based system.
-

### DIFF
--- a/aries_cloudagent/askar/profile.py
+++ b/aries_cloudagent/askar/profile.py
@@ -6,8 +6,8 @@ import time
 
 # import traceback
 
-from typing import Any, Mapping
-from weakref import ref
+from typing import Any, Mapping, Optional
+from weakref import finalize, ref
 
 from aries_askar import AskarError, Session, Store
 
@@ -145,6 +145,18 @@ class AskarProfile(Profile):
         if self.opened:
             await self.opened.close()
             self.opened = None
+
+    def finalizer(self) -> Optional[finalize]:
+        """Return a finalizer for this profile.
+
+        See docs for weakref.finalize for more details on behavior of finalizers.
+        """
+
+        def _finalize(opened: Optional[AskarOpenStore]):
+            if opened:
+                asyncio.get_event_loop().run_until_complete(opened.close())
+
+        return finalize(self, _finalize, self.opened)
 
 
 class AskarProfileSession(ProfileSession):
--- a/aries_cloudagent/askar/profile.py
+++ b/aries_cloudagent/askar/profile.py
@@ -36,11 +36,18 @@ class AskarProfile(Profile):
 
     BACKEND_NAME = "askar"
 
-    def __init__(self, opened: AskarOpenStore, context: InjectionContext = None):
+    def __init__(
+        self,
+        opened: AskarOpenStore,
+        context: InjectionContext = None,
+        *,
+        profile_id: str = None
+    ):
         """Create a new AskarProfile instance."""
         super().__init__(context=context, name=opened.name, created=opened.created)
         self.opened = opened
         self.ledger_pool: IndyVdrLedgerPool = None
+        self.profile_id = profile_id
         self.init_ledger_pool()
         self.bind_providers()
 
@@ -56,8 +63,8 @@ class AskarProfile(Profile):
 
     async def remove(self):
         """Remove the profile."""
-        if self.settings.get("multitenant.wallet_type") == "askar-profile":
-            await self.store.remove_profile(self.settings.get("wallet.askar_profile"))
+        if self.profile_id:
+            await self.store.remove_profile(self.profile_id)
 
     def init_ledger_pool(self):
         """Initialize the ledger pool."""
@@ -151,6 +158,11 @@ class AskarProfile(Profile):
 
         See docs for weakref.finalize for more details on behavior of finalizers.
         """
+        # Askar Profiles (not to be confused with AskarProfiles) don't need
+        # additional clean up
+
+        if self.profile_id:
+            return None
 
         def _finalize(opened: Optional[AskarOpenStore]):
             if opened:
@@ -172,11 +184,10 @@ class AskarProfileSession(ProfileSession):
     ):
         """Create a new IndySdkProfileSession instance."""
         super().__init__(profile=profile, context=context, settings=settings)
-        profile_id = profile.context.settings.get("wallet.askar_profile")
         if is_txn:
-            self._opener = self.profile.store.transaction(profile_id)
+            self._opener = self.profile.store.transaction(profile.profile_id)
         else:
-            self._opener = self.profile.store.session(profile_id)
+            self._opener = self.profile.store.session(profile.profile_id)
         self._handle: Session = None
         self._acquire_start: float = None
         self._acquire_end: float = None
--- a/aries_cloudagent/core/profile.py
+++ b/aries_cloudagent/core/profile.py
@@ -4,6 +4,7 @@ import logging
 
 from abc import ABC, abstractmethod
 from typing import Any, Mapping, Optional, Type
+from weakref import finalize
 
 from .event_bus import EventBus, Event
 from ..config.base import InjectionError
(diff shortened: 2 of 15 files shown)


dbluhm:
I'm still in the process of validating the cache behavior so I've opened this as a draft PR but I would appreciate feedback on what's here so far. Due to resourcing constraints, it's not likely that I'll be able to implement a time based cache (I haven't evaluated what that would look like, though, so if it ends up being super easy, maybe lol).

I would be particularly interested to hear feedback from @TimoGlastra and @andrewwhitehead since the two of you are the original participants on #928 but more than happy to get feedback from anyone.

swcurran:
@paulwen - this would seem a really good PR to test with the Load Generator.  Do you have an easy way to run tests on a PR branch?

dbluhm:
Odd how submitting a PR suddenly brings clarity lol; as implemented, if more than 100 (or whatever the capacity is defined as) is actively in use, this can result in rapid opening and closure of wallets and `WalletAlreadyOpen` errors :thinking: I'm probably not quite on the mark here yet...

PaulWen:
This AcaPy repo is already integrated as a git submodule into the Load Generator repo to build AcaPy images with a debugger. Therefore, it should not be difficult to checkout any branch and build a docker image in a similar manner.

@dbluhm Could you describe to me what a suitable test flow would look like? Feel free to drop me a PM in Discord.

dbluhm:
I've sorted out most of the issues with this PR. I was having a heck of a time identifying where references to profiles were being held open but I've found the culprits (reponders held references to profiles/contexts and then profiles held references to responders...) and my rudimentary tests behave as expected.

I'll do my bit to cover new lines better for the sake of code coverage but I'm open to ideas for how or even if we should have tests exercising this functionality inside of ACA-Py.

Should have this ready for review soon.

aries_cloudagent/askar/profile.py
```diff
@@ -146,6 +153,24 @@ async def close(self):
...
+        # Askar Profiles (not to be confused with AskarProfiles) don't need
+        # additional clean up
+
+        if self.profile_id:
+            return None
+
+        def _finalize(opened: Optional[AskarOpenStore]):
+            if opened:
+                LOGGER.debug("Profile finalizer called; closing wallet")
+                asyncio.get_event_loop().create_task(opened.close())
```
andrewwhitehead, line 170: It's nice but not necessary to close the store this way, as the library has its own finalizers. With multiple profile instances potentially referencing the same Store it would probably not be correct to close the store when a single profile is closed, either.

aries_cloudagent/core/profile.py
```diff
@@ -115,6 +116,18 @@ def inject_or(
     async def close(self):
         """Close the profile instance."""
 
+    def finalizer(self) -> Optional[finalize]:
```
andrewwhitehead, line 119: I think the finalizer can just be set up in the profile's __init__ method when required. We don't have to keep a reference to it, and requiring the caller to initialize it seems error prone.

dbluhm:
Not realizing that Askar already had finalizers and considering @andrewwhitehead's comments regarding requiring the caller to initialize the finalizer being potentially error prone, I've simplified the finalizer additions. Essentially, finalizers for Indy SDK wallet profiles are always created. I don't think there's a scenario where it would be appropriate for the wallet to remain opened after the profile has gone out of scope so I think this is a valid simplification.

In other words, all profiles managed by the profile cache are now assumed to clean up their own resources after falling out of scope and it is up to the profile implementation to ensure this is the case.

aries_cloudagent/config/argparse.py
```diff
@@ -1603,12 +1603,13 @@ def add_arguments(self, parser: ArgumentParser):
...
+            nargs="+",
+            metavar="key=value",
             env_var="ACAPY_MULTITENANCY_CONFIGURATION",
             help=(
-                'Specify multitenancy configuration ("wallet_type" and "wallet_name"). '
-                'For example: "{"wallet_type":"askar-profile","wallet_name":'
-                '"askar-profile-name", "key_derivation_method":"RAW"}"'
+                "Specify multitenancy configuration in key=value pairs. "
+                'For example: "wallet_type=askar-profile wallet_name=askar-profile-name" '
+                "Possible values: wallet_name, wallet_key, cache_size. "
```
dbluhm, line 1612: Key derivation help text is lost

aries_cloudagent/indy/sdk/profile.py
```diff
@@ -116,6 +122,19 @@ async def close(self):
...
             self.opened = None
 
+    def _make_finalizer(self) -> finalize:
+        """Return a finalizer for this profile.
+
+        See docs for weakref.finalize for more details on behavior of finalizers.
+        """
+
+        def _finalize(opened: Optional[IndyOpenWallet]):
+            if opened:
```
andrewwhitehead, line 132: I don't think opened can be None because it receives the original IndyOpenWallet instance. Maybe `if opened.handle`?

aries_cloudagent/multitenant/askar_profile_manager.py
```diff
@@ -15,13 +16,23 @@ class AskarProfileMultitenantManager(BaseMultitenantManager):
 
     DEFAULT_MULTIENANT_WALLET_NAME = "multitenant_sub_wallet"
```
andrewwhitehead, line 17: Maybe we could fix this variable name

aries_cloudagent/multitenant/base.py
```diff
@@ -48,7 +48,10 @@ def __init__(self, profile: Profile):
         if not profile:
             raise MultitenantManagerError("Missing profile")
 
-        self._instances: dict[str, Profile] = {}
+    @abstractproperty
+    def open_profiles(self) -> Iterable[Profile]:
+        """Return iterator over open profiles."""
+        ...
```
andrewwhitehead, line 54: Ellipsis?

aries_cloudagent/indy/sdk/profile.py
dbluhm: This came as a recommendation from pyright which indicates that IndySdkProfile.opened is `IndyOpenWallet | None`. I think it derived this from the `self.opened = None` line in the `close` method.

Following that same vein, I think it was possible before for the finalizer to be created after `close` had been called. Now that it's only intended to be called in `__init__`, I don't think this is a problem any more. I'll fix that.

aries_cloudagent/multitenant/base.py
andrewwhitehead: Apparently it's about the same as writing 'pass', I just haven't seen it much before.

dbluhm:
Digging into the unit test failure; it's not behaving quite the same way when I run `run_tests_indy` locally so a little perplexed. Should have more time to look at this shortly.

PR #1694, opened 2022-03-28
Feature/undelivered events

mepeltier:
This PR adds in a number of events which are emitted when a message is undelivered.

These events can be leveraged by plugins to take certain actions when messages are undelivered, such as writing the message to a persistent queue, sending a push notification to a mobile device that attempted to send the message, or some other action.

The events leverage the existing `OutboundSendStatus` states and don’t supersede the current return value. The `OutboundSendStatus` is still returned within ACA-Py as it has been, so this shouldn’t impact any existing functionality, but I’d be happy to hear any concerns I’ve not thought of. Thanks to @dbluhm for his feedback prior to this PR being opened.

### DIFF
--- a/aries_cloudagent/core/conductor.py
+++ b/aries_cloudagent/core/conductor.py
@@ -658,7 +658,7 @@ class Conductor:
         if self.outbound_queue:
             return await self._queue_external(profile, outbound)
         else:
-            return self._queue_internal(profile, outbound)
+            return await self._queue_internal(profile, outbound)
 
     async def _queue_external(
         self,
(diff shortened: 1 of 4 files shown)


dbluhm (approved):

dbluhm:
I was involved in this PR behind the scenes so probably a good idea to get a review from another ACA-Py commiter :slightly_smiling_face:

ianco:
@mepeltier  can you sync your PR with the latest `main` branch?  Thanks!

ianco (approved):

Issue #1689, opened 2022-03-24
Trouble Running Faber Demo On Apple Silicon


drewcapener:
Hey, I am an Aries newbie, and I am just getting familiar with some of the projects. I wanted to run the [Alice/Faber Demo](https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/README.md) according to the instructions provided under the "Running in Docker" section. I was able to successfully follow the instructions up until `./run_demo faber` where I get the following error: 

`Faber      | Operation not permitted (src/thread.cpp:231)`
`Faber      | qemu: uncaught target signal 6 (Aborted) - core dumped`

As mentioned in the title, I am on Apple Silicon, and my instinct is to say this is where the issue is coming from because I have seen people getting this error on m1 chips [elsewhere](https://gitlab.com/qemu-project/qemu/-/issues/340), but this could also be something else. I would be very grateful to anyone who could help me figure this out!

andrewwhitehead:
Hiya, could you please add the versions of your MacOS install and Docker install for reproduction. Are you using Docker Desktop?

drewcapener:
@andrewwhitehead Thanks for the help.

Monterey 12.2.1
Docker Desktop 20.10.11

drewcapener:
[This issue](https://github.com/bcgov/von-network/issues/182) feels like it might hold some answers?

drewcapener:
@TimoGlastra I saw you were wrestling with this issue on a couple of issues. Did you get it figured out?

drewcapener:
Bit of an update- I was able to replace the base image with an unofficial one (snel/von-image:py36-1.16-arm64) built for arm64. I'm not sure if this will work, but for now it seems to have at least moved me to a different issue. Now, I'm getting the following: 

`#8 16.77 ERROR: Could not find a version that satisfies the requirement aries-askar~=0.2.4 (from versions: 0.1.0, 0.1.2)`

andrewwhitehead:
It has universal packages for Askar, maybe you need to upgrade pip?

drewcapener:
@andrewwhitehead Just to make sure we're on the same page, are you suggesting upgrading pip in the container? I tried adding this line to no avail:

`RUN pip3 install --upgrade pip`

and then tried this one with no success either:

`RUN python3 -m pip install --upgrade pip`

andrewwhitehead:
[...]

PR #1697, opened 2022-03-28
platform target in run tests.

burdettadam:
@dbluhm suggested seeing if adding a platform target would allow docker to run the test scripts, which worked. I expect more changes for complete mac m1 development support but this seems to be a good starting point.


swcurran:
Continues to work on Linux/Ubuntu on Intel processor, so that's good :-)

andrewwhitehead:
See also #1689

dbluhm:
We'll likely want similar changes in the `run_docker` and other scripts

andrewwhitehead (approved):

PR #1701, opened 2022-03-30
Allow deleting a pres exchange item without reading it first

matgnt:
Related to #1687 this allows deleting a presentation exchange
record without reading it from storage.

@shaangill025 Does it make sense to have something like this in addition to your change in #1690 to make sure we always have an option to delete an item without reading it?

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v2_0/routes.py
+++ b/aries_cloudagent/protocols/present_proof/v2_0/routes.py
@@ -32,6 +32,7 @@ from ....messaging.valid import (
     UUID4,
 )
 from ....storage.error import StorageError, StorageNotFoundError
+from ....storage.base import BaseStorage
 from ....storage.vc_holder.base import VCHolder
 from ....storage.vc_holder.vc_record import VCRecord
 from ....utils.tracing import trace_event, get_timer, AdminAPIMessageTracingSchema
@@ -1265,8 +1266,17 @@ async def present_proof_remove(request: web.BaseRequest):
     pres_ex_record = None
     try:
         async with context.profile.session() as session:
-            pres_ex_record = await V20PresExRecord.retrieve_by_id(session, pres_ex_id)
-            await pres_ex_record.delete_record(session)
+            try:
+                pres_ex_record = await V20PresExRecord.retrieve_by_id(
+                    session, pres_ex_id
+                )
+                await pres_ex_record.delete_record(session)
+            except (BaseModelError, ValidationError):
+                storage = session.inject(BaseStorage)
+                storage_record = await storage.get_record(
+                    record_type=V20PresExRecord.RECORD_TYPE, record_id=pres_ex_id
+                )
+                await storage.delete_record(storage_record)
     except StorageNotFoundError as err:
         raise web.HTTPNotFound(reason=err.roll_up) from err
     except StorageError as err:
--- a/aries_cloudagent/messaging/models/base_record.py
+++ b/aries_cloudagent/messaging/models/base_record.py
@@ -19,7 +19,7 @@ from ...storage.record import StorageRecord
 from ..util import datetime_to_str, time_now
 from ..valid import INDY_ISO8601_DATETIME
 
-from .base import BaseModel, BaseModelSchema
+from .base import BaseModel, BaseModelSchema, BaseModelError
 
 LOGGER = logging.getLogger(__name__)
 
@@ -329,7 +329,10 @@ class BaseRecord(BaseModel):
                 positive=False,
                 alt=alt,
             ):
-                result.append(cls.from_storage(record.id, vals))
+                try:
+                    result.append(cls.from_storage(record.id, vals))
+                except BaseModelError as err:
+                    raise BaseModelError(f"{err}, for record id {record.id}")
         return result
 
     async def save(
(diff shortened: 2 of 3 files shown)


aries_cloudagent/protocols/present_proof/v2_0/routes.py
```diff
@@ -1234,8 +1236,10 @@ async def present_proof_remove(request: web.BaseRequest):
     pres_ex_record = None
     try:
         async with context.profile.session() as session:
-            pres_ex_record = await V20PresExRecord.retrieve_by_id(session, pres_ex_id)
-            await pres_ex_record.delete_record(session)
+            storage = session.inject(BaseStorage)
+            storage_record = StorageRecord(type=V20PresExRecord.RECORD_TYPE, value=None, id=pres_ex_id)
```
shaangill025, line 1240: I think you refactor this to something like:
`storage_record = await storage.get_record(record_type=V20PresExRecord.RECORD_TYPE, record_id=pres_ex_id)`

shaangill025:
After #1690 fix, are you still aware of any scenario which would cause an issue like that? We generally rely on Marshmallow schema validation. The logic of retrieving `StorageRecord` is not being used anywhere else in `routes.py` for handling `BaseExchangeRecord` deletion.

matgnt:
> After #1690 fix, are you still aware of any scenario which would cause an issue like that? We generally rely on Marshmallow schema validation. The logic of retrieving `StorageRecord` is not being used anywhere else in `routes.py` for handling `BaseExchangeRecord` deletion.

Good question. To be honest, probably not. I thought deleting should always be possible. Since I don't know marshmallow exactly, does it handle versions of schemas? Could a change like the one in #1690 lead us to old records and new records with problems to read one or the other? If so, I think at least deleting a record should work just by the identifier and not trying to go through validation - which is not required for deleting. This should give us a chance to recover live systems from such not foreseen situations. Or are there other ways to access/delete records from storage to recover?

I guess that is a more generic question not bound to present-proof-v2, but other delete operations, too. E.g. present-proof-V1 also does the loading into V10PresentationExchange before it deletes it.

shaangill025:
I created 2 PRs against [boschresearch:fix-deleting-pres-exchange](https://github.com/boschresearch/aries-cloudagent-python/tree/fix-deleting-pres-exchange), I understand your requirement to recover live systems.
- [Minor update to existing work](https://github.com/boschresearch/aries-cloudagent-python/pull/3)
- [Alternative approach using upgrade command](https://github.com/boschresearch/aries-cloudagent-python/pull/4)

matgnt:
> I created 2 PRs against [boschresearch:fix-deleting-pres-exchange](https://github.com/boschresearch/aries-cloudagent-python/tree/fix-deleting-pres-exchange), I understand your requirement to recover live systems.
> 
> * [Minor update to existing work](https://github.com/boschresearch/aries-cloudagent-python/pull/3)
> * [Alternative approach using upgrade command](https://github.com/boschresearch/aries-cloudagent-python/pull/4)

I did merge those PRs and check the diff. I rebased on main now (which basically removes my original PR...)

I did not yet check the upgrade function in detail, but I think that would be a proper approach.
The part for deleting a record looks very good to me. Thanks for taking up the concerns with the recovering of live systems!

shaangill025:
@matgnt Were you able to check the `upgrade` command approach? I was thinking that maybe we can run that on your end to update the bad exchange records and then delete those changes [`/commands/upgrade.py`] from this PR and try to merge other changes related to the identification and deletion of bad `V20PresExRecord`.

matgnt:
@shaangill025 I already used my original changes to delete those items. Do you have plans to integrate such an 'upgrade' command into a future release? I think in general we should consider those features to recover live systems.

I removed the 'upgrade' part from this PR for now.

matgnt:
It looks like the tests are at least somehow in the part of acapy where we made some changes. Any idea @shaangill025 ?

shaangill025:
I am not sure why those tests are failing as they are not related to the changes made, can you merge `hyperledger/aries-cloudagent-python` into `boschresearch:fix-deleting-pres-exchange`.

shaangill025:
> Do you have plans to integrate such an 'upgrade' command into a future release? I think in general we should consider those features to recover live systems.

`upgrade` command is already implemented and logic to update/fix existing `BaseRecord` can be added depending upon the requirement to `update_existing_records` method as shown [here](https://github.com/shaangill025/aries-cloudagent-python/blob/4cef080c790452793bc6d6a70e03da30ec4b33fb/aries_cloudagent/commands/upgrade.py#L226). But I don't see the benefit of going the `upgrade` route for this use case as the issue was only reported by you and it has since been resolved on your end, the bug which caused it has also been fixed and merged.

shaangill025:
If possible, can you add `https://github.com/hyperledger/aries-cloudagent-python.git` as `upstream` [remote] and then do a `git pull upstream main`

matgnt:
> If possible, can you add `https://github.com/hyperledger/aries-cloudagent-python.git` as `upstream` [remote] and then do a `git pull upstream main`

I already did merge main into this branch, but the triggered test run also failed. I'm lost a bit ;-)
Feel free to close this PR and submit from your end if that is easier for you :-)

shaangill025:
Recreated in PR#1718 as some tests were failing after rebase even though local tests are passing.

PR #1705, opened 2022-04-01
Fix Issue-1682

DaevMithran:
- Added wallet startup parameter: --wallet-allow-insecure-seed
- support seed in wallet/did/create route

Some questions around
1. How do we support importing a key pair into the wallet as mentioned in the Issue
2. Should I add seed in the schema of wallet/did/create ? . I don't think we can add conditional schemas according to startup parameter.  I can have it as an optional field.

### DIFF
--- a/aries_cloudagent/wallet/routes.py
+++ b/aries_cloudagent/wallet/routes.py
@@ -355,13 +355,20 @@ async def wallet_create_did(request: web.BaseRequest):
                 f" support key type {key_type.key_type}"
             )
         )
+    seed = None
+    if context.settings.get("wallet.allow_insecure_seed"):
+        seed = body.get("seed") or None
     info = None
     async with context.session() as session:
         wallet = session.inject_or(BaseWallet)
         if not wallet:
             raise web.HTTPForbidden(reason="No wallet available")
         try:
-            info = await wallet.create_local_did(method=method, key_type=key_type)
+            info = await wallet.create_local_did(
+                method=method,
+                key_type=key_type,
+                seed=seed,
+            )
 
         except WalletError as err:
             raise web.HTTPBadRequest(reason=err.roll_up) from err
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -1424,6 +1424,15 @@ class WalletGroup(ArgumentGroup):
             ),
         )
         parser.add_argument(
+            "--wallet-allow-insecure-seed",
+            action="store_true",
+            env_var="ACAPY_WALLET_ALLOW_INSECURE_SEED",
+            help=(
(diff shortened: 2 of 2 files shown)


DaevMithran:
Should I add seed in the schema of wallet/did/create ? . I don't think we can add conditional schemas according to startup parameter.  I can have it as an optional field.

ianco:
This looks good!

> 1. How do we support importing a key pair into the wallet as mentioned in the Issue
I'm not sure if we can support this.  We need to store the private key as well, which I don't think we can get from the did/verkey pair.  So we have to either use a seed to generate a random did.

> 2. Should I add seed in the schema of wallet/did/create ? . I don't think we can add conditional schemas according to startup parameter.  I can have it as an optional field.
No I think let's leave it out of the schema.  If someone is using this feature then they will know about it (after all they have to enable the feature using the new param), and if we include it always in the schema it will confuse people.

aries_cloudagent/wallet/routes.py
```diff
@@ -355,13 +355,20 @@ async def wallet_create_did(request: web.BaseRequest):
                 f" support key type {key_type.key_type}"
             )
         )
+    seed = None
+    if context.settings.get("wallet.allow_insecure_seed"):
+        seed = body.get("seed") or None
```
TimoGlastra, line 360: Would be good to throw an error if a seed is provided and `wallet.allow_insecure_seed` is not enable to make it more clear to the end user the seed is not used. This can give the false sense the seed is being used to create the did.

aries_cloudagent/wallet/routes.py
```diff
@@ -355,13 +355,20 @@ async def wallet_create_did(request: web.BaseRequest):
                 f" support key type {key_type.key_type}"
             )
         )
+    seed = None
```
TimoGlastra, line 358: The seed property is not added to the `DIDCreateSchema` making it harder to discover from the swagger ui and it won't be included when generating clients from the swagger file.

TimoGlastra (commented):
Sorry for being too late with my review, but added some nits.

aries_cloudagent/wallet/routes.py
DaevMithran: @ianco your thoughts on this?

aries_cloudagent/wallet/routes.py
DaevMithran: @TimoGlastra can you suggest the error message to be displayed for this. Can it be a simple HTTPBadRequest: Seed not supported ?

aries_cloudagent/wallet/routes.py
DaevMithran: agreed

PR #1714, opened 2022-04-06
Add seed command line parameter but use only if also an "allow insecure seed" parameter is set

DaevMithran:
Signed-off-by: DaevMithran <daevmithran1999@gmail.com>

Updated pull request changes suggested by @TimoGlastra 

- Raising an http bad request error when a seed is passed in the body and wallet-allow-insecure-seed is not enabled
- Updated schema for wallet/did/create


aries_cloudagent/wallet/routes.py
```diff
@@ -188,6 +188,12 @@ class DIDCreateSchema(OpenAPISchema):
         description="To define a key type for a did:key",
     )
 
+    seed = fields.Str(
+        required=False,
+        description="Optional seed to use for DID",
```
TimoGlastra, line 193:
```suggestion
        description="Optional seed to use for DID. Must be enabled in configuration before use.",
```

TimoGlastra (approved):
LGTM! Nice @DaevMithran

aries_cloudagent/wallet/routes.py
DaevMithran: Resolved

ianco: I don't think we should advertise the `seed` unless the startup option is set, so if we can't make it conditional in the schema we should add a separate endpoint that we can enable/disable based on the `--wallet-allow-insecure-seed` parameter.
@swcurran  @andrewwhitehead  thoughts?

DaevMithran: We are returning an error if a seed is passed without the startup parameter. So I believe people will not get confused with the schema.

andrewwhitehead: @ianco I don't think we can add conditional endpoints or parameters as then we'd have to look at what options were passed when generating the openapi definition

TimoGlastra: What is the reason we want to hide this? There's a lot of properties that are available based on startup parameters (using public dids for invitations requires both the `--public-invites` and a ledger to be configured to work), but are always available in the swagger ui. If not configured, an error will be thrown when you try to use it.

Sorry for pushing you in the wrong direction @DaevMithran.

ianco: I think we want to discourage this (use of seed), unless there is a specific use case for it, in which case the implementer can figure it out.  If we expose this as a top-level parameter then I suspect everyone will just use it, in which case the extra startup parameter will just become one of the de-facto parameters that everyone uses.

Two separate endpoints seems like a good compromise to me, similar to the discussion we had around the multi-tenant bearer token endpoint.

However if I'm out-voted I'll go along with whatever is the consensus decision.  @swcurran  or @andrewwhitehead   can weigh in ...

andrewwhitehead: I guess the other option would be to require encryption of the seed when it's passed it, but that's another can of worms :)

TimoGlastra: Ok let's keep it out of the docs then. I still think throwing an error would be nice, as this doesn't leave users thinking they've used the secret ACA-Py feature while it's actually not enabled and you don't have the seed anymore

swcurran:
The error in the integration tests is unrelated to this PR (see issue #1832 ).  Once we fix that, we'll trigger another integration run.

swcurran:
@ianco -- can you please review.  Is this OK?  Or should we not do this.  We've not had a request for it for a while.

I'm thinking that there should be two new command line parameters (seed and "...insecure") here, but I don't see any documentation about those parameters.

Thanks

ianco:
The new `seed` parameter is for creating a new DID within an existing wallet (mainly for a new multi-tenant wallet) and is not related to the `--seed` startup parameter.  (The former creates a new DID within a wallet, the latter establishes the public DID for the main wallet on startup).  So I think this PR is ok.

ianco (approved):

swcurran:
True about the seed parameter (doh!!) -- but doesn't there have to be an "allow_insecure_seed" parameter and guidance/help text for how to use that?

ianco:
> True about the seed parameter (doh!!) -- but doesn't there have to be an "allow_insecure_seed" parameter and guidance/help text for how to use that?

Yes that already exists:

https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/config/argparse.py#L1453

PR #1719, opened 2022-04-07
Allow specifying key derivation method on sub-wallet create

dbluhm:
When working in managed multi-tenancy mode, we store the keys for the sub-wallets within a WalletRecord in ACA-Py. These keys are then recalled when needed and used to unlock the wallet for processing messages or admin API requests intended for that sub-wallet. In the process of opening the wallet, the key is pushed through a key derivation algorithm to transform it into a wallet encryption key. This key derivation algorithm is a costly operation.

This PR implements allowing the key derivation method to be specified when creating a sub-wallet. This means that a multi-tenant ACA-Py agent can elect to use the `RAW` derivation method, for instance, and significantly reduce the wallet open cost.

Since these keys are securely stored in ACA-Py, using a `RAW` key does not seem to impact security of the key. Additionally, I don't think the profile of potential attack changes from using the key derivation method when considering wallet tokens.

Credit to @burdettadam.


ianco (approved):
LGTM

andrewwhitehead (approved):

swcurran:
@dbluhm -- can you please update your branch for merging?

swcurran (approved):

Issue #1707, opened 2022-04-02
GET /schemas/created returns duplicates


marcos-carretero:
This issue is similar to #678, but manifests without sending multiple POST /schemas.

This issue was observed in the IDIM (Dev) ACA-Py instance: 
`"serviceEndpoint": "https://idim-agent-dev.apps.silver.devops.gov.bc.ca"`

### Steps to reproduce:
1. Create a new schema using the `POST /schemas` endpoint
2. Invoke the `GET /schemas/created` method

#### Expected results:

- A single schema is returned for each schema created
```
{
  "schema_ids": [
    "XpgeQa93eZvGSZBZef3PHn:2:BC Person Credential (Dev):0.0.1"
  ]
}
```

#### Actual results:

- Two schema_id records are returned for the same value

```
{
  "schema_ids": [
    "XpgeQa93eZvGSZBZef3PHn:2:BC Person Credential (Dev):0.0.1",
    "XpgeQa93eZvGSZBZef3PHn:2:BC Person Credential (Dev):0.0.1"
  ]
}
```

marcos-carretero:
I have also observed that `GET /credential-definitions/created` also returns duplicates.

swcurran:
Closing this issue as completed based on #1721

PR #1721, opened 2022-04-07
Duplicate checking for schema and cred def

ianco:

See issue:  https://github.com/hyperledger/aries-cloudagent-python/issues/1707

### DIFF
--- a/aries_cloudagent/messaging/credential_definitions/routes.py
+++ b/aries_cloudagent/messaging/credential_definitions/routes.py
@@ -194,7 +194,7 @@ async def credential_definitions_send_credential_definition(request: web.BaseReq
             tag_query=tag_query,
         )
         if 0 < len(found):
-            # need to check te 'tag' value
+            # need to check the 'tag' value
             for record in found:
                 cred_def_id = record.value
                 cred_def_id_parts = cred_def_id.split(":")
--- a/aries_cloudagent/messaging/credential_definitions/routes.py
+++ b/aries_cloudagent/messaging/credential_definitions/routes.py
@@ -186,6 +186,23 @@ async def credential_definitions_send_credential_definition(request: web.BaseReq
     tag = body.get("tag")
     rev_reg_size = body.get("revocation_registry_size")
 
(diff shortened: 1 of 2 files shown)


shaangill025 (approved):

TimoGlastra:
FYI: This PR introduced a breaking change and now broke AATH because AATH tries to recreate the schema. Before ACA-Py would swallow this. Do we need to update this in ACA-Py? Or do we need to change this behaviour in ACA-Py (i.e. not throwing a record, but also not re-creating the schema)


On a side note, it seems to query by the `schema_name` and `schema_version` and throw an error if any records are found. Does that mean if I change my public did and try to create a schema with the same name and version I created with a previous did it will throw an error? I think a schema is unique by name, version, and did.

TimoGlastra:
Created a PR that fixes this in AATH: https://github.com/hyperledger/aries-agent-test-harness/pull/473

swcurran:
I'm not sure I understand why ACA-Py should fail if you try to create the same schema multiple times. Shouldn't it just create the schema if it doesn't exist and continue happily if it does?  AFAIK - the problem previously was that it was creating multiple instances of the same schema (definitely a problem) and returning multiple instances.

On the second point -- I agree in principle that an agent should be allowed to create a schema with the same name but a different DID, since that is permitted on Indy. Not sure how crucial it is though -- it's an "unlikely" situation right now -- but eventually someone will want it...  How big a can of worms does it open if the query includes the public DID for the agent in the check for a schema?

ianco:
Good question about "resource already exists" - I returned an error but other options are available.  I think we should return something that specifies that the resource already exists, though, rather than giving the same result whether the resource was just created or not.  `409` or `303` are options, see https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists

Regarding the "same schema different did" also a good point, I didn't think about an issuer with multiple DID's.  I don't know if this is a realistic example though (wouldn't there just be a single schema and then multiple cred defs?) but I think the topic of an aca-py agent with multiple public DID's introduces more issues than just this one.

Issue #1703, opened 2022-03-30
Set Public DID error with Multitenancy and Author/Endorser


frostyfrog:
This came up while fixing test errors in PR #1696. When running `LEDGER_URL=http://test.bcovrin.vonx.io PUBLIC_TAILS_URL=https://tails-test.vonx.io BDD_EXTRA_AGENT_ARGS="{\"wallet-type\":\"askar\"}" ./run_bdd -t @T003-RFC0586` (for both askar and indy), the tests are failing while attempting to promote a did to a public did. The error that is specifically returned is `aiohttp.web_exceptions.HTTPBadRequest: No endorser connection found` which can be found in `aries_cloudagent/wallet/routes.py` on line 510. 

The code suggests that the call to `get_endorser_connection_id(profile: Profile)` is not returning a connection ID like it should.

MajdT51:
I'm trying to use this new feature and getting this errors in the logs

curl -X 'POST' \
  'http://localhost:11000/wallet/did/public?did=42gXHkTTzGBDb28EopJLRU&conn_id=785cd5b6-c915-4576-9213-9d9bef01771d&create_transaction_for_endorser=true

> 2022-04-02 11:45:20,894 aiohttp.server ERROR Error handling request
> Traceback (most recent call last):
>   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 435, in _handle_request
>     resp = await request_handler(request)
>   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aiohttp/web_app.py", line 504, in _handle
>     resp = await handler(request)
>   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aiohttp/web_middlewares.py", line 117, in impl
>     return await handler(request)
>   File "/home/indy/aries_cloudagent/admin/server.py", line 159, in ready_middleware
>     return await handler(request)
>   File "/home/indy/aries_cloudagent/admin/server.py", line 196, in debug_middleware
>     return await handler(request)
>   File "/home/indy/.pyenv/versions/3.6.13/lib/python3.6/site-packages/aiohttp_apispec/middlewares.py", line 45, in validation_middleware
>     return await handler(request)
>   File "/home/indy/aries_cloudagent/admin/server.py", line 339, in check_multitenant_authorization
>     return await handler(request)
[...]

PR #1726, opened 2022-04-11
Use provided connection_id if provided

ianco:

See issue https://github.com/hyperledger/aries-cloudagent-python/issues/1703


andrewwhitehead (approved):

swcurran:
I meant to ask before...I notice the variable is called "connection_id", but should it be "endorser_connection_id" to be clearly about the purpose?  Or am I just not understanding what this is for...  🤡

ianco:
> I meant to ask before...I notice the variable is called "connection_id", but should it be "endorser_connection_id" to be clearly about the purpose? Or am I just not understanding what this is for... 🤡

Yah you're right, same issue on all the api's that support endorsement.  (On another note if you use all the aca-py parameters you don't need to use this field anyways)

PR #1734, opened 2022-04-19
Feat/revocation notification v2

frostyfrog:
This PR implements Revocation Notification V2, as described in https://github.com/hyperledger/aries-rfcs/pull/721. At this time, V1 takes precedence over V2 when a revocation occurs and V2 never gets fired off. After talking with @dbluhm, I plan to implement a config option to disable/exclude the loading of specific plugins so that, in terms of configuration, one can disable V1 notifications to allow V2 notifications to be triggered. Since Indy-Anoncreds are currently the only standard to be implemented at this time, there are a few hard coded values pointing to "indy-anoncreds" until such a time where a second credential revocation format needs to be implemented.


dbluhm:
@andrewwhitehead are the tests that are failing related to the Askar update you mentioned on the ACA-PUG call today?

ianco:
> This PR implements Revocation Notification V2, as described in [hyperledger/aries-rfcs#721](https://github.com/hyperledger/aries-rfcs/pull/721). At this time, V1 takes precedence over V2 when a revocation occurs and V2 never gets fired off. After talking with @dbluhm, I plan to implement a config option to disable/exclude the loading of specific plugins so that, in terms of configuration, one can disable V1 notifications to allow V2 notifications to be triggered. Since Indy-Anoncreds are currently the only standard to be implemented at this time, there are a few hard coded values pointing to "indy-anoncreds" until such a time where a second credential revocation format needs to be implemented.

So, just to clarify, the V1 revocation notification will run by default (and the V2 revocation notification won't get sent) unless the V1 notification protocol is disabled using this PR?  https://github.com/hyperledger/aries-cloudagent-python/pull/1737

frostyfrog:
That is correct. The reason for this is that revocation notification loads in first (due to the definition.py), and they both work off the same DB entries from revocation notifications. V1 deletes the DB entry before V2 can get to it, and thus, is essentially disabled

ianco:
> Would a GUI generator be possible to allow for interactively creating a the config? Or would we just need a config for that? :-)

We have the capability to use config files (`--arg-file <something.yml>`) so I don't think we need to build a gui, we just need to provide some config.yml examples

... and I don't think it's a real issue to have lots and lots of CLI parameters, as long as they all have reasonable defaults

ianco (changes requested):
See comment re adding a version parameter to the endpoint

frostyfrog:
Sounds good! I'll get those changes made.

aries_cloudagent/revocation/routes.py
```diff
@@ -157,11 +157,16 @@ def validate_fields(self, data, **kwargs):
 
         notify = data.get("notify")
         connection_id = data.get("connection_id")
+        notify_version = data.get("notify_version")
 
         if notify and not connection_id:
             raise ValidationError(
                 "Request must specify connection_id if notify is true"
             )
+        if notify and not notify_version:
```
ianco, line 166: Can we default to `v1` notifications if not specified?  That way it's not a breaking change.

ianco (changes requested):

aries_cloudagent/revocation/routes.py
frostyfrog: Yep! I took the word "required" and assumed that it was supposed to be a breaking change. I'll set it up to default to v1.

frostyfrog: I have left the check in and added a default value to the get lines. Verified that this works by removing the argument from the behavior tests

ianco (approved):

etschelp:
Just as a remark, I tested this and it is not backwards compatible. I was using the config from aca-py 0.7.3, meaning:

```yaml
notify-revocation: true
monitor-revocation-notification: true
```

and also in the rest api `"notify": true`

After upgrading to the latest revision from main revocation notifications were not sent any more until I set a version either v1 or v2 in the rest api. Meaning for anyone who is not aware of this revocation notifications will stop working.

ianco:
Could be setting the attribute to an empty string vs null value?

PR #1737, opened 2022-04-19
Feature: Add the ability to deny specific plugins from loading

frostyfrog:
As part of #1734, I found that the Revocation Notification V2 code
failed to properly run due to the fact that the V1 code was deleting the
DB objects before V2 could get to it. @dbluhm and I decided that adding
an option to deny specific plugins from loading was one of the better
solutions to the problem.

Before the plugin has been initialized, any plugins listed within
`--deny-plugins` will be unregistered to prevent them from initializing
and potentially interferring with other plugins that have conflicts.

Signed-off-by: Colton Wolkins (Indicio work address) <colton@indicio.tech>


swcurran:
Interesting idea!  Intrigued with what others think of this -- @ianco @shaangill025 @TimoGlastra @andrewwhitehead

ianco:
Interesting ... looks like this will work for built-in protocols?  Does this remove the endpoints from the swagger page as well?

TimoGlastra:
What if the deny list is integrated with the plugin registry itself so there's no need to unregister, but rather the plugins are never registered? 

@ianco As this is done before intialization I don't think it will ever register the swagger endpoints

frostyfrog:
That's an interesting idea, having a filter in the load_plugin call that prevents the loading in the first place. The plugin registry doesn't have access to the settings during initialization if I remember correctly. Where is it initialized? Just woke up and that's one of the issues that I remember running into

frostyfrog:
> Interesting ... looks like this will work for built-in protocols?  Does this remove the endpoints from the swagger page as well?
> 

Working with built-in protocols was the intent :D

ianco:
OK having given this some thought, I have some likes and dislikes.

Likes = I like introducing the ability to load a sub-set of protocols when we run aca-py.  Aca-py has a lot of protocols, and implementers will likely never use *all* of the protocols aca-py has to offer, so the ability to boot up an agent that supports a limited protocol set (i.e. the specific protocols required for the implementation) is good.

Dislikes = All protocols automatically load, except those that have specific start-up parameters required to enable (such as multitenancy, mediation, I don't recall offhand if there are others) and now we have the option to disable built-ins (which is good) but the overall process seems a bit inconsistent to me.  (With the revocation notification protocols, for example, we now have 2 protocols which both load by default, but we know they don't both work together, so we're by default loading at least one protocol that we know won't work.)

So a possible solution is - in addition to the `--deny-plugins` that this PR introduces, *also* introduce:
- a "load-by-default" flag on the protocol ("true" by default so we don't have to change every existing protocol) - we could set this to "false" on the new "revocation notification v2" protocol
-  an `--allow-plugins` parameter to override "load-by-default=false" and force plugins that don't load by default to load

In this case if you don't specify anything, the "v1 revoc notif" would load.  If you want "v2 revoc notif" then you have to both "deny" the v1 protocol and "allow" the v2 protocol.

More complicated, but also more explicit about what protocols should be running (and avoids the case of loading a protocol that we know won't work).

Thoughts?

frostyfrog:
I like the idea of having the "load by default" flag for packages. If such a flag were introduced, I think the loading checks should be:
1. Check the "--blocked-plugins" argument, return if true
1. Check the "--enabled-plugins" argument, return if true
1. Check the "default" flag, return if false
1. Load the plugin

TimoGlastra:
I like the idea of disabling plugins in general, but ian's comment made me think why we actually can't run both protocols at the same time?

We don't have these restrictions for other protocols with multiple majors versions. Isn't this just an issue in the revocation notification protocol implementation? And shouldn't they just be integrated better? From an agent perspective it would be useful to support both versions for interoperability purposes.

dbluhm:
> I like the idea of disabling plugins in general, but ian's comment made me think why we actually can't run both protocols at the same time?
> 
> We don't have these restrictions for other protocols with multiple majors versions. Isn't this just an issue in the revocation notification protocol implementation? And shouldn't they just be integrated better? From an agent perspective it would be useful to support both versions for interoperability purposes.

You make some good points. I think the natural question that arises is how do we determine which to send (I don't think it makes sense to send one of each version). Revocation Notification differs from other protocols in that it's not directly triggered by another (relevant) versioned action, whether that be an Admin API request to a version 2 of a protocol or receiving a message with a particular version. We have the ability to perform feature discovery with a connection but we're not really using that discovery to adjust behavior in ACA-Py yet. I agree that that is something we should have eventually but I'm not in favor of making the solution to that problem a dependency of getting revocation notification v2 support, which is what led us to the ability to have more control over what builtins are active in ACA-Py.

swcurran:
Ah -- I see.  Notification occurs a result of calling the revocation Admin API, but is not invoked directly by the controller.  I'm going to throw out a couple of other ideas on how to handle this.  Don't be mad :-)  These are the old standbys:

- Add a parameter to the "revocation" Admin API call to say what version of the notification protocol to use
- Add a (yet another) startup parameter to always use a given notification protocol version.

Are these easy enough to handle the revocation notification scenario so that we don't have to be slowed down?

I think a valid question is what would do when revocation v3 comes along.  Is the "manage plugins" approach going to be best practices, or would we need to be able to dynamically choose the protocol when revoking?

swcurran:
Not a solution for now, but I do like the idea of making connection feature discovery a thing to allow for protocol versions selection based on the capabilities of the other side, and to get visibility into the features being used in the wild -- e.g. aggregation of the data to better understand the protocols in use.

That said -- we have to be careful about making that a vector for fingerprinting....

swcurran:
Not a solution for now, but I do like the idea of making connection feature discovery a thing to allow for protocol versions selection based on the capabilities of the other side, and to get visibility into the features being used in the wild -- e.g. aggregation of the data to better understand the protocols in use.

That said -- we have to be careful about making that a vector for fingerprinting....

ianco:
> * Add a parameter to the "revocation" Admin API call to say what version of the notification protocol to use
> * Add a (yet another) startup parameter to always use a given notification protocol version.

I vote we add these options to the revocation v2 PR (https://github.com/hyperledger/aries-cloudagent-python/pull/1734)

This solves the immediate problem where revocation notification V1 and V2 don't work together:

- the new startup parameter is the default version of the notification to use
- the new parameter in the API endpoint can override the notification version

This is consistent with how we've done this kind of thing in other endpoints.

This is really a change to the revocation notification PR (not this PR) so I'll add this comment there.

If we agree to make this change to the other PR then I don't think there's any issues with THIS PR and we can approve/merge it.

ianco:
> * Add a parameter to the "revocation" Admin API call to say what version of the notification protocol to use
> * Add a (yet another) startup parameter to always use a given notification protocol version.

I vote we add these options to the revocation v2 PR (https://github.com/hyperledger/aries-cloudagent-python/pull/1734)

This solves the immediate problem where revocation notification V1 and V2 don't work together:

- the new startup parameter is the default version of the notification to use
- the new parameter in the API endpoint can override the notification version

This is consistent with how we've done this kind of thing in other endpoints.

This is really a change to the revocation notification PR (not this PR) so I'll add this comment there.

If we agree to make this change to the other PR then I don't think there's any issues with THIS PR and we can approve/merge it.

ianco:
Discussed on today's aca-pug call and the consensus is to accept this PR, but not use this mechanism to control the revocation notification version (see comments on the revocation notification PR)
@frostyfrog if you update the branch and fix the tests we can accept and merge this PR

frostyfrog:
I believe the update should be enough. We'll see in a few though.

Issue #1745, opened 2022-04-26
Fetch from --genesis-url likely to fail in composed container setup


tdiesler:
The VON-Network webserver [sleeps 10sec](https://github.com/bcgov/von-network/blob/main/docker-compose.yml#L25) before startup in docker-compose. Presumably, it waits for the nodes to settle on some stable genesis state before it can serve the genesis Txs.

In Aca-Py we are a little less patient and only allow [5 retries with a 1sec](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/config/ledger.py#L34) timeout each - this is the [default of the http fetch](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/utils/http.py#L74).

As a result, if you start VON-Network and Aca-Py in a composed way (i.e. docker compose or k8s) Aca-Py will likely fail to fetch from `--genesis-url`

PR #1746, opened 2022-04-26
[#1745] Fetch from --genesis-url likely to fail in composed container…

tdiesler:
… setup

Signed-off-by: Thomas Diesler <tdiesler@redhat.com>

### DIFF
--- a/aries_cloudagent/config/ledger.py
+++ b/aries_cloudagent/config/ledger.py
@@ -31,7 +31,9 @@ async def fetch_genesis_transactions(genesis_url: str) -> str:
     headers["Content-Type"] = "application/json"
     LOGGER.info("Fetching genesis transactions from: %s", genesis_url)
     try:
-        return await fetch(genesis_url, headers=headers)
+        # Fetch from --genesis-url likely to fail in composed container setup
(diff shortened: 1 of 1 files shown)


TimoGlastra:
As this is more of a VON-network thing, maybe you can make the `acapy` service dependant on the `webserver` service (https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on) and add a healthcheck to the `webserver` service that pings whether the container is online (https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck)

tdiesler:
`depends_on` does not work, because liveness does not imply readiness. IMHO, it is an ACA-Py issue, because this project requires that the genesis information becomes available within 5sec of process startup - which is a bold assumption.

If genesis becomes available within 5sec, the proposed change has no effect. If genesis doesn't become available at all, the process would not fail before the 20sec timeout, which might not be ideal. An improved solution might be an additional config parameter that tells ACA-Py how long to wait for dependent components (i.e. Ledger + Tails Server)

TimoGlastra (approved):

ianco (approved):

PR #1751, opened 2022-04-28
fix: Resolve Revocation Notification environment variable name collision

frostyfrog:
`--monitor-revocation-notification` and `--notify-revocation` have a
name conflict with their corresponding environment variables. Now they
don't have a conflict.

Signed-off-by: Colton Wolkins (Indicio work address) <colton@indicio.tech>

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -680,7 +680,7 @@ class RevocationGroup(ArgumentGroup):
         parser.add_argument(
             "--monitor-revocation-notification",
             action="store_true",
-            env_var="ACAPY_NOTIFY_REVOCATION",
+            env_var="ACAPY_MONITOR_REVOCATION_NOTIF",
             help=(
                 "Specifies that aca-py will emit webhooks on notification of "
                 "revocation received."


aries_cloudagent/config/argparse.py
```diff
@@ -680,7 +680,7 @@ def add_arguments(self, parser: ArgumentParser):
         parser.add_argument(
             "--monitor-revocation-notification",
             action="store_true",
-            env_var="ACAPY_NOTIFY_REVOCATION",
+            env_var="ACAPY_MONITOR_REVOCATION_NOTIF",
```
TimoGlastra, line 683: Nit, but I like it when the env variable follows the argument (so I could guess what the env name is from the arg name)

```suggestion
            env_var="ACAPY_MONITOR_REVOCATION_NOTIFICATION",
```

TimoGlastra (dismissed):

aries_cloudagent/config/argparse.py
dbluhm, line 683: The abbreviation feels slightly out of place but it's also one of the longest monitoring subjects so I'm not feeling particularly nit-picky on that point

dbluhm (dismissed):

aries_cloudagent/config/argparse.py
frostyfrog: That's the main reason why I shortened it, because it's such a long environment variable

TimoGlastra: https://developer.apple.com/documentation/coremedia/1489454-cmmetadataformatdescriptioncreat I mean, could be worse .... 🙃

ianco (approved):

PR #1769, opened 2022-05-11
feat: event and webhook on keylist update stored

dbluhm:
This PR adds emission of an event and webhook on keylist update stored. This event simply completes the feedback loop for mediation related actions and enables the controller to keep track of mediation routing state.


dbluhm:
A review for this one would be appreciated.

@swcurran

PR #1774, opened 2022-05-17
Replace async workaround within document loader

frostyfrog:
After talking with Timo, the document loader uses a separate thread due to the fact that PyLD is synchronous code and calls 
into the document loader (which requires asynchronous code). When swapping out the cache with (for example) a Redis based cache, the exception `got Future attached to a different loop` may arise due to the current separate thread/event loop based implementation.

Switching to `nest_asyncio` over using a separate thread/event loop resolves the issues that were observed.


TimoGlastra (approved):
LGTM! Can't speak however on nest_asyncio, haven't used it before

PR #1777, opened 2022-05-18
fix: add a close statement to ensure session is closed on error

reflectivedevelopment:
Askar, on error, would end up not returning connections to the connection pool because aca-py left the connections open.

### DIFF
--- a/aries_cloudagent/askar/profile.py
+++ b/aries_cloudagent/askar/profile.py
@@ -213,6 +213,7 @@ class AskarProfileSession(ProfileSession):
                 await self._handle.commit()
             except AskarError as err:
                 raise ProfileError("Error committing transaction") from err
+        await self._handle.close()
         self._handle = None
         self._check_duration()
 
--- a/aries_cloudagent/askar/profile.py
+++ b/aries_cloudagent/askar/profile.py
@@ -213,7 +213,8 @@ class AskarProfileSession(ProfileSession):
                 await self._handle.commit()
             except AskarError as err:
                 raise ProfileError("Error committing transaction") from err
-        await self._handle.close()
+        if self._handle:
+            await self._handle.close()
         self._handle = None
         self._check_duration()
 


ianco:
One of the tests is failing:

```
aries_cloudagent/indy/credx/tests/test_cred_issuance.py:222: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
aries_cloudagent/indy/credx/issuer.py:177: in create_and_store_credential_definition
    await txn.commit()
aries_cloudagent/core/profile.py:204: in __aexit__
    await self._teardown()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'NoneType' object has no attribute 'is_transaction'",) raised in repr()] AskarProfileSession object at 0x7fe728b97da0>
commit = None

    async def _teardown(self, commit: bool = None):
        """Dispose of the session or transaction connection."""
        if commit:
            try:
                await self._handle.commit()
            except AskarError as err:
                raise ProfileError("Error committing transaction") from err
>       await self._handle.close()
E       AttributeError: 'NoneType' object has no attribute 'close'

aries_cloudagent/askar/profile.py:216: AttributeError
```

andrewwhitehead:
Are you able to test the changes in this PR? https://github.com/hyperledger/aries-cloudagent-python/pull/1692

The store should still be closed in the case of an exception when it is garbage collected, however there might be an existing reference preventing that from happening. The garbage collector should still handle circular references so I'm not sure how it would accumulate many of these, though.

Does the error still happen with Sqlite as a backend? There was a potentially related fix in sqlx, but it doesn't look like it's being backported to a 0.5.x release which is unfortunate. https://github.com/launchbadge/sqlx/pull/1799

reflectivedevelopment:
@andrewwhitehead I tested against https://github.com/hyperledger/aries-cloudagent-python/pull/1692, and that doesn't solve the problem. It appears the issue may be related to the Python garbage collector not closing the session quickly enough, leaving connection pool objects borrowed from the pool. To keep the pool operating efficiently, we should probably close the connection as quickly as possible.

reflectivedevelopment:
@ianco fixed the tests.

ianco (approved):

ianco:
@reflectivedevelopment if you sync up wth the main branch then we can merge

PR #1792, opened 2022-06-01
Fix: Inbound Transport is_external attribute

### LINKED ISSUE #1790: Issues with aries-acapy-plugin-toolbox when updating to 0.7.4
While updating https://github.com/hyperledger/aries-acapy-plugin-toolbox from 0.7.4 RC0 to 0.7.4 RC2, multiple errors have been encountered in the integration tests. I am still digging in to what these issues are, the only one that I remember off of the top of my head is related to #1659. A new attribute was added to the base inbound transport (and the default was added then removed) and it appears to be a required attribute.

### LINKED ISSUE #1791: Inbound Transport now has an undocumented, required attribute is_external
As mentioned in #1790, PR #1659 has introduced a new required attribute `is_external`. While going through the history, I noticed that in the PR, a default was set and then removed later on in this commit: https://github.com/hyperledger/aries-cloudagent-python/commit/577f34a48daea5b3c87705ab84f4f7e722afeedf#r74716748

This change caused the `HttpWsTransport` in https://github.com/hyperledger/aries-acapy-plugin-toolbox to break on startup (indicating a breaking change) due to the use of `is_external` in other parts of the code.

shaangill025:
Signed-off-by: shaangill025 <gill.shaanjots@gmail.com>
- resolve #1790
- resolve #1791

### DIFF
--- a/aries_cloudagent/transport/inbound/http.py
+++ b/aries_cloudagent/transport/inbound/http.py
@@ -15,8 +15,6 @@ LOGGER = logging.getLogger(__name__)
 class HttpTransport(BaseInboundTransport):
     """Http Transport class."""
 
-    is_external = False
-
     def __init__(self, host: str, port: int, create_session, **kwargs) -> None:
         """
         Initialize an inbound HTTP transport instance.
@@ -27,7 +25,7 @@ class HttpTransport(BaseInboundTransport):
             create_session: Method to create a new inbound session
 
         """
-        super().__init__("http", create_session, **kwargs)
+        super().__init__("http", create_session, is_external=False, **kwargs)
         self.host = host
         self.port = port
         self.site: web.BaseSite = None
--- a/aries_cloudagent/transport/inbound/http.py
+++ b/aries_cloudagent/transport/inbound/http.py
@@ -25,7 +25,7 @@ class HttpTransport(BaseInboundTransport):
             create_session: Method to create a new inbound session
 
         """
-        super().__init__("http", create_session, is_external=False, **kwargs)
+        super().__init__("http", create_session, **kwargs)
         self.host = host
         self.port = port
         self.site: web.BaseSite = None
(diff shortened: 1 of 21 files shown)


aries_cloudagent/transport/inbound/http.py
```diff
@@ -15,8 +15,6 @@
 class HttpTransport(BaseInboundTransport):
     """Http Transport class."""
 
-    is_external = False
-
     def __init__(self, host: str, port: int, create_session, **kwargs) -> None:
```
frostyfrog, line 18: Should this follow the same flow as the `WsTransport` class and include `is_external` on the constructor?

aries_cloudagent/transport/inbound/http.py
```diff
@@ -27,7 +25,7 @@ def __init__(self, host: str, port: int, create_session, **kwargs) -> None:
             create_session: Method to create a new inbound session
 
         """
-        super().__init__("http", create_session, **kwargs)
+        super().__init__("http", create_session, is_external=False, **kwargs)
```
frostyfrog, line 28: If added to the constructor, this should use that value

aries_cloudagent/transport/inbound/ws.py
```diff
@@ -16,9 +16,9 @@
 class WsTransport(BaseInboundTransport):
     """Websockets Transport class."""
 
-    is_external = False
-
-    def __init__(self, host: str, port: int, create_session, **kwargs) -> None:
+    def __init__(
+        self, host: str, port: int, create_session, is_external=False, **kwargs
+    ) -> None:
```
frostyfrog, line 21: `is_external` is being defined here, but it's never used. It's not being passed to the parent class

frostyfrog (changes requested):
Aside from the 2 things I've noted, this change looks good to me. It moves the `is_external` property to a "private variable" and makes sure it's only written upon class initialization, where the variable is now being initialized.

shaangill025:
Closing this PR due to merge issues, PR #1802 has been created with changes made here.

PR #1797, opened 2022-06-08
feat: Add filter param to connection list for invitations

### LINKED ISSUE #1789: Unable to filter for invitation from OOB invitations in multi-tenant environment
When creating an OOB invitation (whether it's with a public DID or not), there is not enough information returned to retrieve the specific invitation connection via the `/connections` API call. The Invitation's ID is returned in both calls and it could be used as a filter.

I plan on adding `invitation_msg_id` as a filter to `GET /connections`. Hoping to get this in for 0.7.4, @ianco

frostyfrog:
As mentioned in #1789, there is no way to currently filter for connections/invited created via the OOB handler. This commit adds a filter query parameter to filter with the invitation message ID.
- resolve #1789

### DIFF
--- a/aries_cloudagent/protocols/connections/v1_0/routes.py
+++ b/aries_cloudagent/protocols/connections/v1_0/routes.py
@@ -205,6 +205,11 @@ class ConnectionsListQueryStringSchema(OpenAPISchema):
         ),
         example=ConnRecord.Protocol.RFC_0160.aries_protocol,
     )
+    invitation_msg_id = fields.UUID(
+        description="Identifier of the associated Invintation Mesage",
+        required=False,
+        example=UUIDFour.EXAMPLE,
+    )
 
 
 class CreateInvitationQueryStringSchema(OpenAPISchema):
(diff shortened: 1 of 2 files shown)


aries_cloudagent/protocols/connections/v1_0/routes.py
```diff
@@ -205,6 +205,11 @@ class ConnectionsListQueryStringSchema(OpenAPISchema):
         ),
         example=ConnRecord.Protocol.RFC_0160.aries_protocol,
     )
+    invitation_msg_id = fields.UUID(
+        description="Identifier of the associated Invintation Mesage",
```
shaangill025, line 209: Typo: Invitation Message

shaangill025:
@frostyfrog I tested this and it is working. Based upon the description you provided in issue #1789, I think `invitation_msg_id` filter is the only one you had intended to add. If so, you can take this out of draft and I can approve it.

frostyfrog:
Yep! You're absolutely right. I just had it in draft because I wanted to make sure the black linting didn't error out on me before marking ready.

shaangill025 (approved):

Issue #1807, opened 2022-06-14
Under Investigation: Unable to create creddef due to error with Tails files


TheTechmage:
I'm currently investigating what is causing this, but I'm seeing the following error occur randomly when creating a credential definition. The code path suggests that we are creating the tails registry, however, we fail to upload the tails file to the tails folder because we aren't able to get the tails file off the local filesystem. The problem may possibly be in https://github.com/hyperledger/indy-sdk, but I have yet to narrow it down that far yet. All I know for certain is that this function is returning None for some reason. ACA-Py is setup with multi-tenant wallets in author mode when the error is being observed. The endorser should be approving all transactions automatically.
https://github.com/hyperledger/aries-cloudagent-python/blob/e7fd7e57faea226b4931840fc2c0048722160a6e/aries_cloudagent/indy/util.py#L42-L53

Tracestack that I'm running into:
```
Traceback (most recent call last):
File "/home/indy/aries_cloudagent/core/event_bus.py", line 120, in notify
    await processor()
File "/home/indy/aries_cloudagent/revocation/routes.py", line 1260, in on_revocation_tails_file_event
    max_attempts=5, # heuristic: respect HTTP timeout
File "/home/indy/aries_cloudagent/tails/indy_tails_server.py", line 51, in upload_tails_file
    max_attempts=max_attempts,
File "/home/indy/aries_cloudagent/utils/http.py", line 161, in put_file
    with open(file_path, "rb") as f:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

andrewwhitehead:
The `tails_path` method should likely raise an exception when it can't determine the file path, rather than just returning None. I suspect it's finding more than one tails file in the directory although it's possible the file is missing too. I can look at this shortly along with another registry creation issue.

swcurran:
Is this still considered an issue and more importantly, a must have for 0.7.4. Please provide an update when you can.

frostyfrog:
I do believe that it is still an issue but I do not have access to an environment to reproduce & debug. Those who do are currently busy with other tasks

jhoar:
Hi,

This seems to be related, the Faber agent in the demo also fails with revocation enabled (with 1.0.0-rc1).

Schema ID: Pb4gEUKLwig3iv3PzAmNbV:2:degree schema:15.16.38
Faber      | Revocation registry: (undetermined) state: init
[...]

PR #1808, opened 2022-06-15
Fix put_file when the server returns a redirect

andrewwhitehead:
I believe this fix only applies to tails file uploads at the moment. Previously, aiohttp would encounter an error because it closed the tails file after the first attempt.


swcurran:
I assume this addresses #1807 ? @frostyfrog , can you please take a look?

Looks like some tests are failing.

andrewwhitehead:
@swcurran  No, it doesn't fix that issue.

shaangill025 (approved):

PR #1710, opened 2022-04-05
feat: support connectionless exchange

TimoGlastra:
Adds support for connectionnless exchange using ICv1, ICv2, PPv1 and PPv2 protocols for both roles (so issuer, holder and verifier).

Can be initiated by creating or receiving an oob invitation without handshake protocols. If a connection already exists for the did in the oob invitation service, the connection will be used instead (as specified in OOB RFC).

Fixes #1636

Some notes on the implementation:
- Added an `OobRecord` to keep track of the oob state. This is used for both connectionfull and connectionless exchanges and will be removed once the interaction is done
- The oob record will also emit events at the moment about reuse of connections etc... To not make breaking changes I kept the old OOB reuse webhook events in place. May be good to remove these in a future PR
- Connectionless is not integrated with mediation yet (when creating an invitation). If this is desired, maybe someone else can take a look at this
- I disabled multi use when attachments were present in the invitation as this would never have worked (there's no support for multi-use requests, offers AFAIK?)
- Adds a cleanup of stale connections if a connection is being reused by sending a direct response to an attachment message in the oob invitation (I think connections weren't being cleaned up in this case)

~Keeping it in draft until I know for sure all CI tests have passed (can't run all tests on my M1 mac)~

PR #1811, opened 2022-06-15
Fix: present-proof v1 send-proposal flow

### LINKED ISSUE #1809: Prover mixes up thread_id in present-proof/send-proposal flow
I have another one, but this time it shouldn't be about duplicated api calls ;)

version: from rc2
wallet-type: indy or askar
role: prover

- Prover sends proof proposal via /present-proof/send-proposal, auto_present=true
- Verifier receives event:
```json
{
    "@type": "https://didcomm.org/present-proof/1.0/propose-presentation",
    "@id": "3b7403c8-bf0c-4718-a7c8-f95fa27e9ed9",
    "presentation_proposal": {
        "@type": "https://didcomm.org/present-proof/1.0/presentation-preview",
        "attributes": [
            {
                "name": "bic",
                "cred_def_id": "F6dB7dMVHUQSC64qemnBi7:3:CL:571:phil-test-83-multi",
                "value": "qwertz-01"
            },
            {
                "name": "iban",
                "cred_def_id": "F6dB7dMVHUQSC64qemnBi7:3:CL:571:phil-test-83-multi",
                "value": "qwertz-01"
            }
        ],
        "predicates": []
    }
}
```
- Verifier request presentation
- Prover sends presentation
- Verifier receives presentation
```json
 {
    "@type": "https://didcomm.org/present-proof/1.0/presentation",
    "@id": "1a632114-4324-4b1f-83ad-fac9287c954c",
    "presentations~attach": [
        {
            "@id": "libindy-presentation-0",
            "mime-type": "application/json",
            "data": "truncated"
        }
    ]
}
```
- Verifier fails with:

```json
aries_cloudagent.storage.error.StorageNotFoundError: V10PresentationExchange record not found for {'thread_id': '1a632114-4324-4b1f-83ad-fac9287c954c'}, {'role': 'verifier', 'connection_id': 'ebf94709-bb1b-488b-93b3-b540fe0bcb10'}
```

It looks like the prover mixes up the thread id, but only when using send-proposal, if the verifier starts the exchange with present-proof/send-request it works.

shaangill025:
- resolve #1809

### DIFF
--- a/aries_cloudagent/protocols/present_proof/v1_0/handlers/presentation_request_handler.py
+++ b/aries_cloudagent/protocols/present_proof/v1_0/handlers/presentation_request_handler.py
@@ -80,6 +80,10 @@ class PresentationRequestHandler(BaseHandler):
                         "connection_id": connection_id,
                     },
                 )  # holder initiated via proposal
+                presentation_exchange_record.presentation_request = indy_proof_request
+                presentation_exchange_record.presentation_request_dict = (
+                    context.message.serialize()
+                )
         except StorageNotFoundError:  # verifier sent this request free of any proposal
             presentation_exchange_record = V10PresentationExchange(
                 connection_id=connection_id,
@@ -94,7 +98,6 @@ class PresentationRequestHandler(BaseHandler):
                 trace=(context.message._trace is not None),
             )
 
-        presentation_exchange_record.presentation_request = indy_proof_request
         presentation_exchange_record = await presentation_manager.receive_request(
             presentation_exchange_record
         )  # mgr only saves record: on exception, saving state null is hopeless


swcurran:
@etschelp -- can you test this with the problem you reported in #1809?  Would like to get this merged and included in the release.  And would REALLY like to get the release out :-).

Thanks, @shaangill025 !

etschelp:
Ok, works again. What worries me though is that #1710 broke this because these lines were removed and now they have been added again, which raises two concerns:

1. Why wasn't there a failing test? - I think it would makes sense to add a couple of lines to: test_presentation_request_handler.py to also check for the thread_id
2. Why were those lines removed, and might adding them back cause issues with the OOB changes in turn?

I will also comment this on said PR so that Timo can chime in.

TimoGlastra:
It's unclear to me how this will fix the issue with the thread id?

Also @etschelp, the code you commented on in my PR is not the same code that is being changed here. 

I changed the functionality because we don't always a connection id anymore (connectionless, or connection but not yet in record because of create-offer/create-request). Were you able to verify the thread id is incorrect? If that's the case I think we need to fix this in another way

etschelp:
@TimoGlastra sorry, you are right, I'm totally confused now. All I know for sure is that if the holder starts with /send-proposal the thread id changes during the exchange leading to a storage error on the verifier side. This was introduced either with or around 1710. It works again with this PR.

shaangill025:
> It's unclear to me how this will fix the issue with the thread id?

Earlier when initiated from proposal, `presentation_request_dict` would not be assigned [None] for the record and then upon calling `assign_thread_from` [holder] it would use presentation message id as thread_id which was incorrect.

This is unrelated to PR#1710, I think this bug has been in there for quite some time.

TimoGlastra:
Found the issue. I changed the `assign_thread_from` to use the presentation request message dict thread instead of the presentation record thread id. This is done because with OOB we now also leverage the pthid which we need to include in the next message.

As the dict wasn't being stored this wouldn't assign anything and it would generate a new thread_id. So it makes sense this fixes the issue

PR #1813, opened 2022-06-16
Refactoring of revocation registry creation

andrewwhitehead:
This updates the revocation registry initialization process.

Previously, an event was fired using `profile.notify`, which led to an `IssuerRevRegRecord` being created and the associated revocation registry being generated. Now, the `IssuerRevRegRecord` is created first, and then the event is fired, which leads to the registry generation. This allows the detection of a registry in progress and helps to avoid multiple revocation registries being created in parallel for the same credential definition.

Also, the tails file is now uploaded after the registry definition is posted and before the first entry is posted, instead of after the first entry is posted. This ensures that the registry doesn't enter the ACTIVE state until after the tails file is publicly available.

### DIFF
--- a/aries_cloudagent/indy/credx/issuer.py
+++ b/aries_cloudagent/indy/credx/issuer.py
@@ -28,8 +28,6 @@ from ..issuer import (
     DEFAULT_CRED_DEF_TAG,
     DEFAULT_SIGNATURE_TYPE,
 )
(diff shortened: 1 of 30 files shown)


ianco:
The integration test passes when I run locally :-(  I'm going to try just re-running the tests

ianco (approved):
LGTM, all tests pass locally

Issue #1815, opened 2022-06-16
Faber demo getting error: "Error: agent faber.agent has no method handle_out_of_band to handle webhook on topic out_of_band"


swcurran:
Steps to reproduce:

* Running ACA-Py based on the current main. 
* Started a von-network instance and a indy-tails-server.
* `cd demo`
* `./run_demo faber`
* The normal startup occurred, but after printing the QR code, the following appears on the terminal:

```
Waiting for connection...
Error: agent faber.agent has no method handle_out_of_band to handle webhook on topic out_of_band
```

swcurran:
Fixed by #1816

PR #1816, opened 2022-06-16
Fix missing webhook handler

ianco:


swcurran:
Awesome!  This fixes #1815 .

@ianco -- the handler is printing on the terminal "handle_out_of_band()" after the QR code prints.  Is that normal for handlers? Do they also display that by default?  It looked odd to me.

Other than that -- looked good.

swcurran (approved):
LGTM and fixes the issue.

PR #1835, opened 2022-06-22
Allow setting a baseurl for the swagger urls

tbruecke:
When deploying different agents on same gateway url in service mesh there occures some problem cause static swagger path und swagger.json has no unique uri. This PR enables a baseurl to be set for the swagger urls mentioned via the environment variable ACA_PY_BASE_URL:
Until now, the static swagger path and swagger.json path is always the default value set in setup_aiohttp_apispec (static/swagger for 'static_path' and /api/docs/swagger.json for 'url'). By setting the ACA_PY_BASE_URL, the path can now be adjusted so that the urls can be set to ${ACA_PY_BASE_URL }/static/swagger or ${ACA_PY_BASE_URL }/api/docs/swagger.json. The urls can then be easily resolved in the virtual service via unique urls.

### DIFF
--- a/aries_cloudagent/admin/server.py
+++ b/aries_cloudagent/admin/server.py
@@ -269,6 +269,7 @@ class AdminServer(BaseAdminServer):
         # This should be enforced during parameter parsing but to be sure,
         # we check here.
         assert self.admin_insecure_mode ^ bool(self.admin_api_key)
+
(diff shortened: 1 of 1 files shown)


tbruecke:
Done in https://github.com/hyperledger/aries-cloudagent-python/pull/1837

PR #1836, opened 2022-06-22
feat: make base wallet route access configurable

dbluhm:
When adding admin routes through plugins, it makes sense for the plugin to be able to specify that those routes should be accessible to the base wallet when appropriate.


burdettadam (approved):

swcurran:
@dbluhm -- can you please look into the Circle-CI failure?  Thanks.

andrewwhitehead (approved):

PR #1844, opened 2022-06-25
Fix a typo in DevReadMe.md

feknall:
There is a typo in the example command to run the project using the docker container.

```
PORTS="5000:5000 8000:8000 1000:1000" ./scripts/run_docker start --inbound-transport http 0.0.0.0 10000 --outbound-transport http --debug --log-level DEBUG
```

It opens the port 1,000 for access instead of 10,000 which is specified as "inbound-transport". As a result, the agent wouldn't be accessible.

### DIFF
--- a/DevReadMe.md
+++ b/DevReadMe.md
@@ -149,7 +149,7 @@ To enable the [ptvsd](https://github.com/Microsoft/ptvsd) Python debugger for Vi
 Any ports you will be using from the docker container should be published using the `PORTS` environment variable. For example:
 
 ```bash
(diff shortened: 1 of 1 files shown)


swcurran:
Thanks for this -- much appreciated!!  We need all commits signed using DCO in order to merge.  Could you update your commit to sign it: DCO - Developer Certificate of Origin - https://github.com/apps/dco -- see the instructions on the "Details" link on the DCO line above  ^^^^^,  Thanks!

feknall:
You're welcome. It's done. Please check it again.

swcurran (approved):

PR #1851, opened 2022-07-01
Enable manually triggering keylist updates during connection

dbluhm:
This PR is a significant refactor of routing for both Mediation and Multi-tenancy, impacting Connections, Out-of-band, and DID Exchange managers, with the goal of eliminating a race condition possible when using mediation. Additionally, this PR also simplifies routing setup significantly for both mediation and multi-tenancy.

### Mediation Race Condition during Connection Formation

During connection formation, the current implementation of mediation support triggers a keylist update message to be sent just prior to sending the connection request or response (or a DID Exchange request or response) **without** waiting for confirmation from the mediator that the keylist update was received and processed successfully. This can result in a scenario where the connection request or response is received and responded to before the mediator has processed the keylist update. Typically, the keylist update process is quick and, especially when using the default HTTP transport mechanisms, the chances of things getting out of order are very slim. However, when introducing outbound queuing or another transport, this race condition has proven to be frequent.

#### Mediation Client as Invitee
image

#### Mediation Client as Inviter
image

### Solution
To prevent this race condition, this PR adds the `/mediation/update-keylist/{conn-id}` Admin endpoint that is called after `receive-invitation` or `receive-request` and before `accept-invitation` or `accept-request`. This endpoint creates connection keys and sends those keys in a keylist update to the specified mediator. When combined with the changes in #1769, the controller can then wait for the keylist update response received webhook, match the thread_id to the id of the keylist update returned from `/mediation/update-keylist/{conn-id}`, and then proceed with the next step in the connection.

image

image

This should look familiar; this is more or less how the old `establish-inbound` endpoint worked. However, a key difference and the next major addition of this PR is the new `RouteManager` abstract base class. Two concretions are included in this PR: one for standard Coordinate Mediation v1 route management and one for Multi-tenancy + Mediation, which handles base wallet route creation and routing through a base wallet mediator and/or a sub wallet mediator.

Using the `RouteManager` to abstract away routing details, the Connection, OOB, and DIDExchange managers have each been significantly simplified (IMHO :slightly_smiling_face:).

Some additional notes:
- This is a backwards compatible change. If `/mediation/update-keylist/{conn-id}` is not used, forming connections through a mediator should behave exactly as before.
- This PR also corrects a slight misbehavior where both the base wallet's and sub wallet's mediator were informed of new connection keys being formed, even though the base wallet mediator was the only mediator in that chain that would have seen those keys.


swcurran:
Best PR ever submitted -- great job!  @andrewwhitehead @shaangill025 @TimoGlastra and others, please review.

I'll try to contribute by testing this PR using the existing AATH tests -- ACA-Py to ACA-Py.

@dbluhm -- we frequently see a problem with the first credential offer after a connection is established. Could this race condition be a cause for that problem?

Thanks!!

dbluhm:
@swcurran if the connection made it all the way to `active`/`completed` state, I would not expect this particular race condition to be the cause of a lost credential offer right after connection establishment :thinking:

The way this one plays out, it's usually exhibited by the connection never fully advancing to `active`/`completed`

swcurran:
All of the Aries Agent Test Harness tests for ACA-Py that are tagged "RFC0211" (Mediation) work with this PR - good stuff. Those only cover the RFC0160 Connections protocol -- not RFC0023 DID Exchange.

I'm a little concerned that there are no mediation tests that use RFC0023 DID Exchange in AATH.  @nodlesh -- is that something you could add?  I'm not sure who initially added the mediator tests.  I can add an issue to AATH so we can track it.

TimoGlastra:
> I'm a little concerned that there are no mediation tests that use RFC0023 DID Exchange in AATH. @nodlesh -- is that something you could add? I'm not sure who initially added the mediator tests. I can add an issue to AATH so we can track it.

I added the mediation tests. There are mediation tests with DIDExchange: https://github.com/hyperledger/aries-agent-test-harness/blob/main/aries-test-harness/features/0211-coordinate-mediation.feature#L7-L26. All mediation tests are ran twice, once for 0160 and once for 0023.

aries_cloudagent/connections/base_manager.py
```diff
@@ -57,6 +62,17 @@ def __init__(self, profile: Profile):
...
+        multitenant_mgr = profile.inject_or(BaseMultitenantManager)
+        wallet_id = profile.settings.get_str("wallet.id")
+        if multitenant_mgr and wallet_id:
+            self._route_manager: RouteManager = multitenant_mgr.get_route_manager(
+                profile, wallet_id
+            )
+        else:
+            self._route_manager: RouteManager = CoordinateMediationV1RouteManager(
+                profile
+            )
```
TimoGlastra, line 74: I've never been a fan of how I implemented the multitenant routing part as it made subwallets aware of the hierarchy (i.e. it was the tenants responsiblity to call the base wallet to inform it about routing). 

With the approach you've taken here I think we could fully remove that dependency. What if we bind the route manager to the injection context and inject it here. That would mean the specific manager don't need to know about multitenancy at all as they're just injecting the route manager. Under the hood that could have a different implementation based on whether multitenancy is enabled or not.

TimoGlastra (approved):
This is great @dbluh and will make it much easier to maintain and integrate new features with multitenancy / mediation in the future.

aries_cloudagent/wallet/routes.py
```diff
@@ -582,7 +582,9 @@ async def promote_wallet_public_did(
         multitenant_mgr = profile.inject_or(BaseMultitenantManager)
         # Add multitenant relay mapping so implicit invitations are still routed
         if multitenant_mgr and wallet_id:
-            await multitenant_mgr.add_key(wallet_id, info.verkey, skip_if_exists=True)
+            await multitenant_mgr.get_route_manager(
+                profile, wallet_id
+            ).route_public_did(info.verkey)
```
TimoGlastra, line 587: To add to my previous comment. If we bind the route manager to the injection context, we could also remove the multitenant integratino in other parts of the codebase. This would just be:

```py
route_manager = profile.inject(RouteManager)
# Doesn't really do anything in non-multitenancy cases, but means we don't have to explicitly integrate with MT here.
await route_manager.route_public_did(info.verkey)

```

dbluhm:
Got a report that this might not be working quite right yet in some multi-tenancy scenarios. I'll investigate more tomorrow! (Independence Day in the US today)

dbluhm:
As an aside, I intend to follow up this PR with a refactor on the locations of a lot of these components. I expect that to be tedious and verbose so I've not done any of that with this PR to make these additions clearer.

That refactor may include:
- Moving route manager base (and provider) to `aries_cloudagent/routing` or similar
- Moving mediation record and route record to `aries_cloudagent/routing` or similar
- Split default mediator storage and retrieval to a separate component from the coordinate medation v1 manager
- Move associated tests

These changes would make it so the route manager is not strictly dependent on the coordinate mediation v1 protocol or the vestiges of the old routing v1 protocol. It also moves records used by multiple components to a common location, which better reflects their common usage, IMO.

aries_cloudagent/multitenant/route_manager.py
```diff
@@ -0,0 +1,105 @@
...
+            except (StorageNotFoundError):
+                pass
+
+        await routing_mgr.create_route_record(
+            recipient_key=recipient_key, internal_wallet_id=wallet_id
+        )
+
+        # External mediation
+        keylist_updates = None
+        if mediation_record:
```
andrewwhitehead, line 72: It looks like this may be performing key updates with the base wallet mediator, which the old code explicitly avoids. Or does the mediation manager skip it when not necessary?

dbluhm: Old add_key method in multitenant manager:
```python
        mediation_record = await mediation_mgr.get_default_mediator()

        if skip_if_exists:
            try:
                async with self._profile.session() as session:
                    await RouteRecord.retrieve_by_recipient_key(session, recipient_key)

                # If no error is thrown, it means there is already a record
                return
            except (StorageNotFoundError):
                pass

        await routing_mgr.create_route_record(
            recipient_key=recipient_key, internal_wallet_id=wallet_id
        )

        # External mediation
        if mediation_record:
            keylist_updates = await mediation_mgr.add_key(recipient_key)

            responder = self._profile.inject(BaseResponder)
            await responder.send(
                keylist_updates, connection_id=mediation_record.connection_id
            )
```

So this old method looks very similar to what's currently implemented. However, the old code only dealt with sending keylist updates to the base wallet mediator when needed. The old code paths through the different connection forming protocols would trigger keylist updates to a sub wallet mediator independent of the multi-tenant `add_key` method. This actually resulted in both the sub wallet and base wallet mediator getting keylist updates when in reality only the base wallet mediator should have been receiving keylist updates.

This new method will always send keylist updates to a base wallet mediator, if present. It will send a keylist update to a sub wallet mediator iff no base wallet mediator is set. And, of course, this results in the correct behavior because all of the keylist update sending behavior has been removed from the connection forming protocols and delegated to the route managers.

Does that clear things up?

andrewwhitehead: I'm thinking of this logic in the connections manager: https://github.com/hyperledger/aries-cloudagent-python/blob/f0b27196614a3d196af82ee6927d6c9bfab58a2f/aries_cloudagent/protocols/connections/v1_0/manager.py#L241

dbluhm: Ah, okay, I must have missed that line. So the old code did send keylist updates to only the base wallet mediator, at least in that instance. The `keylist_updates` getting nulled out at the chunk you linked to ensures keylist updates aren't sent from the connection protocol methods because they would instead be sent by the multi-tenant manager `add_key` method.

I'm pretty confident the same logic is captured correctly in the multi-tenant route manager -- it's just all in one place now

andrewwhitehead (approved):

PR #1854, opened 2022-07-05
python-3.7 update

burdettadam:
# UNDER RAPID DEVELOPMENT!
This draft PR updates ACA-PY to use python-3.7 along with dependency updates. 
For questions during development please reach out to @burdettadam or @dbluhm.


aries_cloudagent/core/dispatcher.py
```diff
@@ -284,9 +253,6 @@ def __init__(
 
         """
         super().__init__(**kwargs)
-        # Weakly hold the context so it can be properly garbage collected.
-        # Binding this DispatcherResponder into the context creates a circular
-        # reference.
```
andrewwhitehead, line 289: Is some auto-formatter removing these comments and newlines?

swcurran: @burdettadam -- please investigate this.  We definitely don't want to lose comments from the code.

burdettadam: I absolutely agree, we don't want to lose any comments or clarity in code. I did notice some comments lost from some tooling I used. We are focused on getting all required interface updates done then updating tests. We will then go back over and add any lost comments and clarity. I also plan to squash my commits to make my git history clearer. 
Nice to see interest in the PR already. 😃.

swcurran:
We also have to think about how this will impact the daily AATH testing. Presumably the "dockerfile-main" file will have to change when we merge this PR.  We can do something locally or on a PR branch for testing this, but for the daily runs, we have to wait until this PR is merged.

dbluhm:
Yeah, good points; my reason for asking isn't quite so forward thinking -- just trying to get an image base to point circleci and the integration tests to :sweat_smile:

That being said, rather than merging to main, it does seem like perhaps a good idea to have a python 3.7 feature branch that lives in parallel as we sort out updates and fixing deprecation warnings and/or producing experimental builds.

swcurran:
I'm not an expert on that approach, but do think it sounds like a reasonable idea.  I would not want it to live long, but perhaps it should be around for a bit.  We do need to get the world used to using 3.7.

dbluhm:
For now, I've pointed circleci and integration tests to ghcr.io/indicio-tech/von-image:py37-1.16-0 but happy to replace this with a more official image when we're ready.

dbluhm:
Using a very rough tool I cobbled together, my most recent commits on this PR update usage of marshmallow fields to correct deprecations made in newer versions. The tool can be found here: https://github.com/dbluhm/acapy-marshmallow-upgrade. This significantly increases the number of lines touched by this PR. I can replay the changes on a different PR if that is preferred.

swcurran:
What is the status of this @burdettadam / @dbluhm ? 

We have a new-to-ACA-Py dev @Jsyro looking at the failing Python 3.7 and 3.10 tests to see if he can get them to run.  Should he take over this?  Any suggestions on next steps?

Thanks!

dbluhm:
This PR sought to update dependencies and to resolve deprecation warnings. We solved a number of issues but this PR has since grown stale (partially out of a shift in focus to get container images with newer python versions first, partially out of just getting busy with other tasks). I think this probably just needs to be closed but if we're interested in solving the many deprecation warnings from marshmallow, the pattern I used and described in my comments above I think are still relevant.

Issue #1313, opened 2021-07-13
Update the resolver for did:sov to use the latest DID Method Rules


swcurran:
Please update the DID Resolver within ACA-Py for "did:sov" to follow the rules as recently updated in the DID Method spec.  This means adjusting the returned DIDDoc to exactly what is documented in the DID Method spec for the cases of their being no endpoint ATTRIB, a "current convention" endpoint ATTRIB and support for the defined new convention endpoint ATTRIB.

The PR for the DID Method that highlights the changes is here:

https://github.com/sovrin-foundation/sovrin/pull/296/files

@dbluhm -- great if you could point out where in the ACA-Py code this needs to go.

Thanks!

dbluhm:
https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/resolver/default/indy.py#L64-L91

domwoe:
Please don't remove support for additional service types beside did-communication/DIDcomm. In the Business Partner Agent project we rely on service type "profile"

swcurran:
That one is new to me.  Could you give an example and where it comes from?

domwoe:
https://hackmd.io/4oZOgwFOQDSFUuu3ruN-_g

swcurran:
Dom -- the "profile" is not a convention that has been supported in Indy AFAIK -- e.g. are there any ATTRIBs on Sovrin MainNet using it?  On the other hand, the "endpoint" ATTRIB is commonly used on Sovrin and likely other Indy networks.

For that and many other use cases, we really need to have "did:indy", so that the controller of the DID can define what they want in their DIDDoc.  That is supported in what we have defined "did:indy" to be.

Let's work on that to achieve this capability.

domwoe:
I agree that we should move forward with did:indy, but it would be great to preserve this functionality in did:sov as well to not break our BPA deployments.

The did-sov driver of the uni resolver has always supported this feature and still is (see https://dev.uniresolver.io/1.0/identifiers/did%3Asov%3Aidu%3A4PDz9iiQqgVXhMA7nWqRzS). Although an ID should be added to the service.

Furthermore, we added the feature to write/read services of different types to ACA-Py more than a year ago.
https://github.com/hyperledger/aries-cloudagent-python/issues/589

If the did:sov resolver in ACA-Py gets implemented exactly like the did:sov driver of the ACA-Py our use case is still supported, but i'd suggest to generate an id field to the service object.

[...]

PR #1863, opened 2022-07-18
fix: resolve dids following new endpoint rules

dbluhm:
Naive solution to the issue raised by @TimoGlastra in #1313.


aries_cloudagent/resolver/default/indy.py
```diff
@@ -43,6 +46,60 @@ def supported_did_regex(self) -> Pattern:
...
+        self,
+        builder: DIDDocumentBuilder,
+        endpoint: str,
+        recipient_key: VerificationMethod,
+    ):
+        builder.service.add_didcomm(
+            ident=self.SERVICE_TYPE_DID_COMMUNICATION,
+            type_=self.SERVICE_TYPE_DID_COMMUNICATION,
+            service_endpoint=endpoint,
+            priority=1,
```
dbluhm, line 59: This one sets priority to 1, other sets priority to 0 -- should probably be consistent.

aries_cloudagent/resolver/default/indy.py
```diff
@@ -43,6 +46,60 @@ def supported_did_regex(self) -> Pattern:
...
+        self,
+        builder: DIDDocumentBuilder,
+        endpoint: Mapping[str, Any],
+        recipient_key: VerificationMethod,
+    ):
+        types = endpoint.get("types", [self.SERVICE_TYPE_DID_COMMUNICATION])
+        routing_keys = endpoint.get("routingKeys", [])
+        endpoint_url = endpoint.get("endpoint")
+        if not endpoint_url:
+            raise ValueError("endpoint url not found in endpoint attrib")
```
dbluhm, line 74: Need a test to hit this.

TimoGlastra (approved):

aries_cloudagent/resolver/default/indy.py
```diff
@@ -43,6 +46,60 @@ def supported_did_regex(self) -> Pattern:
...
+
+        if self.SERVICE_TYPE_DIDCOMM in types:
+            builder.service.add(
+                ident="#didcomm-1",
+                type_=self.SERVICE_TYPE_DIDCOMM,
+                service_endpoint=endpoint_url,
+                recipient_keys=[recipient_key.id],
+                routing_keys=routing_keys,
+                accept=["didcomm/v2"],
```
andrewwhitehead, line 83: Isn't it an issue that it resolves to accept only didcomm/v2, while we don't actually support that yet? Or does it not matter because this endpoint isn't examined anyway.

dbluhm: Yeah, I think in this case, we're just constructing the DID Document as the DID Method spec outlines and whether we use the all of the services outlined in the doc is another question. We will be essentially ignoring them for now since we're only looking for the `did-communication` type.

andrewwhitehead (approved):

PR #1880, opened 2022-07-28
Allow fully qualified class names for profile managers

chumbert:
Fix a small oversight that prevents specifying a qualified class name as
wallet type.


dbluhm (approved):
LGTM

PR #1881, opened 2022-08-01
fix: didx request cannot be accepted

rmnre:
Hi everyone,

this PR fixes a few problems which were apparently introduced by #1710:

- A didx request which is received in response to an explicit oob invitation cannot be accepted because the state of the corresponding connection record is not updated to `'request-received'`

- The admin api incorrectly indicates that the `/out-of-band/receive-invitation` endpoint returns a `ConnRecord` instead of the new `OobRecord` (which is also why the `OobRecord` is not included in the `swagger.json`)

@TimoGlastra: maybe you want to take a look to make sure I didn't break the connectionless exchange feature by saving the updated connection record?


dbluhm (approved):
I haven't explicitly evaluated whether this impacts connection-less exchange but not saving the conn rec in just that one case certainly seems like a bug. This fix looks reasonable to me.

swcurran:
An integration test is failing. Hopefully this is not an intermittent failure, but it needs to be investigated.

PR #1885, opened 2022-08-04
fix: schema class can set Meta.unknown

dbluhm:
This PR enables the Schema class of a model to determine the default `unknown` behavior. Without this, if you have a schema where you expect extra values to be present, every time `deserialize` or `serialize` is called, you must set `unknown=INCLUDE`. Now, in the `Meta` class of the Schema, `unknown` can be set to some value that will be respected as the default if not overrided by method parameter.

While I was at it, I also touched up some of the typing. I'm not fond of using `@overload` but I think it's the best way to address typing a method where a bool flag impacts the return type.

### DIFF
--- a/aries_cloudagent/messaging/models/base.py
+++ b/aries_cloudagent/messaging/models/base.py
@@ -5,7 +5,8 @@ import json
 
 from abc import ABC
 from collections import namedtuple
-from typing import Literal, Mapping, Optional, Type, TypeVar, Union, cast, overload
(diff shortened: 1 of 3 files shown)


dbluhm:
The circleci failure is a flake8 check complaining about missing docstrings on the `@overload` decorated methods. I don't think the overloads should be required to have a docstring since the primary method itself does. I'm also unable to get the check to fail locally :thinking: Does anyone with more knowledge of the circleci config than myself have a suggestion for a fix?

ianco (approved):

Issue #1871, opened 2022-07-22
Upgrade default version of Python from 3.6


swcurran:
We need to update from Python 3.6, and so need a plan to do so.  PR #1854 has been created to start the process.

We discussed this at the [ACA-Pug meeting on July 12](), notes copied below.

* 3.6 is end of life'd, so this is a high priority
* We have a 3.7 PR, but 3.7 is less than a year from end of life
* Other [versions lifespans](https://endoflife.date/python)
   * Why not use 3.10?
* Meeting Discussion:
  * On 3.7 branch, all tests are passing, but with lots of deprecation warnings
  * 3.6 to 3.7 is likely the biggest breaking change (vs. going to 3.8, 3.9, etc.)
     * 3.11 evidently has a significant performance improvement (e.g. 25%), so being ready is important
  * Suggested approach:
     * 3.7 branch for now
     * Add GHAs to run all tests on all Python versions 3.7 - 3.10.
       * Run one version on PRs (same as today)
       * Run other versions daily on a cron
     * Add tag-based publishing of artifacts, and publish to the Hyperledger GitHub Container Repository
     * @burdettadam and team to produce a HackMD doc about the plans to promote comment and community collaboration

swcurran:
Closing this as done. Opened #2072 to track errors on specific versions of python.

PR #1856, opened 2022-07-12
Githubaction publish docker image

PaulWen:
Hi, @swcurran mentioned in today's AcaPug call that you would be interested in a pipeline for building and publishing AcaPy Docker images automatically for every tagged commit.

The GitHub Action below should do this. It is triggered for every tagged commit and will build the docker image using the `docker/Dockerfile.run` file. The images are being published in the GitHub package registry of this repo.

Example: If a commit is tagged with `0.7.4-rc6` an image will automatically be built and published with the following two tags:
1. aries-cloudagent-python:latest
2. aries-cloudagent-python:0.7.4-rc6

You can also see the build image in my forked version of this repo: https://github.com/PaulWen/aries-cloudagent-python/pkgs/container/aries-cloudagent-python

Let me know if this is of any help or if I misunderstood the requirements. I am happy to iterate over it and appreciate any feedback!

PR #1888, opened 2022-08-06
feat: add dockerfiles and github actions for multiple python versions

dbluhm:
Putting this out there to hopefully spark some discussion.

This PR adds Dockerfiles for publishing images for ACA-Py directly from this repo (i.e. no dependencies on bcgovimages). I made some opinionated decisions that I hope will catch some people's attention (so you can tell me why I'm wrong :slightly_smiling_face:):
- Rather than building aries-askar, indy-shared-rs, and indy-vdr directly in the image, I opted to instead let it use the binaries bundled in the python packages. Good chance I'm wrong, but I'm about 70% sure these libraries might have been taking precedence over those included in von-image anyways.
- I am thinking of "indy-less" ACA-Py as the default. `ghcr.io/hyperledger/aries-cloudagent-python:py3.6-0.7.4` image would not have any Indy libraries installed. To get an image with Indy included, I'm thinking the image name+tag might look like: `ghcr.io/hyperledger/aries-cloudagent-python:py3.6-indy-1.16.0-0.7.4`.
- The ACA-Py+Indy image does NOT include Indy CLI (von-image did).
- The default ACA-Py image's user is now `aries` instead of `indy`. The ACA-Py+Indy image's user will still be `indy` (this inconsistency may be undesirable...).
- I introduced an `indy-python` image that can be reused as a base for things needing an image with just python3-indy and built Indy libraries. This is the base image used by the ACA-Py+Indy image. I would like to see this published as `ghcr.io/hyperledger/indy-python`.
- Rather than using pyenv to get a specific python version, I opted to just use the relevant python docker image for the requested version, i.e. `python:3.6.13-slim-buster`. Python packages are installed into the system site packages (rather than the user site packages or a virtual environment).
- ~For now, I've followed the example of the aries-cloudagent-container repo and am installing ACA-Py through the published PyPI package. It might make more sense to be able to publish directly from the repo contents?~ With the GHA additions, the images are now built directly from the repo.

I get the feeling that we can further prune a lot of the installed packages but (for now) I've left in most of those originally installed in von-image.

I've included a simple Makefile purely to demonstrate building the image and I do not anticipate that we'd leave this Makefile around. I'll plan on removing it before this PR moves on to potentially getting merged.

This is relevant to #1854, #1856, and #1871.

cc @WadeBarnes @andrewwhitehead @swcurran @burdettadam


dbluhm:
Okay, some updates. I've added Github Actions for various operations:

- Unit testing
    - Tests for ACA-Py without Indy
        - These tests run directly on the runner.
    - Tests for ACA-Py with Indy
        - These tests run inside of a container (necessary for Indy library to be available). The image for the container is built and pushed, if needed, to the ghcr.io for the current repository. The images are `acapy-test:py{python-version}-{indy-version}-{hash of dependent files}` and `indy-python-test:py{python-version}-{indy-version}-{hash of dependent files}`.
    - These are currently configured to run on each PR but it is likely more appropriate to schedule runs of most versions and just select a preferred python version and run (minimally) one with indy or one with indy and one without.
- Nightly Builds
    - If a nightly build at the current hash of the repo doesn't yet exist, these will build an ACA-Py without Indy and an ACA-Py with Indy images at midnight each day. I included these more as a proof of concept than anything. I'm not strongly opinionated about whether we should or should not do nightly builds but we certainly have that ability.
- Publish images
    - Publish an image of ACA-Py without Indy. This will produce an image with the name+tag `ghcr.io/{repo-owner}/aries-cloudagent-python:py{python-version}-{acapy-version}`. For example `ghcr.io/hyperledger/aries-cloudagent-python:py3.7-0.7.4`.
    - Publish an image of ACA-Py with Indy. This will produce an image with the name+tag `ghcr.io/{repo-owner}/aries-cloudagent-python:py{python-version}-indy-{indy-version}-{acapy-version}`. For example `ghcr.io/hyperledger/aries-cloudagent-python:py3.7-indy-1.16.0-0.7.4`.
    - These are currently only set to be triggered manually but it seems appropriate to have them run automatically on release.

dbluhm:
The `Tests (Indy)` workflows are currently failing (I believe) because of permissions of this repo to publish packages. If we want to pursue this approach, we'd probably need to reach out to Ry to help set things up.

WadeBarnes:
> The `Tests (Indy)` workflows are currently failing (I believe) because of permissions of this repo to publish packages. If we want to pursue this approach, we'd probably need to reach out to Ry to help set things up.

It's the permissions around PRs causing the issue.  PRs don't have access to GitHub secrets including `GITHUB_TOKEN`.  The images can only be published after the PR gets merged.

dbluhm:
> It's the permissions around PRs causing the issue. PRs don't have access to GitHub secrets including `GITHUB_TOKEN`. The images can only be published after the PR gets merged.

Ah, well that makes perfect sense :slightly_smiling_face:

dbluhm:
To see some completed workflows, check here for the runs from my testing on Indicio's fork: https://github.com/Indicio-tech/aries-cloudagent-python/pull/66/checks (I think those are publicly accessible??)

swcurran:
I don't have the background to assess this work.  Can someone help out and work with @dbluhm on this?

@PaulWen, @dbluhm does this related to the PR that Paul submitted earlier on Dockerfiles and the like -- #1856 

I tagged v1.0.0-rc0, and would love to see that published using the new mechanism.  I did the publish an image the old way.

PaulWen:
@swcurran as @dbluhm also added the pipelines for building and publishing the Docker images I would consider #1856 to be obsolete

reflectivedevelopment:
I think it would make sense to move the indy build into indy-sdk instead of having it in aries-cloudagent-python.

dbluhm:
I'm going to take a final pass at this this week, update documentation, make a few updates to the actions that I discovered would be helpful while testing similar setups on other repositories, and then push for this to get merged. I'll make some noise when I have those updates in and seek review from maintainers.

dbluhm:
I found that there were a number of things I could simplify after looking at this with fresh eyes. I'm working through testing those simplifications. Progress is a little slower than I'd hoped due to competing priorities but I've tapped in some of the rest of my team to help move this along.

dbluhm:
@swcurran @ianco @WadeBarnes Anything preventing us from merging in this PR? Thanks!

WadeBarnes:
@dbluhm, I'd like to look over the changes in detail first.  I'm almost at a point were I can do that.

.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: Tests (Indy)
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+
+    uses: ./.github/workflows/tests-indy.yml
+    with:
+      python-version: ${{ matrix.python-version }}
+      indy-version: "1.16.0"
```
WadeBarnes, line 29: Some of the nightly tests are failing as expected without the added support for the additional versions of Python.  Both sets of tests are failing with python 3.7 and 3.10 in a similar fashion.  The "standard" tests are failing for python 3.6 because it is not available for ubuntu:latest, only 18.04 and 20.04.

Refer to https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3517241353 for and example.

.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,18 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  tests:
+    name: Tests
+    uses: ./.github/workflows/tests.yml
+    with:
+      python-version: "3.6"
```
WadeBarnes, line 11: I expect this will fail for the same reason it fails in the nightly tests.  The shared `tests` workflow runs on ubuntu:latest, and python 3.6 is not available for it.

.github/workflows/publish-indy.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Log in to the GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
WadeBarnes, line 56: Using `github.actor` as the username causes issues in the Hyperledger org.  Use `github.repository_owner` instead.  For example; https://github.com/hyperledger/indy-shared-gha/blob/main/.github/workflows/buildimage.yaml#L51-L53

.github/workflows/publish-indy.yml
```diff
@@ -0,0 +1,99 @@
...
+          context: .
+          file: docker/Dockerfile.indy
+          tags: ${{ steps.dispatch-meta.outputs.tags || steps.meta.outputs.tags }}
+          target: main
+          build-args: |
+            python_version=${{ matrix.python-version }}
+            indy_version=${{ inputs.indy_version || env.INDY_VERSION }}
+            acapy_version=${{ inputs.version || github.event.release.tag_name }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
```
WadeBarnes, line 91: Setup of the image tags is not working as expected.  When run manually or triggered by a release a single tag is used for each of the images, resulting in the images being overwritten with only the last matrix job to run as the final image tagged.

Whether run manually or triggered by a release there should be an explicitly tagged image for each version of python listing in the matrix. 

References:
https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3517366030
https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3517511271
https://github.com/WadeBarnes/aries-cloudagent-python/pkgs/container/aries-cloudagent-python

.github/workflows/tests.yml
```diff
@@ -0,0 +1,37 @@
...
+  workflow_call:
+    inputs:
+      python-version:
+        required: true
+        type: string
+
+jobs:
+  tests:
+    name: Test Python ${{ inputs.python-version }}
+    runs-on: ubuntu-latest
```
WadeBarnes, line 13: This fails when used with python 3.6 because there is no python 3.6 package available for this version of ubuntu (>20.04).

ContainerImagesAndGithubActions.md
```diff
@@ -0,0 +1,122 @@
...
+variants and between the BC Gov ACA-Py images.
+
+- Standard Image
+    - Based on slim variant of Debian
+    - Does **NOT** include `libindy`
+    - Default user is `aries`
+    - Uses container's system python environment rather than `pyenv`
+    - Askar and Indy Shared libraries are installed through pip from
+      pre-compiled binaries included in the python wrappers.
+    - Built from repo contents
```
WadeBarnes, line 72: This works for release builds only because the images are built at the same time, and therefore with the same dependencies, as the package that is published to PyPi.

ContainerImagesAndGithubActions.md
```diff
@@ -0,0 +1,122 @@
...
+- Indy Image
+    - Based on slim variant of Debian
+    - Based on `indy-python`
+    - Includes `libindy` but does **NOT** include the Indy CLI
```
WadeBarnes, line 76: This is a good thing!

ContainerImagesAndGithubActions.md
```diff
@@ -0,0 +1,122 @@
...
+    - Default user is `indy`
+    - Based on `indy-python`
```
WadeBarnes, line 78: Duplicate line

ContainerImagesAndGithubActions.md
```diff
@@ -0,0 +1,122 @@
...
+    - Uses container's system python environment rather than `pyenv`
+    - Askar and Indy Shared libraries are installed through pip from
+      pre-compiled binaries included in the python wrappers
```
WadeBarnes, line 81: More specifically they are installed as dependencies of aca-py when the built wheel is used for installation.

ContainerImagesAndGithubActions.md
```diff
@@ -0,0 +1,122 @@
...
+    - Based on slim variant of Debian
+    - Does **NOT** include `libindy`
+    - Default user is `aries`
+    - Uses container's system python environment rather than `pyenv`
+    - Askar and Indy Shared libraries are installed through pip from
+      pre-compiled binaries included in the python wrappers.
+    - Built from repo contents
+- Indy Image
+    - Based on slim variant of Debian
+    - Based on `indy-python`
```
WadeBarnes, line 75: In the Dockerfile.indy, this is referred to as `indy-base`.

ContainerImagesAndGithubActions.md
```diff
@@ -0,0 +1,122 @@
...
+    - Includes `libindy` but does **NOT** include the Indy CLI
+    - Default user is `indy`
+    - Based on `indy-python`
+    - Uses container's system python environment rather than `pyenv`
+    - Askar and Indy Shared libraries are installed through pip from
+      pre-compiled binaries included in the python wrappers
+    - Built from repo contents
```
WadeBarnes, line 82: This works for release builds only because the images are built at the same time, and therefore with the same dependencies, as the package that is published to PyPi.

docker/Dockerfile.indy
```diff
@@ -0,0 +1,260 @@
...
+    cargo clean
+
+# Package python3-indy
+RUN tar czvf ../python3-indy.tgz -C indy-sdk*/wrappers/python .
+
+# grab the latest sdk code for the postgres plug-in
+WORKDIR $HOME
+ARG indy_postgres_url=${indy_sdk_url}
+RUN mkdir indy-postgres && \
+    curl "${indy_postgres_url}" | tar -xz -C indy-postgres
```
WadeBarnes, line 66: The `indy-sdk` is getting downloaded twice, once to build the `sdk` and once to build the `postgres plug-in`.  I'd recommend downloading the source once and building the both packages off the same source, like it's done in `von-image`.

WadeBarnes (changes requested):
@dbluhm, I finally got around to reviewing this PR in detail.  Sorry it took so long.

It appears the `acapy-test` target in `Dockerfile.indy` replaces `Dockerfile.test-indy`, and that `ghcr.io/hyperledger/indy-python` is no longer built and published.  If this is the case remove `Dockerfile.test-indy`.  Should a similar approach be taken with `Dockerfile` to replace `Dockerfile.test`?

Remove the `Makefile`, it's already obsolete.

Specific review comments follow.

dbluhm:
>  Should a similar approach be taken with Dockerfile to replace Dockerfile.test?

Fair question; I think we could add the test image as a stage in the main Dockerfile but I don't think I would recommend this. The test image being a stage for the indy setup was essentially required for us to have the indy-base available without needing to separately commit it to an image. We don't have the same requirements for the standard, non-indy setup. There's no shared base (besides just a python image) between the main docker image and the test docker image.

.github/workflows/nightly-tests.yml
dbluhm: Ah, thanks for pointing this out. I threw python 3.6 back into the mix last minute and didn't consider that it might be unavailable. I'll see if I can add ubuntu version as another matrix parameter to solve the issue. When we've officially moved away from python 3.6, I imagine we can just go back to running with ubuntu:latest.

ContainerImagesAndGithubActions.md
dbluhm: I'm not certain I follow your meaning here. The Dockerfile will build a wheel from the repo contents and install that wheel in the final stage of the image build. As currently structured, the Dockerfile is not pulling from PyPI for the ACA-Py package, just its dependencies.

docker/Dockerfile.indy
dbluhm: I think this is a fair simplification but to call out an implication: this would mean that the postgres storage plugin with a slightly different version from the Indy SDK could not be installed. I think I got the idea that this was desirable by looking at `von-image`'s build scripts, actually lol. So do we need to retain this ability?

.github/workflows/publish-indy.yml
dbluhm: I believe this has been corrected now. See: https://github.com/dbluhm/aries-cloudagent-python/actions/runs/3603865313/jobs/6072572552

WadeBarnes (changes requested):
Looking better.

Example test run; https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3689652122.  Tests against python versions 3.7 and 3.10 for `ACA-Py` and `ACA-Py (Indy)` both fail (as expected) as they did before.

There is still an issue with the images built when triggered from a release.  The python tags are not getting added like they do when you run the pipeline manually.

References (there is only a single image tagged `1.0.0-rc3`):
- https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3694879436
- https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3694879435
- https://github.com/WadeBarnes/aries-cloudagent-python/pkgs/container/aries-cloudagent-python/versions

You'll see the manual runs (`test-acapy-publish` and `test-acapy-publish-indy`) were tagged with the python version number as expected.  However, the release triggered runs of both `Publish ACA-Py` and `Publish ACA-Py (Indy)` are overwriting each other's images AND are not receiving the expected python version prefix.  I suspect the manually triggered publish workflows would overwrite each others images too, if I had used the same tags for each, since the two builds do not use explicit tags to distinguish the resulting `indy` and `askar` images.

The other issue with the image builds is the versions of python being used.  Images are being published with python 3.7, 3.8, 3.9, and 3.10, however, the aca-py tests fail when using python 3.7 and 3.10 for both `ACA-Py` and `ACA-Py (Indy)` tests.  Images should not be built for versions when their associated tests fail.  Otherwise we're publishing images that will not work.

There are also no images build for python 3.6 to keep things consistent with the current `bcgovimages/aries-cloudagent` images.

dbluhm:
@WadeBarnes @swcurran in the interest of keeping this moving forward, here's a quick update: I've attempted to correct the issues and pushed what I hope will fix at least the tag name collision issues. I have some currently broken work on running tests before publishing a release which can be found here: https://github.com/dbluhm/aries-cloudagent-python/tree/fix/release-version-issues

I propose that, unless we have major issues with these as currently structured, we go ahead and merge at least the PR related actions so we can overcome CircleCI issues and then commit to fixing the release workflow ASAP. I'd be happy to prepare a separate PR with just the actions for running tests on nightly and PRs if that would help us feel more comfortable.

WadeBarnes:
@dbluhm, It looks like there are still some bugs to workout with the images published with the workflows are triggered by a release.  The images are still getting overwritten, as they are not getting tagged with the python version.

References:
- https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3743057801
- https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3743058781
- https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3743062864
- https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3743062872
- https://github.com/WadeBarnes/aries-cloudagent-python/pkgs/container/aries-cloudagent-python/versions

The `pr-tests` workflow is working; https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/3743130229

I think the best way forward is to separate the PRs.  Please proceed with preparing a separate PR with just the actions for running tests on nightly and PRs.

dbluhm:
This PR is being broken up into more bite-sized pieces. See:
- #2058 
- #2059 
- Another PR will be opened for docker images and publishing those images.

Issue #1868, opened 2022-07-20
Workaround because the revocation registry fix cannot update accumulator stored in Indy SDK storage


swcurran:
Recently, an endpoint was added to "fix" an out of sync RevReg. Out of sync in this case can be caused by an Issuer trying to revoke a credential, but the update on the ledger failing, leaving the two out of sync. To address this, an ACA-Py endpoint was added that recalculated the accumulator from the list of revoked credentials, and updates the ledger. However, this still does not prevent future revocation errors when using the Indy SDK storage (indy-wallet) because it retains a copy of the accumulator that cannot be updated using the (current) Indy SDK.

A possible workaround to this issue for the Indy SDK is to catch the exception that occurs when a credential is revoked and immediately invoke the "fix RevReg" endpoint.  This results in the ledger being accurate (which is what is important), and the accumulator in the Indy SDK storage remaining out of sync.

A better fix is possible with Askar as it can update the accumulator and will be added as another issue to be addressed.

PR #1892, opened 2022-08-09
Refactor ledger correction code and insert into revocation error handling

ianco:
Addresses issue https://github.com/hyperledger/aries-cloudagent-python/issues/1868

### DIFF
--- a/aries_cloudagent/revocation/models/issuer_rev_reg_record.py
+++ b/aries_cloudagent/revocation/models/issuer_rev_reg_record.py
@@ -7,7 +7,7 @@ from functools import total_ordering
 from os.path import join
 from pathlib import Path
 from shutil import move
-from typing import Any, Mapping, Sequence, Union
+from typing import Any, Mapping, Sequence, Union, Tuple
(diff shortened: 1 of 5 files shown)


aries_cloudagent/revocation/models/issuer_rev_reg_record.py
```diff
@@ -307,6 +340,80 @@ async def send_entry(
 
         return rev_entry_res
 
+    async def fix_ledger_entry(
+        self,
+        profile: Profile,
+        apply_ledger_update: bool,
+        genesis_transactions: str,
+    ) -> (dict, dict, dict):
```
shaangill025, line 348: I see a warning on my end here: Tuple expression not allowed in type annotation. Use Tuple[T1, ..., Tn] to indicate a tuple type

shaangill025 (dismissed):

ianco:
Per feedback from @WadeBarnes - do a TAA check before allowing the revocation to be processed (same applies to schema and cred def operations)

shaangill025 (approved):

Issue #1895, opened 2022-08-12
Stopping the aca-py shell process keeps python process running


tdiesler:
This becomes a problem when controlling the aca-py process from another (e.g. Java) process. The aca-py shell process does not need to create another process.

```
exec [-cl] [-a name] [command [arguments]]
              If command is specified, it replaces the shell.  No new process  is  created.   The
              arguments become the arguments to command.  If the -l option is supplied, the shell
              places a dash at the beginning of the zeroth argument passed to command.   This  is
              what  [login](https://manpages.ubuntu.com/manpages/trusty/en/man1/login.1.html)(1)  does.   The  -c  option causes command to be executed with an empty
              environment.  If -a is supplied, the shell passes name as the  zeroth  argument  to
              the  executed  command.   If  command  cannot  be  executed for some reason, a non-
              interactive shell exits, unless the execfail shell  option  is  enabled.   In  that
              case,  it returns failure.  An interactive shell returns failure if the file cannot
              be executed.  If command is not specified, any  redirections  take  effect  in  the
              current  shell,  and  the return status is 0.  If there is a redirection error, the
              return status is 1.
```

tdiesler:
PR: https://github.com/hyperledger/aries-cloudagent-python/pull/1896

PR #1896, opened 2022-08-12
[#1895] Stopping the aca-py shell process keeps python process running

tdiesler:
Signed-off-by: Thomas Diesler <tdiesler@redhat.com>


swcurran:
What's up with the Circle-CI fail - @andrewwhitehead @ianco .  Per @dbluhm 's suggestion the other day -- is it time to move from CircleCI to GHA, or is this a legit fail?

Oh, and please take a look at the proposed change in the PR :-).

ianco:
> What's up with the Circle-CI fail - @andrewwhitehead @ianco . Per @dbluhm 's suggestion the other day -- is it time to move from CircleCI to GHA, or is this a legit fail?

Not sure, I updated the branch to the latest code from `main` so this should re-run the tests.  (Everything runs fine on my local.)

> Oh, and please take a look at the proposed change in the PR :-).

I think the change is fine, however if you're running aca-py from another process I suggest to run python directly rather than through a shell script.

ianco (dismissed):

swcurran:
@WadeBarnes @ryjones -- suggestions on how we ramp up to the use of SonarCloud?  A one-line, 5 character PR should not need to cleanup 257 files to be merged.  :-). 

What is the best strategy here?  Ideally someone at some point goes through the duplication things and makes decisions (fix, ignore warning), but that won't happen today.

WadeBarnes:
@swcurran, The behavior can be adjusted via config.  For example; https://github.com/hyperledger/blockchain-explorer/blob/8c85d2a6b0f3b7af88c80ce2fd3f252138d09343/.sonarcloud.properties.  Someone just needs to submit the config file.  Also the Sonar Scans are not blocking merges, it's just info at this point; i.e. it's not a required status check.

swcurran (dismissed):

swcurran:
@WadeBarnes it is a block -- see current state of this PR.  There is not a green "merge pull request" button -- which doesn't feel right.  Is it possible to configure the branch check so that would be the case on this PR? 

Configuring the behaviour seems very difficult -- at least at this point.  For example, how do we "ignore duplication" in a way that let's us move forward (e.g. green buttons), but doesn't mean we never look at the issues flagged?  AFAIK -- we don't know what duplication checks are useful and should be fixed vs. those that should be ignored.  For example, can we ignore all duplication checks in the "test" source files?

WadeBarnes:
It's not blocking the merge.  The option to `Merge pull request` is still there.   If it was a block you would not have the option to `Merge pull request`.  It's simply not green because all of the status checks did not "pass".  The button not being green is an indication for you to review the status checks and decide whether to merge the code.

I'm not sure why it's complaining about code that is not part of the changes in this PR though.

@ryjones, Is there a way to relax the quality gates on the server side?

swcurran:
Clarify what I mean by blocking -- AFAIK, if any of the checks fail, I can still merge it as the non-green "merge pull request" still shows.  I don't want us to get in the habit of merging in that state, as all of the other checks should be blocking.  Right now this one shouldn't.

WadeBarnes:
If a `Required` check fails you would not see the option to `Merge pull request`.  There are only two `Required` check set right now.

WadeBarnes:
First step might be to add `sonar.cpd.exclusions=**/tests/**` to a  `.sonarcloud.properties` file to exclude tests from duplicate (copy-paste) detection.

PR #1897, opened 2022-08-12
Fixes a few AATH failures

ianco:
Fixes:
- one 023-did-exchange error
- 0434-oob errors
- one 0183-revocation error

### DIFF
--- a/aries_cloudagent/revocation/routes.py
+++ b/aries_cloudagent/revocation/routes.py
@@ -424,7 +424,8 @@ async def revoke(request: web.BaseRequest):
     body["notify"] = body.get("notify", context.settings.get("revocation.notify"))
     notify = body.get("notify")
     connection_id = body.get("connection_id")
-    notify_version = body.get("notify_version", "v1_0")
+    body["notify_version"] = body.get("notify_version", "v1_0")
+    notify_version = body["notify_version"]
 
     if notify and not connection_id:
         raise web.HTTPBadRequest(reason="connection_id must be set when notify is true")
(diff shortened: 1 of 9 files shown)


swcurran:
Helpful to know which ones are fixed :-).  I'll check it out on main and this branch.

aries_cloudagent/revocation/routes.py
```diff
@@ -424,7 +424,8 @@ async def revoke(request: web.BaseRequest):
     body["notify"] = body.get("notify", context.settings.get("revocation.notify"))
     notify = body.get("notify")
     connection_id = body.get("connection_id")
-    notify_version = body.get("notify_version", "v1_0")
+    body["notify_version"] = body.get("notify_version", "v1_0")
+    notify_version = body["notify_version"]
```
ianco, line 428: Question about this change - basically it changes aca-py behaviour to *always* send a revocation notification if the startup option is set (`--notify-revocation`) vs only sending if a `notify_version` is explicitly set in the revocation payload.

I assume that if you are notifying then you always want to notify.

swcurran: I would think that is right.  And presumably if `notify_version` is not set, it should default to something -- e.g. the latest version?

ianco: > I would think that is right. And presumably if `notify_version` is not set, it should default to something -- e.g. the latest version?

`v1_0` is the default  (it was the default before however the value wasn't getting passed through to the actual notification code so no notification was getting sent)

PR #1899, opened 2022-08-15
Feat/public did endpoints for agents behind mediators

cjhowland:
This PR supports publishing endpoint information with routing keys in accordance with the did:sov specification. Routing keys are retrieved from `mediation_id` if specified, or else a default mediator. If no mediated connection is present, `routingKeys` defaults to an empty array. 

This PR allows for publishing DIDs with routing keys; PR #1863 allows for resolving DIDs with routing keys.

### DIFF
--- a/aries_cloudagent/ledger/base.py
+++ b/aries_cloudagent/ledger/base.py
@@ -7,7 +7,7 @@ import re
 from abc import ABC, abstractmethod, ABCMeta
 from enum import Enum
 from hashlib import sha256
-from typing import Sequence, Tuple, Union
+from typing import List, Sequence, Tuple, Union
 
 from ..indy.issuer import DEFAULT_CRED_DEF_TAG, IndyIssuer, IndyIssuerError
 from ..utils import sentinel
@@ -80,6 +80,23 @@ class BaseLedger(ABC, metaclass=ABCMeta):
         """
 
     @abstractmethod
+    async def construct_attr_json(
+        self,
+        endpoint: str,
+        endpoint_type: EndpointType = None,
+        all_exist_endpoints: dict = None,
+        routing_keys: List[str] = None,
+    ) -> str:
+        """Create attr_json string.
+
+        Args:
+            all_exist_endpoings: Dictionary of all existing endpoints
+            endpoint: The endpoint address
+            endpoint_type: The type of the endpoint
+            routing_keys: List of routing_keys if mediator is present
+        """
+
+    @abstractmethod
     async def update_endpoint_for_did(
         self,
         did: str,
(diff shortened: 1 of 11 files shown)


aries_cloudagent/ledger/base.py
```diff
@@ -75,6 +75,23 @@ async def get_all_endpoints_for_did(self, did: str) -> dict:
             did: The DID to look up on the ledger or in the cache
         """
 
+    @abstractmethod
+    async def construct_attr_json(
```
dbluhm, line 79: The implementation in the concretions for this method are exactly the same (right now, at least -- perhaps there's a reason for there to be differences in the future). It might make sense to move the implementation of this method to the base class here and just have it not be an abstract method. Also, just by convention, since we don't really intend this method to be called outside of the ledger class, it might be more appropriate to consider it "private" and name it `_construct_attr_json`.

dbluhm (changes requested):
Thank you for the thorough tests! I have just one minor change request.

dbluhm:
To further comment on the necessity of this PR, without this, an ACA-Py instance configured to operate behind a mediator is currently unable to publish accurate endpoint information for forming DIDComm connections through its public DID.

dbluhm:
The SonarCloud analysis is new. It's complaining about test values. @cjhowland we could probably appease it by just changing the example endpoint to something like `https://example.com`.

WadeBarnes:
> The SonarCloud analysis is new. It's complaining about test values. @cjhowland we could probably appease it by just changing the example endpoint to something like `https://example.com`.

Yes, it's new.  It just got added to some of the Indy and Aries projects today.

dbluhm:
Should/Can we configure it to ignore tests? I suppose if we leave it as a not required check, maybe it's not a problem but "fixing" the code duplication warnings for the tests would be inappropriate, IMO. Maybe I just have an aversion to seeing red X's on PRs that I'm approving :slightly_smiling_face:

WadeBarnes:
> Should/Can we configure it to ignore tests? I suppose if we leave it as a not required check, maybe it's not a problem but "fixing" the code duplication warnings for the tests would be inappropriate, IMO. Maybe I just have an aversion to seeing red X's on PRs that I'm approving 🙂

Yes, the settings can be adjusted by adding a config file.  Example config file; https://github.com/hyperledger/blockchain-explorer/blob/8c85d2a6b0f3b7af88c80ce2fd3f252138d09343/.sonarcloud.properties

dbluhm:
Char is OOTO so I'll take a look at test failures today.

WadeBarnes:
More suggestions here; https://github.com/hyperledger/aries-cloudagent-python/pull/1896#issuecomment-1218523849

dbluhm (approved):
LGTM!

ianco:
Please see PR https://github.com/bcgov/von-network/pull/233

PR #1906, opened 2022-08-22
Send webhooks upon record/credential deletion

frostyfrog:
There is a ToDo note talking about sending webhooks when records are deleted. There are also no webhooks that get sent when a credential is deleted. This PR aims to fix the ToDo, as well as provide webhooks for when a credential has been deleted.


aries_cloudagent/holder/routes.py
```diff
@@ -274,6 +274,10 @@ async def credentials_remove(request: web.BaseRequest):
         async with context.profile.session() as session:
             holder = session.inject(IndyHolder)
             await holder.delete_credential(credential_id)
+            topic = "acapy::record::credential::delete"
+            await session.profile.notify(
+                topic, {"id": credential_id, "state": "deleted"}
+            )
```
andrewwhitehead, line 280:
```suggestion
        topic = "acapy::record::credential::delete"
        await context.profile.notify(
            topic, {"id": credential_id, "state": "deleted"}
        )
```

Better to close the session first.

frostyfrog: Great catch! I still have trouble understanding what's in the various base objects (like context and session), so this helped a lot

andrewwhitehead:
LGTM!

ianco (approved):

PR #1908, opened 2022-08-23
Add offers_attach to V10 credential schema

JonasNiestroj:
We found out that the `offers_attach` attribute is not present in the credential offer schema. It is possible to send it to the api but its not present in the openapi spec. This is a problem is you generate yourself a client with the openapi spec because offers_attach is not defined.

This pr adds the attribute to the credential offer schemas.
It seems that there were also some other changes that are not reflected in the openapi.json. Therefore the changes in the openapi spec are bigger than the real changes in this pr.

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v1_0/routes.py
+++ b/aries_cloudagent/protocols/issue_credential/v1_0/routes.py
@@ -9,6 +9,8 @@ from aiohttp_apispec import (
     response_schema,
 )
 from json.decoder import JSONDecodeError
+from aries_cloudagent.messaging.decorators.attach_decorator import AttachDecoratorSchema
+
 from marshmallow import fields, validate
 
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
Can you please leave out the changes to openapi.json? Updating that likely deserves another PR.

swcurran:
I don't understand how this is needed.  This API and protocol has had extensive testing for years and works.  Why would this be needed now? I also notice that you have `offers_attach`, but the correct item, per [RFC 0036](https://github.com/hyperledger/aries-rfcs/tree/main/features/0036-issue-credential#offer-credential) is `offers~attach`. Is it possible that this is a misunderstanding and not needed?

JonasNiestroj:
Sorry, i guess it was a little bit of a misunderstanding from my side. I went a step back and created an issue about the use case.
[Issue](https://github.com/hyperledger/aries-cloudagent-python/issues/1918)

PR #1914, opened 2022-08-24
feat: include connection ids in keylist update webhook

dbluhm:
This PR adds the connection IDs of the mediated connections. This should make it easier to trigger the next step in a flow (i.e. accept a connection request for one of the connections, if it's in the right state).


dbluhm:
This PR doesn't yet address multi-tenancy with a mediator connected to the base wallet. I'll mark this as a draft as I work on that.

ianco (approved):

ianco:
@dbluhm  can you catch this one up with the latest main branch?

PR #1915, opened 2022-08-25
chore: update pydid

dbluhm:
Recent changes to the resolution of sov/indy DIDs caused PyDID to parse
service endpoints that should have been recognized as DIDComm services
as plain services. This broke some logic in accepting invitations from
public DIDs. Pulling in these updates from PyDID resolves the issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ pyld~=2.0.3
 pyyaml~=5.4.0
 ConfigArgParse~=1.5.3
 pyjwt~=2.4.0
-pydid~=0.3.3
+pydid~=0.3.6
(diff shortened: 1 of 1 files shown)


dbluhm:
The error this fixes:
```
aries_cloudagent.connections.base_manager.BaseConnectionManagerError: Cannot connect via public DID that has no associated DIDComm services
```

Caused by these lines:
https://github.com/hyperledger/aries-cloudagent-python/blob/a8101e1fef6004c79a8f0fb10b63138fb0ad140f/aries_cloudagent/connections/base_manager.py#L256-L264

andrewwhitehead (approved):

PR #1922, opened 2022-08-30
fix: propagate endpoint from mediation record

cjhowland:
This PR solves a bug in #1899 in which the public DID endpoint of an agent behind a mediator was the agent's own endpoint. These changes ensure that a mediated agent's public DID endpoint is the mediator's endpoint.

Signed-off-by: Char Howland <char@indicio.tech>


andrewwhitehead (approved):
LGTM

swcurran:
Any insight into the integration test(s) that are failing, @cjhowland ?  Caused by this PR or intermittent?

Thanks

andrewwhitehead:
The integration test failure seems to be intermittent, and could be a bug in the test suite. It's submitting the value 'None' as a revocation registry identifier.

andrewwhitehead:
I think the test failure should be fixed by #1924

andrewwhitehead (approved):

PR #1924, opened 2022-08-31
Pre-populate revoc_reg_id on IssuerRevRegRecord

andrewwhitehead:
This avoids having a confusingly blank `revoc_reg_id` on the record in the `init` state.

Registries in the `init` state are also filtered from the `/revocation/registries/created` endpoint.


ianco (approved):

ianco:
Integration tests all pass on my local so I'm merging.

PR #1926, opened 2022-09-02
Endorser doc updates and some bug fixes

ianco:

Fixes a couple of issues with DID ATTRIB transactions, and adds some technical docs.


Endorser.md
```diff
@@ -61,3 +59,44 @@ Endorsement:
...
+- a [manager.py](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py) file implements common functionality that is called from both the routes.py and handler classes (as well as from other classes that need to interact with Endorser functionality)
+
+The Endorser makes use of the [Event Bus](https://github.com/hyperledger/aries-cloudagent-python/blob/main/CHANGELOG.md#july-14-2021) (links to the PR which links to a hackmd doc) to notify other protocols of any Endorser events of interest.  For example, after a Credential Definition endorsement is received, the TransactionManager writes the endorsed transaction to the ledger and uses teh Event Bus to notify the Credential Defintition manager that it can do any required post-processing (such as writing the cred def record to the wallet, initiating the revocation registry, etc.).
+
+The overall architecture can be illustrated as:
+
+![Class Diagram](./docs/assets/endorser-design.png)
```
swcurran, line 74: Not crucial, but I think you can now use mermaid directly in GitHub-flavoured Markdown, so you don't have to have this as a separate asset -- https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

Endorser.md
```diff
@@ -61,3 +59,44 @@ Endorsement:
...
+
+### Create Credential Definition and Revocation Registry
+
+An example of an Endorser flow is as follows, showing how a credential definition endorsement is received and processed, and optionally kicks off the revocation registry process:
+
+![Sequence Diagram](./docs/assets/endorse-cred-def.png)
+
+You can see that there is a standard endorser flow happening each time there is a ledger write (illustrated in the "Endorser" process).
+
+At the end of each endorse sequence, the TransactionManager sends a notification via the EventBus so that any dependant processing can continue.  Each Router is responsibiel for listening and responding to these notifications if necessary.
```
swcurran, line 84: Typo - responsible.

swcurran (changes requested):
I'd say merge the puml into the Readme before merging.  And fix the typo. I will not be able to test directly, so will leave it to @shaangill025 or @andrewwhitehead to provide a code review.

@dbluhm -- not sure how much your team plans on using the endorser, but you might want to do a review as well.

ianco:
@swcurran @frostyfrog @dbluhm there's support for a startup parameter called `--auto-promote-author-did`, which (for authors) will auto-promote their DID to be their public DID on receipt of the endorser response.

However I now feel that this parameter is pretty useless, and I'd like to get rid of it.  I think it may end up causing more problems that it's resolving.

(I mention it in this PR because one of the issues I was addressing - from @cchiu : "i also tried setting the auto-promote-author-did: true in the author's config files but i still get the same issue where the ATTRIB is not written to the ledger when an org is created under the author-endorser flow using v0.7.4." - mentioned this flag and it didn't look like it was working properly.)

The intent was that an author wouldn't need to monitor their endorser responses - if the endorser endorsed their DID transaction, then their aca-py would auto-promote this DID to be their public DID.

However it's only useful in the case where the author *already has* a public DID, so they can sign their endorser request.  If they don't have a public DID yet, then they have to go though some out of band process anyways to register their first public DID (maybe phone a friend?) so this `--auto-promote-author-did` is pretty useless.

Any opinions on if I can just remove it?

swcurran:
Should the scenario of an author creating a DID for itself be handled by the Endorser not signing and returning the DID, but rather creating and executing the transaction?  That is the flow that is expected -- the "out of band" process you are mentioning. Basically, handle requests to create DIDs differently from other transactions.

ianco:
> Should the scenario of an author creating a DID for itself be handled by the Endorser not signing and returning the DID, but rather creating and executing the transaction?

Possibly, but this is a fairly big change.  The author can't *sign" the request, because they have no public DID to sign with.  So the endorser process would have to be customized for this one scenario for the author to send an unsigned request, the endorser to write the transaction, and then to send an acknowledgement back.  (Which kind of goes against the intent to make the endorser protocol a generic "please sign this" protocol.)

So if we're going to do this I suggest it be a separate protocol (or at least a dedicated function within the endorser protocol).

And there's still the question of whether to keep or get rid of the `--auto-promote-author-did` param.

swcurran:
Fair enough.  I think we need that functionality, but agree that it is a separate protocol and Issue/PR from the rest of the endorsing.  I would say remove the start up option, and lets add an issue that will extend the work of the endorser to handle this new type of action, with a new protocol.

Endorser.md
ianco: This is cool, thanks for the link!  I did a quick test and the sequence diagrams don't render ...  I think I'll leave these ones in puml and next time I need to diagram I'll try out mermail.

ianco:
> Fair enough. I think we need that functionality, but agree that it is a separate protocol and Issue/PR from the rest of the endorsing. I would say remove the start up option, and lets add an issue that will extend the work of the endorser to handle this new type of action, with a new protocol.

Added a new issue https://github.com/hyperledger/aries-cloudagent-python/issues/1929

andrewwhitehead (approved):

swcurran (approved):

PR #1940, opened 2022-09-12
Fix: OOB - Handling of minor versions

### LINKED ISSUE #1917: OOB Handling of both 1.0 and 1.1 messages
It appears that ACA-Py is not handling OOB messages for both protocol version 1.0 and 1.1, as it should be.  Per [Aries RFC 0003](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0003-protocols#semver-rules-for-protocols), ACA-Py should be able to handle all minor versions of a protocol.  Please investigate.

See [this comment on an AFJ](https://github.com/hyperledger/aries-framework-javascript/pull/930#issuecomment-1228897006) issue with details about what is happening.

@JamesKEbert -- could you please add steps to reproduce?

Thanks!

shaangill025:
Signed-off-by: Shaanjot Gill [gill.shaanjots@gmail.com](mailto:gill.shaanjots@gmail.com)

- resolve #1917
- Fixes the following error:
```
File "/home/indy/.venv/lib/python3.6/site-packages/aries_cloudagent/utils/classloader.py", line 97, in load_class
   |     if "." in class_name:
   | TypeError: argument of type 'NoneType' is not iterable
```
- Working on adding minor changes for adding OOB 1.1 and a mechanism for ACA-Py to respond with the same supported version, basically, `handshake-reuse-accepted 1.1` for `handshake-reuse 1.1` and `handshake-reuse-accepted 1.0` for `handshake-reuse 1.0`.

### DIFF
--- a/aries_cloudagent/core/tests/test_dispatcher.py
+++ b/aries_cloudagent/core/tests/test_dispatcher.py
@@ -1,7 +1,7 @@
 import json
 
 from async_case import IsolatedAsyncioTestCase
-import mock as async_mock
+from asynctest import mock as async_mock
 import pytest
 
 from marshmallow import EXCLUDE
@@ -111,9 +111,17 @@ class TestDispatcher(IsolatedAsyncioTestCase):
             StubAgentMessageHandler, "handle", autospec=True
         ) as handler_mock, async_mock.patch.object(
             test_module, "ConnectionManager", autospec=True
-        ) as conn_mgr_mock:
+        ) as conn_mgr_mock, async_mock.patch.object(
+            test_module,
+            "get_version_from_message_type",
+            async_mock.CoroutineMock(return_value="1.1"),
+        ), async_mock.patch.object(
+            test_module,
+            "validate_get_response_version",
+            async_mock.CoroutineMock(return_value=("1.1", None)),
+        ):
             conn_mgr_mock.return_value = async_mock.MagicMock(
-                find_inbound_connection=async_mock.AsyncMock(
+                find_inbound_connection=async_mock.CoroutineMock(
                     return_value=async_mock.MagicMock(connection_id="dummy")
                 )
             )
@@ -152,7 +160,15 @@ class TestDispatcher(IsolatedAsyncioTestCase):
 
         with async_mock.patch.object(
             StubAgentMessageHandler, "handle", autospec=True
-        ) as handler_mock:
+        ) as handler_mock, async_mock.patch.object(
+            test_module,
+            "get_version_from_message_type",
+            async_mock.CoroutineMock(return_value="1.1"),
+        ), async_mock.patch.object(
+            test_module,
+            "validate_get_response_version",
+            async_mock.CoroutineMock(return_value=("1.1", None)),
+        ):
             await dispatcher.queue_message(
                 dispatcher.profile, make_inbound(message), rcv.send
             )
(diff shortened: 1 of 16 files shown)


aries_cloudagent/protocols/out_of_band/definition.py
```diff
@@ -3,8 +3,8 @@
 versions = [
     {
         "major_version": 1,
-        "minimum_minor_version": 0,
-        "current_minor_version": 0,
+        "minimum_minor_version": 1,
```
swcurran, line 6: We want to support minor version 0 for OOB.  Unless it is hugely complex, we **always** want to support all minor versions from 0 to `current_minor_version` to ensure backwards compatibility.

swcurran (changes requested):
Needs to support OOB 1.0 and 1.1.

shaangill025:
@swcurran As discussed, the problem reports [`version-with-degraded-features`, `fields-ignored-due-to-version-mismatch`] as specified in [RFC0003](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0003-protocols#recipient-rules) have been implemented but commented out.

ianco (approved):

swcurran (approved):

swcurran:
@WadeBarnes -- looks like the SonarCloud config is processing test code again.  Could you please take a look at that?

Do we need to be concerned about the 4 relevant code smells?  The complexity of the version checking code (I assume that is just the way it is) and the raising of the exceptions.  All good with those?

The unused variables in the test code and the commented out code are fine/intended, I'm sure.

Thanks

aries_cloudagent/core/tests/test_dispatcher.py
```diff
@@ -1,7 +1,7 @@
 import json
 
 from async_case import IsolatedAsyncioTestCase
-import mock as async_mock
+from asynctest import mock as async_mock
```
andrewwhitehead, line 4: This shouldn't be necessary, maybe you need to install requirements.dev.txt?

aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
```diff
@@ -140,12 +143,20 @@ def __init__(
...
 
     @classmethod
     def wrap_message(cls, message: dict) -> AttachDecorator:
         """Convert an aries message to an attachment decorator."""
         return AttachDecorator.data_json(mapping=message, ident="request-0")
 
+    @classmethod
+    def assign_version_to_message_type(cls, version: str):
+        """Assign version to Meta.message_type."""
+        cls.Meta.message_type = Template(cls.Meta.message_type).substitute(
```
andrewwhitehead, line 156: Wouldn't this overwrite message_type for all the class instances going forward?

shaangill025: To update `message_type` in Meta, I found this can only be done in the classmethod before initialization. Regarding overwriting, `$version` is only present in `message_types` for OOB v1_0, so it will overwrite. But when we have a hardcoded version such as in other protocols then it won't overwrite.

aries_cloudagent/core/tests/test_dispatcher.py
shaangill025: Updated to use asyncMock().

shaangill025:
I believe CircleCI test failing is related to their ongoing security situation.

swcurran:
@WadeBarnes --- is there a way other than pushing another commit to trigger a re-execution of Circle/CI when things like this happen?  I hadn't heard about the Circle/CI issues -- any update on that?  Should we take the plunge here and switch to another tool?

ianco (approved):

Issue #1222, opened 2021-06-05
Sending image as part of credential


juvebogdan:
Hello,

I am trying to issue a credential that has image/jpeg as a field. I am using also mobile agent and mediator. but whenever i try to issue offer i get this error:

aries_cloudagent.transport.outbound.manager ERROR >>> Error when posting to: http://192.168.65.3:8022/webhooks/topic/issue_credential/; Error: (<class 'aries_cloudagent.transport.outbound.base.OutboundTransportError'>, OutboundTranspo
rtError('Unexpected response status 413, caused by: Request Entity Too Large',), <traceback object at 0x7f907d3e05c8>); Re-queue failed message ...

My image is 460Kb

Is it possible to change this?

I tried to change default value of --max-message-size in argparse.py but that did nothing

ducadellaforcoletta:
Hi, 
i found the same problem in the past at the following code line,
 https://github.com/hyperledger/aries-cloudagent-python/blob/653a4c2c4d19e1e3a0b6b89467affcea81827434/aries_cloudagent/admin/server.py#L288

The default client size is used, it can be customized passing the size as here app = web.Application(middlewares=middlewares, client_max_size=2048**2)

@andrewwhitehead  Maybe an args params could be usefull for the scope?

juvebogdan:
Hi,

I dont see that line in that file. Is that the latest code?

ducadellaforcoletta:
Sorry, I sent an old link. Looking at the actual code it looks like it is already managed with the "admin.admin_client_max_request_size" parameter. Default 1 mb. It can be customized using this parameter.

juvebogdan:
I saw that but whatever i put there i get the same error. There is no way i am sending more then 20MB

juvebogdan:
Maybe its because i am running faber demo? How can i add that option in that demo?

swcurran:
Closing this issue because it is old, but am open to having it re-opened, especially if you could provide a more complete example of the problem -- the steps you used to create the problem using the Alice/Faber demo.  We'd just need what you did differently from the doc, and an example image you are using.

PR #1941, opened 2022-09-13
Add startup flag --light-weight-webhook to trim down outbound webhook payload

victorlee0505:
When we create a credential offer with image, the image is repeated in following objects (dict) 
"credential_request", "cred_request", "credential_proposal", "cred_proposal", "credential_offer", "cred_offer", "credential_preview", "cred_preview", "values", "credentials~attach", "offers~attach"

a Credential offer with 30KB image resulted in a webhook sized at 227KB in my use case and resulted HTTP 413 as well as unnecessary stress on WAF rule regex processes on those encrypted string and base64 image string.

This startup flag will remove those objects only on outbound webhook and size went down to 7KB at most.
I did preserve 1 object "cred_issue" which has the "rev_reg_id" and "cred_rev_id".

To me this is a more like a nice to have flag if someone ever encounter problem with image and webhook.

Previous related discussion
[PR #725](https://github.com/hyperledger/aries-cloudagent-python/pull/725)
[Issue #1222](https://github.com/hyperledger/aries-cloudagent-python/issues/1222)

Signed-off-by: Victor Lee <victorlee0505@gmail.com>


swcurran:
Interesting problem. Looking forward to folks reviewing this and the best approach to solving the very real issue.  I don't know nearly enough to know what is the right way to handle this, so adding @ianco @dbluhm @frostyfrog @andrewwhitehead to the discussion.

It would be a huge change, but perhaps the webhooks should only send new information to the controller (vs. the full state) and leave it to the controller to hold the data it needs to make decisions as the protocol proceeds.  Or do that, and add in an endpoint that says "send all state for protocol X" whenever it is needed by a controller?

Thanks for the PR!

frostyfrog:
I'll agree that this is an interesting problem and I'm certain it will only get worse when implementors don't take file size into consideration, especially if upper management requests higher resolution imagery. For example, an image from an iPhone that controls embedded 3D data for biometric systems.

This issue, in terms of credentials, is prevalent during not only credential issuance, but verification/proof and endorsement. There are also plugins implementing features (such as the Q and A protocol) which may also cause increased webhook payloads if the system chooses to accept binary or non-fixed width payloads (I'm sure some malicious user is just waiting to send a script from Shakespeare as an answer to "why are you visiting the country").

I believe that what you've got here is a great first step towards solving the issue, but I believe that there's a lot more that can be done overall. Making sure to decrease unnecessarily duplicated data is one thing. Limiting the size of the payload in the face of untrusted data is another. I recommend that, for lightweight payloads, we send the minimal amount of data needed to request the full data where possible. Such as a presentation exchange ID during proof and allowing the system to retrieve the full record manually if they need it. This may be considered a bit extreme, but since we can't quite trust that the third party agent won't send us something massive, I feel as though it's the best compromise to prevent errors.

P.S. I hope that's all readable. Woke up and sent it from my phone. Really interesting problem we've got here :)

swcurran:
@ianco could you take a look at this one?  Seems valuable to me.

ianco:
> @ianco could you take a look at this one? Seems valuable to me.

ok will do.

ianco (changes requested):
Looks good!  I suggest to rename the class (`LightWeightWebhook` -> `LightWeightCredExchWebhook` or something like that) since it's specific to the credential exchange.  (Or possibly even have 2 separate classes for the v1 and v2 cred exch.)  That way if we implement lightweight hooks for other protocols we already have a naming convention in place.

Another possibility would be to put the lightweight hook model into the protocol messaging package?

swcurran:
Would it make sense to make this a breaking change with a “—useOldStyleCredExchangeProtocolStateObject” (or something equally weird) for backwards compatibility?  It seems this is something we would rather have than not.

We are doing a 0.8.0 breaking release change Real Soon Now, so this would fit there.

ianco:
I think the default is the old style verbose webhooks, no?  so not a breaking change ...  (unless you want the default to be the more terse format)

ianco:
PS  failing test:  `FAILED aries_cloudagent/messaging/models/light_webhook.py::FLAKE8`

swcurran:
I’m suggesting that we want the default to be the terse form, and only if needed use the long form. AFAIK, all of the data is available in the long form — it’s just the extra copies of the data that are in the short form.

Put another way, we could be smarter about not repeating content in the Protocol State Object by blindly adding in each message, where the message repeats data from a previous message.

Or is that a bad idea? I guess things could be lost if the content changes across messages — then we would only have the latest version vs. all versions.  How is that handled in this case?  Is there checking that the content hasn’t changed between messages, and if so what is done?

ianco:
> I’m suggesting that we want the default to be the terse form, and only if needed use the long form.

I think this would be fine, however if we do this we should probably implement a terse form for the presentation exchange as well (this PR only addresses the cred exchange).  (I think all the other exchanges are prob ok as is)

swcurran:
So maybe go with this now, and add an Issue to change the presentation model as well, and then change to terse by default  in the next breaking release?  In theory, we could get it in for the final 0.8.0.

victorlee0505:
sorry my bad, still getting use to black and flake8...ha...

ianco (approved):

victorlee0505:
@ianco @swcurran Thank you very much!
I think next step is to create new issue on proof exchange webhook and terse form by default so that we can carry the discussion further.

swcurran:
Thanks @victorlee0505 — agree on that as the next step.

PR #1955, opened 2022-09-21
did method registry

burdettadam:
To support different did methods in plugins we need to support did method pluggability. This PR changes the default methods to be static and registered into a registry of methods at start-up. This allows plugins to define new did methods that can be registered. The bulk of changes in this PR is updating the current code to use a did method registry.

### DIFF
--- a/aries_cloudagent/wallet/did_method.py
+++ b/aries_cloudagent/wallet/did_method.py
@@ -47,7 +47,7 @@ class DIDMethods:
     def __init__(self) -> None:
         self._registry: Dict[str, DIDMethod] = {
             SOV.method_name: SOV,
-            KEY.method_name: KEY
+            KEY.method_name: KEY,
         }
 
     def registered(self, method: str) -> bool:
--- a/aries_cloudagent/wallet/did_method.py
+++ b/aries_cloudagent/wallet/did_method.py
@@ -1,91 +1,68 @@
 """Did method enum."""
 
-from typing import List, Mapping, NamedTuple, Optional
-from enum import Enum
-
-from .key_type import KeyType
+from typing import Dict, List, Mapping, Optional
 from .error import BaseError
-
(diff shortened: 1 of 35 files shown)


dbluhm:
Thought provoking question: If we load a DID from the wallet that was created with a plugged in method but the current instance of ACA-Py does not have the same plugin loaded, should we fail on loading the DID or when we attempt to use the DID?

As currently structured, the DID Method includes info about key types and support for key rotation. This would need to be stored alongside DIDs in order for us to restore a DID without knowing the Method. This doesn't seem favorable so failing to load the DID seems like the right choice.

dbluhm:
Also making note of this here: the Indy wallet does not hold a reference to an injection context. This means it is unable to load the set of supported methods from the context. Many of its operations are hard coded for did:sov and did:key. This PR leaves those operations as hard coded to support did:sov and did:key and does not introduce the injection context on the Indy wallet.

ianco:
@burdettadam can you catch this one up with the latest main branch?

aries_cloudagent/wallet/did_method.py
```diff
@@ -1,91 +1,81 @@
...
 
-    SOV = DIDMethodSpec(
-        method_name="sov", supported_key_types=[KeyType.ED25519], supports_rotation=True
-    )
-    KEY = DIDMethodSpec(
-        method_name="key",
-        supported_key_types=[KeyType.ED25519, KeyType.BLS12381G2],
-        supports_rotation=False,
-    )
+    def __init__(self, name, key_types, rotation) -> None:
```
dbluhm, line 11: These parameters should have type hints on them

dbluhm (changes requested):

burdettadam:
We have a single PR with a related key type registry now, #1986. Closing this one in favor of the new one.

PR #1957, opened 2022-09-22
key type registry

burdettadam:
This PR makes key types pluggable to enable plugins to define and register new ones.


aries_cloudagent/wallet/did_method.py
```diff
@@ -20,11 +20,11 @@ class DIDMethod(Enum):
     """DID Method class specifying DID methods with supported key types."""
 
     SOV = DIDMethodSpec(
```
swcurran, line 22: Just as a matter of interest -- should there be other types of DID methods here now or in the future?  For example, should `did:indy` be here since we want to use it Real Soon Now.  If not, has consideration been given to how it would be added?  Also wondering about `did:peer`.  What if we start supporting another VDR (e.g. cheqd.io)?

I'm clueless at the this level, but I'm surprised that there are hardcoded references in ACA-Py code to specific DID Methods and cryptographic suites (e.g. ED25519 throughout the code) vs. them being variables with late binding during deployments.  I'm guessing I'm confused about this?

dbluhm: Indeed, this is exactly what we're seeking to be able to do. #1955 specifically addresses turning the set of known DID Methods from a static enum (as seen on these lines) into something that can be determined on start up with plugins getting a chance to add more methods. This PR does a similar "pluggification" or late binding as you put it for key types as the other PR does for DID methods.

While we are making methods and key types pluggable values, to limit breaking changes, the previously statically defined methods and key types are retained as statically declared values inserted into a dynamic registry by default.

At some future date, I think it makes sense for ACA-Py itself to only directly "know" about DID Peer and then all other methods are plugged in or configurable as built in plugins. This PR and #1955 are intended as incremental steps in that direction. Additional feedback on the approach is welcome :slightly_smiling_face:

ianco:
@burdettadam can you catch this one up with the latest main branch?  I get conflicts ...

aries_cloudagent/wallet/key_type.py
```diff
@@ -1,78 +1,95 @@
...
-
-    # NOTE: the py_multicodec library is outdated. We use hardcoded prefixes here
-    # until this PR gets released: https://github.com/multiformats/py-multicodec/pull/14
-    # multicodec is also not used now, but may be used again if py_multicodec is updated
-    ED25519 = KeySpec("ed25519", "ed25519-pub", b"\xed\x01")
-    X25519 = KeySpec("x25519", "x25519-pub", b"\xec\x01")
-    BLS12381G1 = KeySpec("bls12381g1", "bls12_381-g1-pub", b"\xea\x01")
-    BLS12381G2 = KeySpec("bls12381g2", "bls12_381-g2-pub", b"\xeb\x01")
-    BLS12381G1G2 = KeySpec("bls12381g1g2", "bls12_381-g1g2-pub", b"\xee\x01")
+    def __init__(self, key_type, multicodec_name, multicodec_prefix: bytes) -> None:
```
dbluhm, line 9: These parameters should have type hints on them

dbluhm (changes requested):

ianco (changes requested):
There are conflicts between this PR and PR #1955

dbluhm:
> There are conflicts between this PR and PR #1955

It does indeed; we have a patch handy to cleanly merge the two after one or the other is merged but maybe we should just go ahead and put it in a single PR.

PR #1960, opened 2022-09-27
fix: Safely shutdown when root_profile uninitialized

frostyfrog:
If an exception was thrown during the initialization of the root_profile, we would throw an exception during shutdown and fail to end the process. This would leave ACA-Py running, without accepting/handling any further requests. Adding guards for shutdown code that uses the root_profile should avoid this situation


swcurran:
PR added to the 0.7.5 milestone and reviews requested.

ianco (approved):
LGTM

PR #1962, opened 2022-10-01
feat: 00B v1.1 support

shaangill025:

### DIFF
--- a/aries_cloudagent/core/protocol_registry.py
+++ b/aries_cloudagent/core/protocol_registry.py
@@ -2,7 +2,7 @@
 
 import logging
 
-from string import Template
+from re import sub
 from typing import Mapping, Sequence
 
 from ..config.injection_context import InjectionContext
@@ -102,8 +102,8 @@ class ProtocolRegistry:
     def _get_updated_tyoeset_dict(self, typesets, to_check, updated_typeset) -> dict:
         for typeset in typesets:
             for msg_type_string, module_path in typeset.items():
-                updated_msg_type_string = Template(msg_type_string).substitute(
-                    version=to_check
+                updated_msg_type_string = sub(
+                    r"(\d+\.)?(\*|\d+)", to_check, msg_type_string
                 )
                 updated_typeset[updated_msg_type_string] = module_path
         return updated_typeset
--- a/aries_cloudagent/core/protocol_registry.py
+++ b/aries_cloudagent/core/protocol_registry.py
@@ -1,14 +1,14 @@
 """Handle registration and publication of supported protocols."""
 
 import logging
+import re
 
-from re import sub
 from typing import Mapping, Sequence
 
 from ..config.injection_context import InjectionContext
 from ..utils.classloader import ClassLoader
 
-from .error import ProtocolMinorVersionNotSupported
+from .error import ProtocolMinorVersionNotSupported, ProtocolDefinitionValidationError
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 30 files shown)


aries_cloudagent/resolver/base.py
```diff
@@ -132,7 +132,12 @@ async def supports(self, profile: Profile, did: str) -> bool:
 
         return bool(supported_did_regex.match(did))
 
-    async def resolve(self, profile: Profile, did: Union[str, DID]) -> dict:
+    async def resolve(
+        self,
+        profile: Profile,
+        did: Union[str, DID],
+        accept: Optional[Sequence[Text]] = None,
```
andrewwhitehead, line 139: I like the idea of the DID resolver receiving the accept options, I wonder if we can push that idea up to other levels (like the universal resolver).

aries_cloudagent/messaging/agent_message.py
```diff
@@ -98,6 +101,18 @@ def _get_handler_class(cls):
         """
         return resolve_class(cls.Meta.handler_class, cls)
 
+    @classmethod
+    def assign_version_to_message_type(cls, version: str):
```
andrewwhitehead, line 105: It seems like a real issue to me that the message type is being assigned at the class level and not the object level. It looks like instances of records like the OOB invitation may have a different version values, but the class is only loaded once and the version can only be set once until the agent is restarted?

shaangill025: I have reworked it to address this issue.

aries_cloudagent/resolver/base.py
shaangill025: [Here](https://github.com/decentralized-identity/universal-resolver/blob/main/README.md), I found the following:
```
curl -H "Accept: application/did+ld+json" https://dev.uniresolver.io/1.0/identifiers/did:sov:WRfXPg8dantKVubE3HX8pw
curl -H "Accept: application/did+json" https://dev.uniresolver.io/1.0/identifiers/did:sov:WRfXPg8dantKVubE3HX8pw
curl -H "Accept: application/did+cbor" https://dev.uniresolver.io/1.0/identifiers/did:sov:WRfXPg8dantKVubE3HX8pw
```

To use `Accept` in header but I don't think it will process something `didcomm/aip1`, `didcomm/aip2;env=rfc19` which `pydid` [Indy ledger] can.

andrewwhitehead: Right, it's using mime types that would describe the whole document, while we are using one that describes the required service endpoint(s). Maybe it should be called `service_accept`.

aries_cloudagent/core/protocol_registry.py
```diff
@@ -91,28 +91,34 @@ def create_msg_types_for_minor_version(self, typesets, version_definition):
...
                     self._get_updated_tyoeset_dict(typesets, to_check, updated_typeset)
                 )
+        else:
+            raise ProtocolDefinitionValidationError(
+                "min_minor_version is greater than curr_minor_version for the"
+                f" following typeset: {str(typesets)}"
+            )
         return (updated_typeset,)
 
     def _get_updated_tyoeset_dict(self, typesets, to_check, updated_typeset) -> dict:
```
andrewwhitehead, line 107:
```suggestion
    def _get_updated_typeset_dict(self, typesets, to_check, updated_typeset) -> dict:
```

aries_cloudagent/protocols/out_of_band/v1_0/messages/problem_report.py
```diff
@@ -64,6 +66,13 @@ class Meta:
         model_class = OOBProblemReport
         unknown = EXCLUDE
 
+    _type = fields.Str(
```
andrewwhitehead, line 69: Should this be part of the base AgentMessageSchema?

shaangill025: `_type` is already part of AgentMessageSchema but that has `dump_only=True` which gets removed in `BaseModel  -> skip_dump_only` when deserializing. So I am overwriting `_type` only for these OOB messages. I updated `BaseModel -> make_model` to adjust for the above. The reason for these updates is to achieve backward compatibility, earlier if we received `1.0` message type and version found in `message_types` was `1.1` then the model `_type` was always `1.1`.

aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py
```diff
@@ -143,20 +143,17 @@ def __init__(
         )
         self.requests_attach = list(requests_attach) if requests_attach else []
         self.services = services
-        self.assign_version_to_message_type(version=version)
+        if msg_type:
+            self._type = msg_type
+        else:
+            self._type = self.get_updated_msg_type(version)
```
andrewwhitehead, line 149: Maybe the AgentMessage constructor can accept a couple more options to avoid reassigning the message type and repeating this code. For example:

```super().__init__(_message_type=msg_type, _message_version=version, **kwargs)```

If `_message_type` is null then it defaults to `Meta.message_type`. If `_message_version` is non-null then it performs the substitution.

File: aries_cloudagent/protocols/out_of_band/v1_0/messages/problem_report.py (hunk 4 above)
Comment (andrewwhitehead) [reply aries_cloudagent/protocols/out_of_band/v1_0/messages/problem_report.py:69]: Ah okay, that makes sense.

File: aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py (hunk 5 above)
Comment (andrewwhitehead) [reply aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py:149]: (Or just `_type` and `_version`)

Comment (shaangill025) [reply aries_cloudagent/protocols/out_of_band/v1_0/messages/invitation.py:149]: Something like below in `AgentMessage -> __init__`:
```
if _type:
    self._message_type = _type
elif _version:
    self._message_type = self.get_updated_msg_type(_version)
else:
    self._message_type = self.Meta.message_type
```

andrewwhitehead: Yes I think that would work

andrewwhitehead (approved):

PR #1977, opened 2022-10-15
fix: public did mediator routing keys as did keys

dbluhm:
This PR fixes an issue @reflectivedevelopment found while using the recent changes made to support routing keys associated with public DID endpoints.

Without this fix, when creating a public DID OOB invitation, creation will fail due to no DIDComm service being found. This stemmed from PyDID expecting that routing keys be expressed strictly as DID URLs. Previously, we were resolving the routing keys as raw base58 encoded public keys. This PR makes it so that routing keys are resolved as did:key url values (if already did:key, simply pass through the values but if not, transform into did:key urls).

Credit to @cjhowland for doing most of the digging on this issue :slightly_smiling_face:

### DIFF
--- a/.github/workflows/blackformat.yml
+++ b/.github/workflows/blackformat.yml
@@ -10,7 +10,9 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
       - name: Black Code Formatter Check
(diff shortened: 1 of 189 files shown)


dbluhm:
@cjhowland and I need to add unit tests to this PR ensuring we've got coverage of the added lines.

We also need to ensure that if the resolved routing key value is a did:key but not a did:key URL that we perform an appropriate transformation.

ianco (dismissed):

andrewwhitehead:
Is this finished testing?

dbluhm:
@andrewwhitehead yes, this is ready for review

ianco (approved):

PR #1978, opened 2022-10-17
chore: fix ACAPY_PROMOTE-AUTHOR-DID flag 

### LINKED ISSUE #1965: Env_var ACAPY_PROMOTE-AUTHOR-DID is invalid with hyphen
Current behaviour:

The argparse `aries_cloudagent/config/argparse.py` defines the ACAPY_PROMOTE-AUTHOR-DID as an env var in ll. 1791. That is invalid when using an env file parsed to a container and the containers fails to start up entirely.

Suggested change:

Rename `ACAPY_PROMOTE-AUTHOR-DID` to `ACAPY_PROMOTE_AUTHOR_DID`

Expected behaviour:
`ACAPY_PROMOTE_AUTHOR_DID` can be set in an env file and `ACAPY_PROMOTE_AUTHOR_DID` get correctly picked up by aca-py.

morrieinmaas:
using underscore instead of hyphen

closes #1965 

Signed-off-by: Moriarty <moritz@animo.id>

### DIFF
--- a/Endorser.md
+++ b/Endorser.md
@@ -56,7 +56,7 @@ Endorsement:
                         For Authors, specify whether to automatically create transactions for a cred def's revocation registry. (If not specified, the controller must invoke the endpoints required to create
                         the revocation registry and assign to the cred def.) [env var: ACAPY_CREATE_REVOCATION_TRANSACTIONS]
   --auto-promote-author-did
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
It looks like maybe it should be `ACAPY_AUTO_PROMOTE_AUTHOR_DID`?

morrieinmaas:
> It looks like maybe it should be `ACAPY_AUTO_PROMOTE_AUTHOR_DID`?

Yeah, thx. That sounds more than reasonable.

PR #1986, opened 2022-10-25
did method & key type registry

burdettadam:
This PR contains did method and key type registry, allowing plugins to register and use new methods and key types.


swcurran:
Ooopppsss...so much for @andrewwhitehead 's requested review. Merge is done...

PR #1994, opened 2022-10-27
Fix/txn job setting

ianco:
Fixes a timing issue where the endorser and author are connecting and trying to set connection met-data before the connection is "active".


swcurran (approved):
LGTM -- removing code, nice!

PR #2009, opened 2022-11-07
Fix for mediator load testing race condition when scaling horizontally

ianco:
Pause and retry if a routing record is not found

### DIFF
--- a/aries_cloudagent/protocols/routing/v1_0/manager.py
+++ b/aries_cloudagent/protocols/routing/v1_0/manager.py
@@ -81,7 +81,7 @@ class RoutingManager:
                     raise RouteNotFoundError(
                         f"No route found with recipient key: {recip_verkey}"
                     )
-                await asyncio.wait(500)
+                await asyncio.sleep(0.5)
                 pause = False
 
     async def get_routes(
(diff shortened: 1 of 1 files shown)


shaangill025:
@ianco We can also use the Redis cache across scaled-up mediators to track if `RoutingRecord` for a recip_key is being created? Mediator_A before committing a new `RoutingRecord` to wallet adds `routing_record::{recip_key}` to cache. When mediator_B calls `retrieve_by_recipient_key` and `StorageNotFoundError` gets raised then we check the cache for  `routing_record::{recip_key}` and retry only if it exists.

ianco:
> @ianco We can also use the Redis cache across scaled-up mediators to track if `RoutingRecord` for a recip_key is being created? Mediator_A before committing a new `RoutingRecord` to wallet adds `routing_record::{recip_key}` to cache. When mediator_B calls `retrieve_by_recipient_key` and `StorageNotFoundError` gets raised then we check the cache for `routing_record::{recip_key}` and retry only if it exists.

@shaangill025 I think using the redis (or kafka) queue is a better approach than the hack-y solution I put in.  However unless we make queues a requirement (as @swcurran suggested) we also have to have a fall-back solution for when there are no queues deployed.  (I think this problem can occur even if there is just a single instance of the mediator, although much less likely.)

For a redis/kafka solution, I think all that aca-py need to do - if there is no routing record found - is just leave the message on the queue with a "retry later" flag, I don't think there's a reason to do any additional checking.  (Plus I'm not sure of the timing - is it possible that Mediator_B will try to find the routing record before Mediator_A even gets the request to add the routing record?  I'm not sure of the root cause of this problem so we would need to assume this is a possibility.)

swcurran (approved):
Confirmed and we agree this should be merged. A subsequent update for adding queuing support in this area may be needed, but that can follow.

PR #2013, opened 2022-11-09
fix claim format designation in presentation submission

### LINKED ISSUE #1819: Wrong claim format designation in presentation submission
When issuing a presentation of a W3C verifiable credential, the `"format"` property in the _Input Descriptor Mapping Object_ of the presentation submission is currently being set to `"ldp_vp"`. It must be set to `"ldp_vc"` instead, as it refers to the format of the enclosed credential. (See https://identity.foundation/presentation-exchange/spec/v1.0.0/#presentation-submission)

The parts that need to be changed to produce correct presentation submissions are here:
https://github.com/hyperledger/aries-cloudagent-python/blob/4240fa9b192ea4cdb4026211ea4bec694aec5506/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py#L1357-L1361

and here (maybe just remove the default entirely?):
https://github.com/hyperledger/aries-cloudagent-python/blob/4240fa9b192ea4cdb4026211ea4bec694aec5506/aries_cloudagent/protocols/present_proof/dif/pres_exch.py#L835-L840

However, verification logic might also need some adaptations when this parameter is changed, I haven't looked into that yet.

rmnre:
fixes #1819 

Signed-off-by: Roman Reinert <roman.reinert@gematik.de>

### DIFF
--- a/aries_cloudagent/protocols/present_proof/dif/pres_exch.py
+++ b/aries_cloudagent/protocols/present_proof/dif/pres_exch.py
@@ -837,7 +837,7 @@ class InputDescriptorMappingSchema(BaseModelSchema):
     fmt = fields.Str(
         description="Format",
         required=False,
-        default="ldp_vp",
+        default="ldp_vc",
         data_key="format",
     )
     path = fields.Str(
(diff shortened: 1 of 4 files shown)


TimoGlastra (approved):
LGTM! Thanks @rmnre.

@shaangill025 can you also take a look?

shaangill025 (approved):

PR #2024, opened 2022-11-23
Add ability to set docker container name

matrixik:
I'm using setting names when running more containers locally so they can easily connect to each other with network names (like `http://aries_holder:8000/`).

### DIFF
--- a/scripts/run_docker
+++ b/scripts/run_docker
@@ -60,8 +60,8 @@ echo ""
 echo "Container name: ${ACAPY_CONTAINER_NAME}"
 
 ACAPY_NETWORK_NAME="${NETWORK_NAME}"
-# Create new Docker network if it does not exist.
 if [ -n "${NETWORK_NAME}" ]; then
+  # Create new Docker network if it does not exist.
   if [ -z "$("$CONTAINER_RUNTIME" network ls --filter name=^"${NETWORK_NAME}"$ --format="{{ .Name }}")" ]; then
     echo "Creating new Docker network: ${ACAPY_NETWORK_NAME}"
     "$CONTAINER_RUNTIME" network create "${NETWORK_NAME}"
--- a/scripts/run_docker
+++ b/scripts/run_docker
@@ -27,13 +27,13 @@ for arg in "$@"; do
     echo "Backing up database before running aca-py upgrade is highly recommended. Do you wish to proceed"
     select yn in "Yes" "No"; do
       case $yn in
-      Yes) break ;;
-      No) exit ;;
+        Yes) break ;;
+        No) exit ;;
       esac
     done
   fi
 done
-ACAPY_NETWORK_NAME="${NETWORK_NAME}"
+
 if [ -n "${ENABLE_PTVSD}" ]; then
   ARGS="${ARGS} -e ENABLE_PTVSD=\"${ENABLE_PTVSD}\" -p $PTVSD_PORT:$PTVSD_PORT"
 fi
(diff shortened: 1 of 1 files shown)


matrixik:
@swcurran @ianco

matrixik:
Would be nice if this could be also included in 0.7.6 (if there will be one).

scripts/run_docker
```diff
@@ -48,13 +48,28 @@ if [ "$OSTYPE" == "msys" ]; then
...
+  RAND_NAME=$(env LC_ALL=C tr -dc 'a-zA-Z0-9' </dev/urandom | fold -w 16 | head -n 1)
+  ACAPY_CONTAINER_NAME="aries-cloudagent-runner_${RAND_NAME}"
+  ARGS="${ARGS} --name ${ACAPY_CONTAINER_NAME}"
+fi
+echo ""
+echo "Container name: ${ACAPY_CONTAINER_NAME}"
+
+ACAPY_NETWORK_NAME="${NETWORK_NAME}"
+# Create new Docker network if it does not exist.
+if [ -n "${NETWORK_NAME}" ] && [ -z "$(docker network ls --filter name=^"${NETWORK_NAME}"$ --format="{{ .Name }}")" ]; then
```
andrewwhitehead, line 64: I believe this should use `$CONTAINER_RUNTIME` instead of `docker`. It might be cleaner to have one outer block for `[ -n "${NETWORK_NAME}" ]` and check for the existence of the network inside.

matrixik: Thank you for pointing this. Fixed.

swcurran:
Error in the linting, but I suspect it is unrelated to this PR.

@WadeBarnes -- can you take a look at this PR?

scripts/run_docker
```diff
@@ -48,13 +48,30 @@ if [ "$OSTYPE" == "msys" ]; then
   CONTAINER_RUNTIME="winpty docker"
 fi
 
-RAND_NAME=$(env LC_ALL=C tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 16 | head -n 1)
-if [ -z "$NETWORK_NAME" ]; then
-  echo "No Docker network specified."
-  $CONTAINER_RUNTIME run --rm -ti --name "aries-cloudagent-runner_${RAND_NAME}" \
-    $ARGS aries-cloudagent-run "$@"
+ACAPY_CONTAINER_NAME="${CONTAINER_NAME}"
```
WadeBarnes, line 51: I find this assignment and then the following mixed use of the two variables confusing.

scripts/run_docker
```diff
@@ -48,13 +48,30 @@ if [ "$OSTYPE" == "msys" ]; then
...
+  ARGS="${ARGS} --name ${CONTAINER_NAME}"
+else
+  RAND_NAME=$(env LC_ALL=C tr -dc 'a-zA-Z0-9' </dev/urandom | fold -w 16 | head -n 1)
+  ACAPY_CONTAINER_NAME="aries-cloudagent-runner_${RAND_NAME}"
+  ARGS="${ARGS} --name ${ACAPY_CONTAINER_NAME}"
+fi
+echo ""
+echo "Container name: ${ACAPY_CONTAINER_NAME}"
+
+ACAPY_NETWORK_NAME="${NETWORK_NAME}"
```
WadeBarnes, line 62: Similarly, I find this assignment and then the following mixed use of the two variables confusing.

WadeBarnes (changes requested):

scripts/run_docker
matrixik: I was just following previous code but you are right, removed.

scripts/run_docker
matrixik: Removed.

WadeBarnes:
The linting errors have nothing to do with the one file updated in this PR.

WadeBarnes:
CircleCI builds today, including this one, have been failing with the following message.
image

WadeBarnes:
The recommendation is to get off CircleCI and move to pure GHA.

cc @swcurran, @andrewwhitehead

andrewwhitehead (approved):

swcurran:
@ryjones -- can you merge this?  I don't have the power given the state, but it has been approved, and the failures are not relevant.

Thanks.

PR #2058, opened 2022-12-22
ci: add gha for pr-tests

dbluhm:
This PR adds GitHub Actions for testing PRs. This is pulled from work on #1888. These tests are ready for easy use with other python versions, when we get to that point.


dbluhm:
Image layers for the Indy Tests are cached. The first run and when dependencies change will be the slowest but subsequent runs will be faster.

Issue #2068, opened 2023-01-06
Allow HEAD requests to default route


johnekent:
https://github.com/hyperledger/aries-cloudagent-python/blob/6f2ef55d2dbd22d7313f3c346180008f2ba93188/aries_cloudagent/admin/server.py#L436

The current default route does not allow_head.  External systems can use the HEAD method to check the health of this port/route successfully because these external systems get a response, but the agent logs an HTTPMethodNotAllowed error and stack trace with each check.  In some cases, the external system's check mechanism can't be changed.

Is there a reason why the default route does not have allow_head=True?  If this change could be made and doesn't impose any risk or concern to ACA-py, then the same check functionality could be met while having cleaner server logs for easier monitoring and troubleshooting.

Steps to reproduce error:
1. start cloud agent
2. go to admin port in browser, and get a redirect to swagger
3. run curl (-I, --head, etc.) with head method request to admin port
4. see error in cloud agent log

Same steps with allow_head=True in line noted above.
1-3 the same as above.
4. no errors in the agent log

swcurran:
For those wondering, as I was, what `HEAD` is, here is a description: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

Who knew? 

@WadeBarnes -- any reason not to do this?

@johnekent -- would you do a PR?  If so, make sure you use DCO: DCO - Developer Certificate of Origin - https://github.com/apps/dco (e.g. use `commit -s ...`).

ianco:
Yes we should allow this, probably just on specific endpoints

johnekent:
Thanks for the review of this issue and for the opportunity to modify.  I will pursue a PR following the contributions guide, likely next week.

WadeBarnes:
@johnekent, Just to confirm, did you know about the k8s compatible health check endpoints that were already available on the aca-py instances?  They provide some additional internal checks to ensure the health and readiness of an aca-py instance that a HEAD query would not provide.  They are `/status/live` and `/status/ready`.

johnekent:
Thanks for ensuring this awareness @WadeBarnes .  I am aware, and this was nicely described on discord.  The issue I'm facing is a lack of control over the health checker, which is constrained by the issue noted.  These more robust internal checks should be preferred when allowed by the checker.

johnekent:
Hi @swcurran  @ianco @WadeBarnes .  PR is https://github.com/hyperledger/aries-cloudagent-python/pull/2077
[...]

PR #2077, opened 2023-01-13
Issue #2068 boolean flag change

johnekent:


johnekent:
Updated the pre-commit as well, since flake8 location changed.  That's why multiple commits.

ianco (changes requested):
This returns `401 Unauthorized` is an admin api key is enabled, which is probably not what you want.

E.g aca-py as `./bin/aca-py start --endpoint 8888 --no-ledger -it http 0.0.0.0 8887 -ot http --admin 0.0.0.0 8889 --admin-api-key 666` and then in another shell `curl -I http://localhost:8889/`

There are a few url's that are unprotected (even when there is an api key), for example you can still `curl http://localhost:8889/api/doc` (as well as the status url's that Wade mentioned)

johnekent:
@ianco - This change was not initially intended to impact security behavior, but you make a good point.  With allow_head = True or False, the server returns a 401 error with admin api key enabled.  With the api key enabled the development proxy would not work which is currently OK, but this may have gaps in the future.
This feature only currently considers HEAD calls with admin-insecure-mode, so that the default route returns a 302 instead of a 405 with the corresponding error in the server log.  In my development proxy situation, I can't configure it to use any routes other than the default.
The security risk of adding this default route to the unprotected set isn't clear to me.

Do you think it would be better to first handle the initial use case (insecure mode) now, and then consider the admin api key scenario in the future?  Or, should more thought be given to adding the default route to the unprotected set (or similar)?  If you provide direction on where in the code base the unprotected set is configured, I can also try to see how that works across the use case variations.

WadeBarnes:
I don't think it would be advisable to add the default route to the set of unprotected routes.

@johnekent, Are you able to provide more details on why your setup requires `allow_head = True`?  Perhaps there is another way to accomplish the same thing.  Would there be an issue in you're scenario if the admin api key were enabled?

ianco:
@johnekent the list of unprotected paths is defined here:  https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/admin/server.py#L300

These include the `/api/doc` endpoint (so you can connect to the swagger page and authenticate) and the status endpoints that we use for our deployments.  (All of our deployments are secured by an api key, but we don't require the api key access the status endpoints.)

It's recommended to enable the api key for all deployments (other than local dev).  Could your do a HEAD query against one of our existing status endpoints?

johnekent:
@WadeBarnes I am using the Coder development environment which provides a [Dev URLs](url) feature for ports (admin, endpoints, etc.) to be available as a reverse proxy - e.g. for accessing the swagger docs from a browser, or for a controller UI to connect with an agent.  The majority of work will be on developing controllers, which shouldn't require admin api keys.  I.e., this is for controller development, not for server administration/configuration.

@ianco the setup described above would be considered local dev, and would only be used for those purposes.  The dev urls configuration can't specify any endpoints, only server and port.

If this feature can be merged, then Coder can be used for controller development.  That seems like a benefit to the aries ecosystem that doesn't cause any harm and also doesn't change the current security.  However, I am open to exploring alternatives.

ianco:
OK, I'm ok with this as long as @WadeBarnes is ok with it lol

It won't cause any hard because the HEAD method can't be used if the admin api key is enabled, so it's only available in a non-secure environment.

(Btw the admin api key is definitely intended for use with controllers, it's all that's protecting aca-py's admin api from malicious use.  If you deploy aca-py into a prod environment with no api key and someone gains access to the admin url they can issue themselves credentials under your identity.)

johnekent:
@ianco and @WadeBarnes what about this addition - adding HEAD to the existing OPTIONS check.  This allows both use cases while extending the existing security for method exceptions.

WadeBarnes:
Allowing `HEAD` requests to bypass security when the admin api key is enabled would falsely indicate the endpoints were available when in fact they would not respond to any other request.  It also opens the door for the API to be more easily scanned when it's supposed to be secured.

Allowing `HEAD` under the regular security constraints would be fine.  Meaning `HEAD` requests could be used in insecure mode, or secure mode provided the api key was also supplied in the request.

aries_cloudagent/admin/server.py
```diff
@@ -323,7 +323,7 @@ async def check_token(request: web.Request, handler):
                 if (
                     valid_key
                     or is_unprotected_path(request.path)
-                    or (request.method == "OPTIONS")
+                    or (request.method in ["OPTIONS","HEAD"])
```
WadeBarnes, line 326: Please revert this change.

WadeBarnes (changes requested):

aries_cloudagent/admin/server.py
johnekent: Reverted.  Put it for review because it was another approach, similar to OPTIONS.  Thanks for reviewing.

WadeBarnes:
@johnekent, Looks good to me.  Could you please squash and rebase your changes into a single commit and add DCO sign-off please?

johnekent:
Appreciate the patience and feedback here @WadeBarnes and @ianco .  Squashed back to what's ultimately the initial commit, and I believe the DCO's on that one.  Will check back after the scans.  Thanks.

WadeBarnes (approved):

WadeBarnes:
@ianco, ready for your review and approval.

ianco (approved):

PR #2089, opened 2023-01-24
Multitenancy demo (docker-compose with postgres and ngrok)

ianco:


demo/multi-demo/README.md
```diff
@@ -0,0 +1,40 @@
...
+docker-compose stop
+docker-compose rm -f
+```
+
+This will leave the agent's wallet data, so if you restart the agent it will maintain any created data.
+
+- to remove the agent's wallet:
+
+```bash
+docker volume rm docker-agent_wallet-db-data
```
shaangill025, line 39: I think this should be `multi-demo_wallet-db-data`

ianco: Good catch

PR #2095, opened 2023-01-26
feat: universal resolver - configurable authentication

chumbert:
Context: we want to allow aca-py instances to connect to privately run universal resolvers such as the one from godiddy. This requires authentication.

* Add `--universal-resolver-bearer-token` to allow authentication against privately run universal resolvers.

BREAKING CHANGE:
Configured url should now include the `/1.0` suffix. The default url reflects that changes.
* This is to have a more flexible configuration of the universal resolver's api route (required in the case of godiddy).
* The alternative would be to make the suffix configurable too. I just found it a bit fiddly both in code and configuration. Let me know if the breaking change in configuration is not acceptable.

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -643,6 +643,14 @@ class GeneralGroup(ArgumentGroup):
                 "resolver instance."
             ),
         )
+        parser.add_argument(
+            "--universal-resolver-bearer-token",
+            type=str,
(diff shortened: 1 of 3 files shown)


dbluhm (approved):

dbluhm:
Requesting a sanity check before I hit merge :slightly_smiling_face: @ianco

ianco (approved):

PR #2097, opened 2023-01-26
Respect `auto-verify-presentation` flag in present proof v1 and v2

dbluhm:
When using the debug flags for auto responding to presentation messages, it's possible for the auto-verify-presentation flag not to properly take effect if starting from a presentation proposal and not taking any action in the Admin API (each step is advanced to automatically by the debug flags). This PR corrects this so that a presentation exchange can be executed from start to finish without Admin API input.

Also, this PR fixes reply behavior on the presentation ack message. The BaseResponder (a DispatcherResponder instance) is passed into the request handler but not passed to the presentation manager methods that resulted in a presentation ack message. Instead, the manager was retrieving the responder from the context, which would be an AdminResponder instance that lacks the context of the inbound message. This causes return routes to not be respected. This PR corrects this by passing the DispatcherResponder to the manager methods when available. Otherwise, the AdminResponder will be used as before.


dbluhm:
I didn't get the overridde semantics quite right on this yet. I'll fix that...

ianco (approved):

swcurran:
Let us know when the fixes are made @dbluhm .  Note the integration test error(s).

PR #2116, opened 2023-02-11
Fix multi-use invitation performance

reflectivedevelopment:
Multiuse invitation performance degrades over time. By adding state into the tag names, performance doesn't degrade over time while using multi-use invitations.

This breaks backwards compatibility, but while load testing it can be seen that performance no longer degrades when using a multi-use invitation. Previously, only 1600 connections could be established while load testing, and on a second run is reduced to 300. After the modification, the load testing exceeded 2400 connections and the testing client machine ran out of ram for clients.

This shows a clear improvement in performance.


ianco:
We can add these new tags on startup.  I don't recall the details but @shaangill025 had to do this at one point

ianco:
> @ianco are you referring to the upgrade command or is there another mechanism for adding the new tags?

Yes the upgrade command, that's right.

reflectivedevelopment:
Add upgrade command to upgrade connections.

ianco:
@reflectivedevelopment  this looks good!  I'm just doing some testing ...  can you sync up with the latest main branch?

ianco (approved):

ianco:
Added "UPGRADE" to the PR title so we remember to add instructions re upgrading the db to the release docs.

swcurran:
@ianco or @reflectivedevelopment — can one of you add the details of what has to be done to upgrade that I can put into the ChangeLog.  Will it happen automagically, or does something have to be done by the deployer to trigger the upgrade?  I’m not clear on exactly what is happening — there isn’t a lot of code changing in the PR :-)

ianco:
I think you just need to run `aca-py upgrade` but @reflectivedevelopment can confirm what are the details

reflectivedevelopment:
It is the aca-py upgrade command that needs to be run.

swcurran:
Shouldn’t the `—from-version` be something earlier than v0.8.0, since that is the version that has this fix?

reflectivedevelopment:
@swcurran this seems okay.

swcurran:
Awesome — thanks.

PR #2130, opened 2023-02-17
Update ACA-Py docker files to produce OpenShift compatible images

WadeBarnes:
- and separate aca-py and indy version labels.

The Python 3.9 versions of both the Standard and Indy variants of the images have been smoke tested on OpenShift to ensure basic functionality.  The Indy variant was smoke tested with both Indy and Askar wallets, and the Standard variant was smoke tested with an Askar wallet.

The images, version `0.8.0-rc0`, can be previewed here; https://github.com/WadeBarnes/aries-cloudagent-python/pkgs/container/aries-cloudagent-python

### DIFF
--- a/docker/Dockerfile.indy
+++ b/docker/Dockerfile.indy
@@ -104,7 +104,7 @@ LABEL summary="$SUMMARY" \
     io.k8s.description="$DESCRIPTION" \
     io.k8s.display-name="indy-python $indy_version" \
     name="indy-python" \
-    version="$indy_version" \
+    indy-sdk.version="$indy_version" \
     maintainer=""
 
 # Add indy user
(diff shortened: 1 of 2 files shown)


WadeBarnes:
cc @jleach, @wadeking98

swcurran (approved):

docker/Dockerfile.indy
```diff
@@ -243,26 +242,22 @@ LABEL summary="$SUMMARY" \
...
-    rm aries_cloudagent*.whl
+    rm aries_cloudagent*.whl && \
+    chmod +rx $(python -m site --user-site)
 
 # Clean-up unneccessary build dependencies and reduce final image size
-# RUN apt-get purge -y --auto-remove build-essential
+USER root
+RUN apt-get purge -y --auto-remove build-essential
+
+USER $user
```
dbluhm, line 261: This being right at the end of the dockerfile stands out to me as not being ideal. Not that it's perfectly optimized anyways but these lines will basically always run. Is it possible to rearrange or produce another stage that can omit the build essentials from the start?

dbluhm (commented):
Small comment, otherwise looks good to me

docker/Dockerfile.indy
WadeBarnes: Depending on the availability of binary packages for the given platform `build-essential` is required during the `aca-py` install to build and install some of the dependencies.  It can only be removed after the install is complete.

WadeBarnes: @dbluhm, does that make sense?

dbluhm: Ah, I see. I retain an itch that there might be a better way but I don't think it's worth holding this PR up; sorry for the slow response.

dbluhm (approved):

PR #2135, opened 2023-02-21
feat: add verification method issue-credentials-2.0/send endpoint

chumbert:
Accommodate credential sending for did methods that are not `did:sov` or `did:key` by adding an optional `verification_method` parameter to the send endpoint options.

Currently, verification method is derived "automatically" for did:sov and did:key, raising error for any other DID method.

The parameter is optional and stored along the credential exchange for recall when issuing the credential.


dbluhm (approved):

dbluhm:
I think this is an important PR as we increase support for other DID methods but I think some more eyes on this change would be good. @ianco @swcurran

ianco (approved):

PR #2144, opened 2023-02-24
feat: allow marking non-SOV DIDs as public

chumbert:
In the continuity of the work with non-indy DIDs: remove the restriction for did:sov when marking a DID as public **at the wallet level**.

These wallet implementations don't include any indy-related logic. All the logic and restrictions are done in the `wallet/set-public-did` route.

This change allows using the wallet to mark DIDs from other methods as being public as opposed to having to re-implement all the logic in an ad-hoc plugin.

### DIFF
--- a/demo/AliceWantsAJsonCredential.md
+++ b/demo/AliceWantsAJsonCredential.md
@@ -376,7 +376,7 @@ To request a proof, submit the following (with appropriate `connection_id`) to F
                                     "directive": "required",
                                     "field_id": [
                                         "1f44d55f-f161-4938-a659-f8026467f126"
-                                    ],
+                                    ]
                                 }
                             ],
                             "fields": [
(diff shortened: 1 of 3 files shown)


ianco (approved):

swcurran:
@chumbert — can you please update the branch so we can merge it?  Thanks!

chumbert:
Can someone with the permissions restart the integration tests ? It looks like the failing test is flaky and not related to these changes.

ryjones:
@swcurran [if you go here](https://github.com/hyperledger/aries-cloudagent-python/actions?query=is%3Afailure), do you see the failed tests? If [you go here](https://github.com/hyperledger/aries-cloudagent-python/actions/runs/4295004886), for example, do you see the "rerun jobs" button at the top?
Screenshot 2023-03-01 at 07 59 10

swcurran:
Thanks @ryjones — I was able to find that button. Cool stuff - didn’t know it was possible.

That will be a big a help.  Of course if we made the tests a little more stable, this wouldn’t happen… :-)

PR #2151, opened 2023-02-28
Fix: Performance Demo [no --revocation]

shaangill025:
- Fixes issue with performance demo [no revocation], where number of issued credentials stops incrementing when received credentials reaches 100% and the demo would eventually timeout.


ianco (approved):
Bueno!!!

Issue #2152, opened 2023-03-02
Nightly tests on main fork only?


dbluhm:
The nightly test github actions we recently merged run on every fork of ACA-Py (unless the owner disables actions on their fork, I suppose). Perhaps it's a good idea to have a check that has them run only if on the main hyperledger fork of ACA-Py (or if some config value is set in the environment on the repo? Opting in to have them run on your fork seems valuable, even if only for testing the action).

cc @WadeBarnes @swcurran

WadeBarnes:
@dbluhm, 100% agree.

swcurran:
Definitely!

dbluhm:
@anwalker293 will take a stab at this :slightly_smiling_face:

PR #2166, opened 2023-03-14
fix: run only on main, forks ok

anwalker293:
Addresses issue #2152 (https://indicio.atlassian.net/browse/ACAP-1004)

I did not include an environment variable, as GitHub seems to throw up its hands when you put it in the workflow file. There seems to be a workaround if you put the variable in .env, but I'm not sure if we want that though...

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -3,9 +3,6 @@ name: Nightly Tests
 on:
   schedule:
     - cron: '0 0 * * *'
-  push:
-    branches:
-      - main
 
 jobs:
   tests:
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   tests:
-    if: github.repository == 'hyperledger/aries-cloudagent-python' || (github.event.inputs.run_flows)
+    if: github.repository == 'hyperledger/aries-cloudagent-python' || github.event_name == 'workflow_dispatch'
     name: Tests
     strategy:
       fail-fast: false
@@ -28,7 +28,7 @@ jobs:
       os: ${{ matrix.os }}
 
   tests-indy:
-    if: github.repository == 'hyperledger/aries-cloudagent-python' || (github.event.inputs.run_flows)
+    if: github.repository == 'hyperledger/aries-cloudagent-python' || github.event_name == 'workflow_dispatch'
     name: Tests (Indy)
     strategy:
       fail-fast: false
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   tests:
-    if: github.repository == 'hyperledger/aries-cloudagent-python' || (github.event.inputs.run_flows)
+    if: ${{github.event.repository.full_name}} == 'hyperledger/aries-cloudagent-python' || (github.event.inputs.run_flows)
     name: Tests
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


anwalker293:
Re: edit: fix: on schedule - Sorry, I put in for the flows to run on push of the main branch for testing purposes. Just removed this now, so this flow should only run nightly

.github/workflows/nightly-tests.yml
```diff
@@ -3,10 +3,10 @@ name: Nightly Tests
 on:
   schedule:
     - cron: '0 0 * * *'
-  workflow_dispatch:
 
 jobs:
   tests:
+    if: github.event.ref == 'refs/heads/main'
```
dbluhm, line 9: This will cause the workflow to run main but will not limit it to the hyperledger fork. See the int test worflow as an example: https://github.com/hyperledger/aries-cloudagent-python/blob/main/.github/workflows/integrationtests.yml#L13

.github/workflows/nightly-tests.yml
```diff
@@ -3,10 +3,10 @@ name: Nightly Tests
 on:
   schedule:
     - cron: '0 0 * * *'
-  workflow_dispatch:
```
dbluhm, line 6: I think we want to retain the workflow dispatch, which enables other forks to run these tests on demand (assuming the conditions are set up similar to the integration tests conditions)

.github/workflows/nightly-tests.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       os: ${{ matrix.os }}
 
   tests-indy:
+    if: github.event.ref == 'refs/heads/main'
```
dbluhm, line 25: Same story here :)

dbluhm (changes requested):

dbluhm (approved):

WadeBarnes (commented):
I've typically used `github.event.repository.full_name` to get the repository name.  I believe you also need to wrap the variable names in `${{ }}`.

Examples:
- https://github.com/hyperledger/indy-node/blob/main/.github/workflows/build.yaml#L63
- https://github.com/bcgov/von-network/blob/main/.github/workflows/build-docker.yml#L16

anwalker293:
@WadeBarnes I wanted to follow up with you regarding these changes. Thank you for your suggestion, it's greatly appreciated! It looks like this issue is more optional than strictly a problem (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif). For the time being, I did keep the original format, but, if desired, I can change the format
However, I do greatly appreciate any advice, so please let me know if there's anything else you would like me to change! 
Cheerio!

WadeBarnes:
@anwalker293, Have you run any tests using this new workflow on your branch?  For example tried to trigger it as is on your branch to make sure it behaves as expected and then changed the github.repository to match yours to ensure it triggers as expected?  Also have you tested the workflow dispatch trigger?

We can see from here, https://github.com/hyperledger/aries-cloudagent-python/actions/runs/4630877399, it's running, but is it behaving as intended on another repo?

anwalker293:
@WadeBarnes I believe so! :) I tested it on my fork of this repo, on the branch `test` (https://github.com/anwalker293/aries-cloudagent-python/actions/runs/4691869092)

WadeBarnes:
> @WadeBarnes I believe so! :) I tested it on my fork of this repo, on the branch `test` (https://github.com/anwalker293/aries-cloudagent-python/actions/runs/4691869092)

That's a different workflow, it's the [codeql.yml](https://github.com/hyperledger/aries-cloudagent-python/blob/main/.github/workflows/codeql.yml) workflow.  You're PR is for the `nightly-tests.yml` workflow.  I don't see any runs associated to it in your repo.

WadeBarnes (approved):

PR #2169, opened 2023-03-14
0.8.0 release

swcurran:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ aries-cloudagent-python].
 [publish-indy.yml]: https://github.com/hyperledger/aries-cloudagent-python/blob/main/.github/workflows/publish-indy.yml
 [Container Images and Github Actions]: https://github.com/hyperledger/aries-cloudagent-python/blob/main/ContainerImagesAndGithubActions.md
 
-## Breaking Changes
+## Breaking Changes and Upgrades
 
 ### PR [\#2034](https://github.com/hyperledger/aries-cloudagent-python/pull/2034) -- Implicit connections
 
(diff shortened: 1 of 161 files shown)


WadeBarnes (dismissed):

swcurran:
@ianco — can you help with the PR Tests Failing?  Error message is:

`FAILED aries_cloudagent/commands/tests/test_upgrade.py::TestUpgrade::test_upgrade_missing_from_version`

Do I need to add something specific about the upgrade and the tag?

swcurran:
Add the integration tests failed — @andrewwhitehead — perhaps this is related to the new CredX release? It’s in the credX area.  Or sunspots?

swcurran:
Ah — integration tests have now passed. Nevermind.

andrewwhitehead:
I think the failing test may require a small update to `aries_cloudagent/commands/default_version_upgrade_config.yml` to add v0.8.0

swcurran:
Thanks @andrewwhitehead — I’ll add that.

swcurran:
v0.8.0 is already in the file:

```
v0.8.0:
  resave_records:
    base_record_path:
      - "aries_cloudagent.connections.models.conn_record.ConnRecord"
  update_existing_records: false
  .
  .
  .
```

swcurran:
Need to add a note about the upgrade. But do integration tests also need to be updated?

swcurran:
@andrewwhitehead -- please investigate the failed integration test and let us know what you find. I think it is important to get this out there.

ianco:
I think the integration tests are ok

swcurran:
What is the semantics of the `aries_cloudagent/commands/default_version_upgrade_config.yml`.  

- I think @reflectivedevelopment was trying to say — “When you update to 0.8.0, this upgrade process should be run”. 
- However, if the integration process is correct, it’s saying — “There is an upgrade process for 0.8.0, but you are already on 0.8.0”. 
- How do we convey in the file or in the integration test: “I’m already on 0.8.0, so no upgrade is needed”?

Also in the PR #2116, there is a description of an upgrade to be run. I’m going to try to add it to the CHANGELOG.md, but I’m not clear on exactly what someone deploying this version of ACA-Py needs to do.  Is it automatic?

ianco:
I think you just need to run `./aca-py upgrade` (or whatever your path to aca-py is).  What the upgrade actually does is just update all the `connection` tags.  It looks like the upgrade process itself (and the unit test) weren't updated, just the config file.  And the stuff the upgrade process does is the same as was done to upgrade to version `0.7.2`.

I'm not sure why that unit test is failing, as @swcurran mentions if the version is already `0.8.0` (and you're upgrading to `0.8.0`) then the upgrade should just exit with a message, not an exception.

swcurran:
So this [RaiseError call](https://github.com/hyperledger/aries-cloudagent-python/blob/5445a3eedcd3db4d1a3a9bf92abe8152a0624538/aries_cloudagent/commands/upgrade.py#L140) should be changed to a `print` -- correct?

andrewwhitehead (approved):

swcurran:
@WadeBarnes — not urgent — I got feedback from @reflectivedevelopment so all is well.  I’ll merge after my next meeting and get the release out the door.

PR #2174, opened 2023-03-21
WIP: Peer DID Method 1

shaangill025:


aries_cloudagent/connections/base_manager.py
```diff
@@ -61,6 +65,99 @@ def __init__(self, profile: Profile):
         self._profile = profile
         self._route_manager = profile.inject(RouteManager)
 
+    async def create_did_peer_document(
```
swcurran, line 68: Is this function going to be the same regardless of the type of peer DID we are using?  E.g. unqualified, did:peer:1, did:peer:2, etc.?

If so, that's good, but if it is specific to, for example, did:peer:1, then I'm thinking it should called accordingly.

Issue #2178, opened 2023-03-23
default_version_upgrade_config.yml file missing from aca-py release


ianco:
`default_version_upgrade_config.yml` is missing from aca-py python release - this file is required to run the `aca-py upgrade` command.
 
See this issue:  https://github.com/bcgov/openshift-aries-mediator-service/issues/25

WadeBarnes:
The file is in the correct location and contains the correct content.  It just needs to be included in the release.

I did find it strange that you can only specify versions that are explicitly listed in the config file and that when upgrading to `v0.8.0` you have to specify `--from-version=v0.8.0` rather than indicating the version you're actually upgrading from.

WadeBarnes:
The file existing in the source code; https://github.com/hyperledger/aries-cloudagent-python/tree/main/aries_cloudagent/commands

But is not included in the PyPi distributions; https://github.com/bcgov/openshift-aries-mediator-service/issues/25#issuecomment-1481453249

and therefore is not installed on the images.

swcurran:
OK — some somewhere there is a manifest of things to include in the PyPi release.  I’ll see what I can figure out.

swcurran:
But that means we need to do a 0.8.1 immediately, right?

swcurran:
I can’t see how to add it to the package. AFAIK, the PyPi release should pickup everything in `aries_cloudagent`, which should include that file.

@ianco — if you know what update is needed, please do it. If not, I’ll get someone else to investigate.  I’m assuming this is trivial.

PR #2179, opened 2023-03-23
Adds the upgrade command YML file to the PyPi Release

swcurran:
@andrewwhitehead -- can you please test the packaging of this to verify that the next release that goes to PyPi will include the added [YML File](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/commands/default_version_upgrade_config.yml).

@ianco -- I don't think you need to do anything on this.

Addresses issue #2178


ianco:
> @ianco -- I don't think you need to do anything on this.

Thanks @swcurran the fix looks good

andrewwhitehead (approved):
Checked the resulting packages

PR #2180, opened 2023-03-24
Add parameter to force upgrade regardless of version checks

ianco:


swcurran:
I don’t think this will work as expected.  For the upgrade to work, the upgrade has know the “from” and “to” versions.  If the two are the same, forcing the upgrade won’t do anything, I don’t think.

I think the right solution is to switch the override when the “—from-version” arg is passed in.  In that case, use the passed in version instead of that in storage.

ianco:
Closing PR, there is WIP to address this

PR #2185, opened 2023-03-29
Update and automate ACA-Py upgrade process

### LINKED ISSUE #2181: Update the logic for performing upgrades between releases
The following issues were identified in trying to upgrade a `0.8.0-rc0` deployment to `0.8.0`.  All the issues are relatively straight forward, so putting them all together into one issue.

### Issue: How to force an upgrade:
 
 - Decision is to have both `--force-upgrade` and `--from-version`
 
### Issue: What if the from-version is not in the `upgrade` yaml?  It would be nice to have it fallback to the one previous to the `--from-version` or in secure storage.
 
 - Decision: Make the code smarter and have it find the first version equal to or greater than the `from` version (from args or storage) and apply all the upgrades including and to the current version.  This becomes the literal from to use.
 - Further -- if there is more than one identical update -- run it only once. See the example above -- all of the updates are to `resave-ConnRec` so it really only needs to be done once.

### Issue: Once a version is in storage, cannot move to a newer version without running the `upgrade` command.

- Decision: Change to only stop if there is an upgrade required between versions -- and if not, update the storage version.

### Issue: On a new deployment, there is no version put into storage.

- Decision: Add to the create storage the additional of the version record.

Issue: No version in storage in existing deployments.

- Decision: Recommend that the deployment have an `upgrade --from-version...` command.

Issue: No warning currently produced when version in storage is not found.

- Decision: Add the warning, with details about what to do.

### Issue: Can we run an upgrade automatically when deploying a new version. Current behaviour is to abort the startup.

- Option: Auto-upgrade.
   - Risk: Multiple instances try to run the upgrade could cause unexpected side effects.
      - Same as the next problem -- must be run offline
      - ~~Need a semaphore or equivalent if the upgrade must be run once and only once.~~
   - Risk: Some updates MAY require running the upgrade offline.
      - Flag in YML file to say MUST be run as an explicit upgrade.

### Issue: The current expectation is that in the YML, the version for the change is the "from" version -- e.g. when upgrading from a designated release.

- Decision: The version associated with an upgrade command should be the version in when the upgrade should be run. E.g. Any version being upgraded to that version must run the upgrade steps.  So:
   - If the entry is `v0.8.0: Update_Connections` -- that command should be run from every pre-0.8.0 version when upgrade to 0.8.0 or beyond.  Further, if going from 0.7.5 to 0.9.0, they would run the `0.8.0` upgrade.

shaangill025:
- resolve #2181

### DIFF
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -172,7 +172,7 @@ async def upgrade(settings: dict):
                     )
                 )
 
-        force_upgrade_flag = root_profile.settings.get("upgrade.force_upgrade") or False
+        force_upgrade_flag = settings.get("upgrade.force_upgrade") or False
         if upgrade_from_version_config and upgrade_from_version_setting:
             if (
                 package_version.parse(upgrade_from_version_config)
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -1,6 +1,7 @@
 """Upgrade command for handling breaking changes when updating ACA-PY versions."""
 
 import asyncio
+import logging
 import yaml
 
 from configargparse import ArgumentParser
@@ -25,6 +26,7 @@ from . import PROG
 DEFAULT_UPGRADE_CONFIG_PATH = (
     "./aries_cloudagent/commands/default_version_upgrade_config.yml"
 )
+LOGGER = logging.getLogger(__name__)
 
 
 class UpgradeError(BaseError):
@@ -125,6 +127,7 @@ async def add_version_record(profile: Profile, version: str):
             )
         else:
             await storage.update_record(version_storage_record, version, {})
+    LOGGER.info(f"{RECORD_TYPE_ACAPY_VERSION} storage record set to {version}")
 
 
 async def upgrade(settings: dict):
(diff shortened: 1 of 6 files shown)


aries_cloudagent/commands/upgrade.py
```diff
@@ -136,68 +168,64 @@ async def upgrade(settings: dict):
...
+        upgrade_version_in_config = get_upgrade_version_list(
+            sorted_versions_found_in_config, upgrade_from_version
+        )
+        force_upgrade_flag = root_profile.settings.get("upgrade.force_upgrade") or False
+        if upgrade_from_version == upgrade_to_version and not force_upgrade_flag:
             print(
                 f"Version {upgrade_from_version} to upgrade from and "
                 f"current version to upgrade to {upgrade_to_version} "
-                "are same."
+                "are same. If you still wish to run upgrade then plese "
```
swcurran, line 179: Typo in “please”. 

It makes no sense to me to force and upgrade when the “from” and “to” are the same, so I don’t think the logic here is right.  I think that it should just print a message a proceed.  An upgrade is only possible when the `from` is before the `to` and there are upgrades in between the two versions.

aries_cloudagent/commands/upgrade.py
```diff
@@ -136,68 +168,64 @@ async def upgrade(settings: dict):
                         f"{upgrade_from_version} as --from-version from "
                         "the config."
                     )
-        if upgrade_from_version == upgrade_to_version:
+        upgrade_version_in_config = get_upgrade_version_list(
+            sorted_versions_found_in_config, upgrade_from_version
+        )
+        force_upgrade_flag = root_profile.settings.get("upgrade.force_upgrade") or False
```
swcurran, line 174: The `force` upgrade should apply when the “—from-version” command line arg is before the `from` that is pulled from storage.  When the `force` is in effect, the command line arg overrides the secure storage, otherwise the opposite is used — the storage `from` overrides the command line `from`.

aries_cloudagent/config/argparse.py
```diff
@@ -2053,11 +2053,20 @@ def add_arguments(self, parser: ArgumentParser):
             ),
         )
 
+        parser.add_argument(
+            "--force-upgrade",
+            action="store_true",
+            env_var="ACAPY_UPGRADE_FORCE_UPGRADE",
+            help="Brute force the upgrade process.",
```
swcurran, line 2060: Should be — “Forces the '—from-version' argument to override the version retrieved from secure storage when calculating upgrades to be run."

aries_cloudagent/core/conductor.py
```diff
@@ -311,26 +312,41 @@ async def start(self) -> None:
...
+                config_path=self.root_profile.settings.get("upgrade.config_path"),
+                from_version=from_version,
+            )
+            if len(config_available_list) >= 1 and (
+                from_version != agent_version
+                or self.root_profile.settings.get("upgrade.force_upgrade")
+            ):
+                await upgrade(self.root_profile.settings)
+        else:
+            LOGGER.exception(
```
swcurran, line 341: This looks to me like if the wallet version is not found in storage, an exception is raised and the app stops — is that right?  If so, that is not what we want.  If the user does not pass a `—from-version`, and there is no `from` in storage, then I think the app should write the current version to storage, so that it will be found in the future.  A note should be written to the log with instructions on what to do if this is unexpected (notably — how to run the upgrade from the command line).

If the user then determines that they `upgrade` was bypassed, then they can run `—from-version x.y.z` and `—force-upgrade` to apply the upgrade.

ianco (approved):

PR #2191, opened 2023-03-31
feat(WIP): anoncreds updates

dbluhm:
This PR is a significant refactor of ACA-Py's handling of Anonymous Credentials, switching to the ledger-agnostic anoncreds-rs library instead of Indy specific libraries we've been using up until this point.

This PR is a work in progress. Given that ACA-Py has been built with Indy-isms and Indy specific libraries since the beginning, there's a lot that needed to be touched in order to move towards a more ledger-agnostic mentality. As of the time of opening this PR, it includes:

- A pluggable interface for defining AnonCreds Method Plugins, enabling resolution or registration of AnonCreds objects on any ledger/network/resolvable location as long as there is a plugin that implements it.
- A new set of Admin API endpoints for creating AnonCreds objects in a ledger-agnostic, plugin-friendly way
- Replaced indy-shared-rs with anoncreds-rs (indy-shared-rs still getting installed but not used)
- Added builtin plugins for:
  - Legacy Indy (Identifiers not using did:indy urls)
    - Supports registering and resolving Schemas and Credential Definitions; revocation objects are WIP
  - did:indy - Currently just stubbed out
  - did:web - Currently just stubbed out
- A simple demo script (to be removed and replaced with proper Integration tests) demoing issuance and verification (without revocation currently) on Legacy Indy.
  - This can be run with `docker-compose run tests && docker-compose down -v`

We are actively working on:
- Implementing revocation support through the new AnonCreds interface

Things that are very broken right now:
- The old `/schemas` and `/credential-definitions` endpoints
- Revocation
- Endorsement


anoncreds_test.py
```diff
@@ -0,0 +1,86 @@
+import os
+
+from controller.controller import Controller
```
andrewwhitehead, line 3: I don't see this `controller` package, does it need to be added?
This demo might be better placed under `examples/`

healthcheck.py
```diff
@@ -0,0 +1,25 @@
+import sys
```
andrewwhitehead, line 1: I'm not sure why this file was added

scripts/run_docker
```diff
@@ -3,7 +3,7 @@
 cd "$(dirname "$0")" || exit 1
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
 
-$CONTAINER_RUNTIME build -t aries-cloudagent-run -f ../docker/Dockerfile.run .. || exit 1
+$CONTAINER_RUNTIME build --platform linux/amd64 -t aries-cloudagent-run -f ../docker/Dockerfile.run .. || exit 1
```
andrewwhitehead, line 6: This looks like it was adjusted for the dev's system? It could be useful to support an environment variable for setting the platform, though.

andrewwhitehead:
I think this is generally looking pretty good. You'll want to define the `anoncreds` test marker in setup.cfg. Before this is added to main I would like to remove the indy-credx support code, as there's no need to support both implementations. I see some references to `anoncreds.sdk` which seems like a string replacement error, those would be indy-sdk tests which we won't be removing just yet.

One general question I'm not sure about. When publishing Anoncreds objects, do we assume that the ID can be generated locally, or can it support a registry that determines the ID for you?

swcurran:
To answer @andrewwhitehead ’s question:

> One general question I'm not sure about. When publishing Anoncreds objects, do we assume that the ID can be generated locally, or can it support a registry that determines the ID for you?

The AnonCreds v1.0 Spec. specifies that the objects DO NOT contain IDs that reference themself on the assumption that the IDs cannot always be generated locally — it depends on the AnonCreds Method.  So, for example, a CredDef contains a SchemaId, but does not contain a CredDefId for itself.

So — the implementation must support the case of the registry determining the ID for you.  It may be that the AnonCreds method does the actual calculation (such as with Indy), but the component calling the AnonCreds Method does not do that.

anoncreds_test.py
dbluhm: Sorry for the confusion, this is part of a quick test script that I intend to remove before merging to main, replacing it with some proper testing. This script is run by a container built from `docker/Dockerfile.anoncreds-test`, which does a quick `pip install` of the needed library. This gets run by the `tests` service in `docker-compose.yml` (which will also be removed before merging). We just found having a quick high-level test helpful while things are in flight.

healthcheck.py
dbluhm: This is also part of that test script -- this is used as a healthcheck in the docker-compose script so the tests run only after the test agents have reached a healthy state. I think this file could be legitimately useful to add to the project (don't have to rely on coming up with a curl command to do the same kind of check from a container orchestrator), though it probably belongs in the `docker` directory.

scripts/run_docker
dbluhm: Indeed, this was added to support running the script on M1, in the absence of support for images specifically built for aarch64.

dbluhm:
Closing in favor of #2276

PR #2200, opened 2023-04-04
Add Upgrading ACA-Py document

swcurran:


UpgradingACA-Py.md
```diff
@@ -0,0 +1,88 @@
...
+taken off line temporarily, the database upgraded explicitly, and then
+ACA-Py re-deployed as normal. As yet, we do not know of any cases of this, but
+those deploying ACA-Py should be aware of this possibility. For example,
+we may at some point need an upgrade that **MUST NOT** be executed by more
+than one ACA-Py instance. In that case, a "normal" upgrade could be dangerous
+for deployments on container orchestration platforms like Kubernetes.
+
+If the Maintainers of ACA-Py recognize a case where ACA-Py must be upgraded while
+offline, a new Upgrade feature will be added that will prevent the "auto upgrade"
+process from executing. See [Issue XXXX] for the status of that feature.
```
shaangill025, line 47: This should #2201

UpgradingACA-Py.md
```diff
@@ -0,0 +1,88 @@
...
+
```
shaangill025, line 48: I think we should also mention:
There are 2 flavours of this explicit upgrade mechanism, `"critical"` and `"warning"`. With `critical`, it will stop the "auto upgrade" process, raise an exception and force users to do the upgrade offline first. With `warning`, it will stop the "auto upgrade" process, log a warning but allows ACA-Py to startup normally. In this case, the version is not updated in secure storage and it will only update when doing the upgrade offline.

UpgradingACA-Py.md
```diff
@@ -0,0 +1,88 @@
...
+running version 0.7.5, and execute the upgrade from that version to the current
+one. The choice of 0.7.5 as the default is safe because the same upgrades will
+be run of any version 0.7.5 and earlier, as can be seen in the [Upgrade
+Definition YML file]. Thus, even if you are really upgrading from (for example)
+0.6.2, the same upgrades are needed as from 0.7.5.
+
+### Forcing an upgrade
+
+If you would like to force an upgrade from a given version, a pair of ACA-Py
+configuration arguments can be used together. If you specify "`--from-version
```
shaangill025, line 79: Maybe mention an example for clarity:
For instance, if the version in secure storage has been set as `v0.8.0` but for some reason you want to apply upgrades from `>=v0.7.2`, then you can specify "`--from-version v0.7.2`" and "`--force-upgrade`".

UpgradingACA-Py.md
swcurran: I've referenced the docs. We'll update this once when the merge of the #2204 is done -- perhaps as part of that PR, even.

UpgradingACA-Py.md
swcurran: Done.  Confirming @shaangill025 that the "v" is needed in "--from-version".  I've put it in, but is that right?

shaangill025: Yes, as we use "v" in config too.

shaangill025 (approved):

PR #2204, opened 2023-04-05
Add Explicit/Offline marking mechanism for Upgrade

### LINKED ISSUE #2201: Add mechanism for marking an Upgrade as offline only
As of release 0.8.1, we are making upgrades automatic. When you deploy a version of ACA-Py (as stated in `version.py`) that is higher than what is stored in the ACA-Py secure storage database, an upgrade is automatically run, based on the contents of the [Upgrade Definition YML]() file. However, in some cases, an upgrade may be particularly risky, and we want to make sure it is run only offline.  For example, if the upgrade has side effects that it must only run exactly once, we don't want to an automated upgrade in a Kubernetes environment where the upgrade might be run more than once.

To prevent that, please add a parameter that is optional but that can be added to YML upgrade definitions that prevents an "auto-update".  I think there should be two flavours of the parameter -- one that triggers and error and stops the operation, and one that triggers only an entry in the log and continues with processing.   The "error and stop" can be used when the upgrade **MUST** be run before the new version can be deployed.  The "log" can be used when the upgrade must be run offline, but is not needed for the new version to be run.  In the latter case, the version in storage MUST NOT be updated to the new current ACA-Py version -- it needs to be left as is.

shaangill025:
- resolve #2201 
- Pending extensive testing with docker-test scripts

In the config, `explicit_upgrade` is the flag or key and it's value can be `critical` [meaning error and stop] or `warning` [log and proceed]. ~~With `warning`, if there are versions which are not marked as explicit and are in scope for applying the upgrade then upgrade for these versions will proceed [after that it will continue the ACA-Py start up process].~~ With warning, if there are version which are not marked as explicit and are in scope for applying the upgrade then no upgrade will be performed and version in storage not updated.

```
"v0.7.2": {
    ...,
    "explicit_upgrade": "warning",
},
"v0.7.3": {
    ...,
    "explicit_upgrade": "critical",
},
"v0.7.1": {
    ...,
},
```


swcurran:
@shaangill025 — what is left for this one to leave “Draft” state?  Thanks

swcurran:
As I think about it — probably best to leave this to post-0.8.1 as it is not needed.

usingtechnology:
Tested the two new configuration values and got the expected results (Error and stop when "critical", Logged warning when "warning").
 
Approved.

swcurran (approved):
Tested and approved by @usingtechnology before he was a maintainer.

PR #2206, opened 2023-04-05
BREAKING: Auto-remove flags for Presentation Exchanges

### LINKED ISSUE #2126: Add an "auto delete" capability for presentation exchange records
Presentation exchange records don't automatically get deleted.

There is currently a `--preserve-exchange-records` parameter but that seems to apply only to credential exchange records.

Add the ability to "auto-delete" presentation exchange records, either through a startup param (like the cred exch flag), or through an extra param on the "verify" function (e.g. auto-delete after verify).

usingtechnology:
The current `--preserve-exchange-records` configuration argument supports credential exchanges only. This PR introduces a new configuration for presentation exchanges and roles within those exchanges. 

We want to preserve existing default behaviour: credential exchanges are removed, and presentation exchanges are stored. The credential exchanges could be saved by setting the'- preserve-exchange-records' configuration. Further override (to the issuer) specifies `auto_remove` in specific API payloads.

The default behaviour remains (credential exchanges removed, presentation exchanges stored) with addition of specifying default removal behaviour for each exchange and role. Since the API payloads use `auto_remove=True/False`, the new configurations are named `--auto-remove-*` and require a "boolean" value: true means remove, and false means store.

Truthy values are: `yes`, `true`, `t`, `y`, and `1`.
Falsey values are: `no`, `false`, `f`, `n`, and `0`.

Values are not case-sensitive.

**Command line Arguments**

```
--auto-remove-cred-exch-records-issuer <boolean value>
--auto-remove-cred-exch-records-holder <boolean value>
--auto-remove-pres-exch-records-verifier <boolean value>
--auto-remove-pres-exch-records-prover <boolean value>
```

**Environment Variables**

```
ACAPY_AUTO_REMOVE_CRED_EXCH_RECORDS_ISSUER=<true/false>
ACAPY_AUTO_REMOVE_CRED_EXCH_RECORDS_HOLDER=<true/false>
ACAPY_AUTO_REMOVE_PRES_EXCH_RECORDS_VERIFIER=<true/false>
ACAPY_AUTO_REMOVE_PRES_EXCH_RECORDS_PROVER=<true/false>
```

The `--preserve-exhange-records` argument is deprecated but still accepted. It is equivalent to:

```
--auto-remove-cred-exch-records-issuer false
--auto-remove-cred-exch-records-holder false
```

If none of the arguments are specified (system default) is equivalent to:

```
--auto-remove-cred-exch-records-issuer true
--auto-remove-cred-exch-records-holder true
--auto-remove-pres-exch-records-verifier false
--auto-remove-pres-exch-records-prover false
```

Once arguments are parsed, the settings are:

```
settings.get("auto_remove_cred_exch_records_issuer")
settings.get("auto_remove_cred_exch_records_holder")
settings.get("auto_remove_pres_exch_records_verifier")
settings.get("auto_remove_pres_exch_records_prover")
```

These settings should **always** be `True` or `False`, not `None`.

**Important**

While testing the previous behaviour, some stray credential exchanges remained when the system was set with `auto_remove = True` and API payload with no field named `auto_remove` to override the default behaviour.  Hopefully, no systems were built to take advantage of this unintended behaviour.

Fixes #2126


usingtechnology:
I have run this over an existing instance to test out if the new field on the presentation exchange models work over existing databases. Seems to work as expected...

dbluhm:
Minor comment here: back when the `--preserve-exchange-record` flag was introduced, it was intended to affect both credentials and presentations. I'm not sure where along the way that was lost but I think I'd consider this PR as fixing a regression unless we are aware of situations where people are explicitly expecting the presentation exchange records. In the interest of not adding more flags that are not frequently used, does it make sense to just fix the `--preserve-exchange-record` flag behavior instead of adding more flags?

swcurran:
I agree with @dbluhm — we have more than enough flags already… :-(

@usingtechnology — could you please update accordingly?

usingtechnology:
just to confirm, we can do 2 things:

1. add presentation exchange to `--preserve-exchange-record`
2. remove the 4 new "role" based flags `--auto-remove-<exchange-type>-<role>`

let me know if we are doing 1 and 2 or just 1. and if we are doing 2, then probably best to close this PR and make a new one to address item 1.

swcurran:
Arrggh…. Just read through all of the documentation.  Sorry it got this far before the question was asked.

I think we should leave it as is.  I’m tempted to do 1 (because it should have been done way back when that flag was added), but I’m worried about backwards compatibility. Despite flagging the change as breaking, I’m worried someone will be relying on presentations existing and they will be in trouble if they unexpectedly (for them) start to disappear.  I like the by role approach.

@dbluhm — are you OK with it?  The description in the top note covers the new functionality nicely.

usingtechnology:
since there wasn't a push from any particular user/group to have this kind of granularity, I think it's fine to remove the 4 new flags and add presentation exchange to the original flag.

@dbluhm @swcurran - ok to just remove the 4 and enhance the old one?

swcurran:
@dbluhm — good summary. I 100% agree that the ACA-Py storage should be used as a stream as you describe for protocol execution. Only key/crypto-related things like connections, issuer objects, credentials should be retained in ACA-Py storage. Things needed for agent operations.

To be very clear, I think you are saying we do 1 (add presentation exchange to --preserve-exchange-record) and we do not do 2 (remove the 4 new "role" based flags `--auto-remove-<exchange-type>-<role>`). Is that correct?

Doing 1 breaks backwards compatibility and risks (badly) breaking existing implementations (my argument), but your experience says that is the right way to go, correct?

I’m going to add a separate comment to propose another change...

swcurran:
To promote the use of ACA-Py as a protocol stream (and discourage saving transient data in secure storage), I think we should make a slight tweak to the new flags added in this PR.  I suggest we rename the flags to use the form `--auto-remove-<protocol>-<role>`, such that we can later add more protocols/role, and perhaps even a flag `--auto-remove-<protocol>` and perhaps even `--auto-remove-all`.  In other words, we use document the flags as expressions vs. specific values and allow them to be used for all protocols.

That would mean tweaking this PR to use the flags:

```
--auto-remove-issue-credential-issuer <boolean value>
--auto-remove-issue-credential-holder <boolean value>
--auto-remove-present-proof-verifier <boolean value>
--auto-remove-present-proof-prover <boolean value>
```

Too crazy?

@ianco @dbluhm @usingtechnology @andrewwhitehead @TimoGlastra

swcurran:
We should definitely be defaulting to using Askar for the testing.  Leaving it open to run the demo with Indy is nice to have, but Askar must be the default.  The sooner the indy-sdk is not used, the better.

dbluhm:
> To promote the use of ACA-Py as a protocol stream (and discourage saving transient data in secure storage), I think we should make a slight tweak to the new flags added in this PR. I suggest we rename the flags to use the form `--auto-remove-<protocol>-<role>`, such that we can later add more protocols/role, and perhaps even a flag `--auto-remove-<protocol>` and perhaps even `--auto-remove-all`. In other words, we use document the flags as expressions vs. specific values and allow them to be used for all protocols.
> 
> That would mean tweaking this PR to use the flags:
> 
> ```
> --auto-remove-issue-credential-issuer <boolean value>
> --auto-remove-issue-credential-holder <boolean value>
> --auto-remove-present-proof-verifier <boolean value>
> --auto-remove-present-proof-prover <boolean value>
> ```
> 
> Too crazy?

I agree with the idea that we should probably be more deliberate about clearing or not storing transient data in secure storage. Removing records should be the default and then not removing them can be a debug option. If you need to debug one record, I think it's okay to preserve all records -- given that it should be temporary or never used outside of development and testing. I'd rather have `--preserve-exchange-recrods` affect all record types than to have individual flags by protocol and role. Overall, though, my YAGNI senses are tingling.

usingtechnology:
Could we do something simple like @dbluhm suggests above: default is purge on terminal stage, `--preserve-exchange-records` can override that default for ALL exchange types.  

Could we add more complex behaviour with a config file instead of many switches/flags/env vars? Just in case some instance is not using `--preserve-exchange-records` but is relying on completed presentation exchanges being persisted?

My gut feeling would be to just expand `--preserve-exchange-records` to include credential and presentation (and future?) exchanges... but implemented in such a way that a development team could (more) easily write a plugin to alter that behaviour.

For instance, the Traction team will want to preserve completed records (in an altered, pared-down form).

ianco:
I think it makes sense to treat exchange records differently for different protocols.

For a credential exchange:
- the issuer has a record of the source data, and they need to track the fact that they issued a credential (and when) in case the source data changes and they need to revoke and/or re-issue
- the holder has the credential - they may want to track who they received it from (and when) but probably not a critical requirement, the credential has the issuer's did after all
- so it makes sense to delete by default

For a proof exchange the requirements are a bit different:
- the verifier knows they issued the request, and likely will record the result, but they may want to track the actual proof for audit purposes (to prove that they received a proof (this was a requirement for one project I was involved in)
- the prover may want to track what information they shared and with who
- i think it makes sense to *preserve* by default

Since aca-py doesn't (didn't) remove proof exchange records by default some applications may be relying on this behaviour.  So probably best to confirm with the community before changing.

This issue was originally raised because OrgBook wasn't deleting proof exchange records (and was generating a lot of them) but I think OrgBook is an anomaly, and most apps won't generate proof exchange records with such abandon.

swcurran:
Thanks @ianco.  That gives us some rationale for keeping the behaviour as is in this PR.

Unfortunately, we don’t really have a way to check with the broader community — it’s a lot bigger than just those that show up for ACA-Pug meetings.  That said, I don’t think it is urgent that we merge this, and having a conversation about this vs. just the PR comments would be better, so let’s leave it for the next ACA-Pug.  I’m happy to lead the discussion and at the end we’ll make the call.  Unfortunately, I don’t think there is a clear right/wrong so it’s going to be a compromise.

Regards your statement @ianco about keeping the records. I think best practices should be that anything you want to keep, you keep OUTSIDE the ACA-Py secure storage.  At minimum, we should have a standard that all protocols have a minimize function to shrink the size of a protocol state object after execution completes. That could be a plugin, as @usingtechnology suggests.

usingtechnology:
after ACA-PUG meeting, just leaving notes...

We did discuss keeping default ACA-Py configuration simple (enhance the preserve flag) with ensuring that `auto-remove` API flags are used. (option 1 above)

Any further customization (role based, exchange type based logic) can be put in by a plugin developer. That could include what data is retained when hitting terminal/completed state.

PR #2211, opened 2023-04-15
Enhancements regarding the support of non Unix-like operating systems

sk13:
This would support execution on operating systems which uses a path seperator character different from '/' (e.g Windows).


WadeBarnes:
Updated the code to be inline with `main`.  Waiting to see the results from the checks and tests.

WadeBarnes:
@sk13, Can you look into the formatting and test errors to see if they have anything to do with your changes.  Thanks

dbluhm:
@sk13 are you available to look at these test failures? If not, I might be able to volunteer some time from someone on my team to fix this up and resolve conflicts.

dbluhm:
See #2327

PR #2220, opened 2023-04-27
Update AriesOpenAPIDemo.md clarify callback can be one of webhook or websocket depending how agent is configured

miroslav-inc:


demo/AriesOpenAPIDemo.md
```diff
@@ -705,11 +705,11 @@ The following table lists endpoints that you need to call ("REST service") and c
 | Protocol Step        | Faber (Verifier)       | Alice (Holder/Prover)     | Notes |
 | -------------------- | ---------------------- | ------------------------- | ----- |
 | Send Proof Request | **`POST /present-proof-2.0/send-request`** | | REST service |
-| Receive Proof Request | | <agent_cb>/present_proof_v2_0 | callback |
+| Receive Proof Request | | <agent_cb>/present_proof_v2_0 | callback(webhook or websocket) |
```
swcurran, line 708: I think it makes sense to add `webhook`, but not sure about websocket.  The use of websockets or http is a transport level concern -- how the webhook is received -- not an alternative to the webhook.  It is always a webhook.

dbluhm:
I've been working on ACA-Py for literal years and only recently realized that we supported delivering webhooks over websocket :smile: Not sure how I overlooked it for so long... I think adding emphasis that webhooks can be posted to a configured webhook URL or sent via websocket somewhere in the docs would be helpful. There's an allusion to this support in the Multitenancy docs but it's more of a side note.

I agree with @swcurran on terminology but I am also interested in updates to this doc that clarify websockets can be used.

dbluhm:
It's been a while since we heard from OP so I opened a new PR with feedback implemented. Thanks @miroslav-inc for your contributions! I'll close this PR in favor of #2287.

PR #2221, opened 2023-05-02
Fix broken link in README

ff137:
Link to the Trust Over IP framework PDF was broken, now fixed. All other links confirmed working


ff137:
Anybody here? It's just a link fix

swcurran (approved):
Thanks — sorry for the slow response.

PR #2233, opened 2023-05-16
Allow Configuration Settings on a per-tenant basis

### LINKED ISSUE #2209: Multitenancy: expose agent configuration settings on a per-tenant basis
When using an agent running on a multi-tenant instance of ACA-Py, each tenant might want/need to specify a different behaviour.
A lot of the the ACA-Py startup parameters/environment variables would benefit from being configurable at a tenant level (all of the "auto" flags for once). The following is a subset of settings that was identified as it is currently used across agent deployments for BCGov (see https://github.com/bcgov/traction/issues/547):
- `ACAPY_LOG_LEVEL`
- `ACAPY_INVITE_PUBLIC`
- `ACAPY_PUBLIC_INVITES`
- `ACAPY_AUTO_ACCEPT_INVITES`
- `ACAPY_AUTO_ACCEPT_REQUESTS`
- `ACAPY_AUTO_PING_CONNECTION`
- `ACAPY_MONITOR_PING`
- `ACAPY_AUTO_RESPOND_MESSAGES`
- `ACAPY_AUTO_RESPOND_CREDENTIAL_OFFER`
- `ACAPY_AUTO_RESPOND_CREDENTIAL_REQUEST`
- `ACAPY_AUTO_VERIFY_PRESENTATION`
- `ACAPY_NOTIFY_REVOCATION`
- `ACAPY_AUTO_REQUEST_ENDORSEMENT`
- `ACAPY_AUTO_WRITE_TRANSACTIONS`
- `ACAPY_CREATE_REVOCATION_TRANSACTIONS`

Settings that are set as part of the main/agency agent startup command would likely be used as defaults, but each tenant would/should be able to override them based on specific needs.

*Note:* one setting that might require extra attention is the one related to the writable ledger. See #2168  for related information.

shaangill025:
- resolve #2209
- Some changes related to getting `BaseLedger` instance from `ledger_id` won't be used right now but will be used soon with some proposed Traction features.

### DIFF
--- a/Multitenancy.md
+++ b/Multitenancy.md
@@ -417,7 +417,7 @@ To allow configurablity of ACA-Py startup parameters/environment variables at a 
       "extra_settings": {
           "ACAPY_LOG_LEVEL": "INFO",
           "ACAPY_INVITE_PUBLIC": true,
-          "public_invites": true
+          "public-invites": true
       },
   }
   ```
--- a/Multitenancy.md
+++ b/Multitenancy.md
@@ -190,6 +190,71 @@ WebSocket example:
 }
 ```
 
+## Per tenant settings
+
+To allow configurablity of ACA-Py startup parameters/environment variables at a tenant/subwallet level. [PR#2233](https://github.com/hyperledger/aries-cloudagent-python/pull/2233) will provide the ability to update the following subset of settings when creating or updating the subwallet:
+
+| Label  | Setting  |
+|---|---|
+| ACAPY_LOG_LEVEL  |  log.level |
+| ACAPY_INVITE_PUBLIC  |  debug.invite_public  |
+| ACAPY_PUBLIC_INVITES  | public_invites  |
+| ACAPY_AUTO_ACCEPT_INVITES  | debug.auto_accept_invites  |
+| ACAPY_AUTO_ACCEPT_REQUESTS  | debug.auto_accept_requests  |
+| ACAPY_AUTO_PING_CONNECTION  | auto_ping_connection  |
+| ACAPY_MONITOR_PING  | debug.monitor_ping  |
+| ACAPY_AUTO_RESPOND_MESSAGES  | debug.auto_respond_messages  |
+| ACAPY_AUTO_RESPOND_CREDENTIAL_OFFER  | debug.auto_resopnd_credential_offer  |
+| ACAPY_AUTO_RESPOND_CREDENTIAL_REQUEST  | debug.auto_respond_credential_request  |
+| ACAPY_AUTO_VERIFY_PRESENTATION  | debug.auto_verify_presentation  |
+| ACAPY_NOTIFY_REVOCATION  | revocation.notify  |
+| ACAPY_AUTO_REQUEST_ENDORSEMENT  | endorser.auto_request  |
+| ACAPY_AUTO_WRITE_TRANSACTIONS  | endorser.auto_write  |
+| ACAPY_CREATE_REVOCATION_TRANSACTIONS  | endorser.auto_create_rev_reg  |
+| ACAPY_ENDORSER_ROLE  | endorser.protocol_role  |
+
+- `POST /multitenancy/wallet`
+
+  Added `extra_settings` dict field to request schema. `extra_settings` can be configured in the request body as below:
+
+  ```
+  Example request body
+  {
+      "wallet_name": " ... ",
+      "default_label": " ... ",
+      "wallet_type": " ... ",
+      "wallet_key": " ... ",
+      "key_management_mode": "managed",
+      "wallet_webhook_urls": [],
+      "wallet_dispatch_type": "base",
+      "extra_settings": {
+          "ACAPY_LOG_LEVEL": "INFO",
+          "ACAPY_INVITE_PUBLIC": "true",
+          "ACAPY_PUBLIC_INVITES": "True",
+      },
+  }
+  ```
+
+- `PUT /multitenancy/wallet/{wallet_id}`
+
+  Added `extra_settings` dict field to request schema.
+
+  ```
+  {
+    "wallet_webhook_urls": [ ... ],
+    "wallet_dispatch_type": "default",
+    "label": " ... ",
+    "image_url": " ... ",
+    "extra_settings": {
+        "ACAPY_LOG_LEVEL": "INFO",
+        "ACAPY_INVITE_PUBLIC": "True",
+        "ACAPY_PUBLIC_INVITES": "false",
+    },
+  }
+  ```
+
+All the currently suppoorted settings in `extra_settings` are either accepting str or boolean. But the bool attributes in the request body should be passed along as a string such as `"true", "True", "false", "False"`.
+
 ## Authentication
 
 When multi-tenancy is not enabled you can authenticate with the agent using the `x-api-key` header. As there is only a single wallet, this provides sufficient authentication and authorization.
--- a/Multitenancy.md
+++ b/Multitenancy.md
@@ -29,6 +29,7 @@ This allows ACA-Py to be used for a wider range of use cases. One use case could
 - [Tenant Management](#tenant-management)
   - [Update a tenant](#update-a-tenant)
   - [Remove a tenant](#remove-a-tenant)
+  - [Per tenant settings](#per-tenant-settings)
 
 ## General Concept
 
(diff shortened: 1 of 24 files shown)


swcurran:
Thanks, @shaangill025 — let us know about the test failures and we’ll look at moving this forward.

shaangill025:
@swcurran I have added the documentation, this PR is ready for review.

aries_cloudagent/multitenant/admin/routes.py
```diff
@@ -23,6 +24,32 @@
...
+    "ACAPY_AUTO_RESPOND_CREDENTIAL_REQUEST": "debug.auto_respond_credential_request",
+    "ACAPY_AUTO_VERIFY_PRESENTATION": "debug.auto_verify_presentation",
+    "ACAPY_NOTIFY_REVOCATION": "revocation.notify",
+    "ACAPY_AUTO_REQUEST_ENDORSEMENT": "endorser.auto_request",
+    "ACAPY_AUTO_WRITE_TRANSACTIONS": "endorser.auto_write",
+    "ACAPY_CREATE_REVOCATION_TRANSACTIONS": "endorser.auto_create_rev_reg",
+    "ACAPY_ENDORSER_ROLE": "endorser.protocol_role",
+}
+
+ACAPY_ENDORSER_FLGAS_DEPENDENT_ON_AUTHOR_ROLE = [
```
esune, line 46: Small typo: `FLGAS`->`FLAGS`

swcurran:
The goal was to get this into 0.8.2, but since the developer has not been available to work on this, it may not make it.

Multitenancy.md
```diff
@@ -375,3 +376,81 @@ curl -X POST "${ACAPY_ADMIN_URL}/multitenancy/wallet/{wallet_id}/remove" \
...
 {}
 ```
+
+### Per tenant settings
+
+To allow configurablity of ACA-Py startup parameters/environment variables at a tenant/subwallet level. [PR#2233](https://github.com/hyperledger/aries-cloudagent-python/pull/2233) will provide the ability to update the following subset of settings when creating or updating the subwallet:
+
+| Label  | Setting  |
+|---|---|
+| ACAPY_LOG_LEVEL  |  log.level |
```
usingtechnology, line 386: I believe that changing the log level will only take affect when writing to a physical log file is configured. It's not an issue, but probably wouldn't be clear to anyone. Maybe we just need a note about that limitation?

https://github.com/hyperledger/aries-cloudagent-python/blob/421c4ed6a3f750c7f30af7b85d1481a762a5acc4/aries_cloudagent/config/logging.py#L587

Multitenancy.md
```diff
@@ -375,3 +376,81 @@ curl -X POST "${ACAPY_ADMIN_URL}/multitenancy/wallet/{wallet_id}/remove" \
 ```jsonc
 {}
 ```
+
+### Per tenant settings
```
usingtechnology, line 380: Awesome documentation. 👍

aries_cloudagent/multitenant/admin/routes.py
```diff
@@ -23,6 +23,32 @@
 from ..error import WalletKeyMissingError
 
 
+ACAPY_LIFECYCLE_CONFIG_FLAG_MAP = {
```
usingtechnology, line 26: I wonder if we should have another map too, one that better matches config.yml ( so`auto_accept_invites`: `debug.auto_accept_invites).  I am OK with not having it and staying with the ENV_VAR format but could see the case where they are migrating an existing config.yml file contents.

usingtechnology (changes requested):
I did not test the multiledger portion of this, just the tenant extra settings.

This is probably up for discussion. Should this (tenant extra settings) be a plugin? 

In any case, what is missing is the tenant managing their settings. Currently, the base wallet/admin is the one creating the settings and updating them. A tenant/sub-wallet cannot access the `/multitenancy` APIs. I think the current changes can stay, but we should have another API `/settings` (?) that requires a tenant/sub-wallet token and they can manage those settings.

shaangill025:
@usingtechnology The requested changes have been added.

usingtechnology:
@shaangill025 - I would expect the tenant `/settings` API to update/return the small set of values not ALL of the settings. I realize your logic will only update the given set, but the API itself should NOT return all the settings. There are definitely things in there a tenant should not know. Can we limit to the set you provided? Also, I don't think those changes are persistent over restarts of the server. It doesn't appear that we are reading or writing from the wallet record (where the extra settings are initially stored).

aries_cloudagent/settings/routes.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        async with root_profile.session() as session:
+            multitenant_mgr = session.inject_or(BaseMultitenantManager)
+            if multitenant_mgr:
+                wallet_id = context.metadata.get("wallet_id")
+                wallet_record = await multitenant_mgr.update_wallet(
+                    wallet_id, extra_settings
+                )
+                wallet_settings = wallet_record.settings
+                settings_dict = _get_filtered_settings_dict(wallet_settings)
```
usingtechnology, line 86: This is great, having the changes saved with the wallet settings. Only change is that the return of the update should match the return of the get. So return all of the profile settings not just the wallet settings. It should make it easier for the caller to know what all the settings are. For instance, they may cache these settings in their frontend code and returning all of the settings saves them from doing a merge...

usingtechnology (changes requested):
Again, I haven't tested the ledger/indy_vdr changes, but the per tenant settings is almost ready to go. One simple change for the update API response.

Multitenancy.md
```diff
@@ -383,22 +383,22 @@ To allow configurablity of ACA-Py startup parameters/environment variables at a
...
-| ACAPY_AUTO_RESPOND_MESSAGES  |  auto_respond_messages | debug.auto_respond_messages  |
-| ACAPY_AUTO_RESPOND_CREDENTIAL_OFFER  |  auto_respond_credential_offer | debug.auto_resopnd_credential_offer  |
-| ACAPY_AUTO_RESPOND_CREDENTIAL_REQUEST  |  auto_respond_credential_request | debug.auto_respond_credential_request  |
-| ACAPY_AUTO_VERIFY_PRESENTATION  |  auto_verify_presentation | debug.auto_verify_presentation  |
-| ACAPY_NOTIFY_REVOCATION  |  notify_revocation | revocation.notify  |
-| ACAPY_AUTO_REQUEST_ENDORSEMENT  |  auto_request_endorsement | endorser.auto_request  |
-| ACAPY_AUTO_WRITE_TRANSACTIONS  |  auto_write_transactions | endorser.auto_write  |
-| ACAPY_CREATE_REVOCATION_TRANSACTIONS  |  auto_create_revocation_transactions | endorser.auto_create_rev_reg  |
-| ACAPY_ENDORSER_ROLE  |  endorser_protocol_role | endorser.protocol_role  |
+| ACAPY_LOG_LEVEL  |  log-level |  log.level |
```
usingtechnology, line 386: love this documentation!

aries_cloudagent/settings/routes.py
```diff
@@ -79,11 +79,17 @@ async def update_profile_settings(request: web.BaseRequest):
...
                     wallet_id, extra_settings
                 )
+                wallet_record, profile = await multitenant_mgr.get_wallet_and_profile(
+                    root_profile.context, wallet_id, wallet_key
+                )
+                profile_settings = profile.settings.to_dict()
                 wallet_settings = wallet_record.settings
-                settings_dict = _get_filtered_settings_dict(wallet_settings)
+                all_settings = {**profile_settings, **wallet_settings}
+                settings_dict = _get_filtered_settings_dict(all_settings)
```
usingtechnology, line 92: works great.

usingtechnology (dismissed):
great work @shaangill025

Multitenancy.md
```diff
@@ -375,3 +376,81 @@ curl -X POST "${ACAPY_ADMIN_URL}/multitenancy/wallet/{wallet_id}/remove" \
...
+      "default_label": " ... ",
+      "wallet_type": " ... ",
+      "wallet_key": " ... ",
+      "key_management_mode": "managed",
+      "wallet_webhook_urls": [],
+      "wallet_dispatch_type": "base",
+      "extra_settings": {
+          "ACAPY_LOG_LEVEL": "INFO",
+          "ACAPY_INVITE_PUBLIC": true,
+          "public_invites": true
```
andrewwhitehead, line 420: Might be `public-invites` according to above

andrewwhitehead (approved):

PR #2234, opened 2023-05-16
Allow Log Access at Tenant Level

shaangill025:
- resolve https://github.com/bcgov/DITP/issues/23
- Built on top of #2233
- Allows writing of logs with `logger_alias` [tenant identifier] to a specified file, in addition to console output [same as before]
- It is responsibility of the controller to ingest/consume this logs in ELK stack and then be able to query by tenant identifier.


esune:
As per our conversation:
- let's not expose the tenant alias to the settings and use an identifier that is alerady available in ACA-Py (e.g.: the tenant's DID). This is to avoid log highjacking - intentional or not - by tenants with same identifier.
- log statement patterns should be extracted to a configuration file, rather than being hard-coded in the code.
- use python logging configuration file to define a common-sense log rotation and retention configuration, and to hold any other logging settings that might be supported.
- evaluate how to use/specify JSON logging, as it would make it easier for log analytics to ingest and search the data.

aries_cloudagent/core/dispatcher.py
```diff
@@ -170,7 +175,7 @@ async def handle_message(
         # if warning:
```
esune, line 175: Is there a reason for this code block to be commented-out rather than removed?

aries_cloudagent/multitenant/admin/routes.py
esune, line 1: Are these changes meant to be in this PR? It looks like something from #2233 got pulled in, possibly inadvertently..

Logging.md
```diff
@@ -12,23 +12,25 @@ Other log levels fall back to `WARNING`.
 
 * `--log-level` - The log level to log on std out.
 * `--log-file` - Path to a file to log to.
+* `--log-handler-config` - Specifies `when`, `interval`, `backupCount` for the   `TimedRotatingFileMultiProcessHandler`. These 3 attributes are passed as a `;` seperated string. For example, `when` of d (days), `interval` of 7 and `backupCount` of 1 will be passed as `D;7;1`. Note: `backupCount` of 0 will mean all backup log files will be retained and not deleted at all. More details about these attributes can be found [here](https://docs.python.org/3/library/logging.handlers.html#timedrotatingfilehandler). `TimedRotatingFileMultiProcessHandler` supports the ability to cleanup logs by time and mantain backup logs and a custom JSON formatter for logs.
```
esune, line 15: Might want to explain the JSON formatting setting as well.

esune (commented):
Looking good! A couple small-ish comments/notes/suggestions. I cannot approve PRs either way so would ask someone else (maybe @WadeBarnes?) to take a look as well.

aries_cloudagent/core/dispatcher.py
shaangill025: The commented block was added with https://github.com/hyperledger/aries-cloudagent-python/pull/1940, it supports handling of minor version of protocols. This implements what was mentioned in the RFC but a decision was made at that time to comment this out and reconsider in future.

Logging.md
shaangill025: Added `--log-fmt-pattern` and `--log-json-fmt`

aries_cloudagent/config/logging.py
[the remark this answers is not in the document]
shaangill025: - For `mpfhandler`, I have added context in code. The interval and backupCount were not working as intended in 
  the library. Also the old backup files were not being deleted on rotation. This required the following custom 
  implementation.

- For the json formatter, I tested now again python-json-logger library and it is working as intended. Initially when I was developing, `mpfhandler` and `python-json-logger` together wasn't working, so I was making code changes in both to fix it. The issues were only limited to `mpfhandler` so code changes to it fixes it. I have removed the code for json formatter and using `python-json-logger` instead.

WadeBarnes (dismissed):

shaangill025:
At this stage, the implemented utility function [`get_logger_inst` to support per tenant log access] is just used in the following files:
- `aries_cloudagent/connections/base_manager.py`
- `aries_cloudagent/core/dispatcher.py`
- `aries_cloudagent/core/oob_processor.py`

Once the PR is merged and more testing to validate this implementation, other 50+ files will need to be updated in a separate PR along the lines above.

PR #2239, opened 2023-05-24
Improve faber agent with ngrok selection and invitation_url output

DrumRobot:
- Update NGROK_ENDPOINT to select the tunnel with the specified AGENT_PORT
- Add output of invitation_url using log_msg function


swcurran:
Thanks for the PR! Kicked off the merge tests (first time contributor), and we’ll take a look at the tweak.

Jsyro (commented):
Invitation print looks ok. Two quick questions.

How do you start ngrok for the run_demo to interact with?
Can you make the equivalent change in `ngrok-wait.sh` is uses the same jq command that I presume should be updated if this is a better command.

PR #2241, opened 2023-05-25
Feat: Added support for Ed25519Signature2020 signature type and Ed25519VerificationKey2020

dkulic:


aries_cloudagent/vc/ld_proofs/suites/ed25519_signature_2020.py
```diff
@@ -0,0 +1,103 @@
...
+            proof (dict): The proof to add the signature to
+
+        Returns:
+            dict: The proof object with the added signature
+
+        """
+        signature = await self.key_pair.sign(verify_data)
+        pprint("Signature")
+        pprint(signature)
+        pprint(len(signature))
```
TimoGlastra, line 60:
```suggestion
```

TimoGlastra (approved):
LGTM, some primt statements. There are quite some places where suites are registered. Might be good to add a registry at some point for signature suites, which would also allow for dynamic registration of more suites

dkulic:
I will fix these errors, sorry for that, but one of them is multiformats library not supporting python 3.6. Python 3.6 is EOL, maybe it should be dropped?

WadeBarnes:
@swcurran, Yes we can drop Python 3.6.  The `py3.9` images have shown to be drop-in compatible where the BC Gov Python 3.6 based images were used before in all cases where I've upgraded to `0.8.1`.

dbluhm:
I opened up #2247 to make the version bump from 3.6 to 3.9

swcurran:
To be discussed at ACA-Pug Meeting — May 30, 2023.

swcurran:
Per the May 30, 2023 ACA-Pug meeting — we are dropping 3.6 support in ACA-Py #2244 / #2247 . As soon as that is done, this should be updated, processed and merged.

dbluhm:
@dkulic is OOTO so I'll take care of the conflicts on this PR.

dbluhm (approved):

dbluhm:
Requesting input from @usingtechnology and/or @andrewwhitehead; would like to see this one keep moving along.

aries_cloudagent/connections/base_manager.py
```diff
@@ -294,6 +299,8 @@ async def resolve_invitation(
     def _extract_key_material_in_base58_format(method: VerificationMethod) -> str:
         if isinstance(method, Ed25519VerificationKey2018):
             return method.material
+        elif isinstance(method, Ed25519VerificationKey2020):
+            return bytes_to_b58(multibase.decode(method.material))
```
dbluhm, line 303: I think this also needs an "unwrap" on the material to remove the `0xed` multicodec prefix added when using the `publicKeyMultibase`

dbluhm (changes requested):

aries_cloudagent/connections/base_manager.py
dkulic: I added support for both (with and without multicodec). Do you think it should be like that or should we support multibase keys only with multicodec?

dbluhm: Given the ambiguity, I think accepting either is probably a good choice for now at least.

dbluhm (approved):

PR #2243, opened 2023-05-26
fix: route multitenant connectionless oob invitation

TimoGlastra:
This addresses an issue where the keys used in connectionless oob invitations were not correctly set-up for routing through the base wallet when multi-tenancy is enabled, which resulted in those invitation not being usable.

cc @lohanspies @ff137

### DIFF
--- a/aries_cloudagent/multitenant/base.py
+++ b/aries_cloudagent/multitenant/base.py
@@ -199,7 +199,7 @@ class BaseMultitenantManager(ABC):
                 public_did_info = await wallet.get_public_did()
 
             if public_did_info:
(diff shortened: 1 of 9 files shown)


lohanspies:
Thank you @TimoGlastra, much appreciated

dbluhm (commented):
These changes look good though I'm a bit confused how the added config files are used. Are they just examples? Used in testing?

TimoGlastra:
Those are my local config files for testing and were pushed by accident 🤦

dbluhm (approved):

andrewwhitehead (approved):

Issue #2244, opened 2023-05-26
Drop support for Python 3.6


swcurran:
It’s now time to drop support for Python 3.6.  What needs to be done to make that happen?  I’m guessing the list is something like:

- Change the requirements for ACA-Py to use a minimum of 3.7
- Update the documentation (Readme, elsewhere?)
- Stop running the automated 3.6 tests.  If they are the per PR tests, change those to use Python 3.9 (right?)
- Update any container image publishing to use Python 3.9.

FYI BC Gov has updated all deployments from Python 3.6 to 3.9 and there were no issues identified.

Adjustments/additions @WadeBarnes @dbluhm ?

WadeBarnes:
I think we should go straight to using a minimum of Python 3.9.  All of the tests are passing for that version and we're already building and publishing images for py3.9 here https://github.com/hyperledger/aries-cloudagent-python/pkgs/container/aries-cloudagent-python.  BC Gov, as @swcurran pointed out, is using these images to upgrade all services and we've had great success using the `py3.9` images as drop-in replacement/upgrade for all previous versions.  Moving to `py3.9` as the minimum means we just need to drop the support for `py3.6` and `py3.7` in a few GitHub workflows from a CI/CD perspective.

So:
 
- Change the requirements for ACA-Py to use a minimum of Python 3.9
- Update the documentation (Readme, elsewhere?)
- Drop support for 3.6 and 3.7 in all automated tests.  Switch to using 3.9 where needed (though I think it's included in most workflows already).
- Stop publishing container images containing Python 3.6.  (We're already publishing images for Python 3.9)

@swcurran, as a separate topic we need to discuss what to do with `bcgovimages/aries-cloudagent` due to it's 50M+ download profile.

swcurran:
I think we need to transition the `bcgovimages/aries-cloudagent` as well, but open to suggestions on the best way.  I guess since the image is a replacement, we should be able to just move the `latest` tag to a Python 3.9 one?

PR #2247, opened 2023-05-27
chore!: drop python 3.6 support

dbluhm:
Update setup.py minimum python version requirement.
Update workflows to use Python 3.9 by default.
Update container image docs.

@swcurran @WadeBarnes I believe these are the relevant changes in workflows and setup files in order to officially drop Python 3.6. I think once we've made the transition, there are a number of dependencies that ought to be updated. I think those updates are best suited to a follow up PR.

closes: #2244

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -14,9 +14,6 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        include:
-          - os: "ubuntu-20.04"
-            python-version: "3.6"
     uses: ./.github/workflows/tests.yml
     with:
       python-version: ${{ matrix.python-version }}
@@ -30,9 +27,6 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        include:
-          - os: "ubuntu-20.04"
-            python-version: "3.6"
     uses: ./.github/workflows/tests-indy.yml
     with:
       python-version: ${{ matrix.python-version }}
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
     uses: ./.github/workflows/tests.yml
     with:
       python-version: ${{ matrix.python-version }}
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
     uses: ./.github/workflows/tests-indy.yml
     with:
       python-version: ${{ matrix.python-version }}
--- a/ContainerImagesAndGithubActions.md
+++ b/ContainerImagesAndGithubActions.md
@@ -47,12 +47,8 @@ Below is a table of all generated images and their tags:
 
 Tag                     | Variant  | Example                  | Description                                                                                     |
 ------------------------|----------|--------------------------|-------------------------------------------------------------------------------------------------|
-py3.7-X.Y.Z             | Standard | py3.7-0.7.4              | Standard image variant built on Python 3.7 for ACA-Py version X.Y.Z                             |
-py3.8-X.Y.Z             | Standard | py3.8-0.7.4              | Standard image variant built on Python 3.8 for ACA-Py version X.Y.Z                             |
 py3.9-X.Y.Z             | Standard | py3.9-0.7.4              | Standard image variant built on Python 3.9 for ACA-Py version X.Y.Z                             |
 py3.10-X.Y.Z            | Standard | py3.10-0.7.4             | Standard image variant built on Python 3.10 for ACA-Py version X.Y.Z                            |
-py3.7-indy-A.B.C-X.Y.Z  | Indy     | py3.7-indy-1.16.0-0.7.4  | Standard image variant built on Python 3.7 for ACA-Py version X.Y.Z and Indy SDK Version A.B.C  |
-py3.8-indy-A.B.C-X.Y.Z  | Indy     | py3.8-indy-1.16.0-0.7.4  | Standard image variant built on Python 3.8 for ACA-Py version X.Y.Z and Indy SDK Version A.B.C  |
 py3.9-indy-A.B.C-X.Y.Z  | Indy     | py3.9-indy-1.16.0-0.7.4  | Standard image variant built on Python 3.9 for ACA-Py version X.Y.Z and Indy SDK Version A.B.C  |
 py3.10-indy-A.B.C-X.Y.Z | Indy     | py3.10-indy-1.16.0-0.7.4 | Standard image variant built on Python 3.10 for ACA-Py version X.Y.Z and Indy SDK Version A.B.C |
 
--- a/ContainerImagesAndGithubActions.md
+++ b/ContainerImagesAndGithubActions.md
@@ -47,7 +47,6 @@ Below is a table of all generated images and their tags:
 
 Tag                     | Variant  | Example                  | Description                                                                                     |
 ------------------------|----------|--------------------------|-------------------------------------------------------------------------------------------------|
-py3.6-X.Y.Z             | Standard | py3.6-0.7.4              | Standard image variant built on Python 3.6 for ACA-Py version X.Y.Z                             |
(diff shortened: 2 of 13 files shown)


dbluhm:
The integration test action used von-image as a base which had python 3.6 installed. I switched the base image to a locally built indy base image (`indy-base` target in `Dockerfile.indy`). However, this is not currently caching the image build so this adds a significant amount of time to the action (since building Indy takes a while). We could switch to running the integration tests on Askar (maybe it already is?) to reduce build time in the integration test action.

swcurran:
Could we not add a new von-image that is based on Python 3.9 so everything is essentially the same?   Definitely think we should be use Askar as the default.

swcurran:
LGTM, but deferring to @WadeBarnes for a final answer and merge.

dbluhm:
Integration tests didn't finish cleanly -- I'll have to take a closer look

dbluhm:
I'll try to keep this timely but if someone beats me to it, contributions are welcome

.github/workflows/nightly-tests.yml
```diff
@@ -14,9 +14,6 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
```
WadeBarnes, line 16:
```suggestion
        python-version: ["3.9", "3.10"]
```

.github/workflows/nightly-tests.yml
```diff
@@ -30,9 +27,6 @@ jobs:
...
```
WadeBarnes, line 29:
```suggestion
        python-version: ["3.9", "3.10"]
```

docker/Dockerfile.test
```diff
@@ -1,4 +1,4 @@
-ARG python_version=3.6.13
+ARG python_version=3.9.16
 FROM python:${python_version}-slim-buster
```
WadeBarnes, line 2:
```suggestion
FROM python:${python_version}-slim-bullseye
```

ContainerImagesAndGithubActions.md
WadeBarnes, line 1: I'd recommend removing the references to Python 3.7 and 3.8 as well.

WadeBarnes (changes requested):
Some minor recommendations

WadeBarnes:
Once the integration tests are working, I'll have to switch over the Python 3.9 Tests over to the `Required` ones:
image

demo/run_bdd
```diff
@@ -126,6 +126,7 @@ AGENT_PORT=8020
 AGENT_PORT_RANGE=8020-8079
 
 echo "Preparing agent image..."
+docker build -q -t indy-base -f ../docker/Dockerfile.indy --target=indy-base .. || exit 1
```
WadeBarnes, line 129:
```suggestion
docker build -q -t indy-base -f ../docker/Dockerfile.indy --target=main .. || exit 1
```
Should this be a full ACA-Py image, rather than just the `indy-base` image?

WadeBarnes:
The integration tests are running through to completion locally for me.  It seems like they are getting hung up somewhere in GHA.

swcurran:
So good to merge then, @WadeBarnes ?

swcurran:
@dbluhm — can you update the branch so we can merge?

WadeBarnes:
> So good to merge then, @WadeBarnes ?

@swcurran , Not yet, there are some changes that need to be made.  @dbluhm, I can make the changes if you agree to them.

dbluhm:
I agree with the recommended changes. I wondered if it made sense to continue to test against 3.7 and 3.8 but given that I bumped the minimum version to 3.9, that wouldn't really make sense.

demo/run_bdd
dbluhm: I gave this only a few minutes thought but my thought process was that the `indy-base` target is roughly equivalent to `von-image`. I haven't looked closely at what is layered on top of von-image in the demo and bdd images so it may make sense to just use the full ACA-Py image.

swcurran:
Who’s got the ball on this one?  I’d really like to get this merged...

WadeBarnes:
In my court.  I'll get things done first thing in the morning.

WadeBarnes:
@dbluhm, I've made the changes, however I'm unable to push them to the [Indicio-tech:chore/drop-py3.6](https://github.com/Indicio-tech/aries-cloudagent-python/tree/chore/drop-py3.6) branch.

dbluhm:
@WadeBarnes I merged your branch locally and pushed but I'll see if I can push the right buttons to allow edits from maintainers in case any other changes are needed. It might be because I opened it from Indicio's fork and permissions on our org, not sure.

WadeBarnes:
> @WadeBarnes I merged your branch locally and pushed but I'll see if I can push the right buttons to allow edits from maintainers in case any other changes are needed. It might be because I opened it from Indicio's fork and permissions on our org, not sure.

Thanks,  You should see a checkbox on the PR for the edit option.
image

WadeBarnes:
Lots of timeout errors in the integration tests.  I'll have to look into those.

dbluhm:
image
No such box, unfortunately :disappointed: I suspect it is an org permissions thing. I'm happy to close this PR in favor of one you open. Also happy merge in changes as before. Whatever is most convenient for you!

WadeBarnes:
Lets figure out the issue with the test first and then decide how to proceed.

WadeBarnes:
Quick update on the integration test issue.  It's definitely something to do with the differences in the base images used for the bdd tests.  When using the original base image the tests run to completion.  When using the updates that use the indy-base image there are timeout errors.  What's strange is the issues only occur on GHA, when the same setup is used locally on an Ubuntu devContainer the tests run just fine. 

Run using the original base image setup:
https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/5147239795

Runs using the updated base image setup:
https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/5146808656/jobs/9266369701
- Using this setup, the tests run locally without issue.

WadeBarnes:
Suspecting a startup timing issue.  Testing that theory.

WadeBarnes:
The issues are related to the Python version of the base, `indy-base` container.  I've tested with 3.6, 3.7, 3.8, and 3.9.  The crashing and hanging issues start occurring with python 3.8 on GHA, yet locally using a docker-in-docker devContainer I can run the tests successfully with python 3.8 and 3.9.  With the `@T001-RFC0160` tests the `Mediator.Agent.8035` agent fails to start in a timely fashion, errors occur and the tests end up hanging.

WadeBarnes:
I need a little with this issue.  I'm not having much luck tracking down why the tests will run locally but not in GHA.  My most recent run of the tests with logging set to `info` is here; https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/5214202473/jobs/9410171574.  In the scenario were a mediator is created for both Acme and Bob, Bob's mediator never starts, and there are not logs to indicate any possible errors.  There is also this run with logging set to `debug` https://github.com/WadeBarnes/aries-cloudagent-python/actions/runs/5213743145/jobs/9409134016.

I'm hope I'm missing something simple.

dbluhm:
I'll see if I can squeeze in some time to analyze!

demo/multi-demo/Dockerfile.acapy
```diff
@@ -1,4 +1,4 @@
-FROM bcgovimages/aries-cloudagent:py36-1.16-1_1.0.0-rc0
+FROM FROM ghcr.io/hyperledger/aries-cloudagent-python:py3.9-indy-1.16.0-0.8.1
```
dkulic, line 1: Double FROM?

dkulic, line 1:
```suggestion
FROM ghcr.io/hyperledger/aries-cloudagent-python:py3.9-indy-1.16.0-0.8.1
```

dbluhm:
I attempted increasing timeouts on the off chance it simply wasn't ready in time and observed the same failures. I've reverted the timeout change.

dbluhm:
For posterity's sake, the issue was resolved by increasing the number of workers on the thread pool used to spawn the subprocesses for the mediators in the integration tests. On github action runners, the default pool size was 10 (number of processors * 5). This default changed in 3.8 and witnessed in our tests in 3.9:

> Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.


Found here: https://docs.python.org/3.9/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

So max number of workers dropped from 10 to 6 by default, preventing the mediator process from spawning because the pool was full.

dbluhm:
@swcurran What is the order of operations required to get this merged? We need to finalize 0.8.2, right?

WadeBarnes (approved):
LGTM

WadeBarnes:
Required checks have been updated to 
image

PR #2253, opened 2023-06-01
1.0.0-rc2 Updates

swcurran:
@WadeBarnes -- advice needed.

With this PR, we update the "version.py" and "openapi" versions to "1.0.0-rc2", which I don't we want to leave in place.

Here are some things we can do:

- After this PR is merged, add another that changes the "main" branch version to what it usually is -- "0.8.1" (the last released version).  That's the state we are usually in between releases.
   - Side Issue: Should we change this practice and after a release, change the version references in the code to what we expect the next release to be -- e.g. "0.8.2-latest" or something like that.
- Merge this PR into a new branch "1.0.0" and we use that for future "1.0.0-rcX" tags.  As part of making subsequent 1.0.0-rcX tags, we merge main into the 1.0.0 branch.  That means that 1.0.0 is a clone of main for tags, updated with the types of changes that are in this PR.

I kind of lean towards the "1.0.0" branch, but you know this stuff way better than I.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,57 @@
+# 1.0.0-rc2
+
+## April 5, 2023
+
+ACA-Py 1.0.0 will be a breaking update to ACA-Py with the version number
+indicating the maturity of the implementation. The final 1.0.0 release will be
+Aries Interop Profile 2.0-complete, and based on Python 3.9 or higher. The
+decision to start tagging early `1.0.0` releases was probably a bad one given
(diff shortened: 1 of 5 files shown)


swcurran:
@WadeBarnes -- branch, merge and release look good?  I called the branch `1.0.0-rc`, everything else is pretty typical for a release.

WadeBarnes:
@swcurran, Looks good.  The images were built off the correct commit.

PR #2261, opened 2023-06-09
Assign ~thread.thid with thread_id value

### LINKED ISSUE #2259: Messages with an empty ~thread decorator
0.8.1 is sending messages with an empty `~thread`. 

This can confuse callers as [RFC 0008](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0008-message-id-and-threading) appears ambiguous; it states what to do if `~thread` has not `thid` and what to do if no `~thread`, but not what to do with `~thread` with no `thid`.

If ACA-Py has no data in the `~thread`, then let's not include it.

See https://github.com/hyperledger/aries-vcx/issues/875

usingtechnology:
The previous assignment used a value from a field that had not been assigned yet. So, use the correct `thread_id` value for the `thid`.

Fixes #2259 

The fix originally discussed in #2259 is too aggressive and after reviewing the code it is clear that there can be empty `~thread` and they are expected to be dealt with by the receiver.  In this case, we were attempting to set the `thid` but just using an unassigned value.

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v1_0/manager.py
+++ b/aries_cloudagent/protocols/issue_credential/v1_0/manager.py
@@ -309,7 +309,7 @@ class CredentialManager:
             offers_attach=[CredentialOffer.wrap_indy_offer(credential_offer)],
         )
 
-        credential_offer_message._thread = {"thid": cred_ex_record.thread_id}
+        credential_offer_message._thread = {"thid": credential_offer_message._thread_id}
         credential_offer_message.assign_trace_decorator(
             self._profile.settings, cred_ex_record.trace
         )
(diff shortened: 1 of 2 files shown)


dbluhm (approved):
The new dev dependency is interesting. I usually just use the random package directly, telling it to select values from ascii (or something like that). But I'm happy with this change.

usingtechnology:
I can remove the dev dependency, I really only added it to kick off the tests again (failed in GH) as I do not have access to manually restart the checks.

swcurran:
I’d say remove the dependency.  We don’t want them lightly added.  And until you are a maintainer, please just get me to re-kickoff the tests. I can do that.

andrewwhitehead:
Should we be updating `cred_ex_record.thread_id` if it's unset?

usingtechnology:
> Should we be updating `cred_ex_record.thread_id` if it's unset?

It was always being set AFTER being used to set `thid`.

andrewwhitehead (approved):

PR #2262, opened 2023-06-09
feat(did creation route): reject unregistered did methods

chumbert:
Change a small oversight in did creation admin route:

* Missing `method` parameter will default to `sov`
* Unknown (ie: not registed in `DIDMethods`) methods will be rejected


dbluhm (approved):

swcurran:
This needs an update of the base-branch, please.

A question from a dummy looking at the code. It looks like the list of acceptable methods is kind of hard-coded (e.g. "SOV" in the code).  Is there somewhere (that I'm missing) where there are other permitted methods, and a way to add more?  Or is that not important in this context?

chumbert:
> A question from a dummy looking at the code. It looks like the list of acceptable methods is kind of hard-coded (e.g. "SOV" in the code). Is there somewhere (that I'm missing) where there are other permitted methods, and a way to add more? Or is that not important in this context?

A very good question. Within ACA-Py the methods are hardcoded and are `SOV` and `KEY`.

However, ACA-Py provides `DIDMethods`, a registry of supported DID methods. Plugins can use this registry to register additional methods. 

With this PR, the did creation endpoint checks the request method against the content of `DIDMethods` and rejects unknown methods. Without this PR, the endpoint would default any unknown method to SOV.

Issue #2251, opened 2023-05-31
For developers - onboarding documentation


usingtechnology:
For new (and old?) developers, clearly document (and keep up to date) best practices for developing, building, and testing locally. 

For instance, when opening a PR, there are a series of checks. These should be documented with instructions on how a developer would (if and where possible) run through these checks locally. We should use pre-commit hooks, VS Code dev containers etc, where possible, but those should not replace documentation and manual steps.

hiteshgh:
Test

Jsyro:
A few items on my radar

- [ ] Demos may need easy options to fully remove ngrok and run locally with two agents, rather than one agent that assumes a mobile wallet will play the other part. NGROK is taken for granted however certain corporate security policies block any traffic to the ngrok domain. 

- [ ] Unit test suite runs well as a whole, but using the unittest suite for the testing of individual files or tests are either very slow ,because they are configured to always run with `coverage`, or do not run at all. We should explore and refine a configuration that is fully supported by a given development tool (like VSCode)

- [ ] Find windows based python packages that allow for aca-py to be run natively on developers machine if possible.

swcurran:
Helpful to list the demos that you tried that had ngrok requirements, so that we can target them.

Our experience has been trying to maintain a useful environment on every platform (and variation of platform) is very hard, so we suggest that developers find ways to work on containers vs. native. Maintaining things to run on all developer platforms is a lot of overhead.  Doable, but not where we would like to spend our resources.

swcurran:
I would open a ticket asking how the last can be done.  How can a dev run a unit test in 2 seconds?  Someone must know :-)

usingtechnology:
I would like to see documentation simplified (lots and lots of different markdowns) and possibly gathered into streams: ACA-Py developer, ACA-Py user, ... business use cases?  

The demos folder has a lot of stuff in it; possibly move things out (playground). Also, we should really clarify what the purpose is and who the demographic is... I feel like the demographic is more for business case, but was constantly getting pointed to the run_demo because it does X. Then it is a learning curve to go through the shell scripts to see how ACA-Py is configured and started, then work through the agent code to actually find out what is happening to connect two agents.

Do we need multi-demo? If so, it needs a better "script".

[...]

PR #2267, opened 2023-06-14
Add devcontainer for ACA-Py

usingtechnology:
Addressing an issue where it is difficult to get set up to debug and begin coding by adding a devcontainer.

See Issue #2251 - this is only one small part of developer documentation. I wanted this out for feedback while we identify other areas of concern.

### DIFF
--- a/devcontainer.md
+++ b/devcontainer.md
@@ -10,7 +10,25 @@ By no means is ACA-Py limited to these tools; they are merely examples.
 
 The primary use case for this `devcontainer` is for developing, debugging and unit testing (pytest) the [aries_cloudagent](./aries_cloudagent) source code.
 
-There are limitations running this devcontainer, such as not being able to select the docker network. Also, we are not expecting developers to run the [demos](./demo) or other docker based scripts within this container.
+There are limitations running this devcontainer, such as all networking is within this container. This container has [docker-in-docker](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/docker-in-docker.md) which allows running demos, building docker images, running `docker compose` all within this container.
+
+### Files
+The `.devcontainer` folder contains the `devcontainer.json` file which defines this container. We are using a `Dockerfile` and `post-install.sh` to build and configure the container run image. The `Dockerfile` is simple but in place for simplifying image enhancements (ex. adding `poetry` to the image). The `post-install.sh` will install all the ACA-Py requirements and any additional steps (ex. adding additional development libraries).
+
+### Running docker-in-docker demos
+The following is an example of running the demos in this container using a local [von-network](https://github.com/bcgov/von-network/tree/main). You will have to connect to the local von-network using `host.docker.internal` not `localhost`.
+
+Assumes von-network is running outside of VS Code and this container at the default port (9000).
+
+```sh
+# open a terminal in VS Code...
+cd demo
+LEDGER_URL=http://host.docker.internal:9000 ./run_demo faber
+# open a second terminal in VS Code...
+cd demo
+LEDGER_URL=http://host.docker.internal:9000 ./run_demo alice
+# follow the script...
+```
 
 
 ## Further Reading and Links
--- a/devcontainer.md
+++ b/devcontainer.md
@@ -1,0 +1,109 @@
+For information on running demos and tests using provided shell scripts, see [DevReadMe](/DevReadMe.md) readme.
+
+# ACA-Py Development with Dev Container
+The following guide will get you up and running and developing/debugging ACA-Py as quickly as possible. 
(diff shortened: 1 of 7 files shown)


swcurran:
@amanji please check this out. @dbluhm, is there anyone on your team that regularly use Dev Containers?

amanji:
So far so good, I was able to load the devcontainer in VSCode without any issues.

devcontainer.md
```diff
@@ -0,0 +1,109 @@
...
+
+2. Open the root of this repository and you should be prompted to `Reopen in Container`.
+
+*NOTE* follow any prompts to install `Python Extension` or reload window for `Pylance` when first building the container.
+
+#### devcontainer.json
+
+When the [.devcontainer/devcontainer.json](.devcontainer/devcontainer.json) is opened, you will see it building... what it is building is a Python 3.9 image and loading it with ACA-Py requiments (and black). Since this is a Docker container, we will also open ports `9060` and `9061` which will allow you to run ACA-Py with those ports available to your `localhost`. We also load a few Visual Studio settings (for running Pytests and formatting with Flake and Black).
+
+In VS Code, open a Terminal, you should be able to run the following commands:
```
loneil, line 50: Defaults a bash terminal (assume this comes from a default from the devcontainer) for me, think that's correct but not sure if bash specifically is required for the instructions? IE, should this say "open a Bash Terminal"

devcontainer.md
```diff
@@ -0,0 +1,109 @@
...
+            "module": "aries_cloudagent",
+            "justMyCode": true,
+            "args": [
+                "start",
+                "--arg-file=${workspaceRoot}/.vscode/multitenant.yml"
+            ]
+        },
+```
+
+This will start your source code as a running ACA-Py instance, all configuration is in the `multitenant.yml` file. This is just a sample of a configuration, and we chose multi-tenancy so we can easily create multiple wallets/agents and have them interact.  Note that we are not using a database and are joining the ` http://test.bcovrin.vonx.io` ledger. Feel free to change to a local VON Network (by default, it would be `http://host.docker.internal:9000`) or another ledger. This is purposefully a very simple configuration.
```
loneil, line 100: Could note ways to run this launch config (discussed in RC)

loneil (commented):
These are great! couple small comments
I started from scratch, didn't even have the dev container extension installed and can get things going with the instructions and container config here

devcontainer.md
```diff
@@ -0,0 +1,111 @@
...
+- Dev Container Extension Version: v0.295.0
+
+### Open ACA-Py in the devcontainer
+
+There are 2 ways to open ACA-Py into the devcontainer and start developing.
+
+1. Open Visual Studio Code, and use the Command Palette to open the devcontainer.
+    
+    ```
+    > Dev Containers: Open Folder in Container...
```
swcurran, line 39: I would suggest adding what folder to select.  I think you select the root of ACA-Py, vs. `.devcontainer` which is kind of what I expected.  Real devs might have a better intuition...

devcontainer.md
usingtechnology: addressed

devcontainer.md
usingtechnology: yeah, it'll be bash for all. noted in the readme now.

devcontainer.md
usingtechnology: done.

swcurran:
Something that might be worth trying. Evidently, docker-in-docker is possible with devcontainers, as per this: https://github.com/devcontainers/features/tree/main/src/docker-in-docker.  It might be interesting to see if with that added to the devcontainer config if the ACA-Py demos can be run.

usingtechnology:
Yes, I took a quick look at docker-in-docker, but didn't explore it fully (or try to implement it). I could spend more time on it if we want.

swcurran:
How about for now, we just add a note about it in the readme, and an issue with a “help wanted” sign.

dbluhm:
I don't have anyone on my team actively using devcontainers right now but I think this would be a great addition. I agree with Stephen on getting this merged sooner than later and plan on making further enhancements in the future.

dbluhm (approved):

usingtechnology:
added docker-in-docker and updated the devcontainer readme to show how to run the demos within the container.

WadeBarnes:
For the devContainer, I'd recommend using `bullseye` rather than `buster` to keeps things consistent with the published images.

usingtechnology:
> For the devContainer, I'd recommend using `bullseye` rather than `buster` to keeps things consistent with the published images.

Thanks for that... updated

swcurran (approved):
Tested out with running the demo and worked nicely. A couple of notes we can add, but overall looks good.  Great work!

PR #2270, opened 2023-06-15
Update requests lib 2.31.0

usingtechnology:
See https://github.com/hyperledger/aries-cloudagent-python/security/dependabot/3


usingtechnology:
Closing this, update is not available for Python 3.6 and we do not use the vulnerable code.

swcurran:
Thanks for addressing this one.

PR #2283, opened 2023-06-27
Update anoncreds-rs to v0.1.0

usingtechnology:
updating references to the anoncreds lib to use the v0.1.0 release.

### DIFF
--- a/docker/Dockerfile.indy
+++ b/docker/Dockerfile.indy
@@ -197,7 +197,6 @@ RUN pip3 install --no-cache-dir \
     -r requirements.bbs.txt \
     -r requirements.dev.txt \
     -r requirements.anoncreds.txt
-RUN curl -sL https://github.com/hyperledger/anoncreds-rs/releases/download/v0.1.0/library-linux-x86_64.tar.gz | tar -xz -C /usr/local/lib/python3.9/site-packages/anoncreds/
 
 ADD --chown=indy:root . .
 USER indy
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.indy
```diff
@@ -197,7 +197,7 @@ RUN pip3 install --no-cache-dir \
     -r requirements.bbs.txt \
     -r requirements.dev.txt \
     -r requirements.anoncreds.txt
-RUN curl -sL https://github.com/Indicio-tech/anoncreds-rs/releases/download/v0.1.0-dev.9/library-linux-x86_64.tar.gz | tar -xz -C /usr/local/lib/python3.9/site-packages/anoncreds/
+RUN curl -sL https://github.com/hyperledger/anoncreds-rs/releases/download/v0.1.0/library-linux-x86_64.tar.gz | tar -xz -C /usr/local/lib/python3.9/site-packages/anoncreds/
```
dbluhm, line 200: Seeing that we now have a build on PyPI that includes bundled binaries (as of June 2; I had missed that until now :slightly_smiling_face:), this separate step for retrieving the binary is no longer necessary and can be removed from each of the Dockerfiles

usingtechnology: Thanks, that answers a question I was going to pose later. Will remove those commands.

dbluhm (approved):

PR #2292, opened 2023-07-10
feat(anoncreds): Implement automated setup of revocation

dbluhm:
This PR implements automated setup of revocation by introducing an `AnonCredsRevocationSetupManager` base class and `DefaultRevocationSetup` concrete class and adding the appropriate hooks to the artifact registration steps.

Overall, I'm very pleased with the simplicity of this setup manager as compared to our original mechanism. It works in more or less the same way, listening for events and triggering the next step in the flow; however, all the event listeners are defined in one place (rather than across 2 or 3 modules) and the event objects themselves have been formalized to make accessing the information in the event simpler. Additionally, encapsulation of the anoncreds interface is better than what we had previously so the event listeners themselves are fewer lines with significantly more straightforward logic.

I've left the door open for enabling plugins to implement their own `AnonCredsRevocationSetupManager` if there is ever a need to customize the setup behavior. I'm considering moving the `handle_full_registry` logic to this same component for the same reason -- giving plugin authors the ability to customize if needed -- but I'm not sure how likely it is anyone will actually need or want to do customizations.

This PR invalidates the following temporary endpoints we added for testing:

- `POST /anoncreds/revocation-registry-definition`
- `PUT /anoncreds/registry/{rev_reg_def_id}/tails-file`
- `PUT /anoncreds/registry/{rev_reg_def_id}/active`
- `POST /anoncreds/revocation-list`

I haven't removed these endpoints just yet in this PR though.

This PR adds `max_cred_num` to the expected options of the `POST /anoncreds/credential-definition` endpoint.

### DIFF
--- a/aries_cloudagent/anoncreds/events.py
+++ b/aries_cloudagent/anoncreds/events.py
@@ -1,0 +1,81 @@
+"""Events fired by AnonCreds interface."""
+
+import re
+from typing import NamedTuple, Optional
+
+from aries_cloudagent.anoncreds.models.anoncreds_revocation import RevRegDef
+from ..core.event_bus import Event
+
+
+CRED_DEF_FINISHED_EVENT = "anoncreds::credential-definition::finished"
+REV_REG_DEF_FINISHED_EVENT = "anoncreds::revocation-registry-definition::finished"
+REV_LIST_FINISHED_EVENT = "anoncreds::revocation-list::finished"
+
+CRED_DEF_FINISHED_PATTERN = re.compile("anoncreds::credential-definition::finished")
+REV_REG_DEF_FINISHED_PATTERN = re.compile(
+    "anoncreds::revocation-registry-definition::finished"
+)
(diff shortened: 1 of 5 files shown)


dbluhm:
Updated gist with tests used to validate these changes: https://gist.github.com/dbluhm/4d33d8946d23dbaced9be825c49a9d5c/revisions

We really need to get these tests into the bdd tests on the anoncreds-rs branch but this is helping to fill the gap at the moment.

swcurran (approved):

swcurran:
I’m going to go ahead and merge this.  Are the failing tests because of Python 3.6, or because we need to remove some tests that are no longer applicable?

dbluhm:
The latter; getting testing setup to a better state is high on the to-do list now that the automatic setup is completed

swcurran:
@usingtechnology has an item on his list right now, but we were thinking implementing the BDD tests would be next on his list.

PR #2295, opened 2023-07-11
Allow any did to be public

mkempa:
Having a universal resolver set in the configuration, I want to be able to create and receive invitations containing a public DID other than did:sov (Indy DID). In order to do that I must promote the DID to public after it has been created in the wallet.

This PR adds checks of the DID format to associate ledger operations to the did:sov only in the said operation, thus allowing any DID format to be set as public.

### DIFF
--- a/aries_cloudagent/ledger/base.py
+++ b/aries_cloudagent/ledger/base.py
@@ -10,6 +10,7 @@ from hashlib import sha256
 from typing import List, Sequence, Tuple, Union
 
 from ..indy.issuer import DEFAULT_CRED_DEF_TAG, IndyIssuer, IndyIssuerError
+from ..messaging.valid import IndyDID
 from ..utils import sentinel
 from ..wallet.did_info import DIDInfo
 
@@ -294,6 +295,11 @@ class BaseLedger(ABC, metaclass=ABCMeta):
         public_info = await self.get_wallet_public_did()
         if not public_info:
             raise BadLedgerRequestError("Cannot publish schema without a public DID")
+
+        if not bool(IndyDID.PATTERN.match(public_info.did)):
+            raise BadLedgerRequestError(
+                "Cannot publish schema when public DID is not an IndyDID"
+            )
 
         schema_info = await self.check_existing_schema(
             public_info.did, schema_name, schema_version, attribute_names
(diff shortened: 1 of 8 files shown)


dbluhm:
@chumbert Interested in your thoughts on these changes

chumbert:
> @chumbert Interested in your thoughts on these changes

Hello ! The changes themselves are fine. My personal preference would be quite outside of the scope of this PR.

Lately people have been contributing code that got us rid of the
```
if this did_method:
...
elif this did_method:
```
logic for DIDs and verification methods. I think it is time we also consider a pluggable way to handle did registration.

Now back to reality: in our deployments we have side-stepped the `publish did` endpoint completely for our own `did:web` needs:

```
@docs(
    tags=[OPENAPI_TAG],
    summary="Set public DID of the wallet."
    "This is limited to the wallet, no ledger interaction.",
)
@match_info_schema(DIDSchema())
@response_schema(DIDResultSchema, 200, description="The updated did.")
async def set_public_did(request: web.Request):
    did = request.match_info.get("did")
    if not did:
        raise web.HTTPBadRequest(reason="Request query must include DID")

    async with context.profile.transaction() as transaction:
        did_info = await transaction.inject(BaseWallet).set_public_did(did)
        await transaction.commit()

    return web.json_response(
        data={
            "did": did_info.did,
            "verkey": did_info.verkey,
            "posture": DIDPosture.get(did_info.metadata).moniker,
            "key_type": did_info.key_type.key_type,
            "method": did_info.method.method_name,
        }
    )
```

I'm working internally to be able to publish and share with you our whole plugin soon. In addition to marking DIDs public, it also has an endpoint to configure wallet routing for public did in multitenancy scenarios (impossible to receive connection requests based on implicit invitations otherwise), and other things less relevant to this discussion.

chumbert:
Quick update on the plugin we currently use: https://github.com/sicpa-dlab/aries-acapy-plugin-did-management

swcurran:
@mkempa -- could you please resolve the conflicts?  @dbluhm -- could you review?  Looks like @chumbert likes it, so hopefully easy to do.

dbluhm (dismissed):

PR #2302, opened 2023-07-11
Update to indy-credx 1.0

andrewwhitehead:


dbluhm (dismissed):

dbluhm (dismissed):

andrewwhitehead:
The `--refuse-legacy-revocation` flag could still use some testing, I'm looking at that now.

swcurran:
@andrewwhitehead -- are you able to update this?  Status on what's keeping it in Draft state?

Thanks!

dbluhm:
I put together a quick test setup to validate these changes and I think it looks like everything is working. My setup has an "Alice" and a "Bob" agent:

- Alice is acting as an issuer and verifier
- Bob is a holder and prover
- Alice is running on an image built from this branch
- Bob is running on 0.8.2
- Alice has `--anoncreds-legacy-revocation=reject`

In my scenario:

- Alice goes through credential issuance setup (TAA, public DID, schema, cred def, rev reg, etc.)
- Alice issues a credential to Bob
- Alice requests presentation from Bob without `non_revoked`; verified as `true`, not sure what to actually expect here
- Alice revokes the credential
- Alice requests presentation from Bob with `non_revoked` after revocation; verified as `false`, would normally verify as `false` but expected `false` anyways because of `--anoncreds-legacy-revocation=reject`
- Alice requests presentation from Bob with `non_revoked` before revocation; verified as `false`, would normally verify as `true`, expected `false` because of `--anoncreds-legacy-revocation=reject`

dbluhm:
I'll run some more tests to hit the other cases:
- accepts old when set to accept
- accepts new when set to reject

andrewwhitehead:
I think this is ready then, but I'm currently getting a server error when trying to move it out of draft.

dbluhm (approved):

PR #2305, opened 2023-07-12
Fix alice/faber demo execution

andrewwhitehead:
This PR adds the `--debug-webhooks` flag for all demo agents, as it is required to distinguish between credential offers and presentation requests for different credential types.

It also works around an issue with #2234 which causes the agents to exit with a RuntimeError due to the call to `loop.run_until_complete` within the logging configuration. I think a more complete fix might be to use a ContextVar to store the current DID identifier and set that when a tenant context is initialized, to avoid any async operations in the logger config.


swcurran:
@andrewwhitehead — how do I create the problem on the current main branch?  I can run the demo, issue a credential, request a proof, and all works fine.  What am I missing?

swcurran (approved):

Issue #2304, opened 2023-07-12
Add a new endpoint to use a new RevReg


swcurran:
From time to time an Issuer may want to issue credentials from a new RevReg. That can be done by changing the CredDef, but that could impact verifiers. An alternative is to mark the existing RevRegs as "full" and trigger the creation of new RevRegs.  This task is to add an endpoint to enable that.  As with all RevReg handling in ACA-Py, we want the effort on the Issuer controller to be minimized, so this endpoint provides a simple interface, and ACA-Py must handle the details of "making it happen."

Recall that with ACA-Py, there are is always one active and one ready RevReg, so that there is never a delay in issuing because a RevReg is full. The following is a high-level description of what happens -- please consult the code for the technical details of how it happens.

* On creation of a CredDef that is revocable, two RevRegs are created, one after the other, with first marked as Active, and the second Ready.
* Whenever the currently Active is full (credentials have been issued up to the size of the RevReg), the RevReg is marked as "Full", the "Ready" RevReg becomes "Active", and a new RevReg is created and becomes the "Ready" RevReg. This occurs automatically, and the controller does not have to do anything.

For this task, the controller will invoke an endpoint -- say "/revocation/decommission" (please make the endpoint name fit the pattern in ACA-Py).  The term decommission came to me on the fly as I wrote this, so definitely not stuck on it.  The task should logically do the following -- although again, the technical details must be worked out. It make sense to me that the mechanisms rely as much as possible on the existing logic around when a new RevReg needs to be created.

* Mark the first RevReg as full, which might automatically (or via another call) trigger the creation of another RevReg.
* Mark the second RevReg as full, which again, might automatically (or via another call) trigger the creation of another RevReg.

The result should be that the two existing RevRegs are no longer used, and any new credentials issued come from the first of the new RevRegs -- the "Active" one, and the second is marked as "Ready".

It would be nice, but not required, that credential issuance stop until the first of the new RevRegs is active, so that no additional credentials are issued with the old RevReg. Since the creation of RevRegs and the Issuances of Credentials are not synchronized (unless there are no active RevRegs -- which we are working hard to avoid), if an issuance occurs with either of the "deactivated" RevRegs, that is OK.

[...]

PR #2309, opened 2023-07-18
API endpoint to decommission revocation registry

usingtechnology:
Closes issue #2304 

There are two active registries at a given time, this API endpoint - given a `cred_def_id` - will find and decommission the active revocation registries and will kick off the procedure to create new active registries (same as when a registry is full).

Note that the `pytest` doesn't have a handler to transition the `init` state registries to `active`, so we simulate that before testing the decommission code.

### DIFF
--- a/aries_cloudagent/revocation/indy.py
+++ b/aries_cloudagent/revocation/indy.py
@@ -19,6 +19,7 @@ from ..storage.base import StorageNotFoundError
 
 from .error import (
     RevocationError,
+    RevocationInvalidStateValueError,
     RevocationNotSupportedError,
     RevocationRegistryBadSizeError,
 )
@@ -105,25 +106,42 @@ class IndyRevocation:
 
         return record
 
-    async def handle_full_registry(self, revoc_reg_id: str):
+    async def handle_full_registry(self, revoc_reg_id: str) -> IssuerRevRegRecord:
         """Update the registry status and start the next registry generation."""
+        return await self._set_registry_status_registry(
+            revoc_reg_id, IssuerRevRegRecord.STATE_FULL
+        )
+
+    async def decommission_registry(self, revoc_reg_id: str) -> IssuerRevRegRecord:
+        """Update the registry status and start the next registry generation."""
+        return await self._set_registry_status_registry(
+            revoc_reg_id, IssuerRevRegRecord.STATE_DECOMMISSIONED
+        )
+
+    async def _set_registry_status_registry(self, revoc_reg_id: str, state: str):
+        """Update the registry status and start the next registry generation."""
+        if state not in IssuerRevRegRecord.STATES:
+            raise RevocationInvalidStateValueError(
+                reason=f"{state} is not a valid Revocation Registry state value."
+            )
         async with self._profile.transaction() as txn:
             registry = await IssuerRevRegRecord.retrieve_by_revoc_reg_id(
                 txn, revoc_reg_id, for_update=True
             )
-            if registry.state == IssuerRevRegRecord.STATE_FULL:
+            if registry.state == state:
                 return
             await registry.set_state(
                 txn,
-                IssuerRevRegRecord.STATE_FULL,
+                state,
             )
             await txn.commit()
 
-        await self.init_issuer_registry(
-            registry.cred_def_id,
-            registry.max_cred_num,
-            registry.revoc_def_type,
-        )
+        if state in IssuerRevRegRecord.TERMINAL_STATES:
+            return await self.init_issuer_registry(
+                registry.cred_def_id,
+                registry.max_cred_num,
+                registry.revoc_def_type,
+            )
 
     async def get_active_issuer_rev_reg_record(
         self, cred_def_id: str
--- a/aries_cloudagent/revocation/indy.py
+++ b/aries_cloudagent/revocation/indy.py
@@ -1,5 +1,5 @@
 """Indy revocation registry management."""
-
+import logging
 from typing import Optional, Sequence, Tuple
 from uuid import uuid4
 
@@ -26,6 +26,8 @@ from .error import (
 from .models.issuer_rev_reg_record import IssuerRevRegRecord
 from .models.revocation_registry import RevocationRegistry
 from .util import notify_revocation_reg_init_event
+
+LOGGER = logging.getLogger(__name__)
 
 
 class IndyRevocation:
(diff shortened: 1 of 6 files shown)


swcurran:
Maintainers — @dbluhm @andrewwhitehead @esune — please take a look. This stands between us and the 0.9.0-rc0 release.  I’d like to get that out there today.

aries_cloudagent/revocation/indy.py
```diff
@@ -107,23 +110,57 @@ async def init_issuer_registry(
...
+        await self._set_registry_status_registry(
+            revoc_reg_id, IssuerRevRegRecord.STATE_FULL
+        )
+
+    async def decommission_registry(self, cred_def_id: str):
+        """Decommission active registries and start the next registry generation."""
+        async with self._profile.session() as session:
+            active = sorted(
+                await IssuerRevRegRecord.query_by_cred_def_id(
+                    session, cred_def_id, IssuerRevRegRecord.STATE_ACTIVE
```
dbluhm, line 122: I would have to dig a bit to back this up but I believe there's only one "active" registry at a time and then the hot spare is in state "posted" and switched to active when the previous active is full. What I think that means is that all registries of either active or posted state would need to be retrieved and decommissioned.

dbluhm (changes requested):

aries_cloudagent/revocation/indy.py
dbluhm: I wrote this comment before seeing the others above :sweat_smile: I'll do that digging to confirm

usingtechnology:
i was running my smoke tests in multitenant mode, but wouldn't think that had any bearing. When creating a cred_def, the two registries were `active`, getting the currently active was `# active record is oldest published but not full`.

dbluhm:
I should have looked closer:

```
    STATE_POSTED = "posted"  # definition published
    STATE_ACTIVE = "active"  # initial entry published, possibly subsequent entries
```

So active is the expected state when the initial entry has been published. The initial entry is sent immediately after the definition is final. However, if using an endorser, there could be an indeterminate amount of time between definition and initial entry. Likewise, there is an indeterminate amount of time between generated and posted state. We should account for this. I think the correct answer is to consider all registries that have been generated as invalid and should be decommissioned.

If we decommissioned registries of all states, finally receiving the endorsed transaction could cause an event handler that clobbers the decommissioned state. So we should watch out for that too.

usingtechnology:
@dbluhm - looking at [indy_rev_reg_record.send_entry](https://github.com/hyperledger/aries-cloudagent-python/blob/ad415700338ff81ff6da54c3e26e1c1dbb367a0e/aries_cloudagent/revocation/models/issuer_rev_reg_record.py#L285)...
I think if we set active/posted to decommissioned and then set the above to include decommissioned we will cover our bases? that will allow "closed" registries to send entries.

dbluhm:
> @dbluhm - looking at [indy_rev_reg_record.send_entry](https://github.com/hyperledger/aries-cloudagent-python/blob/ad415700338ff81ff6da54c3e26e1c1dbb367a0e/aries_cloudagent/revocation/models/issuer_rev_reg_record.py#L285)... I think if we set active/posted to decommissioned and then set the above to include decommissioned we will cover our bases? that will allow "closed" registries to send entries.

Yeah, I think that would enable us to continue to revoke creds on old registries as needed but prevent issuing any new creds against those registries.

To be perhaps overly specific, I was referring to these event handlers: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/revocation/routes.py#L1299-L1500

These govern the advancement of registry through the states, accounting for needing to wait on endorser signature. So in addition to decommissioning all registries >= generated, we need to make sure any pending transaction endorser exchagnes don't result in a registry erroneously advancing from decommissioned back to `posted` or `active`.

usingtechnology:
Updated to decommission all (except init). I can't see any state transitions that will affect registries in the `decommissioned` state. I did remove `decommissioned` selection from `set-state` - so we don't have callers changing the state inadvertently.

Still need to update the documentation.

aries_cloudagent/revocation/routes.py
```diff
@@ -1570,6 +1600,10 @@ async def register(app: web.Application):
                 get_active_rev_reg,
                 allow_head=False,
             ),
+            web.post(
+                "/revocation/active-registry/{cred_def_id}/decommission",
```
esune, line 1604: I am wondering if the name of the endpoint should be changed to something like `rotate` rather than decommision, since this is the actual outcome of this action. It would make the intent clearer in my opinion.

esune (changes requested):

aries_cloudagent/revocation/routes.py
usingtechnology: I like that. I think `decommissioned` is a good state name, but I like `rotate` for the API call.

swcurran: I agree — the state is right, but I like `rotate` for the API call.

usingtechnology:
Added a blurb to docs/GettingStartedAriesDev/CredentialRevocation.md

Update the endpoint to be `rotate`; the state is still `decommissioned`.

aries_cloudagent/revocation/tests/test_indy.py
```diff
@@ -154,6 +154,74 @@ async def test_list_issuer_registries(self):
 
         assert len(await self.revoc.list_issuer_registries()) == 2
 
+    async def test_decommision_issuer_registries(self):
```
swcurran, line 157: Typo in decommission — missing s

usingtechnology: d'oh... I don't know how many times I've typed that incorrectly... thanks for the catch.

swcurran:
I think this was covered in the discussion above, but a decommissioned registry can still have its credentials revoked and published, correct?  That should be supported.

usingtechnology:
> I think this was covered in the discussion above, but a decommissioned registry can still have its credentials revoked and published, correct? That should be supported.

Yes, [see here](https://github.com/hyperledger/aries-cloudagent-python/blob/82b4c4967fe0342fcf40d3de0ee4eaaf520c72d1/aries_cloudagent/revocation/models/issuer_rev_reg_record.py#L299), we added decommissioned to the set of states that allow revoke/publishing.

aries_cloudagent/revocation/indy.py
```diff
@@ -107,23 +110,60 @@ async def init_issuer_registry(
...
+            await self._set_registry_status_registry(
+                rec.revoc_reg_id, IssuerRevRegRecord.STATE_DECOMMISSIONED, init
+            )
+            init = False  # only call init once.
+
+        return recs
+
+    async def _set_registry_status_registry(
+        self, revoc_reg_id: str, state: str, init: bool = True
+    ):
```
dbluhm, line 143: It doesn't look like `init` is used anywhere inside this method? Also, a nitpick: I feel like there's one too many "registry" in the name of this method.

usingtechnology: haha, there certainly is one too many "registry" in that name. great catch on both accounts... i think the init got lost in a code merge/rebase.

touching up now.

dbluhm (approved):

andrewwhitehead (approved):

swcurran:
Ah…I didn’t know you could dismiss an objection!  :-).  Thanks.  I’ll get going on 0.9.0-rc0.

PR #2313, opened 2023-07-19
Add workaround for ARM based macs

### LINKED ISSUE #2311: Dependency bugs when starting ACA-py agents using Docker
Hi all, I am new to the Hyperledger space and trying to get my first Aries agents up and running using docker. I have been running into several issues while starting and I was hoping to get some help from everyone. Firstly, I am running all of this on an M1 MacBook Pro with macos Ventura 13.2.1.

Following the documentation [here](https://github.com/hyperledger/aries-cloudagent-python/blob/main/DevReadMe.md#docker) I have been attempting to run ACA-py agents using the `scripts/run_docker start` command as detailed in the README above. A von-network server is already running with an 'Alice' DID written to the ledger. The exact command used is:

```
scripts/run_docker start \
--inbound-transport http 0.0.0.0 8000 \
--outbound-transport http \
--admin 0.0.0.0 11000 \
--admin-insecure-mode \
--genesis-url http://host.docker.internal:9000/genesis \
--seed Alice000000000000000000000000001 \
--endpoint http://localhost:8000/ \
--debug-connections \
--auto-provision \
--wallet-type indy \
--wallet-name Alice1 \
--wallet-key secret
```

Bugs:

1. No matching distributions can be found for the `indy-vdr~=0.3.3` and `ursa-bbs-signatures~=1.0.1` modules when installing the requirements. Commenting out these dependencies fix these issues but would this have further downstream affects and is there a better solution?

2. Using the 'fix' above the agent fails to start due to `aries_cloudagent.utils.classloader.ModuleLoadError: Unable to import module aries_cloudagent.indy.sdk.profile: No module named 'indy'`, however, I checked the requirements and `python3-indy` is already installed.

Any and all help is much appreciated thank you very much!

finnformica:
- ARM based macs caused dependency issues with indy-vdr and ursa-bbs-signatures
- Solution checks for architecture and updates DOCKER_DEFAULT_PLATFORM
- Closes #2311


swcurran:
Thanks for this PR!  Sorry you ran into this — we try to keep the getting started guides working.  Obviously we need to take another look and make sure that all are working. Thanks for your patience.

@andrewwhitehead — can you please review this PR and give it a go on your Mac?  Thanks!

andrewwhitehead:
Looks reasonable, although I think anybody running Linux on an ARM Mac would have the same issue. Hopefully we'll have a compatible release of indy-vdr soon (0.4), although we'll need to get ursa-bbs-signatures published as well.

swcurran:
@andrewwhitehead — could you please look at the test failure?  I assume this is unrelated, but just to be sure.

andrewwhitehead:
Confirmed that it runs on my M1 Mac. The integration test failure I saw was due to a timeout connecting to storage, probably something in Github's backend.

swcurran (approved):

PR #2327, opened 2023-07-20
Fix support for non-unix-like OS

dbluhm:
Supersedes #2211. Thanks @sk13. There was a small typo that was putting a wrench in things. I don't have a good way to test these changes on Windows (my dev environment is on Linux). I'm also not sure about the workaround for signal handling.

On top of that, I get the impression there might be deeper challenges when it comes to Windows support. I'm not sure that's a can of worms we want to open right now. Open to thoughts and discussion but wanted to at least get this changed fix for consideration.

cc @WadeBarnes


WadeBarnes:
Honestly, even though I use a PC running Windows, I've never run or developed ACA-Py in a native Windows environment.  I've never even considered that as a option.  For runtime I've always used containers and for any development I've used a Linux devContainer with VSCode (or WSL or a Linux box).  That said, I do agree we shouldn't be making platform specific assumptions (where reasonably possible), such as which path separators are used.

I had a quick look through the project and I see a few other places (mainly around tests associated with tails files) where the Linux path separator (`/`) is hard coded in the paths.  So we can expect some issues with the tests at least, but likely with other code associated with handling tails files too.

I too am unsure about the workaround for signal handling.

If we are serious about officially providing native support for an OS like Windows, I think the best way to start is to update the tests (perhaps staring with the nighty ones, since they are non-blocking) to run on Windows.  From there we can fix any issues that arise and when ready update the PR blocking tests to also run on Windows.  That way we're ensuring any development is cross platform compatible.

swcurran:
Are the tests failing because of this issue or intermittent failures.  If this issue, then I would say we close it and put in a “Help Wanted” issue for someone to do what @WadeBarnes proposes — create a Windows job for testing and work through getting the tests to work on all platforms.

That said — I agree with Wade that bare metal development is a questionable choice — especially when we have full debugging support available in a container environment.

dbluhm:
Looks like test failures are a result of minor formatting issues; however, in order to really validate these changes, I think we'd basically need those nightly tests anyways. I'm in favor of moving in the direction Wade suggests (or rather opening an issue documenting it at least) while strongly encouraging contributors to use things like dev containers or WSL. I consider Windows support to be a low priority right now.

I'm going to close this PR. Thanks for the discussion and thoughts!

WadeBarnes:
The Black Formatter error is related, but I don't think the test failures are related.

dbluhm:
I believe the test failure was flake8 complaining about whitespace but then pytest-flake8 choked. I think that issue should be fixed by dependency updates.

PR #2335, opened 2023-07-21
⬆️ upgrade `pyjwt` to latest; introduce leeway to `jwt.decode`

ff137:
Latest release (2.8) introduces support for python 3.10+

Release notes: https://github.com/jpadilla/pyjwt/releases

Note: `jwt.decode` can fail spuriously due to breaking changes introduced in https://github.com/jpadilla/pyjwt/pull/797, where an exception is raised if the 'issued at' time of the JWT is in the future when decoding, with default 0 leeway. This means that a miniscule mismatch in clock synchronization, between generation and decoding of JWT, can now cause a failure when decoding. This arose in the integration tests, and is resolved with a leeway of 1 second.


ff137:
Some work required here, I can have a look after the weekend

ff137:
Test failures with `pyjwt~=2.6.0`:
```
Failing scenarios:
  features/0160-connection.feature:19  establish a connection between two agents -- @1.4 
  features/0160-connection.feature:20  establish a connection between two agents -- @1.5 
  features/0453-issue-credential.feature:40  Issue a json-ld credential with the Issuer beginning with an offer -- @1.4 
  features/0453-issue-credential.feature:58  Issue a credential with revocation, with the Issuer beginning with an offer, and then revoking the 
```

ff137:
Breaking change is from: https://github.com/jpadilla/pyjwt/pull/797

This introduces stricter validation for the "iat" (issued at) claim in JWT tokens. If the "iat" value is in the future, even just a few seconds, PyJWT will raise an `ImmatureSignatureError`.

ff137:
@swcurran  Any chance to include with 0.9.0 release / #2344 ?

dbluhm:
To understand the fix, we're permitting leeway of 1 second? 1 ms?

swcurran:
I think we need to get 0.9.0 out.  We’ve got several hot items to follow quickly on the tail of 0.9.0 so we’ll put this there (and we can do an `rc0` early if this is needed). But I think we need to get 0.9.0 ahead of this one.

Sorry about that.  We’ll be talking about the next release at the ACA-Pug meeting tomorrow.

dbluhm (approved):
@swcurran I think this would be a low risk addition but happy to defer to your decision (I was a split second behind on my approval lol).

ff137:
All good! It just resolves a python dependency conflict on a local test environment, but not a breaking one

And the 1s leeway could be shorter, but it's a reiterative process to halve until it breaks again - with builds taking 35 minutes, I figured 1s is good enough :-) can't think of any exploit for this

swcurran:
@dbluhm convinced me :-).  Merged and will be in 0.9.0.

PR #2339, opened 2023-07-21
Feat: Support Selectable Write Ledger

### LINKED ISSUE #2168: multi-ledger write failure on register-nym
version: aries-cloudagent-python-1.0.0-rc1
multitenant = true
wallet type = askar AND indy

tests were performed locally using a local endorser that is preconfigured. the production and write ledger is `http://test.bcovrin.vonx.io/genesis`.

Cannot successfully call `ledger/register-nym` for an agent/wallet, with a valid endorser connection when running in multiledger mode (`ACAPY_GENESIS_TRANSACTIONS_LIST=ledgers.yml`) but the same code will work when running in single ledger mode (`ACAPY_GENESIS_URL=http://test.bcovrin.vonx.io/genesis`)

I have a branch in [traction](https://github.com/bcgov/traction/tree/feature/multiledger) that I was using to run the tests and there are a lot of notes in this [hack md](https://hackmd.io/C3vImVHIR5OyR3Mjr4m28A?edit).

I ran both scenarios (multi ledger vs single ledger) with indy and askar wallets and both wallets gave the same results (at the api level). Both indy and askar were successful with single ledger and both failed with multi ledger: `Cannot update endpoint at ledger without a public DID`.  See [hackmd - where error seems to be](https://hackmd.io/C3vImVHIR5OyR3Mjr4m28A?view#Where-error-seems-to-be…)

I dug in a lot deeper and event adjusted some of the acapy code to get as far into the call stack as I could, and then both fail similarly with `WalletItemNotFound` (indy) `WalletNotFoundError` (askar). See [hackmd - focused failure](https://hackmd.io/C3vImVHIR5OyR3Mjr4m28A?view#Focused-failure-askar-and-indy)

It appears that there is some context switching or changing of sessions when we get into multi ledger. This is pure speculation on my part... But in the same call stack, we can see the wallet record containing the public did information, but further down the stack, the same call returns `None`.  

To get as deep into the code as I could, I passed in the `public_info` (DIDInfo) into wallet and ledger calls just so they wouldn't fail on lookup, but of course, eventually we hit a point where I couldn't change the code.

Please refer to [hackmd - deeper testing](https://hackmd.io/C3vImVHIR5OyR3Mjr4m28A?view#Deeper-testing) to see where indy and askar eventually end up failing for me.

Bottom line is there is some kind of low-level difference between multi ledger and single ledger that is preventing an agent/wallet from successfully calling `ledger/register-nym`

shaangill025:
- resolve #2168

### DIFF
--- a/Multiledger.md
+++ b/Multiledger.md
@@ -44,6 +44,8 @@ If `--genesis-transactions-list` is specified, then `--genesis-url, --genesis-fi
   is_production: true
   is_write: true
   genesis_url: 'http://test.bcovrin.vonx.io/genesis'
+  endorser_did: '9QPa6tHvBHttLg6U4xvviv'
+  endorser_alias: 'endorser_test'
 - id: greenlightDev
   is_production: true
   is_write: true
@@ -84,6 +86,10 @@ Optional properties:
 - `keepalive`: how many seconds to keep the ledger open
 - `socks_proxy`
 - `is_write`: Whether this ledger is writable. It requires atleast one write ledger specified. Multiple write ledgers can be specified in config.
+- `endorser_did`: Endorser public DID registered on the ledger, needed for supporting Endorser protocol at multi-ledger level.
+- `endorser_alias`: Endorser alias for this ledger, needed for supporting Endorser protocol at multi-ledger level.
+
+Note: Both `endorser_did` and `endorser_alias` are part of the endorser info. Whenever a write ledger is selected using `PUT /ledger/{ledger_id}/set-write-ledger`, the endorser info associated with that ledger in the config updates the `endorser.endorser_public_did` and `endorser.endorser_alias` profile setting respectively.
 
 
 ## Multi-ledger Admin API
--- a/Multiledger.md
+++ b/Multiledger.md
@@ -36,6 +36,37 @@ If `--genesis-transactions-list` is specified, then `--genesis-url, --genesis-fi
   genesis_url: 'http://test.bcovrin.vonx.io/genesis'
 ```
 
+```
+- id: localVON
+  is_production: false
+  genesis_url: 'http://host.docker.internal:9000/genesis'
+- id: bcorvinTest
+  is_production: true
+  is_write: true
+  genesis_url: 'http://test.bcovrin.vonx.io/genesis'
+- id: greenlightDev
+  is_production: true
+  is_write: true
+  genesis_url: 'http://dev.greenlight.bcovrin.vonx.io/genesis'
+```
+
+Note: `is_write` property means that the ledger is write configurable. With reference to the above config example, both `bcorvinTest` and `greenlightDev` ledgers are write configurable. By default, on startup `bcorvinTest` will be the write ledger as it is the topmost write configurable production ledger, [more details](#write-requests) regarding the selection rule. Using `PUT /ledger/{ledger_id}/set-write-ledger` endpoint, either `greenlightDev` and `bcorvinTest` can be set as the write ledger.
+
+```
+- id: localVON
+  is_production: false
+  is_write: true
+  genesis_url: 'http://host.docker.internal:9000/genesis'
+- id: bcorvinTest
+  is_production: true
+  genesis_url: 'http://test.bcovrin.vonx.io/genesis'
+- id: greenlightDev
+  is_production: true
+  genesis_url: 'http://dev.greenlight.bcovrin.vonx.io/genesis'
+```
+
+Note: For instance with regards to example config above, `localVON` will be the write ledger, as there are no production ledgers which are configurable it will choose the topmost write configurable non production ledger.
+
 ### Config properties
 For each ledger, the required properties are as following:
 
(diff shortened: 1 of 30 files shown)


aries_cloudagent/ledger/routes.py
```diff
@@ -740,6 +811,12 @@ async def register(app: web.Application):
...
             ),
+            web.put("/ledger/multiple/{ledger_id}/set-write-ledger", set_write_ledger),
+            web.get(
+                "/ledger/multiple/get-write-ledgers",
+                get_write_ledgers,
+                allow_head=False,
+            ),
             web.get("/ledger/multiple/config", get_ledger_config, allow_head=False),
         ]
     )
```
esune, line 822: Do we really need the `/multiple/` path added? Whether the agent is connected to one ledger or more than one, I think the endpoints should still return information. The `ledger_id` can be used as path parameter where the operation requires one ledger instance to be identified.

esune (commented):
I like how the changes generalize how we handle the configured ledgers and open the door to -potentially - having more than one ledger set to write mode in the future.

I am wondering though if, given the current assumption that there will only ever be one write ledger active at any given time (and there is no way to specify which ledger to target when creating a transaction that needs to be written, as far as I know), we should keep some checks to prevent users from having more than one writeable ledger at the same time.

aries_cloudagent/config/ledger.py
```diff
@@ -89,9 +89,7 @@ async def load_multiple_genesis_transactions_from_config(settings: Settings):
             False if config.get("is_write") is None else config.get("is_write")
         )
         ledger_id = config.get("id") or str(uuid.uuid4())
-        if is_write_ledger and write_ledger_set:
-            raise ConfigError("Only a single ledger can be is_write")
-        elif is_write_ledger:
+        if is_write_ledger:
```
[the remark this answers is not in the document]
shaangill025: The meaning of `is_write` has changed with this PR. For example:
```
- id: localVON
  is_production: false
  genesis_url: 'http://host.docker.internal:9000/genesis'
- id: bcorvinTest
  is_production: true
  is_write: true
  genesis_url: 'http://test.bcovrin.vonx.io/genesis'
- id: greenlightDev
  is_production: true
  is_write: true
  genesis_url: 'http://dev.greenlight.bcovrin.vonx.io/genesis'
```
`is_write` property means that the ledger is write configurable. With reference to the above config example, both `bcorvinTest` and `greenlightDev` ledgers are write configurable. By default, on startup `bcorvinTest` will be the write ledger as it is the topmost write configurable production ledger. Using `PUT /ledger/{ledger_id}/set-write-ledger` endpoint, either `greenlightDev` and `bcorvinTest` can be set as the write ledger.

aries_cloudagent/ledger/routes.py
shaangill025: `/multiple/` has been removed. In case, multi-ledger config is not provided and single-ledger is applicable then the endpoint's response will return `ledger_id` as `default`. This is not applicable for `GET /ledger/config`, it will return a `Multiple ledger support not enabled` error.

esune (dismissed):
Great work @shaangill025 , this looks good!
@andrewwhitehead , @dbluhm and @usingtechnology could you please take a look and confirm I did not miss somethign important?

aries_cloudagent/indy/sdk/profile.py
```diff
@@ -87,8 +88,59 @@ def bind_providers(self):
...
+                "endorser_alias" in write_ledger_config
+                and "endorser_did" in write_ledger_config
+            ):
+                self.settings["endorser.endorser_alias"] = write_ledger_config.get(
+                    "endorser_alias"
+                )
+                self.settings["endorser.endorser_public_did"] = write_ledger_config.get(
+                    "endorser_did"
+                )
+        elif self.ledger_pool:
```
dbluhm, line 143: I know Indy SDK wallet is on its way out the door but this is an almost exact copy-paste of what's in the askar profile. Is it possible to reuse at least some of the logic so if issues are discovered later we don't have to change it in both places?

dbluhm (dismissed):

dbluhm (dismissed):

Multiledger.md
```diff
@@ -36,6 +36,39 @@ If `--genesis-transactions-list` is specified, then `--genesis-url, --genesis-fi
...
+```
+
+Note: `is_write` property means that the ledger is write configurable. With reference to the above config example, both `bcorvinTest` and `greenlightDev` ledgers are write configurable. By default, on startup `bcorvinTest` will be the write ledger as it is the topmost write configurable production ledger, [more details](#write-requests) regarding the selection rule. Using `PUT /ledger/{ledger_id}/set-write-ledger` endpoint, either `greenlightDev` and `bcorvinTest` can be set as the write ledger.
+
+```
+- id: localVON
+  is_production: false
+  is_write: true
+  genesis_url: 'http://host.docker.internal:9000/genesis'
+- id: bcorvinTest
```
usingtechnology, line 62: Spelling bcorvinTest -> bcovrinTest

Multiledger.md
```diff
@@ -36,6 +36,39 @@ If `--genesis-transactions-list` is specified, then `--genesis-url, --genesis-fi
...
+  genesis_url: 'http://test.bcovrin.vonx.io/genesis'
+  endorser_did: '9QPa6tHvBHttLg6U4xvviv'
+  endorser_alias: 'endorser_test'
+- id: greenlightDev
+  is_production: true
+  is_write: true
+  genesis_url: 'http://dev.greenlight.bcovrin.vonx.io/genesis'
+```
+
+Note: `is_write` property means that the ledger is write configurable. With reference to the above config example, both `bcorvinTest` and `greenlightDev` ledgers are write configurable. By default, on startup `bcorvinTest` will be the write ledger as it is the topmost write configurable production ledger, [more details](#write-requests) regarding the selection rule. Using `PUT /ledger/{ledger_id}/set-write-ledger` endpoint, either `greenlightDev` and `bcorvinTest` can be set as the write ledger.
```
usingtechnology, line 55: Spelling bcorvinTest -> bcovrinTest

usingtechnology (approved):
Spelling in documentation (bcorvinTest -> bcovrinTest) to update. 

I have a bigger question... I don't see anything about multitenancy? Is that coming later? I would assume that each tenant could potentially write to a different ledger.

esune:
> 
> I have a bigger question... I don't see anything about multitenancy? Is that coming later? I would assume that each tenant could potentially write to a different ledger.

I believe `/ledger/{ledger_id}/set-write-ledger` would be used by each tenant to set which ledger they want to write to.

shaangill025:
> > I have a bigger question... I don't see anything about multitenancy? Is that coming later? I would assume that each tenant could potentially write to a different ledger.
> 
> I believe `/ledger/{ledger_id}/set-write-ledger` would be used by each tenant to set which ledger they want to write to.

There will be a default write ledger set based on the multi-ledger config provided for all tenants. The tenants can then use the above endpoint to set different write ledger.

usingtechnology:
> 

Thanks @shaangill025 and @esune.  Is that tenant selection permanent? Maybe I am missing something, but it appears that the selection isn't persisted.

shaangill025:
> > 
> 
> Thanks @shaangill025 and @esune. Is that tenant selection permanent? Maybe I am missing something, but it appears that the selection isn't persisted.

It should be, it is basically binds `BaseLedger` instance based upon selection to the Tenant profile.

shaangill025:
@usingtechnology Tenant write ledger selection persistence should be working now.

dbluhm (approved):

PR #2150, opened 2023-02-28
Problem report generation for Out of band protocol

ramreddychalla94:
Creating this PR to add an endpoint to generate problem-report for out of band protocol on the basis of ConnectionId. Looking for the feedback from the community.
-Signed off by: ramreddychalla94 [ram.challa@ontario.ca](mailto:ram.challa@ontario.ca)
-Funded by the Ontario Government

This PR resolves https://github.com/hyperledger/aries-cloudagent-python/issues/2101 @swcurran please review and let us know of your comments, thanks.

PR #2350, opened 2023-07-25
Oob problem report

### LINKED ISSUE #2101: Add Problem Report API to Out of Band
Currently manual problem reports creation is available only under Issue Credential `POST /issue-credential-2.0/records/{cred_ex_id}/problem-report` and Present Proof `POST
 /present-proof-2.0/records/{pres_ex_id}/problem-report` APIs with `cred_ex_id` as input. 

Creating this issue to add a new API endpoint under OOB which will be `POST /out-of-band/{connection_id}/problem-report` to create a problem report and send it to another agent using `connection_id` as input, as currently there is no way to send a problem report when the connection is not established successfully or within a specific the time limit. The logic will be implemented in the controller side, but an API is required from AcaPy. 

Looking for feedback from the community before working on this, thanks.

dbluhm:
Supersedes #2150. Closes #2101. I am volunteering to help get this change in; looks like things are nearly there, just need to fix some issues with tests.

### DIFF
--- a/aries_cloudagent/protocols/out_of_band/v1_0/__init__.py
+++ b/aries_cloudagent/protocols/out_of_band/v1_0/__init__.py
@@ -9,9 +9,7 @@ from .messages.problem_report import (
 )
 
 
-def problem_report_for_record(
-        record: ConnRecord, desc_en: str
-) -> OOBProblemReport:
+def problem_report_for_record(record: ConnRecord, desc_en: str) -> OOBProblemReport:
     """
     Create problem report for record.
 
@@ -32,11 +30,11 @@ def problem_report_for_record(
 
 
 async def report_problem(
-        err: BaseError,
-        desc_en: str,
-        http_error_class,
-        record: ConnRecord,
-        outbound_handler: Coroutine,
+    err: BaseError,
+    desc_en: str,
+    http_error_class,
+    record: ConnRecord,
+    outbound_handler: Coroutine,
 ):
     """
     Send problem report response and raise corresponding HTTP error.
--- a/aries_cloudagent/protocols/out_of_band/v1_0/__init__.py
+++ b/aries_cloudagent/protocols/out_of_band/v1_0/__init__.py
@@ -1,0 +1,58 @@
+from typing import Coroutine
+
+from ....connections.models.conn_record import ConnRecord
+from ....core.error import BaseError
+
+from .messages.problem_report import (
+    OOBProblemReport,
+    ProblemReportReason,
+)
+
+
+def problem_report_for_record(
+        record: ConnRecord, desc_en: str
+) -> OOBProblemReport:
+    """
+    Create problem report for record.
+
+    Args:
+        record: connection or exchange record
+        desc_en: description text to include in problem report
+
+    """
+    result = OOBProblemReport(
+        description={"en": desc_en, "code": ProblemReportReason.ABANDONED.value}
+    )
+    if record:
+        thid = getattr(record, "thread_id", None)
+        if thid:
+            result.assign_thread_id(thid)
+
+    return result
+
+
+async def report_problem(
+        err: BaseError,
+        desc_en: str,
+        http_error_class,
+        record: ConnRecord,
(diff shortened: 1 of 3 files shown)


dbluhm:
To add some context, this PR seeks to implement a way to send a problem report as an explicit rejection to an OOB invitation. The reciprocating code, handling a problem report when we are the inviter, was already implemented in ACA-Py but there was no way to actually do the rejecting from ACA-Py as an invitee.

dbluhm:
Actually, I think there is a flaw in this reasoning. The out of band protocol RFC dictates that errors from this protocol express things like an inappropriately reused invitation or not supporting the suggested handshake protocols. I may have misinterpreted the intent behind the changes but I think a better way to reject a connection would be to send a DIDExchange problem report or Connections problem report, depending on what was reported in the handshake protocol of the invitation. All other instances of errors in OOB I think should be automatically generated by ACA-Py during message handling and would not require an Admin API endpoint to manually trigger it.

What were your intentions with these changes, @ramreddychalla94?

(I don't know Tim Bloomfield's Github handle else I would tag him as well since he reported that this change was stalled on their side; @swcurran, would you happen to know his handle?)

dbluhm:
> there is no way to send a problem report when the connection is not established successfully or within a specific the time limit. The logic will be implemented in the controller side, but an API is required from AcaPy.

From the issue I linked to early (without reading it...). If the connection is not established successfully and there's enough info known to send a problem report, I think this should be ACA-Py's responsibility, not the controller. If there is a time limit imposed on the invitation, I think the DID Exchange/Connection request should be rejected instead. ACA-Py doesn't implement this for DID Exchange yet.

tbloomfi:
@dbluhm thanks for taking a look at this issue. 

This particular requirement came out of an enterprise deployment where the issuer developing the controller identified a number of corner cases that needed to be handled - most having to do with holder eligibility for the credential changing during the issuance process.

In this example the offer could time out or the holder may become ineligible to receive the credential by the time they accept the invitation. In general, the issuer needs a way to generate standard error messages that can be presented in the wallet. 

Essentially the thinking was to use problem reports (RFC 0035) and possibly ~l10n catalogs to create a standard set of error messages usable by any issuer in the enterprise. 

The OOB errors section includes an example of an invitation expiry as a problem report - referenced as a courtesy message:

{
  "@type": "https://didcomm.org/out-of-band/%VER/problem_report",
  "@id": "5678876542345",
  "~thread": { "pthid": "<@id of the OutofBand message>" },
  "description": {
    "en": "The invitation has expired.",
    "code": "expired-invitation"
  },
  "impact": "thread"
}

Having said that I have no strong opinion as to where the problem report comes from. 

I like the idea of ACA-Py handling the timeout. One thing we learned from this project is that controller developers have little interest in the details of the protocols and want the API to be as simple as possible. The conversation around the controller managing revocation registries did not go well.

swcurran:
Good explanation, @tbloomfi — helps me understand it.  As I write this I’ve gone back and forth on this…

The only way to send such a message (“expired”) is after the other party has responded to the invitation. Since by definition an OOB message is sent out of band, you can’t send the message proactively. So this will only be sent if the invitation has expired and then the other party does respond.

It is reasonable to respond with a problem report in each protocol for which you can send an OOB invitation — DID Exchange, Connections and Presentation Request. In theory, other protocols can be added, but each would need to explicitly support being invoked via an invitation.  Given that, _I think_ it would be best to respond with a problem report in the protocol used by the other side.

I can see that using an OOB problem report could also work, so it’s not 100% obvious what the right thing to do is, but my leaning is not to use OOB. 

I’d say go ahead with one way or the other and let’s put in a clarification to the OOB RFC.

dbluhm:
These changes are superseded by #2394 which implemented rejection in the DID Exchange protocol. The DID Exchange protocol is a better place for the intended "invitation expired" since the invitation kicks off another protocol. #2394 doesn't implement expiration of invitations but does implement all the necessary components to add expiration in a future change. This change should be tracked in a new issue.

PR #2352, opened 2023-07-25
Extend `--preserve-exchange-records` to include Presentation Exchange.

### LINKED ISSUE #2126: Add an "auto delete" capability for presentation exchange records

usingtechnology:
Add `auto_remove` flags to appropriate APIs for override of default behaviour.

Verifiers/Provers can override default behaviour in `send-request` (both free presentation and bound presentation), `send-presentation`, `send-proposal`.

Note: added `auto_remove` to issue credentials (1 and 2) `send-request`, allowing holders to override the default behaviour.

Closes #2126 
Closes #2206


aries_cloudagent/indy/models/proof.py
```diff
@@ -674,3 +674,11 @@ class IndyPresSpecSchema(AdminAPIMessageTracingSchema):
...
         example=False,
     )
+    auto_remove = fields.Bool(
+        description=(
+            "Whether to remove the presentation exchange record on completion "
+            "(overrides --preserve-exchange-records configuration setting)"
+        ),
+        required=False,
+        default=False,
+    )
```
dbluhm, line 684: This feels misplaced. Why is this on an indy specific model?

usingtechnology: maybe there is a bad/inappropriate inheritance structure...

but that schema is the request schema for `/present-proof/records/{pres_ex_id}/send-presentation`...

yeah, some ugly stuff going on there.

usingtechnology: great catch though.  i don't want to get into a complete re-engineering here but will create a new schema using the IndyPresSpecSchema as a parent and add the `auto_remove` to the child.  Should be a step in the right direction and put the `auto_remove` where it is actually used.

dbluhm (approved):

Issue #2345, opened 2023-07-24
Produce peer:did:3 from existing connections' did_documents. 


Jsyro:
In the transition to using peer did's for connections, acapy has unqualified dids and the target of those connections may being to use `did:peer:3` without notice. This is directly related to #2249.

Acapy needs to be able to produce a did:peer:3 (which may required a did:peer:2), from a did_document of an established connection.

This should be added to this repository if possible. https://github.com/sicpa-dlab/peer-did-python

NOTE: ACAPY's DIDDoc class is not complaint with the actual did_document spec if references (i'm currently working on that). So the developer of this task should be working to produce a did:peer:3 from an instance of the DIDDocument class from pydid found here https://github.com/Indicio-tech/pydid/blob/main/pydid/doc/doc.py.

Jsyro:
I needed to tackle this in #2249. did:peer:3 is the identifier, did:peer:2 should be treated as did:peer:3 + did_doc, did:peer:2 documents are not saved as they can be resolved at anytime. 

did:peer:3's need to be resolved to a document, which should one that was previously seen and stored in the db. the id and controller' entries of that did_doc should have the did:peer:3, requiring manipulating the DIDDocument object.... circumventing the 'immutable' behaviour of the class....

PR #2353, opened 2023-07-25
did:peer:3 optional implementation, DIDDoc -> DIDDocument migration.

Jsyro:
Backwards compatible with older agents and existing unqualified dids.

Added flows that accept and send `did:peer:2` within DIDExchange protocol, however sends by default. 

Custom DIDDoc class from acapy exists, however upon deserialization will converted into a pydid DIDDocument instance. the existing Custom DIDDoc class is not compliant with the CONTEXT attribute it provides. 

Remaining work: 

- [ ] Clean up debugging and WIP shrapnel. 
- [ ] promote pydid and peer-did-python alpha changes to actual releases.
- [x] Ensure it will work with upgrading acapy when the agent already has an existing connection
- [x] Ability to promote/convert existing connections' did documents to did:peer:3 (see https://github.com/hyperledger/aries-cloudagent-python/issues/2345)
- [ ] Tests


swcurran:
Is there a start up flag that controls what kind of DID (unqualified, did:peer:2/3) is used when preparing an OOB invitation or sending an unsolicited (implict invitation) DID Exchange “Request” message to another agent?  I think there should be, and we’ll need to document that in the release notes, and perhaps elsewhere in the documentation.

docker/Dockerfile.demo
```diff
@@ -5,9 +5,11 @@ ENV ENABLE_PYDEVD_PYCHARM 0
 ENV PYDEVD_PYCHARM_HOST "host.docker.internal"
 ENV ACAPY_DEBUG_WEBHOOKS 1
 
-RUN mkdir bin && curl -L -o bin/jq \
+RUN mkdir --p bin && curl -L -o bin/jq \
```
dbluhm, line 8: Typo? It's just `-p`, isn't it?

aries_cloudagent/connections/base_manager.py
```diff
@@ -196,12 +197,23 @@ async def store_did_document(self, did_doc: DIDDoc):
...
-        await self.remove_keys_for_did(did_doc.did)
-        for key in did_doc.pubkey.values():
-            if key.controller == did_doc.did:
-                await self.add_key_for_did(did_doc.did, key.value)
+        await self.remove_keys_for_did(did_doc.id)
+        if hasattr(did_doc, "pubkey"):
+            for key in did_doc.pubkey.values():
+                if key.controller == did_doc.id:
+                    await self.add_key_for_did(did_doc.id, key.value)
+        if hasattr(did_doc, "verification_method"):
```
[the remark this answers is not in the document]
Jsyro: I do not know. https://github.com/hyperledger/aries-cloudagent-python/blob/4de1a12586456feef1fed231610f1391ba96bf99/aries_cloudagent/connections/models/diddoc/diddoc.py#L84

This maybe be completely obsolete within my branch, but i have not investigated that yet.

aries_cloudagent/connections/base_manager.py
```diff
@@ -196,12 +197,23 @@ async def store_did_document(self, did_doc: DIDDoc):
...
+            for vm in did_doc.verification_method or []:
+                if vm.controller == did_doc.id:
+                    if vm.public_key_base58:
+                        await self.add_key_for_did(did_doc.id, vm.public_key_base58)
+                    elif vm.material:
+                        # TODO: accept/decode multibase
```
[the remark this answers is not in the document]
Jsyro: `publicMultibase` appears to be the more widely used one. Added support for that.

swcurran:
From discussions in the last couple of days on this, the scope of this work will be:

- For DID Exchange, enable receiving an unqualified DID or did:peer:2 in the "request" message.
- When receiving a did:peer:2 in the "request" message, send a did:peer:2 in the "response" message.
- Add a flag (suggest `--use-did-peer-23` such that when it is true (default: false), send a did:peer:2 in the "request" message.
- When receiving a did:peer:2:
  - Add an ability to create the did:peer:3 identifier for the DID from the did:peer:2 value.
  - Resolve the did:peer:2 to its DIDDoc.
    - In doing that, extend the resolution to include an "AlsoKnownAs" field that is the "did:peer:3" value.
  - Convert the did:peer:2 DIDDoc to be a "did:peer:3" DIDDoc
     - In doing that -- swap the places where the DID identifiers are used -- did:peer:3 is the primary, did:peer:2 is the AlsoKnownAs.
  - Store the DIDDoc
  - In the connections object, set the "their_did" and "my_did" (as needed) to the "did:peer:3" identifier.
- Extend the general "store DIDDoc" capability to detect and use the "AlsoKnownAs" identifier.
   - Presumably that means enabling finding a DIDDoc by either name.

Note what is not being done here. For now, we won't touch the "RFC 0160 Connections" protocol, and we won't do a conversion of existing unqualified DIDs to be did:peer:3 DIDs.  For now...  :-)

It seems to me that some of this might best be done in the PyDID and DIDPeer libraries. I leave it to you to decide where to put the functionality.  Perhaps in both -- a PR to the appropriate libraries, implemented in ACA-Py with a note to update when the PRs are accepted in the libraries.

In all of this, try to align with the use of a generalized DID Resolver functionality.

@Jsyro @dbluhm -- sound good?

aries_cloudagent/connections/base_manager.py
```diff
@@ -170,24 +174,31 @@ async def create_did_document(
...
-    async def store_did_document(self, did_doc: DIDDoc):
+    async def store_did_document(self, did_doc: DIDDocument):
+        return await self._store_did_document(did_doc,did_doc.id)
+
+    
+    async def store_did_document_with_different_id(self, did_doc: DIDDocument, unqualified_did:str):
+        """TO HANDLE LEGACY CONNECTION UNQUALIFIED DIDS WITH NEW DIDDocuments """
+        return await self._store_did_document(did_doc,unqualified_did)
+
+    async def _store_did_document(self, did_doc: DIDDocument, storage_id: Union[str,DID]):
```
Jsyro, line 187: How do we feel about this? @swcurran @dbluhm It allows us to get rid of the old DIDDoc class and always use the pydid DIDDocument classes for diddocuments while allowing old unqualified dids for now.

swcurran: Works for me, although I’m perhaps not the best judge. We want to use pydid as much as possible (completely). As the issues with Peer DID get sorted out, we can evolve usage to get things cleaner.

dbluhm:
I'd like to make a proposal that I hope will help to make this change more digestible and reviewable. I think I'd like to see this work split into at least three separate chunks:

- Add Peer DID Resolution
- Add optionally sending Peer DIDs
- Eliminate DIDDoc class

As it stands, these changes mix these three things together. I think they are actually orthogonal changes (albeit all related in the sense that they build toward our desired end state).

Splitting it down this way will reduce the volume of code that needs to be considered while reviewing each PR. I think that is particularly important for this change given how fundamental a change it is and the fact that some of those chunks of changes will span many modules that are quite sensitive to changes. As it stands, I'm having a hard time detecting what changes are related to which of these chunks. Some of the changes that have been made I think are actually not strictly necessary. That's fine -- all part of the process -- it's just hard to tell what's what in it's current state.

Jsyro:
> I'd like to make a proposal that I hope will help to make this change more digestible and reviewable. I think I'd like to see this work split into at least three separate chunks:
> 
> * Add Peer DID Resolution
> * Add optionally sending Peer DIDs
> * Eliminate DIDDoc class
> 
> As it stands, these changes mix these three things together. I think they are actually orthogonal changes (albeit all related in the sense that they build toward our desired end state).
> 
> Splitting it down this way will reduce the volume of code that needs to be considered while reviewing each PR. I think that is particularly important for this change given how fundamental a change it is and the fact that some of those chunks of changes will span many modules that are quite sensitive to changes. As it stands, I'm having a hard time detecting what changes are related to which of these chunks. Some of the changes that have been made I think are actually not strictly necessary. That's fine -- all part of the process -- it's just hard to tell what's what in it's current state.

I'm not opposed to this decomposition of the changes I'm exploring here and agree this PR is larger than ideal. However I would propose decommissioning the old DIDDoc class as the first step. Then peer-did-resolution, then the handling and integrations. Thoughts?

dbluhm:
> However I would propose decommissioning the old DIDDoc class as the first step. Then peer-did-resolution, then the handling and integrations.

I think that's fair -- I suggested the order I did because I think the DID resolution is the smallest, adding the integration is more complicated, and then decommissioning `DIDDoc` feels like we could take it or leave it, to be honest. I agree it needs to be done eventually but I don't think it necessarily has to be tied to the implementation of DID Peer. I can be convinced otherwise on that though :slightly_smiling_face:

Jsyro:
won't merge this, work to be decomposed into smaller PR's. https://github.com/hyperledger/aries-cloudagent-python/pull/2472
is the first component. 

Expect at least one follow up that modifies the managers to utilize the resolution capabilities

Jsyro:
Won't merge this, work to be decomposed into smaller PR's. https://github.com/hyperledger/aries-cloudagent-python/pull/2472
is the first component. 

Expect at least one follow up that modifies the managers to utilize the resolution capabilities

PR #2356, opened 2023-07-26
Replace json with orjson through JsonUtil class

ff137:
The [orjson](https://github.com/ijl/orjson) library provides superior performance over the built-in json module. Please see their readme for an overview of features and performance/runtime comparisons.

One note is: `orjson.dumps` serializes to `bytes`, instead of `str`, so it's not a drop in replacement.
Edit: it also omits whitespace between keys and values. i.e. json.dumps: `{"k": [1, 2, 3]}`. orjson.dumps:  `{"k":[1,2,3]}`

There are 449 usages of `json.dumps` throughout the codebase. To refactor, I would just want to regex replace all of those with `orjson.dumps`, but then it would require appending `.decode()` as well (for bytes -> string).

That inspired me to create a JsonUtil class in `utils.json.py`. So, I could replace all instances of `json.loads` and `json.dumps` with `JsonUtil.*`.

This also has the benefit of providing an abstraction layer to simplify future transitions between different JSON libraries.

Note: I had to replace `import json` with `from aries_cloudagent.utils.json import JsonUtils`. Re-added `import json` where needed. However, that means the import organisation is no longer consistent, because json is a top-level import, and the utils import shouldn't be. Nitpick, bit I would want to do a import sort across all files to fix this ...


ff137:
Example of failing tests:
```
FAILED aries_cloudagent/ledger/merkel_validation/tests/test_domain_txn_handler.py::TestDomainTxnHandler::test_parse_attr_txn - assert ('raw', 'name...me":"Alice"}') == ('raw', 'name...e": "Alice"}')
  At index 2 diff: '{"name":"Alice"}' != '{"name": "Alice"}'
  Full diff:
  - ('raw', 'name', '{"name": "Alice"}')
  ?                          -
  + ('raw', 'name', '{"name":"Alice"}')
FAILED aries_cloudagent/ledger/merkel_validation/tests/test_domain_txn_handler.py::TestDomainTxnHandler::test_prepare_attr_for_state - assert b'{"lsn":10,"...99d3a42026a"}' == b'{"lsn": 10,...3d93de859b3"}'
  At index 7 diff: b'1' != b' '
  Full diff:
    (
  -  b'{"lsn": 10, "lut": 1513945121, "val": "6d4a333838d0ef96756cccc680af2531075c5'
  -  b'12502fb68c5503c63d93de859b3"}',
  +  b'{"lsn":10,"lut":1513945121,"val":"3cba1e3cf23c8ce24b7e08171d823fbd9a4929aafd'
  +  b'9f27516e30699d3a42026a"}',
    )
FAILED aries_cloudagent/protocols/coordinate_mediation/v1_0/tests/test_mediation_invite_store.py::TestMediationInviteRecord::test_to_json_should_dump_record - assert '{"invite":"s...,"used":true}' == '{"invite": "..."used": true}'
  - {"invite": "some_invite", "used": true}
  ?           -              -       -
  + {"invite":"some_invite","used":true}
...
```

In a nutshell, orjson is omitting whitespace, which the tests are hardcoded to expect. Will refactor tests to accommodate.

ff137:
Aah, here's an interested problem with this migration... 
we have a failing test, where the diff between actual and expected serialised json is:
```
  -                      'data': {'base64': 'eyJzY2hlbWFfaWQiOiAiR01tNHZNdzhMTHJMSmpwODFrUlJMcDoyOmFob3k6MTU2MDM2NDAwMy4wIiwgImNyZWRfZGVmX2lkIjogIkdNbTR2TXc4TExyTEpqcDgxa1JSTHA6MzpDTDoxMjp0YWcifQ=='},
  +                      'data': {'base64': 'eyJzY2hlbWFfaWQiOiJHTW00dk13OExMckxKanA4MWtSUkxwOjI6YWhveToxNTYwMzY0MDAzLjAiLCJjcmVkX2RlZl9pZCI6IkdNbTR2TXc4TExyTEpqcDgxa1JSTHA6MzpDTDoxMjp0YWcifQ=='},
```

Now, a lot of that string is the same. They both start with `eyJzY2hlbWFfaWQiOi`, but then it changes, before ending the same...

It turns out, that the two tokens decode from base64 to:
```
{"schema_id": "GMm4vMw8LLrLJjp81kRRLp:2:ahoy:1560364003.0", "cred_def_id": "GMm4vMw8LLrLJjp81kRRLp:3:CL:12:tag"}
```
vs
```
{"schema_id":"GMm4vMw8LLrLJjp81kRRLp:2:ahoy:1560364003.0","cred_def_id":"GMm4vMw8LLrLJjp81kRRLp:3:CL:12:tag"}
```

So the removable of whitespace by orjson will impact base64 encoding ...

swcurran:
Thanks @ff137 — good stuff!!  This is a fun one!

OK — maintainers — and anyone else that is interested/experienced in this area.

Is this a good idea?  I think it looks good (and their README and commit history trend looks good), but on an issue like this, I (correctly) have no voice.

What do you think — @dbluhm @usingtechnology @andrewwhitehead @frostyfrog @Jsyro @esune ?

esune:
I think the speed improvements are compelling enough to give this a try. Having some before/after tests for some scenarios would be interesting (not required though). The library seems to be fairly well maintained, one of the risks of moving to an external dependency is in the event of the library not being maintained anymore causing the need to switch back or onto another dependency. Unsure on how likely that is, and how risky to consider this move though. 

What I really think would be helpful is some unit tests of the utility since we are altering the output of the library to restore spaces. I like how the implementation isolates the different functions, it should make it fairly easy to test the methods in isolation as well.

ff137:
If the integration-tests runtime is any indication (with failing tests), it's almost 40% faster!

To try fix tests, I added a format_json method to JsonUtil to post-process the spaces back into the `orjson.dumps().decode()` result.

For some reason orjson doesn't support it as an optional flag yet (the maintainer closed a request for it without answer 😂) 

It's of course sheer overhead to regex process a json string to re-add whitespace, but the speed improvement from orjson still makes it worth it, and at least for now it makes sense to keep spaces for backward compatibility (as it affects base64 tokens)

You can see my last commits trying to reiteratively handle different cases in `format_json`, trying to re-add spaces correctly, but something is still up with the jws tokens not matching, which is omitted from logs. If someone can understand what I'm trying there, and maybe help think of a different way? Cuz that seems to be it, if formatting matches, it'll be backward compatible.

swcurran:
Are the whitespace issues only with the tests, or are those needed “in the wild”?

usingtechnology:
I kind of lean toward @esune on this as far as introducing a new dependency library and all that entails (end of life, conflicts with other library dependencies etc).
However, I also would love to see the performance gains - especially in the tests. They do take a considerable time.

If the changes in this PR were only applied to test files, would the errors be removed? Perhaps we can get away with using `orjson` as a test dependency only and use that for all json dumps and comparisons.

ff137:
> one of the risks of moving to an external dependency is in the event of the library not being maintained anymore causing the need to switch back or onto another dependency

> I kind of lean toward @esune on this as far as introducing a new dependency library and all that entails (end of life, conflicts with other library dependencies etc).

I really don't think that should be a concern. You're saying what if orjson is not maintained anymore -- then you use the version that works! Cross that bridge when you want python 4. Just like for the 50 other dependencies 😂

Thanks to the JsonUtil interface you'll be able to easily switch to a different json library, when that time comes. 

Besides, it's not as though python's built in json module is getting any improvements that would justify sticking to it.

> Perhaps we can get away with using orjson as a test dependency only

This doesn't make sense to me ... do you want faster tests or faster production? I'd say bite the bullet and just do it :-)

With that said, I can't carry this PR alone. Feedback is indeed welcome. If I could see the jws token mismatches (it's hashed out in the test logs), then I can give it another shot.. but I really hope someone more familiar is eager to assign it to themselves! This shouldn't be a difficult upgrade.

ff137:
> Are the whitespace issues only with the tests, or are those needed “in the wild”?

@swcurran So, the json.dumps method adds whitespace when it serialises a Dict object into a string.

json.dumps would give the string `'{"k": [1, 2, 3]}'`, while orjson.dumps gives: `'{"k":[1,2,3]}'`. 
From a computer's perspective, it represents exactly the same dictionary - it's just the string representation that's different. Including whitespace should be completely unnecessary, as it doesn't add info, while using more bytes.

The only problem imo is that previously some base64 tokens are generated from the json.dumps serialised string, which will now generate "visually different" tokens if whitespace is missing. The representation of base64 tokens would be different if whitespace is removed, but the content of the message is identical, from a key-value struct perspective.

Some PR tests are hardcoded to expect the whitespace in the serialisation. That's a trivial fix. 

The real question would be, how important is the backward compatibility with the base64 tokens? It's not clear to me, because decoding the base64 token would still yield the same key-value struct. It'll just be more efficient to not include unnecessary spaces. Hopefully someone has more context on if this will matter or not.

If there are any other mismatches between orjson and json, such as how it serializes without escaping unicode to ASCII,  hopefully that may be fixed by configuring orjson options.

ff137:
Atm the integration tests seem to fail because some keyword args are passed to json.dumps - which I implemented in JsonUtil to maintain and pass through to orjson - but some keyword args can be unsupported, like `indent`:
```sh
...
         File "/home/indy/demo/runners/support/agent.py", line 86, in __repr__
          return JsonUtil.dumps(self.val, indent=4)
        File "/home/indy/aries_cloudagent/utils/json.py", line 49, in dumps
          return JsonUtil.format_json(orjson.dumps(obj, *args, **kwargs).decode())
      TypeError: dumps() got an unexpected keyword argument
```

frostyfrog:
I'm just waking up, but I distinctly remember that @dbluhm and I ran into an issue with compatibility issues with hyperledger/aries-framework-javascript. What we found in our investigation is that AFJ receives the JSON, deserialized it into a native structure, grabs a signature, re-serializes it back to JSON, followed by validating the resulting JSON against the signature.

The change in whitespace broke this validation and caused AFJ to consider the message invalid.

frostyfrog:
> Atm the integration tests seem to fail because some keyword args are passed to json.dumps - which I implemented in JsonUtil to maintain and pass through to orjson - but some keyword args can be unsupported, like `indent`:
> ```sh
> ...
>          File "/home/indy/demo/runners/support/agent.py", line 86, in __repr__
>           return JsonUtil.dumps(self.val, indent=4)
>         File "/home/indy/aries_cloudagent/utils/json.py", line 49, in dumps
>           return JsonUtil.format_json(orjson.dumps(obj, *args, **kwargs).decode())
>       TypeError: dumps() got an unexpected keyword argument
> ```

Just from a quick glance at the files those errors come from, I believe the indent here *is* used for display purposes. The demo agents tend to dump JSON to the command line and, when run manually, are meant for human inspection

ff137:
@frostyfrog Understood. If the signature was generated from a serialised string that contains redundant spaces, then it won't match the signature of a serialised string without spaces. 
The solution is consistency. The space characters aren't necessary, so ideally that can be optimised across the board.

ff137:
Ah, now after fixing the indent keyword error, the only errors I see are generic 500 error / exceptions:
```sh
        File "/home/indy/demo/runners/agent_container.py", line 762, in initialize
          await self.agent.register_did(cred_type=CRED_FORMAT_INDY)
        File "/home/indy/demo/runners/support/agent.py", line 523, in register_did
          raise Exception(
      Exception: Error registering DID {'alias': 'Faber.agent', 'role': 'TRUST_ANCHOR', 'seed': 'd_000000000000000000000000529019'}, response code 500
```

swcurran:
I agree that we should use this in production.  The spaces thing is annoying — it would be good to know where it manifests in production use -- Is it in the envelope handling of every message?  That said, perhaps a test of with/without the extra manipulation would tell us if it was worth worrying about. From experience I know that such cross-community updates are hard.

The speed up is compelling...

dbluhm:
> The change in whitespace broke this validation and caused AFJ to consider the message invalid.

Indeed, this was an issue. The version of AFJ we were testing against (0.2.5) had the following lines:
https://github.com/hyperledger/aries-framework-javascript/blob/v0.2.5/packages/core/src/modules/connections/DidExchangeProtocol.ts#L483-L487

So AFJ was re-encoding the payload as a json string and then a base64url string after decoding it and then checking the signature over that value. The problem was the re-encoded values didn't include whitespace but the original payload signed by ACA-Py did. This might just be a plain old bug; the verifier can't manipulate the signed data and expect it to verify. However, in 0.4.0 release, I found these lines:
https://github.com/hyperledger/aries-framework-javascript/blob/v0.4.0/packages/core/src/modules/connections/DidExchangeProtocol.ts#L527-L529

The most recent AFJ release, 0.4.0, is still manipulating the payload before checking the signature. This should never be a requirement for verifying signed data. I'm not 100% certain I understand why this is happening but I think it's because of our base64 padded vs unpadded issues we're required to accommodate according to spec; barring that, I fear it's a hack to accommodate something ACA-Py is doing.

ff137:
Woot! The PR tests pass :-)

The integration-tests seem to be failing on my other PR with the same errors:
```sh
Exception: Error registering DID {'alias': 'Acme.agent', 'role': 'ENDORSER', 'seed': 'd_000000000000000000000000578403'}, response code 500
```

So seems unrelated to the json changes. Is the test ledger working at `http://test.bcovrin.vonx.io/register` ?

ff137:
On my other open PR, the PR Tests took 8m 24s. Here it was 6m flat - about 30% faster :smiley:

usingtechnology:
> The most recent AFJ release, 0.4.0, is still manipulating the payload before checking the signature. This should never be a requirement for verifying signed data. I'm not 100% certain I understand why this is happening but I think it's because of our base64 padded vs unpadded issues we're required to accommodate according to spec; barring that, I fear it's a hack to accommodate something ACA-Py is doing.

If @dbluhm and @frostyfrog are explicitly manipulating returned payloads to make them compliant, and `orjson` addresses that issue, it's a compelling argument in favour. The performance gain is a bonus. 

Would we have to label this as a breaking change? I'd imagine that other clients have their own workarounds and are now expecting the "invalid" responses?

@ff137 - great work digging in on this.

esune:
> I really don't think that should be a concern. You're saying what if orjson is not maintained anymore -- then you use the version that works! Cross that bridge when you want python 4. Just like for the 50 other dependencies 😂
Thanks to the JsonUtil interface you'll be able to easily switch to a different json library, when that time comes.
Besides, it's not as though python's built in json module is getting any improvements that would justify sticking to it.

Agreed, I was just trying to have an explicit opinion/agreement to adding a dependency from other maintainers - speed and compatibility improvements are compelling enough to want to use `orijson`.

I would still want to have unit tests JUST for the `JsonUtil` class: we are manipulating the output of the underlying library, so we should have unit tests validating our results. If for whatever reason `orijson` introduces spaces in the output, it might break the result of our manipulation and it would be much easier to see that the utility unit tests broke than hunting for a reason in the integration tests...

usingtechnology:
> I don't remember exactly what was happening in our experiments. I just remember that it was a problem we ran into, implying that we should exercise caution in-case it breaks compatibility with other agents.

yeah, i am concerned about just breaking things in the wild. perhaps we can feature flag this? if json access is through a the utils class, we can have default behaviour with `json` and an option to deploy\run with `orjson`? i would think multitenancy would have to adopt that as a tenant flag too...

andrewwhitehead:
> I'm just waking up, but I distinctly remember that @dbluhm and I ran into an issue with compatibility issues with hyperledger/aries-framework-javascript. What we found in our investigation is that AFJ receives the JSON, deserialized it into a native structure, grabs a signature, re-serializes it back to JSON, followed by validating the resulting JSON against the signature.
> 
> The change in whitespace broke this validation and caused AFJ to consider the message invalid.

I ran into one issue in either the connections protocol or did-exchange I believe, where ACA-Py signs the base-64 encoded JSON and then pads it (or vice versa). It should be sufficient to check against the padded and unpadded versions of the input. I'm fairly certain it's not necessary to adjust for whitespace.

swcurran:
There is a problem right now with BCovrin Test — hence that test failing.

It would be good to use at least Aries Agent Test Harness to try the AFJ compatibility.

On the spaces, perhaps where that is being checked we have a fallback that if the signature verification check fails, try it again after adding the spaces?  We do that in the different Aries libraries, and then never send out the spaces.

ff137:
> I would still want to have unit tests JUST for the JsonUtil class

I 100% agree, it's worth it. 

So - the `format_json` method is effectively post-processing the orjson.dumps result to add whitespace back to the string. I only wrote that for the backward compatibility. A unit test can be added to at least verify that the JsonUtil.dumps, as its implemented with the post-processing, gives the same result as the previous json.dumps - across many different examples.

> Would we have to label this as a breaking change?

So long as the post-processing is included, and proven to add the whitespace properly (i.e. not in the middle of values - only between keys and values, which is how it's set up), then it won't be a breaking change. The regex pattern I wrote seems to cover the cases properly, at least well enough for PR tests to pass without making changes to them - but yes, unit tests will add some more certainty to this.

As soon as the post-processing is removed, and the serialized strings no longer carry the redundant spaces (which I believe should be the end goal), then it's probably a breaking change, because expected payloads would be different.

swcurran:
Thoughts from the maintainer meeting on this:

- If we are going to provide a backwards compatibility path on this, best to use a configuration parameter vs. introducing some other technique, such as allowing an instance without orjson, as that would mean a deployer would not be able to use the published container image and would basically be setting a config option by another means.  Hold the nose and accept Yet Another Configuration option 😢 
- Suggest `—slow-json-serde` and that when active, the “standard” JSON processor is used instead of orjson.
- We think that a suggestion in the aries community could be used in face of the issue raised in the comments. That is — when verifying an encoded JSON payload. If feels like a waste, but in the normal path, the 2nd and 3rd decodes will never happen, so all good.
   - Verify the payload as delivered and if OK - end happy.
   - Decode the payload and remove spaces, encode and if OK - end happy.
   - Decode the payload add spaces, encode and if OK - end happy.
   - end sad.
- One consideration for a test to see if this is likely a big improvement.  Since presumably there is a lot of serializing/deserializing in the tests, would it be easy to do a code update such that the test code uses one way (new or old) and the “real” code uses the other way (old or new) and then we see the impact on the test. If the vast majority of the benefit is in the test code, then perhaps this is not a worthwhile change (new dependency, behaviour tweak).  On the other hand — if the “real” code benefit is there — we’d love to see this merged!

Thanks — looking forward to hopefully have this merged.

ff137:
@swcurran Thank you for the feedback! I understand - it'll require a configurable implementation. And tests will have to be refactored as well. So it's a bit more work than I first thought.

For the next week I'll be busy with other priorities, but I'll be sure to return to this soon. Will probably be nice to include this with the v1 release

dbluhm (approved):
This is a good step forward, I think, given the discussion. We can iron out the details of how we deal with the whitespace and associated configuration in a future ticket. Kicking that can down the road keeps this set of changes cleaner.

usingtechnology:
sorry, was right in the middle of looking it over. (i agree about dealing with extra configuration later).

ff137:
Trunk based development 👍

It's at least a step in making it easier to modify and customise any and all json processing - something which seems to happen a lot!

One "further work" suggestion is that, afaict, it's the `.serialize()` and `.deserialize()` calls that are likely contributing more to latency than the json library improvements.
There are optimisations available there. I'd rather do that than adding the config settings, which is new to me :-)

There's of course debugging any downstream impacts, of deciding not to include spaces in the json strings. And fixing tests that are hardcoded to expect specific strings. That's probably a good exercise in reviewing old code and making it more maintainable! Definitely something everyone is welcome to contribute to 😄

aries_cloudagent/utils/json.py
```diff
@@ -0,0 +1,100 @@
...
+            fp: a .read()-supporting file-like object containing a JSON document
+            *args: Extra arguments to pass to the json.load() function
+            **kwargs: Extra keyword arguments to pass to the json.load() function
+
+        Returns:
+            The Python representation of fp
+
+        """
+
+        return json.load(fp, *args, **kwargs)
```
usingtechnology, line 100: shouldn't we be using orjson? something like:

```
orjson.loads(fo.read())
```

I realize this is only used in test code (currently) but feels somehow wrong to leave it as `json`.

usingtechnology (changes requested):
I would like to see `JsonUtils.load` updated to `orjson`.

I do have concerns that this would make the change to `orjson` without the users/teams making that decision consciously.  I know we don't want to get into additional configuration but until this is truly proven, I would feel better that using `orjson` is an explicit choice.  The main reason for my concern is not just the whitespace (that to me is a "fancy" config), but how `orjson` marshalls data. Like datetimes... do we really know if that will impact systems?  `orjson` has A LOT of configuration options that determine how it marshals so that is where my concern is. And without an explicit decision to enable `orjson` (and quickly revert to `json`) I feel like we are  potentially messing up an external system. Just seem like a big thing to not have an escape plan built in. My $0.02, but not going to hold up the PR for that.

aries_cloudagent/utils/json.py
ff137: Aah, I noticed the same this morning. Turns out, orjson doesn't have a load method! I think it makes sense to remove it here and just use json.load directly in the one test that depends on it

usingtechnology: but `orjson.loads` can take file-like objects. so my thought was keep all json code under JsonUtils and update the implementation to use `orjson`.

ff137:
> The main reason for my concern is ... how `orjson` marshalls data. Like datetimes... 

I agree, it's not clear how this might create incompatibility with an external system. I mean, I only know so much of the codebase by now.

Quick inspection is showing me:
#core/dispatcher.py:
```py
        if isinstance(message, AgentMessage) and context.settings.get("timing.enabled"):
            # Inject the timing decorator
            in_time = context.message_receipt and context.message_receipt.in_time
            if not message._decorators.get("timing"):
                message._decorators["timing"] = {
                    "in_time": in_time,
                    "out_time": datetime_now(),
                }

        return await super().create_outbound(message, **kwargs)
...

async def create_outbound(
        self,
        message: Union[BaseMessage, str, bytes],
        ...
    ) -> OutboundMessage:
        """Create an OutboundMessage from a message payload."""
        if isinstance(message, BaseMessage):
            # TODO DIDComm version selection
            serialized = message.serialize()
            # TODO serialized format selection?
            payload = JsonUtil.dumps(serialized)
```

So there is one instance of a datetime object being passed to a field, and being serialized / JsonUtil.dumps

The only thing I can remark on datetimes from orjsons page is:
> serializes datetime, date, and time instances to RFC 3339 format, e.g., "1970-01-01T00:00:00+00:00"

Best I can figure out in short time is what GPT4 tells me: 
> As you mentioned, orjson serializes datetime, date, and time instances to RFC 3339 format by default. The standard json library does not serialize these objects by default; you'd typically handle these manually with a custom serialization method. 

I think it's valid to keep an eye out for cases like this. Any suggestions for possible test cases to add? I can maybe at least add a datetime conversion to the tests and validate it matches the json datetime conversion ... Thoughts?

aries_cloudagent/utils/json.py
ff137: By the way, orjson.loads doesn't work on file-like objects. `loads` expects a string / byte array.

Here's how the json.load method can be replaced with orjson.loads:

Using json.load

```python

import json

with open('file.json', 'r') as f:
    data = json.load(f)
```
Equivalent using orjson.loads

```python

import orjson

with open('file.json', 'rb') as f:
    content = f.read()
    data = orjson.loads(content)
```

Is it fine if I leave `load` out of JsonUtil, and just update the test to use orjson.loads / JsonUtil.loads to read the file?

usingtechnology:
> think it's valid to keep an eye out for cases like this. A part of me wants to say just see where the leaks appear, and fix them, but of course it's good to pre-empt them. Any suggestions for possible test cases to add? I can maybe at least add a datetime conversion to the tests and validate it matches the json datetime conversion ... Thoughts?

It's those unknowns, and perhaps untestable cases (I've been on many projects where test environments do not have the same data and edge cases as production) that make me think we need a way for `orjson`/`json` be selectable. I would prefer that the answer to someone's issue is - deploy with this flag/env var versus "wait for us to fix it".  As much as I don't want to get into configuration, after reading `orjson` documentation red flags pop up that we don't really know how data is affected.

dbluhm:
As a potential middle ground, can we first introduce the `JsonUtil` class and follow up with configuration and orjson?

ff137:
@usingtechnology Understood. If it's being added, it makes sense for it immediately to be configurable. So, I'd propose a flag name `--default-json` or something like that. Which in the JsonUtil I'll read and switch to using python's json

Now I'll just have to know how to add flags :smile: ... it's late here so I'll tackle it tomorrow! But any pointers for where to look on setting/reading config, will be appreciated!

ff137:
> As a potential middle ground, can we first introduce the `JsonUtil` class and follow up with configuration and orjson?
@dbluhm or, yes! Considering I'm new to the config settings, perhaps someone else can hotfix it after it's merge to main

aries_cloudagent/utils/json.py
usingtechnology: This is maybe me being a completist but I don't think leaving that as `json.load` is correct. Currently it is only used in two places, so you could just as well update the calling code to better match up with `orjson.loads`. But more importantly, the impetus for this PR was to gain speed, which was measured by how much quicker the tests ran. The main use case for that `json.load` is in the BDD tests. Perhaps we will see an even greater improvement if everything is using `orjson`, which also supports indent, so we could use that too.  Of more importance is if we need to flag `orjson`/`json`, we can do it in JsonUtils.  I think if it is ever needed in the runtime code, having it in JsonUtils (hopefully) will make it more apparent to that developer.

usingtechnology:
> As a potential middle ground, can we first introduce the `JsonUtil` class and follow up with configuration and orjson?

Yes, we certainly can do that. 
You will have more experience with how the user base tests release candidates. If you feel like that is something that would be uncovered before "prod", then it is okay - let's move forward. Mostly just voicing concern but not needing to hold anything up.

usingtechnology:
> As a potential middle ground, can we first introduce the `JsonUtil` class and follow up with configuration and orjson?

Sorry, maybe I didn't quite understand and missed your point... Are you suggesting that there is a PR that simply uses JsonUtil without `orjson`? So that we can introduce `orjson` or other in a later PR?

dbluhm:
Either way works for me. Configuration options are setup here: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/config/argparse.py

The general pattern is you add an argparse argument and then add an assignment to a `settings` value in the `get_settings` of the group. The configuration option is then accessible from `profile.settings["my.config.option"]` or same but from a `context.settings`.

I think the main tricky part is we aren't currently using the profile/context in the `JsonUtil` class already. To enable this, we'd probably want to turn the `JsonUtil` class into something we bind on startup and obtain an instance of on use.

... This would actually be a lot simpler as an optional dependency. We specify `orjson` as an extra and if it can't be imported, we don't use it, if it can, we use it. Initially, we can make this opt-in -- your deployment would actually need to install the extra (i.e. `aries-cloudagent[orjson,askar,bbs]`) -- and then later we could include it by default in images.

ff137:
> ... This would actually be a lot simpler as an optional dependency.

I had the same thought, and that's indeed much simpler! Perhaps the config flag would be better suited for the whitespace behavior. The dependency usage can be controlled by just trying to import it, and defaulting to `json` :+1:

usingtechnology:
> Either way works for me. Configuration options are setup here: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/config/argparse.py
> 
> The general pattern is you add an argparse argument and then add an assignment to a `settings` value in the `get_settings` of the group. The configuration option is then accessible from `profile.settings["my.config.option"]` or same but from a `context.settings`.
> 
> I think the main tricky part is we aren't currently using the profile/context in the `JsonUtil` class already. To enable this, we'd probably want to turn the `JsonUtil` class into something we bind on startup and obtain an instance of on use.
> 
> ... This would actually be a lot simpler as an optional dependency. We specify `orjson` as an extra and if it can't be imported, we don't use it, if it can, we use it. Initially, we can make this opt-in -- your deployment would actually need to install the extra (i.e. `aries-cloudagent[orjson,askar,bbs]`) -- and then later we could include it by default in images.

hmmm, your probably right about making a provider/service type interface. default implementation could be `json`.  sigh, makes the BDD tests a bit of a pain (used in backchannel so no profile), although we could wire up another param in tests to use a `json` or `orjson`. may be a good idea anyway so maybe we would find differences?

aries_cloudagent/utils/json.py
ff137: @usingtechnology So, my idea is just to move the `read_json_data` method to JsonUtil ... With the dependency check, it's better for that to be configured in JsonUtil. Then the demo/bdd_support/ module just uses `JsonUtil.read_json_file`. Does that seem okay to you? I'll push that in a minute

usingtechnology: > @usingtechnology So, my idea is just to move the `read_json_data` method to JsonUtil ... With the dependency check, it's better for that to be configured in JsonUtil. Then the demo/bdd_support/ module just uses `JsonUtil.read_json_file`. Does that seem okay to you? I'll push that in a minute

That sounds good. I think the bigger thing is what @dbluhm pointed out and that configuration will rely on this being bound to a profile, which means not a utility class but more like a service/provider class.

ff137:
@dbluhm @usingtechnology So, I've added orjson as an optional dependency. 

Now, one catch is that tests have to be run in one mode or the other ... I added `orjson` to the Tests workflow ... and I shudder to think we'll need an extra test workflow just to test default json as well!

What are your thoughts on this? Json parsing is obviously happening everywhere, I feel like it's kinda necessary to validate it works properly in both modes: default and with orjson ... Is 2 workflows an alright compromise? In parallel / github matrix strategy. And I'm thinking only PR tests .... what about integration tests? Of course don't wanna make a mountain out of a molehill, but I do wanna follow best practice. And parallel tests is probably better than saying "well the default json tests passed on my local" :-) 

Thoughts/suggestions? I'm fine with just introducing the JsonUtil class, and adding orjson separately ... or, keep this PR open for longer, allow more time for discussion and refinement

aries_cloudagent/utils/json.py
```diff
@@ -0,0 +1,119 @@
...
+
+        """
+        if orjson is None:
+            return json.loads(s, *args, **kwargs)
+        else:
+            return orjson.loads(s, *args, **kwargs)
+
+
+@staticmethod
+def read_json_file(file_name: str):
```
usingtechnology, line 95: i think you need to indent this, the tests are failing because can't find JsonUtil.read_json_file

ff137: Lol ... thank you. Had no idea why it was failing! That's what working at night (+ during load-shedding) gets me 😂

usingtechnology:
Should we move this into its own branch for collaboration?

There are a lot of little issues to work out that will require some thinking (not all need to be tackled immediately):
- some calls are not in a profile and may need some static (but configurable) utility (ex. in argparse)
- binding a service/provider to each tenant (with tenant configuration)
- unit tests
- BDD tests that run json and any other known implementation (orjson)

It would be interesting / useful to run some load testing to document the speed improvements.

I'm not sure we should make orjson a buildtime option, as that impacts all tenants (no one could choose it if its not on the image), and would we be publishing multiple images?

Anyway, things like that to consider...

ff137:
This PR has gone stale, so will require proper rebasing to account for project refactoring.

I think the outcome is that this change should be applied in more granular steps. Perhaps one PR to create a JsonUtil class, just to abstract json methods away into its own module. Later the json parsing library can be made optionally switchable. Might return to this sometime soon.

dbluhm:
I'm rooting for this change -- sorry to see it overcome by events!

PR #2394, opened 2023-08-02
feat: add DID Exchange specific problem reports and reject endpoint

dbluhm:
This PR makes problem reports sent during the DID Exchange protocol have the DID Exchange protocol adopted problem report rather than the generic problem report. This is defined in RFC 23 but has not been implemented by ACA-Py to this point. In other words, this is a correction to bring ACA-Py more in line with the RFC. These problem reports were automatically sent if there was an issue in processing the DID Exchange messages.

In addition to the above changes, this PR also introduces a `POST /didexchange/{conn-id}/reject` endpoint, enabling the controller to explicitly reject an OOB invitation that specifies DID Exchange as the handshake protocol or a DID Exchange request. Rejecting a response is not supported; don't send a request if you're going to reject the response. This will send a problem report to the originator of the invitation/request and will mark the (local) connection record as abandoned. It won't delete the record though (maybe it should)?

This at least partially supersedes #2350, covering the rejection of invitations to exchange DIDs. However, it does not introduce automatic handling of timeouts for invitations, as was discussed in the comments.


dbluhm:
Adding more tests; I'll keep this in draft until I get those in.

dbluhm:
Oh, also need to add a DIDX Problem Report handler to mark the connection as abandoned if we receive a `request_not_accepted`, etc.

swcurran:
FYI @tbloomfi — will this address your team’s needs?

burdettadam (approved):

dbluhm:
Calling attention to this again. This PR is ready to go, just needs review @usingtechnology @swcurran @shaangill025

aries_cloudagent/protocols/didexchange/v1_0/routes.py
```diff
@@ -362,6 +372,38 @@ async def didx_accept_request(request: web.BaseRequest):
...
+        async with profile.session() as session:
+            conn_rec = await ConnRecord.retrieve_by_id(session, connection_id)
+        report = await didx_mgr.reject(conn_rec=conn_rec, reason=reason)
+    except StorageNotFoundError as err:
+        raise web.HTTPNotFound(reason=err.roll_up) from err
+    except (StorageError, WalletError, DIDXManagerError, BaseModelError) as err:
+        raise web.HTTPBadRequest(reason=err.roll_up) from err
+
+    await outbound_handler(report, connection_id=conn_rec.connection_id)
+    result = conn_rec.serialize()
```
usingtechnology, line 403: Should we return the updated connection record with the abandoned state? I would wonder why my returned connection didn't indicate that it was abandoned...

dbluhm: Been a minute but I believe the `reject` call in the DID Exchange manager should modify the `conn_rec` so it should include the abandoned state.

usingtechnology (approved):

aries_cloudagent/protocols/didexchange/v1_0/routes.py
dbluhm: I'll do a quick double check on this but will merge if it returns an abandoned state as expected. Thanks for the review!

dbluhm: Okay, confirmed that the state "abandoned" is returned.

PR #2399, opened 2023-08-06
Upgrade pre-commit and flake8 dependencies; fix flake8 warnings

ff137:
Edit: flake8 must remain pinned to v4.0 as pytest-flake8 is unmaintained and doesn't support flake8 > 5.0 (latest = 6.1)
I propose migrating to [ruff](https://beta.ruff.rs/docs/) and [pytest-ruff](https://github.com/businho/pytest-ruff). Created issue: https://github.com/hyperledger/aries-cloudagent-python/issues/2400

___

Upgrades `.pre-commit-config.yaml` to use latest versions for pre-commit-hook, black, and flake8.
This fixes pre-commit not being in sync with black/flake8 enforcement in the github actions.

Additionally we upgrade flake8, flake8-docstrings, and pydocstyle to latest. 

Includes fixes for all flake8 warnings in ./aries_cloudagent, because if PR tests enforces flake8, then the existing code should abide by it too :-)

The remaining flake8 warnings are primarily for missing docstrings in ./demo


dbluhm:
@swcurran I just noticed that the `acapy-integration-tests` action isn't marked as required; was this a stopgap with the recent outages on the test ledger? I was going to mark this for auto-merge so it would merge after the tests finished but am unable to.

swcurran:
I don’t think anything changed — all of the BCovrin tests are running again now that the ledger has been reset. Definitely, we need to discuss what tests are required — tomorrow’s ACA-Pug.

PR #2404, opened 2023-08-09
feat: add legacy peer did resolver

### LINKED ISSUE #2161: The DID resolver and DID Admin Endpoint should also resolve local only DIDs
The ACA-Py DID resolver and the Admin API endpoint `GET /resolver/resolve/{did}` evidently does not resolver local only DIDs -- e.g. peer DIDs used for DIDComm connections.  This issue is to extend the resolver to include checking for and resolving such DIDs.  For example, I should be able to grab `theirdid` from a Connection object and resolve it to see the DIDDoc related to that DID. The resolution process should only check locally for properly qualified `peer:did` DIDs. For unqualified DIDs, it's a little trickier since there are both unqualified local (peer) and ledger-based (`did:sov`) DIDs. If there is not a way to tell if the type (local or ledger-based) of such a DID, the approach should probably be to try to resolve it as a local DID, and if it is not, to try to resolve it on a ledger.

Comments from @shaangill025 @dbluhm and others welcome.

dbluhm:
This PR adds support for resolving legacy peer DIDs to DID Documents previously stored in our wallet.

This is one part of some additional changes I have in mind; using these changes, I'd like to make "resolving" connection targets for a DID consistent regardless of whether the DID we're resolving is did:web, did:indy, did:sov, unqualified local did, did:peer, etc. I'm hopeful those changes will also help ease the transition to proper peer DIDs. I'm still fleshing out those changes but I think this resolver is useful regardless of the outcome.

As noted in the docstrings of the methods in the resolver, due to the nature of ACA-Py's use of did:sov and unqualified DIDs for both public (posted to a ledger) and pairwise/peer DIDs, this resolver requires a bit of trial and error. If the DID has a doc associated with it in the wallet, it will report that it supports resolving the DID. The cache is employed to help make this check slightly more efficient but there's still some open questions. For instance, how long should the cache be considered valid?

As implemented, this resolver will only return the DID Documents we've received through the Connections or DID Exchange protocols. In other words, we can only resolve the DID of the remote party, or "their DID." We could also store the docs for our own DIDs, making them resolvable as well. I'm not sure if this is valuable or not yet.

For Indy "style" DIDs that are created outside of the context of a connection, such as Public DIDs, there is no DID Document associated with it stored in our wallet. This means that this resolver will not be selected and the document for the DID will be resolved by the `IndyDIDResolver`, as usual.

In the interest of that consistent connection target goal I mentioned above, I currently have this resolver configured to transform documents retrieved from our wallet to be better aligned with updates to DIDComm/DID Document conventions.

Resolves #2161. I think. lol

### DIFF
--- a/aries_cloudagent/resolver/default/legacy_peer.py
+++ b/aries_cloudagent/resolver/default/legacy_peer.py
@@ -40,7 +40,7 @@ P = ParamSpec("P")
 class LegacyDocCorrections:
     """Legacy peer DID document corrections.
 
-    These correctionsto align the document with updated DID spec and DIDComm
+    These corrections align the document with updated DID spec and DIDComm
     conventions. This also helps with consistent processing of DID Docs.
 
     Input example:
--- a/aries_cloudagent/resolver/default/legacy_peer.py
+++ b/aries_cloudagent/resolver/default/legacy_peer.py
@@ -1,0 +1,272 @@
+"""Resolve legacy peer DIDs.
+
+Resolution is performed by looking up a stored DID Document.
+"""
+
+from collections.abc import Awaitable
+from copy import deepcopy
+from dataclasses import asdict, dataclass
+import functools
+import logging
+from typing import Callable, Optional, Sequence, Text, TypeVar
+from typing_extensions import ParamSpec
+
+from ...cache.base import BaseCache
+from ...config.injection_context import InjectionContext
+from ...connections.base_manager import BaseConnectionManager
+from ...core.profile import Profile
+from ...did.did_key import DIDKey
+from ...messaging.valid import IndyDID
+from ...storage.error import StorageNotFoundError
+from ...wallet.key_type import ED25519
+from ..base import BaseDIDResolver, DIDNotFound, ResolverType
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class RetrieveResult:
+    """Entry in the peer DID cache."""
+
+    is_local: bool
+    doc: Optional[dict] = None
+
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+class LegacyDocCorrections:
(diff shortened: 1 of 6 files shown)


aries_cloudagent/resolver/default/legacy_peer.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+class LegacyDocCorrections:
+    """Legacy peer DID document corrections.
+
+    These correctionsto align the document with updated DID spec and DIDComm
```
burdettadam, line 43: correctionsto or corrections?

burdettadam (dismissed):

dbluhm:
@usingtechnology @swcurran A quick nudge to review these changes. They're ready to go and I think it's a pretty self-contained change.

swcurran:
Sorry for the delay @dbluhm.  Quick question about this from an outsider.  Where/how is the resolver invoked?  I’m asuming there is an existing resolver that is invoked, and this is a new implementation?

dbluhm:
Yes, this hooked into the "global" DID Resolver interface. This resolver will be invoked by the Admin API on `GET /resolver/resolve/{did}` and in a few other locations (json-ld cred verification, OOB invite DID resolution). The DID Resolver has been used more in new development in ACA-Py but in the interest of incremental changes, it hasn't become the core component it ought to be yet. #2409 for instance makes the DID Resolver an integral part of the connection info lookup. So this change is just making these local legacy peer DIDs available to the DID Resolver interface.

dbluhm:
For some added context, this resolver is registered with the "global" resolver here: https://github.com/hyperledger/aries-cloudagent-python/blob/2122fdd123bcacd0373af496e07bb002282ad87f/aries_cloudagent/resolver/__init__.py#L20-L24

swcurran (approved):

PR #2412, opened 2023-08-11
Anoncreds API BDD Tests

usingtechnology:
Add BDD tests for issue v2 and proof presentation v2.
The tests include create schema, create cred def and revocation through the new `/anoncreds` API.

Note that I encountered some weirdness running the tests in the devcontainer... some messages didn't appear to be fully fleshed out (see comments about "by_format"). Running using the `run_bdd` script was fine.

Also, on proof presentation, I could not get it working using restriction with "schema_name" and "schema_version" like other presentations. I used restrictions by `cred_def_id` to pass the tests.  Not sure if that is something that needs to be looked at further as it would impact migrations to `anoncreds` if current presentation requests don't work.

### DIFF
--- a/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
+++ b/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
@@ -168,7 +168,7 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
                     issuer_id=schema["id"].split(":")[0],
                     attr_names=schema["attrNames"],
                     name=schema["name"],
-                    version=schema["ver"],
+                    version=schema["version"],
                 )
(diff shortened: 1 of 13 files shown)


dbluhm:
> Also, on proof presentation, I could not get it working using restriction with "schema_name" and "schema_version" like other presentations.

This is not expected so there's probably something that needs to be fixed

usingtechnology:
Found the issue... the comparison for schema_version failed as we returned the "ver" value when searching our schemas.

swcurran:
Cool — can we merge this?

When we were working on the spec, we noted that potential issue.  It was in the implementation from way back, but…

Nice catch.

usingtechnology:
I'll merge it in, we still have work to do on the branch, so if there are dissenting opinions, we have ample opportunity to correct.

Issue #2087, opened 2023-01-24
Add container vulnerability scanning for the built/published ACA-Py images.


WadeBarnes:
PR https://github.com/hyperledger/aries-cloudagent-python/pull/2076 added support for building and publishing ACA-Py images.  The initial implementation just builds and publishes the images.

It would be a great enhancement to have these containers scanned on a regular basis (and possible during the builds) for vulnerabilities.

WadeBarnes:
@ryjones, Does Hyperledger have any existing tools and processes for this?

ryjones:
@WadeBarnes [I think you can choose from a lot of free ones](https://github.com/hyperledger/aries-cloudagent-python/actions/new?category=security&query=), we don't have a policy

swcurran:
Narrowing to container images seems to give [this list](https://github.com/hyperledger/aries-cloudagent-python/actions/new?category=security&query=Container). Sysdig and Snyk are the names I recognize, but not sure about which to use.  Opinions?

Need to verify their use/pricing for open source projects.

pradeepp88:
I created an acapy image with the new Dockerfile from the docker folder, uploaded it to Dockerhub and scanned it with Snyk tool. It was showing to have lot of vulnerabilities arising from Debian-buster base image. The number of vulnerabilities went down drastically, when changing the base image from `python:{python_version}-slim-buster` to `python:{python_version}-slim`. Used Python version 3.8. The same is the case for all the slim-buster images.

WadeBarnes:
@pradeepp88, it would also be interesting to compare the results when using `3.9-slim-bullseye`.  Thanks for trying this out.

pradeepp88:
@WadeBarnes I have tested the images with `python:3.9-slim-bullseye` and found 0 critical vulnerabilities. I tested in both Qualys and Synk and got the same results. So created a PR #2105 to update the base images from slim-buster to slim-bullseye version. Please review.
Looking for feedback from the community on the vulnerability scanning tool to be used. I checked Snyk, it is good and matches the report generated by enterprise scanning tools. Thanks.

swcurran:
Thanks @pradeepp88 — great stuff.  Much appreciation for the work and conclusions.

WadeBarnes:
thanks @pradeepp88, the related PR updating the base images has been merged.

WadeBarnes:
Addressed with https://github.com/hyperledger/aries-cloudagent-python/pull/2418

Issue #2310, opened 2023-07-18
Slim down docker images


dbluhm:
The current images we're building include utilities like curl, git, less, vim-tiny, etc. I think inclusion of at least some of these utilities is a leftover from von-image days. While these utilities are helpful for debugging, it's generally considered best practice to keep container images minimal to reduce attack surface.

We should remove anything that isn't critical. Some of these utilities may be necessary while fetching and installing dependencies but are not critical after they've been retrieved. In that case, we should split the image into a multipart build so at least the resulting image is minimized.

Might be a good one to tackle alongside changes needed for #2289

PR #2418, opened 2023-08-14
Enable Snyk scanning

ryjones:
addresses #2087

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,0 +1,26 @@
+name: Snyk Container
+on: push
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build a Docker image
+      run: docker build  -t aries-cloudagent -f docker/Dockerfile .
+    - name: Run Snyk to check Docker image for vulnerabilities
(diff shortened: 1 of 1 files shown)


swcurran:
Ouch…236 open issues.  That’s painful.

The three “high” issues are the same thing and are likely fine.  We’ll need guidance on how to mark “uninteresting” issues — presumably that is the docs.

Thanks for doing this, @ryjones — awesome stuff.

Aside — I had to edit the “github-code-scanning” comment, as the link to the overview was bad — unusable in email.  Anything we need to do about that?

dbluhm:
Several of the low severity ones would be taken care of by #2310, I think

ryjones:
@swcurran I think the reason it didn't work was it didn't exist yet

swcurran:
> @swcurran I think the reason it didn't work was it didn't exist yet

I don’t think so.  When I fixed the link, I could navigate to the content, so it existed. The link started with “hyperledger/aries-cloudagent-python…” — it didn’t have github.com in the URL. Might have been related to the link going into an email.  Anyway, we can deal with that later.

ryjones:
@swcurran sounds good. Please approve and merge

dbluhm (approved):

PR #2421, opened 2023-08-15
Issue #2250 Nightly publish workflow

### LINKED ISSUE #2250: Provide nightly builds for ACA-Py
For developers that require access to bug fixes, we should provide nightly builds (or build on merge to main?) of the ACA-Py images.

Investigate if we can leverage a nightly build in AATH or other testing situations.

Gavinok:
This issue resolves #2250

Here I have created a GitHub workflow that will execute the tests just as they are handled in `nightly-tests.yml` excluding the indy tests as mentioned in this comment.

https://github.com/hyperledger/aries-cloudagent-python/issues/2250#issuecomment-1674918887

The workflow acts as follows
```python
run nightly tests
If (tests are successful):
   build a nightly release
   publish this release with the `nightly` tag
```


dbluhm:
I believe this will result in an image tagged `nightly` but no other image tags, is that correct? So each day, the previous nightly would be untagged and a new nightly would get tagged, replacing the old. I think it would be helpful to have a tag to continue to reference the old images by (commit hash or date or something). I think it's okay for the main `nightly` tag to follow the most recent image (like how the `latest` tag is often used) but being able to "roll back" changes by checking out a nightly build from last week would be really useful. Likewise, being able to pin to a specific nightly build and have your colleagues pull the same image (and not just the lastest nightly) will be important for reproducibility.

On a similar vein, if there's no changes merged to main since the last image, it would be nice to skip building and replacing.

swcurran:
Converted to draft while @Gavinok works on @dbluhm's feedback.

Gavinok:
@dbluhm I have now added a check to ensure publications only happen when a commit has occurred in the last 24 hours. In addition, I now publish both a `nightly-(today's date)` release as well as a rolling `nightly` release as you recommended.

dbluhm (approved):

Issue #2383, opened 2023-07-31
Remove/Adapt old endpoints


dbluhm:
The following endpoints will either be adapted to use the AnonCreds AnonCreds endpoints, or removed entirely. Adapted means that the calls will continue to work, but the functionality will be implemented by invoking the AnonCreds endpoints processing. Inputs and results may be adjust if absolutely necessary. Ideally, the endpoint will adjust the inputs, call the AnonCreds processing, and adjust the outputs.

The noted revocation related endpoints will be removed entirely. ACA-Py will handle revocation, and direct control by a controller will be removed.

##### Schemas
- `POST /schemas`
- `GET /schemas/created`
- `GET /schemas/{schema_id}`
- `POST /schemas/{schema_id}/write_record`

##### Credential Definitions
- `POST /credential-definitions`
- `GET /credential-definitions/created`
- `GET /credential-definitions/{cred_def_id}`
- `POST /credential-definitions/{cred_def_id}/write_record`

##### Revocation
- `GET /revocation/active-registry/{cred_def_id}`
  - Not needed and may be removed.
- `POST /revocation/clear-pending-revocations`
  - Error handling — not sure this is needed
- **REMOVE**: `POST /revocation/create-registry`
- `GET /revocation/credential-record`
- `POST /revocation/publish-revocations`
- `GET /revocation/registries/created`
  - Not needed — consider removing
- `DELETE /revocation/registry/delete-tails-file`
   - Delete tails file will occur after the successful publishing of the tails file, as it is no longer needed. The call will remain but will do nothing.
- `GET /revocation/registry/{rev_reg_id}`
  - Not needed — consider removing
- `PATCH revocation/registry/{rev_reg_id}`
   - We’ll investigate if this endpoint is still needed.
- **REMOVE** `POST /revocation/registry/{rev_reg_id}/definition`
- **REMOVE** `POST /revocation/registry/{rev_reg_id}/entry`
- `PUT /revocation/registry/{rev_reg_id}/fix-revocation-entry-state`
   - We’ll investigate if this endpoint is still needed.
- `GET /revocation/registry/{rev_reg_id}/issued`
- `GET /revocation/registry/{rev_reg_id}/issued/details`
- `GET /revocation/registry/{rev_reg_id}/issued/indy_recs`
- `PATCH /revocation/registry/{rev_reg_id}/set-state`
   - We’ll investigate if this endpoint is still needed.
[...]

PR #2427, opened 2023-08-17
Update /schemas to use anoncreds

usingtechnology:
WIP needs feedback to determine if this is the right direction.

Have updated:
- [All /schemas routes](https://github.com/hyperledger/aries-cloudagent-python/blob/2565fe84e2716fcda44a3b1bb32095e9c6fd62bd/aries_cloudagent/messaging/schemas/routes.py)
- [POST /credential-definitions](https://github.com/hyperledger/aries-cloudagent-python/blob/2565fe84e2716fcda44a3b1bb32095e9c6fd62bd/aries_cloudagent/messaging/credential_definitions/routes.py#L167)
- [GET /credential-definitions/<id>](https://github.com/hyperledger/aries-cloudagent-python/blob/2565fe84e2716fcda44a3b1bb32095e9c6fd62bd/aries_cloudagent/messaging/credential_definitions/routes.py#L483)
- [POST /revocations/revoke](https://github.com/hyperledger/aries-cloudagent-python/blob/1f6cbf6635fc53e5c945b95353bff48706ed9638/aries_cloudagent/revocation/routes.py#L440)

Enabled all "broken" BDD tests (except in sign-transaction due to endorser).

More credential definition routes to fix as well as revocation. The BDD tests pass now, so that means BDD does not do complete coverage.

Will have to address the pytests at some point too.

### DIFF
--- a/aries_cloudagent/anoncreds/revocation.py
+++ b/aries_cloudagent/anoncreds/revocation.py
@@ -967,7 +967,7 @@ class AnonCredsRevocation:
             failed_crids = set()
             max_cred_num = rev_reg_def.value.max_cred_num
             rev_info = rev_list_entry.value_json
-            cred_revoc_ids = rev_info["pending"] + (additional_crids or [])
+            cred_revoc_ids = (rev_info["pending"] or []) + (additional_crids or [])
             rev_list = RevList.deserialize(rev_info["rev_list"])
 
             for rev_id in cred_revoc_ids:
--- a/aries_cloudagent/messaging/credential_definitions/routes.py
+++ b/aries_cloudagent/messaging/credential_definitions/routes.py
@@ -3,8 +3,6 @@
 import json
 from time import time
 
-# from asyncio import ensure_future, shield
-# from asyncio import shield
 
 from aiohttp import web
 from aiohttp_apispec import (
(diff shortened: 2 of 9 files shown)


dbluhm:
I think this is generally the right direction. I'm not terribly fond of the commented out portions lingering but I acknowledge this isn't final yet. The only thing that bears comment is that the endorsement specific code will not persist at this layer any longer and will be a part of the Indy AnonCreds Registry/Registrar+Resolver. I'm inclined to say that we should just delete the endorsement code since the main branch will persist as a source of info on the endorser code when we get to implementing endorsement behind the pluggable registry interface.

usingtechnology:
Yes, the commented out code was just short term for easier reference to see what the differences are... will be removed as we proceed.

Thanks for the information/direction about endorser.

swcurran:
Agree about removing Endorser code and have it elsewhere.  I can’t remember if there are Endorser endpoints, but like Revocation, we need to eliminate as much of the controller related code as possible for Endorsing, and have it entirely within ACA-Py and (other than configuration), entirely within the Indy code.

usingtechnology:
I made some more changes, enough to get through the BDD tests. More work to do on cred def and revocation, but think we should do it in another PR.

swcurran:
@dbluhm — over to you for review/merging.

@usingtechnology — please add an additional issue (or issues) for the remainder of the work.  When doing revocation, make sure that there are no admin API endpoints to create/update revocation registries directly, as noted in the #2383 .

dbluhm:
Can we remove the commented and endorser code?

dbluhm (approved):

PR #2187, opened 2023-03-29
3.7 and 3.10 unittests fix

Jsyro:
Disabled two tests related to this issue. Not sure on the resolution in the future. 
https://github.com/hyperledger/aries-cloudagent-python/issues/2072#issuecomment-1489279654

Updating two other dependencies resolved the other issues. See successful test run in my fork.
https://github.com/Jsyro/aries-cloudagent-python/actions/runs/4568725894/jobs/8064139393

PR #2435, opened 2023-08-18
Fix for nightly tests failing on Python 3.10

Gavinok:
Before the recent support for nightly release tests for Python 3.10 (see the logs [here](https://github.com/hyperledger/aries-cloudagent-python/actions/runs/5861979241/job/15893033620)) were broken this fix corrected the tests by simply replacing  [async_case](https://pypi.org/project/async-case/) with the original [unittest](https://docs.python.org/3/library/unittest.html) library.

A similar issue was resolved for the same incompatibility in #2187

As [async_case](https://pypi.org/project/async-case/) was simply a backport of functionality included since Python 3.8 and acapy supports Python 3.9+ there is no longer a need for [async_case](https://pypi.org/project/async-case/).

Finally, this means we also restore the functionality of 2 additional tests that were previously skipped in #2187

## Including Tests as a Dependency for Nightly Releases
In addition, the current release does not check that the tests for Python 3.10 were successful before building a nightly release. This PR also adds this as a dependency. 

## Additional context
The current source of the error with Python 3.10 is due to [async_case](https://pypi.org/project/async-case/) using 
```python
            loop.run_until_complete(
                asyncio.gather(*to_cancel, loop=loop, return_exceptions=True))
```
as of Python 3.10 the `loop` keyword has been depreciated in [asyncio.gather](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather). 

This leads to the following exception
image

### DIFF
--- a/.github/workflows/nigthly.yml
+++ b/.github/workflows/nigthly.yml
@@ -38,7 +38,7 @@ jobs:
 
   publish:
     name: Publish
-    needs: [setup_and_check_pub]
+    needs: [tests, setup_and_check_pub]
     if: needs.setup_and_check_pub.outputs.commits_today > 0
     uses: ./.github/workflows/publish.yml
     strategy:
--- a/aries_cloudagent/admin/tests/test_admin_server.py
+++ b/aries_cloudagent/admin/tests/test_admin_server.py
@@ -2,8 +2,7 @@ import json
 
 import pytest
 import mock as async_mock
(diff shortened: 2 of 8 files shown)


swcurran:
Looks good to me, but I think it better that @dbluhm or @usingtechnology review, approve and merge this.

Thanks!

dbluhm (approved):
LGTM

PR #2436, opened 2023-08-19
#2289 Migrate to Poetry

### LINKED ISSUE #2289: Transition ACA-Py to use poetry
With #2247, we're now running on Python 3.9 and there are a number of dependency updates that I'm planning to make. This could be a good opportunity to make some improvements to our dependency management.

I am personally a fan of poetry -- it's dependency resolution is superior to pip's and it's got several nice-to-haves like builtin virtual environment handling and build and publish commands. There are still some limitations; it doesn't support platform specific builds and bundling binaries so it is unsuitable for projects like Askar right now but I think ACA-Py fits well into poetry's strengths. Externally, this transition should be seamless for ACA-Py users, whether they're consuming ACA-Py via PyPI or container image.

I'm content with our current `setup.py` and `requirements.*` files but my team has been pleased with poetry in our projects over the past couple of years.

Thoughts?

@swcurran @usingtechnology @ianco @TimoGlastra

https://python-poetry.org/

Gavinok:
This PR resolves #2289

I migrated the use of requirements.txt and setup.py over to poetry. The use of multiple requirements.*.txt is replaced with dependency groups in the `pyproject.toml`. I have also added a dedicated `[tool.poetry.group.test.dependencies]` as opposed to the original requirements.dev.txt which contained both developer dependencies and testing dependencies.


dbluhm:
I believe we want to install the dependencies for `askar`, `bbs`, and `indy` (though this one should go away very soon) as [extras](https://python-poetry.org/docs/master/pyproject/#extras) rather than as groups. From the poetry documentation:

> Dependency groups, other than the implicit main group, must only contain dependencies you need in your development process. Installing them is only possible by using Poetry.
>
> To declare a set of dependencies, which add additional functionality to the project during runtime, use [extras](https://python-poetry.org/docs/master/pyproject/#extras) instead. Extras can be installed by the end user using pip.

pyproject.toml
```diff
@@ -0,0 +1,115 @@
...
+web-py = ">=0.62,<1.0"
+pygments = ">=2.10,<3.0"
+
+# [tool.poetry.group.test.dependencies]
+asynctest= { version = "0.13.0", optional = true }
+pytest= { version = "~7.4.0", optional = true }
+pytest-asyncio= { version = "0.14.0", optional = true }
+pytest-cov= { version = "2.10.1", optional = true }
+pytest-flake8= { version = "1.1.1", optional = true }
+mock= { version = "~4.0", optional = true }
```
dbluhm, line 61: These should still be in the dev dependency group. They're used only in development so should be treated as a dependency group rather than an extra.

Gavinok: That makes perfect sense I have just moved them into the dev.dependancy group. Thanks for the feedback

bin/aca_py.py
```diff
@@ -0,0 +1,26 @@
...
+        python_executable = "python"
+    return python_executable
+
+
+def main():
+    python_executable = find_python_executable()
+    aries_cloudagent_args = sys.argv[1:]
+    subprocess.run(
+        [python_executable, "-m", "aries_cloudagent"] + aries_cloudagent_args
+    )
```
dbluhm, line 22: I don't know that this is actually a problem but just calling it out: this will result in two python processes running for the lifetime of the program. The original shell script would have been a shell process and a python process. Perhaps it would be better to `os.execv()`, resulting in only a single process?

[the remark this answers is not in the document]
dbluhm: This might actually solve the issue in my previous comment about multiple long-lived processes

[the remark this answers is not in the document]
Gavinok: I agree. Wasn't aware of all the use cases for the original script so I figured it was best to stick as close to the original as possible. I will just access the module manually rather than calling the executable. Thanks for the feedback

pyproject.toml
```diff
@@ -0,0 +1,116 @@
...
+     "python3-indy"
+]
+test = [
+     "asynctest",
+     "pytest",
+     "pytest-asyncio",
+     "pytest-cov",
+     "pytest-flake8",
+     "mock"
+]
```
dbluhm, line 109: Probably not needed?

pyproject.toml
```diff
@@ -0,0 +1,116 @@
+[tool.poetry]
+name = "aries_cloudagent"
+version = "0.10.0-rc0"
+description = ""
+authors = ["Your Name <you@example.com>"]
```
dbluhm, line 5: Is it permitted to just not include an `authors` entry? Or leave it as an empty list? Maybe we should put something like `Hyperledger Aries <aries@lists.hyperledger.org>`?

Gavinok: Unfortunately, an author is required. I just added Hyperledger as the author

usingtechnology (changes requested):
Works great - awesome job.

Can we add a Poetry.md file in the root of the project that provides a cheat sheet for developers that have not used poetry before? Particularly around virtual envs and usage for running tasks through poetry or not using virtual envs. Also, should explain where and how poetry is used to build and publish the library.

usingtechnology:
pinging @dbluhm for a final approval.

dbluhm (approved):

swcurran:
@Gavinok -- a question for you about this PR.  We discovered that the problem with the Aries Agent Test Harness that was triggered by this PR was the removal of the "bin/aca-py" shell script.  From scanning the commits, it looks like that was done deliberately, but I wanted to confirm that, and the rationale for the change. It is a breaking change we'll need to document in the changelog in case anyone (like AATH) is using the file. Do you think it would be a bad thing to put the file back or in changing to poetry should it be removed?

I see these two commits where the change was made:

* [Dropped dedicated aca-py script for poetry calling of main()](https://github.com/hyperledger/aries-cloudagent-python/pull/2436/commits/ff377ec4158f7de21946815c34a39bc2c9bf7963) 
[ff377ec](https://github.com/hyperledger/aries-cloudagent-python/pull/2436/commits/ff377ec4158f7de21946815c34a39bc2c9bf7963)
* [bin directory is no longer necessary](https://github.com/hyperledger/aries-cloudagent-python/pull/2436/commits/8347c0bd3af0dcc4061a50417e6a8383779763b9)

Thanks!

Gavinok:
> @Gavinok -- a question for you about this PR.  We discovered that the problem with the Aries Agent Test Harness that was triggered by this PR was the removal of the "bin/aca-py" shell script.  From scanning the commits, it looks like that was done deliberately, but I wanted to confirm that, and the rationale for the change. It is a breaking change we'll need to document in the changelog in case anyone (like AATH) is using the file. Do you think it would be a bad thing to put the file back or in changing to poetry should it be removed?
> 
> I see these two commits where the change was made:
> 
> * [Dropped dedicated aca-py script for poetry calling of main()](https://github.com/hyperledger/aries-cloudagent-python/pull/2436/commits/ff377ec4158f7de21946815c34a39bc2c9bf7963) 
> [ff377ec](https://github.com/hyperledger/aries-cloudagent-python/pull/2436/commits/ff377ec4158f7de21946815c34a39bc2c9bf7963)
> * [bin directory is no longer necessary](https://github.com/hyperledger/aries-cloudagent-python/pull/2436/commits/8347c0bd3af0dcc4061a50417e6a8383779763b9)
> 
> Thanks!

This may be a misunderstanding if what poetry did with its script functionality. We should be able to add the original script back into the repository. The idea was that the [tool.poetry.scripts] entry in the pyproject.toml would replace it once installed.

swcurran:
OK — could you do a quick and dirty PR to add it back? If it is safe enough, let’s just put it back and not have to explain it to someone.  It was tougher than it should have been to find in AATH.

Gavinok:
Sure thing, I can't say with 100% confidence that it will work in all cases but clearly the current solution isn't helping filling it's place

PR #2438, opened 2023-08-21
Swap out flake8 in favor of Ruff

### LINKED ISSUE #2400: Replace `flake8` / `pytest-flake8` with `ruff` / `pytest-ruff`
Flake8 is being surpassed by an extremely fast linter: https://github.com/astral-sh/ruff. 
See the readme for speed comparison and testimonials.

This is primarily suggested because `pytest-flake8` is no longer maintained, binding aca-py to flake v4 instead of the latest v6.
To shed technical debt, I propose replacing the `pytest-flake8` dependency (used in PR tests) with [`pytest-ruff`](https://github.com/businho/pytest-ruff)

dbluhm:
Resolves #2400.

This PR replaces flake8 with Ruff as suggested by @ff137. I've selected a "reasonable" set of selected and ignored error codes for now. I think we should add/no longer ignore more in the future.

One in particular that I think should have it's own issue raised over is C901 or cyclomatic complexity checks. By default, Ruff sets the threshold at 10. This is a pretty common threshold. There are 81 instances of ACA-Py exceeding that threshold. These are ignored for the time being as it's likely each instance will require individual attention.

I've also moved miscellaneous configuration for dev tools out of `setup.cfg` and translated them into `pyproject.toml`. We'll need to make sure this meshes well with changes in #2436.


dbluhm:
There are a lot of changed files in this PR. Most are auto fixes from ruff. The auto fixes are all contained in 4f02300. I recommend looking at the other commits individually to see changes I actually made by hand (which is where the issues are likely to be if there are any).

dbluhm:
This is now ready for review. @swcurran @usingtechnology. Also @Gavinok since you recent worked with the pyproject.toml and dependencies, a check from you would be good, too.

usingtechnology (approved):
looks good!
I am adding a follow up ticket to update the devcontainer documentation and sample tasks to use ruff. I don't think we need it in the PR and it will be quicker for me to take it on (not much effort really).

PR #2440, opened 2023-08-22
migrate credential definition routes to anoncreds

### LINKED ISSUE #2431: Update /credential-definitions to use anoncreds
Sub-task for #2383 

Update the remaining `/credential-definitions` endpoints to use `anoncreds`. Some work has been done while updating `/schemas` (see #2427) and associated BDD Tests.

usingtechnology:
Set existing `/credential-definitions` endpoints to use `anoncreds`.

As per the discussion, the `write_record` methods and endpoints were removed for both schema and cred defs.

Closes #2431

### DIFF
--- a/aries_cloudagent/messaging/credential_definitions/tests/test_routes.py
+++ b/aries_cloudagent/messaging/credential_definitions/tests/test_routes.py
@@ -320,7 +320,7 @@ class TestCredentialDefinitionRoutes(AsyncTestCase):
                 self.request
             )
 
-    async def test_created(self):
+    async def anoncreds_break_test_created(self):
         self.request.match_info = {"cred_def_id": CRED_DEF_ID}
 
         with async_mock.patch.object(test_module.web, "json_response") as mock_response:
--- a/aries_cloudagent/anoncreds/issuer.py
+++ b/aries_cloudagent/anoncreds/issuer.py
@@ -275,6 +275,45 @@ class AnonCredsIssuer:
                 "Error checking for credential definition"
             ) from err
 
+    async def fix_cred_def_wallet_record(
+        self,
+        cred_def_id: str,
+        issuer_did: str,
+    ) -> CredDef:
+        """
+        Store an existing credential definition if on the ledger and not in the wallet.
+
+        Args:
+            cred_def_id: the ID of the credential definition
+            issuer_id: the ID of the issuer creating the credential definition
+        Returns:
+            CredDef: the result of the credential definition creation
+
+        """
+        anoncreds_registry = self.profile.inject(AnonCredsRegistry)
+        ledger_creddef = await anoncreds_registry.get_credential_definition(
+            self.profile, cred_def_id
+        )
+        ledger_schema = await anoncreds_registry.get_schema_for_cred_def(
+            self.profile, ledger_creddef
+        )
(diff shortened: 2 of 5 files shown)


swcurran:
The description you give seems wrong.  There is no way to “recover” an issuers cred_def from the ledger —  or even recreate one if it is lost. Unlike a DID that is created from a seed, all cred_defs are created from a randomly generated number, so they cannot be recreated.

Perhaps this is for a holder or a verifier to store a cred_def for a credential they hold or receive?  Or perhaps this is for an issuer that has created a cred_def and as it in memory to store it?  I’m not sure…

Just read the description of the docstrings for the call — very weird.  “Request handler for fixing a credential definition wallet non-secret record.”  I think the next step is to go into the blame history to see if there is a PR and a description of why that is there.  I checked in ACA-Py and the demo does not use the function AFAIK.

dbluhm:
I think I can provide some clarity on the `write_record` method.

Back in Indy SDK days, you could create and submit a cred def to the ledger using the Indy CLI. The cred def private values would exist in the wallet where they needed to but ACA-Py also wrote a separate record for making it easier to keep track of schemas and cred defs it had created. This is what is referred to here as the wallet non-secret record. So this operation was used to sync up ACA-Py's "cred def created" records with the Indy SDK's cred def records.

In the AnonCreds implementation, there is no longer a separate non-secret record for cred defs and schemas to make them more searchable. Instead, the schema and cred def records created and stored on creation of the schema and cred defs are annotated with additional tags to make those records themselves more searchable.

swcurran:
So we drop that endpoint from the interface?

@ianco — looks like you can ignore the email I just sent you :-).

dbluhm:
Yep, it is no longer needed.

usingtechnology:
Thanks for the feedback and insight @swcurran / @dbluhm. OK, so to me it would make sense to drop it since we can not replicate the behaviour.  

I think I should have raised the issue sooner, as it didn't make sense to me that there was a way to replicate the data and this was more than search metadata. 

Since this preceeds the revocation APIs... are we just removing endpoints or will we be leaving them and returning a (transitional) deprecation status? Then we can argue what status code would best represent deprecation. (410 - Gone?)

aries_cloudagent/messaging/credential_definitions/tests/test_routes.py
```diff
@@ -320,7 +320,7 @@ async def anoncreds_break_test_send_credential_definition_ledger_x(self):
                 self.request
             )
 
-    async def test_created(self):
+    async def anoncreds_break_test_created(self):
```
dbluhm, line 323: Minor comment, we can also annotate tests like this with `@pytest.mark.skip("some reason")` to skip tests but to leave an indication of a test that is temporarily not useuful or needs attention.

dbluhm (approved):

aries_cloudagent/messaging/credential_definitions/tests/test_routes.py
usingtechnology: that's actually a much better idea than what i did. i have another PR (for revocation endpoints) before we get to re-implementing tests. I will mark the tests in that PR so they are obvious for subsequent PRs. I know I will probably be working on the tests PR but good to have that marked and direction provided in case it isn't me.

PR #2451, opened 2023-08-28
chore: relax connections filter DID format

chumbert:
Relax the connection filter on `my_did`, `their_did` and `their_public_did` fields of the `/connections` endpoints.

These values were restricted to `did:sov` and are now open to the generic DID format.


dbluhm:
@swcurran I think we should get this one in for 0.10.0

swcurran:
Needs to be updated, please @chumbert

PR #2456, opened 2023-08-31
Update Python image version to 3.9.18

WadeBarnes:
- Reduces the number of vulnerabilities in the image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-ARG python_version=3.9.16
+ARG python_version=3.9.18
 FROM python:${python_version}-slim-bullseye AS build
 
 WORKDIR /src
--- a/docker/Dockerfile.indy
+++ b/docker/Dockerfile.indy
@@ -1,4 +1,4 @@
-ARG python_version=3.9.16
+ARG python_version=3.9.18
 ARG rust_version=1.46
 
 # This image could be replaced with an "indy" image from another repo,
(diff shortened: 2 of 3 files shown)


dbluhm (approved):

WadeBarnes:
> I noticed that Synk is saying this is making “no change detected” — as we are still using Bullseye Slim 3.9.16. Is there a Bullseye Slim 3.9.18 that we should also update to as part of this?

The Synk scan for the PR is not scanning the docker file.  I'm basing this change off the Synk scan that includes the dockerfiles.

WadeBarnes:
> Slim-bookworm, perhaps?

I recall that breaking things when we tried it, but I could try again.

dbluhm:
> > Slim-bookworm, perhaps?
> 
> I recall that breaking things when we tried it, but I could try again.

I was mostly speculating on why the Snyk scan didn't report any changes. Your other comment explained that :slightly_smiling_face: Would be interesting to see if that change further resolved reported vulnerabilities and what breaks as a result but that doesn't seem like an urgent need right now.

WadeBarnes:
> > Slim-bookworm, perhaps?
> 
> I recall that breaking things when we tried it, but I could try again.

On the surface building with `slim-bookworm` succeeds.  However I think we should introduce this through the nightly tests as part of the matrix first to make sure there are no surprises.  I'll enter a ticket for that.

Issue #2433, opened 2023-08-18
Update pytests that were "disabled" while updating endpoints.


usingtechnology:
Sub-task for #2383 

Pytests were commented out as (methods prefixed with `anoncreds_break_`) while updating the endpoints. We need to get coverage back so fix these tests. Note that some tests may not have a fix... endorser logic is moved around. 

See related tasks:

- schemas (#2427)
- credential definitions (#2431)
- revocation (#2432)

Issue #2432, opened 2023-08-18
Update /revocation endpoints to anoncreds


usingtechnology:
Sub-task for #2383 

Update the remaining `/revocation/` endpoints to use `anoncreds`. 

Some work has been done while updating `/schemas` (see #2427) and associated BDD Tests.

usingtechnology:
I am having an extremely difficult time completing the mappings for this task.

Some are very straightforward but stumbling on something as basic as `/revocation/registry/{rev_reg_id}`. Attempting tp transform an anoncreds `RevRegDef` to a `IssuerRevRegRecord` should be straightforward but I am not seeing where fields line up. For instance, in anoncreds, we do have a RevRegDefState, that I could use to set the IssuerRevRegRecord.state, except I don't see the anoncreds value persisted.  Also, no idea what IssuerRevRegRecord.record_id is.

There are other things like is the anoncreds `RevRegDef.value.tails_location` the `tails_local_path` or the `tails_public_uri`?

And very lost on what/where I find values for IssuerRevRegRecord.revoc_reg_entry. Only thing close I could find would be calling registry.get_revocation_list (not sure what timestamp) and getting entries from that list?

I can probably attribute this to never using the revocation api and not knowing that those data calls contain or what / how they are used. But it looks like a lot of data was persisted before that is no longer persisted?

swcurran:
This is the trickiest part, as revocation has changed between the CredX implementation and AnonCreds.  Summarized:

- AnonCreds now takes the “full state” of the registry vs. the previous “deltas”.  Full state meaning for a registry, the state of every credential is listed (Revoked/Not Revoked), whereas with deltas, the list of revoked, and the list of unrevoked since the last entry are given.
- the “timestamps” are the identifier for the specific RevRegEntry written to the ledger — e.g. each time the Issuer updates the revocation state of one or more credentials. They are implemented as timestamps from the ledger, but they are used for the holder to say to the verifier “I used THIS RevRegEntry when I created by non-revocation proof”, so that the verifier can find the accumulator from the RevRegEntry that the holder used and use that to verify the proof.
- On the issuer side, one of the changes that Andrew did in ACA_Py for CredX that needs to be done in ACA-Py for AnonCreds is that the Tails file is no longer needed for issuing and revoking. As a result those calls now have a different signature. We’ve discussed those before — that you would need help on those.

[...]

PR #2458, opened 2023-09-01
Revocation API using anoncreds-rs

### LINKED ISSUE #2493: Anoncreds-RS update revocation endpoints for TAA
See #2383 and #2432

There are some endpoints used in the TAA tests that need to be updated to use anoncreds-rs.

- GET `/revocation/registry/<id>/issued`
- GET `/revocation/registry/<id>/issued/details`
- GET `/revocation/registry/<id>/issued/indy_recs`
- PUT `/revocation/registry/<id>/fix-revocation-entry-state`

These are mentioned in [Troubleshooting.md](https://github.com/hyperledger/aries-cloudagent-python/blob/09b874eea6c20d13351914e2861b04a5bd955747/Troubleshooting.md).

usingtechnology:
See issue #2432.

Trim down `/revocation` API to match up with `/anoncreds/revocation` API. 

Removed APIs related to [`Manually Creating Revocation Entries`](https://github.com/WadeBarnes/aries-cloudagent-python/blob/fed9590fe98920ea83cf0530c4f65aaac2223aae/docs/GettingStartedAriesDev/CredentialRevocation.md#manually-creating-revocation-registries). 

Updated the pytests to use `@pytest.mark.skip(reason="anoncreds-rs breaking change")` instead of renaming the tests. Still need to come back to these and fix them (see issue #2433).

There are remaining related issues - need to update TAA related APIs - see #2493 - but want to close this out and tackle that separately. (Update, these endpoints are now implemented/updated). Closes #2493

We will also have to discuss if we need to add back the `Manually Creating Revocation Entries` API as they were very handy during the IDIM reset/issues.

Since we do not have endorser enabled, I simply couldn't update/use the BDD tests in `0586-sign-transaction.feature` so I did hack together something similar that did not require an endorser. I ran against `main` and all the scenarios passed, but the included file has 1 scenario skipped because the Manual steps/API do not exist here.

### DIFF
--- a/aries_cloudagent/anoncreds/base.py
+++ b/aries_cloudagent/anoncreds/base.py
@@ -1,6 +1,6 @@
 """Base Registry."""
 from abc import ABC, abstractmethod
-from typing import Generic, Optional, Pattern, Sequence, TypeVar
+from typing import Generic, Optional, Pattern, Sequence, Tuple, TypeVar
 
 from ..config.injection_context import InjectionContext
 from ..core.error import BaseError
@@ -135,6 +135,12 @@ class BaseAnonCredsResolver(BaseAnonCredsHandler):
         self, profile: Profile, revocation_registry_id: str, timestamp: int
     ) -> GetRevListResult:
         """Get a revocation list from the registry."""
+
+    @abstractmethod
+    async def get_revocation_registry_delta(
+        self, profile: Profile, rev_reg_def_id: str, timestamp: None
+    ) -> Tuple[dict, int]:
+        """Get a revocation registry delta."""
 
 
 class BaseAnonCredsRegistrar(BaseAnonCredsHandler):
--- a/aries_cloudagent/anoncreds/base.py
+++ b/aries_cloudagent/anoncreds/base.py
@@ -11,7 +11,6 @@ from .models.anoncreds_cred_def import (
     GetCredDefResult,
 )
 from .models.anoncreds_revocation import (
-    GetRevListResult,
     GetRevRegDefResult,
     RevRegDef,
     RevRegDefResult,
@@ -129,12 +128,6 @@ class BaseAnonCredsResolver(BaseAnonCredsHandler):
         self, profile: Profile, revocation_registry_id: str
     ) -> GetRevRegDefResult:
         """Get a revocation registry definition from the registry."""
-
-    @abstractmethod
-    async def get_revocation_list(
-        self, profile: Profile, revocation_registry_id: str, timestamp: int
-    ) -> GetRevListResult:
-        """Get a revocation list from the registry."""
 
     @abstractmethod
     async def get_revocation_registry_delta(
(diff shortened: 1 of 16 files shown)


aries_cloudagent/anoncreds/base.py
```diff
@@ -136,6 +136,12 @@ async def get_revocation_list(
     ) -> GetRevListResult:
         """Get a revocation list from the registry."""
 
+    @abstractmethod
+    async def get_revocation_registry_delta(
+        self, profile: Profile, rev_reg_def_id: str, timestamp: None
+    ) -> Tuple[dict, int]:
+        """Get a revocation registry delta."""
+
```
dbluhm, line 144: Delta's are an Indy AnonCreds Method specific implementation detail. I don't think this belongs in the abstract base.

aries_cloudagent/anoncreds/registry.py
```diff
@@ -147,6 +147,15 @@ async def get_revocation_list(
...
 
+    async def get_revocation_registry_delta(
+        self, profile: Profile, rev_reg_def_id: str, timestamp: None
+    ) -> Tuple[dict, int]:
+        """Get a revocation registry delta."""
+        resolver = await self._resolver_for_identifier(rev_reg_def_id)
+        return await resolver.get_revocation_registry_delta(
+            profile, rev_reg_def_id, timestamp
+        )
+
```
dbluhm, line 158: Looks like this is the method ultimately calling the method I tagged in my last comment and that this is called from revocation routes. Since the deltas are specific to Indy, I recommend having the revocation routes grab an instance of the `LegacyIndyRegistry` directly and call the `get_revocation_registry_delta` method on it.

usingtechnology: OK, thanks for this. I have refactored base and moved the logic directly into routes. The same approach was used to implement the endpoints used in the TAA BDD tests (see #2493).

usingtechnology:
@dbluhm - bumping this if you have time to look over the updates since your last review.

aries_cloudagent/anoncreds/default/did_web/registry.py
```diff
@@ -93,6 +93,12 @@ async def get_revocation_list(
         """Get a revocation list from the registry."""
         raise NotImplementedError()
 
+    async def get_revocation_registry_delta(
+        self, profile: Profile, rev_reg_def_id: str, timestamp: None
+    ) -> Tuple[dict, int]:
+        """Get a revocation registry delta."""
+        raise NotImplementedError()
+
```
dbluhm, line 101: Ditto

dbluhm (changes requested):
Couple of corrections

aries_cloudagent/anoncreds/default/did_web/registry.py
usingtechnology: thanks!  will look over my gaffs on monday.

usingtechnology:
thanks for the extra pair of eyes @dbluhm. tests and everything work but having the wrong methods in the base class sure would have been confusing for other implementations.

dbluhm (approved):

Issue #2434, opened 2023-08-18
Regression: DID Exchange webhook state with "completed"


dbluhm:
Connection records have two different state attributes: one reflecting the old states associated with the connection protocol and one for the DID Exchange protocol. These are `state` and `rfc23_state` respectively. The terminal state for the `state` attribute is `active`. The terminal state for `rfc23_state` is `completed`. Recently something changed and we now get a `state` of `completed` under certain circumstances.

A minimal example can be found here: https://github.com/Indicio-tech/acapy-minimal-example/tree/main/examples/simple

The example can be run by:

```sh
git clone https://github.com/Indicio-tech/acapy-minimal-example
cd examples/simple
docker-compose run example
# Don't forget to down when you're done
docker-compose down -v
```

The webhooks are printed to stdout under the DEBUG log level. Right at the end of the exchange, you will see:

```
[DEBUG] Bob: Event(topic='connections', payload={'created_at': '2023-08-18T18:35:51.295588Z', 'their_role': 'inviter', 'invitation_msg_id': '99d69030-2aaf-4371-8ff7-41031ee7f325', 'invitation_mode': 'once', 'their_did': 'LnqBDAaFbVHtrZcg3XV5KC', 'state': 'response', 'request_id': '66780853-ee45-4373-8efc-3b310b8fc48c', 'routing_state': 'none', 'connection_protocol': 'didexchange/1.0', 'updated_at': '2023-08-18T18:35:51.588161Z', 'invitation_key': 'A74AJiSaMdUkEbqhAVxJJt5vfD55Gz5fjNDGczHkJHrd', 'connection_id': '850787cf-73c4-4bbe-9de6-c4873596ec2a', 'their_label': 'Alice', 'rfc23_state': 'response-received', 'my_did': 'R4mLoWnMXLT2WYS2WS3wVy', 'accept': 'manual'}, wallet_id=None)
[...]

PR #2461, opened 2023-09-05
fix: issue #2434: Change DIDExchange States to Match rfc160

anwalker293:
Includes fixes to https://github.com/hyperledger/aries-cloudagent-python/issues/2434. Changes all `ConnRecord.States` to match that of rfc160.

For additional information, here's the identical PR with some feedback from @dbluhm ~ 
https://github.com/Indicio-tech/aries-cloudagent-python/pull/151

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -533,7 +533,7 @@ class ConnRecord(BaseRecord):
     async def abandon(self, session: ProfileSession, *, reason: Optional[str] = None):
         """Set state to abandoned."""
         reason = reason or "Connectin abandoned"
-        self.state = ConnRecord.State.ABANDONED.rfc23
+        self.state = ConnRecord.State.ABANDONED.rfc160
         self.error_msg = reason
         await self.save(session, reason=reason)
 
--- a/aries_cloudagent/protocols/didexchange/v1_0/manager.py
+++ b/aries_cloudagent/protocols/didexchange/v1_0/manager.py
@@ -822,7 +822,7 @@ class DIDXManager(BaseConnectionManager):
         if responder:
             await responder.send_reply(complete, connection_id=conn_rec.connection_id)
 
-            conn_rec.state = ConnRecord.State.COMPLETED.rfc23
+            conn_rec.state = ConnRecord.State.COMPLETED.rfc160
             async with self.profile.session() as session:
                 await conn_rec.save(session, reason="Sent connection complete")
                 if session.settings.get("auto_disclose_features"):
--- a/aries_cloudagent/protocols/didexchange/v1_0/manager.py
+++ b/aries_cloudagent/protocols/didexchange/v1_0/manager.py
@@ -903,9 +903,9 @@ class DIDXManager(BaseConnectionManager):
     ) -> DIDXProblemReport:
         """Abandon an existing DID exchange."""
         state_to_reject_code = {
-            ConnRecord.State.INVITATION.rfc160
+            ConnRecord.State.INVITATION.rfc23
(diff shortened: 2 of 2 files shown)


dbluhm (dismissed):

dbluhm:
@swcurran @usingtechnology @Jsyro Alex is on my team so, in the interest of transparency, a review from one of you guys would be appreciated. To be explicit, this technically constitutes a breaking change if your controller was dependent on the incorrect state showing up on the final webhook for DID exchange. If your implementation was using the `rfc23_state`, there should be no change.

dbluhm:
Relevant comment from the original issue: https://github.com/hyperledger/aries-cloudagent-python/issues/2434#issuecomment-1701292309

> @anwalker293 spotted the issue; in the DID Exchange protocol manager, the conn_record.state is getting set to ConnRecord.State.COMPLETED.rfc23 which is equal to the string completed. This follows the pattern used elsewhere in the manager but since the states happen to align (i.e. request for rfc160 is also request for rfc23), it wasn't obvious that this was incorrect.
> 
> conn_record.state should ALWAYS be set to the string for rfc160, whether it's getting set in the connections or did exchange protocols. I think we probably should have handled the state management for these protocols differently but we're here so this is the right fix for now.

swcurran (dismissed):
An interesting change.  Looking through “blame” — this goes back to the initial implementation of DID Exchange and then was updated later.  I’m slightly concerned with the long history, but leave that to those that know more about the impact.  The change itself seems right.

dbluhm:
@usingtechnology good question and good catch -- I believe the `abandoned` method ought to use the `rfc160` value as well (@anwalker293, one more spot to change here). I believe the `routes.connections_sort_key` is fine though; it's turning the string into the Enum and then checks against that, which is a good methodology.

anwalker293:
I think this is the only place to update! But please let me know if you spot any others! :)

usingtechnology (approved):
Great catch, diagnosis and work by all on this issue!

usingtechnology:
> I think this is the only place to update! But please let me know if you spot any others! :)

Those were the only 2 places that I found, and the sorting one makes sense to leave alone.

dbluhm:
Before we merge, I'm going to check in with some other members of our team at Indicio about the impact of this change on controller work. I'll report back with their input.

dbluhm:
Heard back from them: as it happens, because of this inconsistency between the connections and DID Exchange webhooks, our controller work has treated `active` and `completed` the same way. So this change will not be breaking for us and it seems likely that those who have walked a similar path as us may have a similar equivalency in their controllers. Do we have any other controller authors on hand that we can ping for input? @swcurran

swcurran:
As I understand this, the impact on Controllers will be that the `state` of a connection as reported in webhooks will be different than previously.   We have one controller in the BC Gov world (@marcos-carretero) that I suspect might break.  Most controllers use `—auto-accept` and so will not see it, but that one is not using `—auto…` so is explicitly watching for state changes.

Questions:

- From what I read, the only difference is `completed` will become `active`.  Will any other state changes be seen?
- What release/PR did this regression occur?

Thanks!

dbluhm:
> Questions:
> 
> * From what I read, the only difference is `completed` will become `active`.  Will any other state changes be seen?
> * What release/PR did this regression occur?

Correct, no other state changes, just `completed` -> `active` and only when doing DID Exchange. The connections protocol webhooks are correct.

This was introduced at the same time DID Exchange was introduced. At least according to your previous message investigating the blame :smile: Given that the webhooks existed previous to that and had the correct states associated with it for the connections protocol, I'm still inclined to call this a regression.

swcurran (approved):

PR #2464, opened 2023-09-06
feat: add timeout to did resolver resolve method

dbluhm:
This PR adds a `timeout` parameter to the DID Resolver interface, defaulting to 30 seconds. Without it, a poorly behaved resolver plugin can "resolve" documents that aren't actually DID Docs and cause ACA-Py to run out of memory.

There's space for additional tuning on the timeout default. Some DID Methods are, in my experience, quite slow (did:btcr and did:ion resolvers can be pretty slow sometimes). 30 seconds might be overly charitable.

### DIFF
--- a/aries_cloudagent/resolver/did_resolver.py
+++ b/aries_cloudagent/resolver/did_resolver.py
@@ -30,6 +30,8 @@ LOGGER = logging.getLogger(__name__)
 class DIDResolver:
     """did resolver singleton."""
 
+    DEFAULT_TIMEOUT = 30
+
     def __init__(self, resolvers: Optional[List[BaseDIDResolver]] = None):
         """Create DID Resolver."""
         self.resolvers = resolvers or []
@@ -64,7 +66,7 @@ class DIDResolver:
                         did,
                         service_accept,
                     ),
-                    timeout if timeout is not None else 30,
+                    timeout if timeout is not None else self.DEFAULT_TIMEOUT,
                 )
                 return resolver, document
             except DIDNotFound:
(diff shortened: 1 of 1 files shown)


aries_cloudagent/resolver/did_resolver.py
```diff
@@ -42,20 +43,28 @@ async def _resolve(
         profile: Profile,
         did: Union[str, DID],
         service_accept: Optional[Sequence[Text]] = None,
+        *,
+        timeout: Optional[int] = None,
```
burdettadam, line 47: why not initialize to 30 instead of None? That would make it more evident that the time-out default is 30 seconds.

burdettadam (dismissed):

aries_cloudagent/resolver/did_resolver.py
dbluhm: Fair question; using the optional enables expression that the timeout wasn't set. Since we call `_resolve` from two other methods, and we want the timeout to apply to both and we want a default of 30 seconds when the timeout isn't set, we would have to set the default timeout to 30 on both the `resolve` and `resolve_with_metadata` methods or else make sure the timeout wasn't included in the call to `_resolve`. It ends up being cleaner to set the default when timeout is None. But I think there's a middle ground. I'll put a constant on the class that has the default timeout value so it's clearer what it is and how it's used at a glance.

usingtechnology (approved):
LGTM.
Your logic around `None` vs `30` makes sense, so works for me.

PR #2466, opened 2023-09-06
Remove old routing protocol code

### LINKED ISSUE #2430: Remove old route protocol
The route protocol was a precursor to the coordinate mediation protocol. The `routing/1.0` protocol itself will retain the `forward` message and its handler. However, the old `route-query-request` and response and `route-update-request` and response have been out of use for at least a couple of years. Before that, I don't think it had ever achieved common use to begin with. I think it's time to remove the handlers and any other references from the codebase.

dbluhm:
Resolves #2430.

This PR removes the old deprecated routing protocol messages pertaining to coordinating a "route" with another connection. This functionality was superseded by the coordinate mediation protocol. It's been well over two years since it was superseded so this PR is just taking care of the clean up that probably ought to have been done a while ago.

In this PR you will see:

- Messages and handlers for route queries and updates have been removed
- Methods on the route manager to facilitate route updates have been removed
- The Coordinate Mediation Manager now directly implements logic for performing route updates. It was previously borrowing this logic from the route manager.
- Route related code has been removed from the connection manager and connection record. This was done in such a way that the `routing_state` in the wallet records will still be accepted but will (continue to) be unused.
- The `RouteManager` continues to exist and is used to manage creation of route records. Its responsibilities are now appropriately narrowed to this. It is used by the MediationManager and MultitenancyManager to track routes.


usingtechnology (approved):
I wish I could provide more feedback on the actual logic changes/updates but not super familiar with Routing and RoutingManager. So my feedback is very much surface level. Looks good and makes sense.

PR #2470, opened 2023-09-06
Fix verkey validation to support did:key with BBS+

### LINKED ISSUE #2437: Verkey validation only supports ed25519 key type
Working with the [aries-cloudcontroller](https://github.com/didx-xyz/aries-cloudcontroller-python) (which is a client generated from the ACA-Py openapi spec), creating a DID with Bbs `bls12381g2` key type, yields a 422 unprocessable entity, due to the verkey not passing the regex defined in the DIDSchema:
```py
class DIDSchema(OpenAPISchema):
    verkey = fields.Str(
        validate=INDY_RAW_PUBLIC_KEY_VALIDATE,
...

INDY_RAW_PUBLIC_KEY_VALIDATE = IndyRawPublicKey()

class IndyRawPublicKey(Regexp):
    PATTERN = rf"^[{B58}]{{43,44}}$"
```

so, `INDY_RAW_PUBLIC_KEY_VALIDATE` is requiring it to be 43/44 base58 characters. This is only true for ed25519 key type.

Creating a DID with the Bbs key type does indeed work through ACA-Py, but doing so through an openapi generated client, will fail due to the verkey not passing validation defined for the schema.

We patched it in cloudcontroller just by removing the validation, but naturally we'd want to add it with support for Bbs. Verkey validation is happening in a few other schemas too, so they'll all need to be updated.

I can't find documentation re what the regex pattern should be for a Bbs generated verkey ... it seems to be 132 base58 characters, so it might be as simple as changing the `{43,44}` to `{43,44,132}`.  But of course I just wanna make sure of that before patching it. 

Are these requirements (re Bbs key type support) on the radar yet? And does anyone know what the Bbs verkey pattern should be?

ff137:
Resolves #2437

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -11,8 +11,8 @@ from ...messaging.models.base_record import BaseRecord, BaseRecordSchema
 from ...messaging.valid import (
     INDY_DID_EXAMPLE,
     INDY_DID_VALIDATE,
-    INDY_RAW_PUBLIC_KEY_EXAMPLE,
-    INDY_RAW_PUBLIC_KEY_VALIDATE,
+    INDY_OR_KEY_PUBLIC_KEY_EXAMPLE,
+    INDY_OR_KEY_PUBLIC_KEY_VALIDATE,
     UUID4_EXAMPLE,
 )
 from ...protocols.connections.v1_0.message_types import ARIES_PROTOCOL as CONN_PROTO
@@ -693,10 +693,10 @@ class ConnRecordSchema(BaseRecordSchema):
     )
     invitation_key = fields.Str(
         required=False,
-        validate=INDY_RAW_PUBLIC_KEY_VALIDATE,
+        validate=INDY_OR_KEY_PUBLIC_KEY_VALIDATE,
         metadata={
             "description": "Public key for connection",
-            "example": INDY_RAW_PUBLIC_KEY_EXAMPLE,
+            "example": INDY_OR_KEY_PUBLIC_KEY_EXAMPLE,
         },
     )
     invitation_msg_id = fields.Str(
--- a/aries_cloudagent/connections/models/connection_target.py
+++ b/aries_cloudagent/connections/models/connection_target.py
@@ -8,8 +8,8 @@ from ...messaging.models.base import BaseModel, BaseModelSchema
 from ...messaging.valid import (
     GENERIC_DID_EXAMPLE,
     GENERIC_DID_VALIDATE,
-    INDY_RAW_PUBLIC_KEY_EXAMPLE,
-    INDY_RAW_PUBLIC_KEY_VALIDATE,
+    INDY_OR_KEY_PUBLIC_KEY_EXAMPLE,
+    INDY_OR_KEY_PUBLIC_KEY_VALIDATE,
 )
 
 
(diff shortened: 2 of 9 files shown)


swcurran:
Great stuff — thanks!  @andrewwhitehead , can you please review?

What is the BBS+ dependency being used?  AFAIK — we are/were using Ursa, which in turn was using Mattr’s BBS+ implementation.  The Ursa BBS+ support was moved to [aries-bbssignatures-rs](https://github.com/hyperledger/aries-bbssignatures-rs).  However, the transition to that repo is still in progress such that the ACA-Py dependency is still from the (archived) Ursa project.

A further practical pain was that the BBS+ support in Mattr does not publish an Arm/M1 compatible artifact, so we can’t produce a “pure” M1 friendly ACA-Py artifact.  Does this change help with that?

dbluhm:
@ff137 For many of the updated schemas, Ed25519 keys are strictly expected because the DIDComm pack and unpack operations only support Ed25519/X25519 keys.

- ConnRecord - `invitation_key` is a key used to pack/unpack initial connection messages
- ConnectionTarget - `recipient_keys`, `routing_keys`, and `sender_key` - each of these is also a key to be used as part pack/unpack
- `verkey` fields in RegisterLedgerNymQueryString and GetDIDVerkeyResponse - These are referencing Indy ledger specific data structures; the verkey associated with nyms are always Ed25519 keys
- `signer` in SignatureDecorator - This one could probably be more flexible. The signature decorator is essentially deprecated though and the only defined signature type supported in the RFC is `ed25519Sha512_single`.
- `recipient_keys` and `routing_keys` in ServiceDecorator, in ConnectionInvitation, and in CreateInvitationRequest - these are pack/unpack keys again
- `my_verkey` and `their_verkey` in ConnectionStaticResult - Same; pack/unpack keys
- `invitation_key` in ConnectionsListQueryString - Same as `invitation_key` above
- `verkey` in DIDSchema and DIDListQueryStringSchema - This one is probably okay but there are deeper problems in how ACA-Py uses the DID objects since it always expects it to be an Ed25519 key, base58 encoded.

ff137:
@dbluhm Thank you! That overview is very helpful.

So, in summary, it'll be best for all of the models to stick with validating for an Ed25519 key?
Except for maybe DIDSchema and DIDListQueryStringSchema.

Initially it was only DIDSchema that I wanted to update, because that's the response schema for the create-did route. Using did:key + BBS, the response doesn't pass validation / yields 422 unprocessable entity (when using OpenAPI clients)

Naturally, if we can create it, then we'd want to be able to use the BBS+ verkey everywhere, so I figured update all the models, and get feedback on what needs to stay with Ed25519. But, it looks like everything does!

So, I'm wondering, how big of an undertaking is it for me to try contribute BBS+ support for some of the methods associated with these schemas? I take it that it mainly comes down to updating the DIDComm pack/unpack operation. How complex is that? And is it already on the roadmap / a goal in the short term? If it's only a long term goal / somebody not assigned to it, then I'd be willing to take a look and help to expedite it.

A simple compromise for now would be to only fix the schema for the create-did response, and then add BBS verkey validation for the other models once their methods have support for it.

dbluhm:
As a quick but somewhat incomplete response:

The pack/unpack crypto is pretty rigidly defined at this point. I think it's unlikely that it would be worth the effort to change it for DIDComm v1. DIDComm v2 I think is a bit more flexible but I haven't looked too deeply at the crypto in use there.

dbluhm:
I think this concept would need to be revisited at some point in the future but from the perspective of DIDComm v2 support for different crypto suites. I think this PR is probably DOA until there is further discussions. Thanks for your contributions!

PR #2471, opened 2023-09-06
fix: version should be set by pyproject.toml

dbluhm:
`aries_cloudagent.version.__version__` was out of sync with the version listed in the pyproject.toml file. The pyproject.toml file should be the source of truth as it will be used when publishing using poetry. However, `__version__` is used in code for some operations. To ensure that the pyproject file and the `__version__` can't get out of sync, `aries_cloudagent.version` will now be read from the package metadata.

Update: I was previously trying to read from pyproject.toml. This was a naive approach that wouldn't have worked after installing ACA-Py as a pip package. The better solution is the one I've switched to, just reading the package metadata.

### DIFF
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -100,7 +100,7 @@ Once you have the list of PRs:
    errors should be fixed in the code.
 
 5. Update the version number listed in
-   [aries_cloudagent/version.py](aries_cloudagent/version.py) and, prefixed with
+   [pyproject.toml](pyproject.toml) and, prefixed with
    a "v" in [open-api/openapi.json](open-api/openapi.json) and
    [open-api/swagger.json](open-api/swagger.json) (e.g. "0.7.2" in the
    version.py file and "v0.7.2" in the openapi.json file). The incremented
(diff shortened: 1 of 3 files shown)


swcurran:
Dang — I didn’t get the memo about that addition. I’ll update the “Publishing.md” doc to add that.  However, in that document, I did say **double check that there is not the version somewhere else** so I should have caught it…

🦤 


Bigger question — does this mean we need to do a 0.10.2 to get this properly published?

dbluhm:
@swcurran No, 0.10.1 is valid since it was published before we merged the poetry PR. So no changes required for the release.

swcurran:
Of course. Phewww…. I didn’t mess up :-)

swcurran (dismissed):

usingtechnology (approved):
great catch and fix.

dbluhm:
@swcurran @usingtechnology Sorry, fixing the conflicts on poetry.lock dismissed your reviews -- this should be ready to merge now though!

usingtechnology (dismissed):

dbluhm:
I need to resolve issues in integration tests. My path to the pyproject.toml is bad. This isn't urgent (until we start preparing for another release) so I'll mark this as a draft and we can get some of the other pending PRs merged first.

dbluhm:
@usingtechnology @swcurran this is ready for another review.

swcurran (approved):

usingtechnology (approved):

PR #2481, opened 2023-09-07
Allow direct reading of pyproject.toml version when package not installed

usingtechnology:
Slight tweak to the package version read. See #2471 

When running locally, without `poetry` (ie `devcontainer`), the pytests need to read the version but the `aries-cloudagent` package is not installed. This will read the value from the `pyproject.toml` directly if package not found.

Built images will read from the installed package metadata, but pytests (run locally) read from pyproject.toml.

### DIFF
--- a/aries_cloudagent/version.py
+++ b/aries_cloudagent/version.py
@@ -1,6 +1,31 @@
 """Library version information."""
 
-from importlib import metadata
+from contextlib import suppress
+import importlib.metadata
+from pathlib import Path
 
-__version__ = metadata.version("aries-cloudagent")
+
+def extract_version() -> str:
+    """Return package version.
+
+    Returns version of the installed package or the one found in nearby pyproject.toml
+    for cases when package is not installed (ie. local development and testing).
+    """
+    try:
+        return importlib.metadata.version("aries-cloudagent")
+    except importlib.metadata.PackageNotFoundError:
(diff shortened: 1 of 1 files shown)


dbluhm:
I'm okay with this workaround but can't help but wonder if it would be better to install the package with poetry in the devcontainer?

usingtechnology:
> 

That would be ideal, but can't tell the container to use the poetry virtual env to be the containers python (at least I couldn't figure it out). You can make any command/task you want use poetry but the built-in facilities - running the pytests - don't/can't use the virtual env :(

dbluhm:
Apologies if this is naive (I don't use the devcontainer setup in my environment (yet lol)): is this perhaps a workaround? https://www.richtman.au/blog/poetry-package-development-with-vscode-devcontainers/

usingtechnology:
Thanks for that. Let's pin this for now and I will explore what you've found. Looks promising! Going to set this to draft (in case anyone needs this for immediate help).

usingtechnology:
I'm going to close this, I did find a more acceptable workaround that is limited to devcontainer code.

Basically, do a separate call to `pip3 install aries-cloudagent` ( we do the same to install behave and black ) to the devcontainer and the pytests pick up the version since the package can be resolved. It still runs the local source for the tests...

PR #2487, opened 2023-09-08
Feat/sd jwt implementation

cjhowland:
Adds SD-JWT support for sign and verify endpoints, including optional key binding.

By default, all claims at all levels of the payload can be selectively disclosable (with the exception of essential verification data such as iss, iat, cnf, etc.), unless indicated otherwise by the issuer.

Info from @dbluhm: This PR, like the JWT sign and verify endpoints added previously, adds basic support for creation and verification of SD-JWTs. The methods/endpoints do not place any expectations on the payload beyond those required by the SD-JWT specification; in other words, the payload is not required to be a VC. As it stands, this enables other services to take advantage of ACA-Py's secure storage and DID Resolution capabilities to create and verify SD-JWT VCs but does not enable ACA-Py to do the same on its own. This is the foundational work required to later add support for SD-JWT VCs to the rest of ACA-Py's Issuance and Verification stacks.


swcurran:
What a treat on a Friday afternoon!!  Way cool!

Would be great to see a doc file included that outlined the functionality.

pyproject.toml
```diff
@@ -51,6 +51,7 @@ unflatten="~0.1"
 asyncpg = ">=0.25.0,<0.26.0"
 web-py = ">=0.62,<1.0"
 pygments = ">=2.10,<3.0"
+sd_jwt = {git = "https://github.com/openwallet-foundation-labs/sd-jwt-python.git"}
```
dbluhm, line 54: We should nudge the OWF to publish a package to PyPI.

dbluhm: Do we have any contacts over there?

dbluhm: https://github.com/openwallet-foundation-labs/sd-jwt-python/issues/6

aries_cloudagent/wallet/sd_jwt.py
```diff
@@ -0,0 +1,320 @@
+"""Operations supporting SD-JWT creation and verification."""
+
+import re
+from typing import Any, List, Mapping, Optional, Union
+from marshmallow import fields
+from jsonpath_ng.ext import parse
```
andrewwhitehead, line 6: I had to look up what `parse` was referring to, it might be better to use an alias here (`jsonpath_parse` for instance).

aries_cloudagent/wallet/tests/conftest.py
```diff
@@ -0,0 +1,66 @@
+import pytest
```
andrewwhitehead, line 1: This file doesn't seem related to the rest?

dbluhm (approved):

andrewwhitehead (approved):

andrewwhitehead:
Looks comprehensive, thanks! Hopefully that library does get released on PyPI shortly.

Issue #2488, opened 2023-09-10
KeyError raised when Subject ID is not a URI


Ennovate-com:
In valid.py, the line:
```subjects = value if isinstance(value, list) else [value]```
converts a single `value` to a single-element list in `subjects`, which ensures that `subjects` will always be a list. However, later on, `value` is treated as a dictionary rather than a list:
```uri_validator(value["id"])```

And in another place, it is treated as a list subscripted with the integer 0:
```f"credential subject id {value[0]} must be URI"'```
In this case, if `value` is not a list, a `KeyError 0` error is raised.

Once `subjects` has been assigned and is known to be a list, using the `subject` makes more sense than continuing to use the `value` having an ambiguous type.

```2023-09-10 03:31:33,175 aiohttp.access INFO 35.208.141.135 [10/Sep/2023:03:31:33 +0000] "GET /issue-credential-2.0/records?connection_id=85c3b74f-1ca7-4bcb-982b-f2493745ef2a HTTP/1.1" 200 172 "-" "Python/3.10 aiohttp/3.8.4"
2023-09-10 03:31:33,181 aries_cloudagent.admin.server ERROR Handler error with exception: 0
=================
Traceback (most recent call last):
  File "/aries_cloudagent/messaging/valid.py", line 753, in __call__
    uri_validator(value["id"])
  File "/usr/local/lib/python3.9/site-packages/marshmallow/validate.py", line 499, in __call__
    raise ValidationError(self._format_error(value))
marshmallow.exceptions.ValidationError: Value 123-45-6789_Dating is not URI

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  ...
  File "/aries_cloudagent/messaging/valid.py", line 756, in __call__
    f"credential subject id {value[0]} must be URI"
KeyError: 0
2023-09-10 03:31:33,183 aiohttp.server ERROR Error handling request
Traceback (most recent call last):
  File "/aries_cloudagent/messaging/valid.py", line 753, in __call__
    uri_validator(value["id"])
  File "/usr/local/lib/python3.9/site-packages/marshmallow/validate.py", line 499, in __call__
    raise ValidationError(self._format_error(value))
marshmallow.exceptions.ValidationError: Value 123-45-6789_Dating is not URI

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  ...
  File "/aries_cloudagent/messaging/valid.py", line 756, in __call__
[...]

PR #2490, opened 2023-09-10
Issue #2488 KeyError raised when Subject ID is not a URI

Ennovate-com:
Changed from `value`, which may or may not be a list, to `subject`, derived from `subjects` which is always a list.


dbluhm:
Thanks for the fix! At a glance, based on the details you provided in the issue, this seems like a pretty straightforward change. I'll take a closer look once checks complete :slightly_smiling_face:

dbluhm (approved):

PR #2494, opened 2023-09-15
WIP: DID Rotate Protocol

dbluhm:
This PR is an early implementation of the [RFC 0794 DID Rotate protocol](https://github.com/hyperledger/aries-rfcs/tree/main/features/0794-did-rotate). This was an exercise to feel out any potential issues with the protocol as defined.

This is not yet a complete implementation. Notably missing right now is a way to actually trigger a DID rotation through an Admin API call. I also need to finish wiring up the handlers to the manager. And who could forget about tests. :slightly_smiling_face: 

Overall, the protocol is solid. I'll report my thoughts on the protocol on the PR to the Aries RFCs.

An ACA-Py specific concern I discovered in this process that probably deserves some discussion: invalidating the connection target cache when the DID Rotation is committed. As currently structured, we never really expect the connection targets for a connection to change after the protocol establishing it concludes. DID Rotation shakes that expectation up.

Right now, in my changes, I've added a `clear_connection_targets_cache` helper method to partially address this. It's partial because it will only work for ACA-Py when run as a single instance with the default in memory cache or if you're using Indicio's redis cache plugin for whole cluster shared caching.

I think it would be better for ACA-Py to use a value including both DIDs of the connection in the cache key as a more complete solution. The problem with this and why I've not done it on a first pass is that there would need to be some restructuring of how we use the `Responder` classes so that the `send` methods expect a whole `ConnRecord` and not just a `connection_id` (or individually a `their_did` and `my_did`, I suppose) so we can extract the DIDs from it. 9 times out of 10, I would say that connection record is already available to the caller of the `send` method and my hypothesis is that in the remaining 1 time out of 10, it will not dramatically impact performance of critical operations (it is not common for us to know the connection_id and not also have a connection record -- maybe only a few places in the Admin API might be like this).

I'll do some more investigation on these points. Might be good to talk implications at the next ACA-PUG call.

cc @swcurran @TelegramSam
FYI @Jsyro this has ties back to updating the DIDs associated with connections that we've talked about recently.

### DIFF
--- a/aries_cloudagent/connections/base_manager.py
+++ b/aries_cloudagent/connections/base_manager.py
@@ -335,7 +335,7 @@ class BaseConnectionManager:
             [self._extract_key_material_in_base58_format(key) for key in routing_keys],
         )
 
-    async def record_keys_for_public_did(self, did: str):
+    async def record_keys_for_resolvable_did(self, did: str):
         """Record the keys for a public DID.
 
         This is required to correlate sender verkeys back to a connection.
--- a/aries_cloudagent/connections/base_manager.py
+++ b/aries_cloudagent/connections/base_manager.py
@@ -626,6 +626,21 @@ class BaseConnectionManager:
             targets = await self.fetch_connection_targets(connection)
         return targets
 
+    async def clear_connection_targets_cache(self, connection_id: str):
+        """Clear the connection targets cache for a given connection ID.
+
+        Historically, connections have not been updatable after the protocol
+        completes. However, with DID Rotation, we need to be able to update
+        the connection targets and clear the cache of targets.
+        """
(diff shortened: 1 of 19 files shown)


dbluhm:
Oh, also this commit warrants calling out: https://github.com/hyperledger/aries-cloudagent-python/pull/2494/commits/85470cdb13b9e4387448192e4c7f93ccdd26224a

Some of the expected attributes in the problem report message schema were out of alignment with the spec. Specifically:
- `problem-items` which should be `problem_items`
- `who-retries` which should be `who_retries`
- and `time-noticed` which should be `noticed_time`

I don't believe these were being used anywhere so I went ahead and corrected the schema.

amanji:
Have opened up #2816. I will add the description from here to that and close this

PR #2495, opened 2023-09-18
fix: run tests script copying local env

dbluhm:
This PR fixes the scripts/run_tests script and its dockerfiles to make sure it doesn't copy local .venv directories into the container image. This was causing the script to fail.


swcurran (approved):
Works nicely — thanks!

Issue #2498, opened 2023-09-20
Dockerfile RUST_LOG field has wrong value


loneil:
I was inquiring about suppressing some askar INFO level logs and after some feedback from Andrew and Wade discovered that setting the RUST_LOG level to `warn` should do that.

In our ACA-Py deployment I could see this environment variable is set to `warning` instead. Which doesn't seem to be right.
![image](https://github.com/hyperledger/aries-cloudagent-python/assets/17445138/d9f451fe-c407-4d21-89de-b58c4020e3d0)

We're not setting this in any Traction deployment vals for what I'm testing it in, so I'm fairly sure it's coming from a mistake in the Dockerfile here.
https://github.com/hyperledger/aries-cloudagent-python/blob/d705ca2fe9132fc5bdc07e427e10f74201bbfac2/docker/Dockerfile#L25

Andrew mentioned that the value could be `warn` or fine tuned with `aries_askar=warn`

I tried both of those manually in my env and they both did what I wanted (suppressing all the askar INFO logs about pool connections and session starts etc). 
So depends on what is wanted for this Dockerfile, but existing `warning` is not having intended effect.

(Note: same issue in Indy Dockerfile if that one matters)

loneil:
Note this is also being used in the devcontainer but that is setting it differently
https://github.com/hyperledger/aries-cloudagent-python/blob/d705ca2fe9132fc5bdc07e427e10f74201bbfac2/.devcontainer/devcontainer.json#L47

PR #2499, opened 2023-09-21
Use correct rust log level in dockerfiles 

loneil:
From issue https://github.com/hyperledger/aries-cloudagent-python/issues/2498

It looks like `warning` is not a valid Rust log level string https://docs.rs/log/latest/log/enum.Level.html
From what I've seen in traction (which deploys using the dockerfile here), this means logs default to INFO level here even though dockerfile is trying to have it at the warn level. (see details in ticket for what I tried out)


loneil:
I suppose if anyone is using this dockerfile and counting on having the INFO level logs come out (even though they're not supposed to) they would be missing those after updating to a version including this fix. But in that case one should be setting their own deployment to INFO anyways...

WadeBarnes (approved):

Issue #2497, opened 2023-09-19
Failing to deserialize DIDDocumentRoot error when using the Indicio mediator and 0.10.2 ACA-Py instance


esune:
[Updated this description to clarify the issue...]

We were testing a deployment of 0.10.2, and were getting errors when connections were established using the Indicio Public Mediator.  When we switched to use the BC Gov Mediator instance, things worked again.

This was observed using ACA-Py `0.10.2rc0` in a Traction multi-tenant instance, targeting a BC Wallet emulator - @cvarjao please add context if you have more information.

c.c.: @loneil who has been looking into this, @dbluhm based on expressed interest from today's ACA-Pug meeting  
   
  ```log
  2023-09-18 22:34:53,164 aries_askar.native.aries_askar.backend.db_utils INFO src/backend/db_utils.rs:168 | Dropped pool connection
2023-09-18 22:34:53,166 aries_cloudagent.core.dispatcher ERROR Handler error: BaseConnectionManager.get_connection_targets
Traceback (most recent call last):
File "/home/aries/.venv/lib/python3.9/site-packages/pydid/validation.py", line 13, in wrap_validation_error
yield
File "/home/aries/.venv/lib/python3.9/site-packages/pydid/resource.py", line 68, in deserialize
return parse_obj_as(cls, value)
File "pydantic/tools.py", line 38, in pydantic.tools.parse_obj_as
File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 4 validation errors for ParsingModel[DIDDocumentRoot]
__root__ -> service -> 0 -> serviceEndpoint
didcomm:transport/queue is not a valid absolute or relative DID URL (type=value_error.invaliddidurl)
__root__ -> service -> 0 -> serviceEndpoint
invalid or missing URL scheme (type=value_error.url.scheme)
__root__ -> service -> 0 -> serviceEndpoint
unexpected value; permitted: '' (type=value_error.const; given=didcomm:transport/queue; permitted=('',))
__root__ -> service -> 0 -> serviceEndpoint
value is not a valid list (type=type_error.list)
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
File "/usr/local/lib/python3.9/asyncio/tasks.py", line 256, in __step
result = coro.send(None)
File "/home/aries/.venv/lib/python3.9/site-packages/aries_cloudagent/connections/base_manager.py", line 609, in get_connection_targets
[...]

PR #2500, opened 2023-09-21
fix: update pydid

dbluhm:
This will correct the errror reported in #2497.


dbluhm:
Multiformats has an issue with typing-extensions that is causing some minor pains: https://github.com/hashberg-io/multiformats/issues/10

I'll do some tweaks to work around this.

swcurran (approved):

Issue #2492, opened 2023-09-14
Routing keys for public DIDs should be published as did:keys


dbluhm:
When using a mediator, routing keys are published in the same attrib transaction that includes the service endpoint of the mediator. Those routing keys are currently being published as base58 encoded values. They should be did:keys.

Example: https://indyscan.indiciotech.io/tx/IND_TESTNET/domain/15635

Issue #2357, opened 2023-07-26
Mediation webhook routing keys aren't did:keys


dbluhm:
Just about everywhere else, in the coordinate mediation protocol and manager, we're now handling routing and recipient keys in did:key format. However, webhooks with topic `mediation` still have routing keys reported as raw base58 encoded public keys. This is occurring at least on the mediation client side.

dbluhm:
```
aries_cloudagent.core.event_bus DEBUG Notifying subscribers: <Event topic=acapy::record::mediation::granted, payload={'endpoint': 'http://mediat
or:3000', 'role': 'client', 'state': 'granted', 'created_at': '2023-07-26T19:07:12.663391Z', 'connection_id': '059b2c28-9ba0-476c-8184-718b7af9a676', 'mediator_terms': [], 'mediation_id': '2
c4bee7f-8ba0-49b1-9d4c-09a802a80b2a', 'updated_at': '2023-07-26T19:07:12.774293Z', 'routing_keys': ['DdMf4dzeASFyQ8M16tjKXv6epQMCyeURov7orS9xRk9T'], 'recipient_terms': []}>
```

PR #2502, opened 2023-09-21
fix: mediation webhook routing keys

anwalker293:
Re: #[2357](https://github.com/hyperledger/aries-cloudagent-python/issues/2357) and #[2492](https://github.com/hyperledger/aries-cloudagent-python/issues/2492)

Indicio PR: https://github.com/Indicio-tech/aries-cloudagent-python/pull/153

Tagging and credit due to @dbluhm

### DIFF
--- a/aries_cloudagent/commands/default_version_upgrade_config.yml
+++ b/aries_cloudagent/commands/default_version_upgrade_config.yml
@@ -1,3 +1,5 @@
+v0.11.0:
+  update_routing_keys: true
 v0.8.1:
   resave_records:
     base_record_path:
--- a/aries_cloudagent/commands/upgrade.py
+++ b/aries_cloudagent/commands/upgrade.py
@@ -1,6 +1,7 @@
 """Upgrade command for handling breaking changes when updating ACA-PY versions."""
 
 import asyncio
+import json
 import logging
 import os
 import yaml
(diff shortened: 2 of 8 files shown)


aries_cloudagent/commands/default_version_upgrade_config.yml
```diff
@@ -1,3 +1,5 @@
+v0.11.0:
+  update_routing_keys: true
```
dbluhm, line 2: Assuming this is merged in for 0.11.0, I believe this is the correct way to specify that our upgrade routine should be run? @shaangill025

aries_cloudagent/commands/upgrade.py
```diff
@@ -394,6 +404,61 @@ async def update_existing_records(profile: Profile):
...
+    from raw base58 encoded values.
+
+    Steps:
+        for each mediation record stored in the wallet:
+            ensure the stored routing_keys list is formated as did:keys
+            save the record if modified
+        for each routing record stored in the wallet:
+            ensure the stored recipient keys are formated as did:keys
+            save the record if modified
+    """
```
dbluhm, line 420: This is the routine we've added for updating the routing keys. The goal of doing the upgrade this way was to ensure that we don't have to keep base58 -> did:key translation code on the `MediationRecord` and `RouteRecord` hydration methods from now until the rest of forever. Is this the right approach? Do we anticipate the upgrade script will continue to be used in this way in the future? Does this logic belong here or should it go somewhere else?

cc @shaangill025 @swcurran

dbluhm:
Marking as draft temporarily: I need to review a couple of instances of these values being used.

dbluhm:
I believe I was overzealous with my recommended changes. To address the linked issues, we need only modify the mediation record. It is probably a good idea to limit this PR to just addressing the mediation record routing key representation. Changing route record would support changing over to sending and receiving `did:key`s in `routing/1.0/forward` messages but I think that's a discussion for another time.

dbluhm:
Closing this PR; the upgrade process was an interesting exercise but ultimately the two issues are better addressed by a smaller change set found in #2516. We'll hold onto these changes in this branch in case they become relevant (if we needed to update the `to` field behavior in `routing/1.0/forward`).

PR #2515, opened 2023-09-25
refactor: drop mediator_terms and recipient_terms

dbluhm:
Term negotiation was dropped from the coordinate mediation protocol early in the proposal process. Our implementation helped to determine that it should be dropped but predates the decision. The terms support has hung around unused for years. Time to remove it.

Terms will no longer be stored in mediation records but old mediation records should continue to work (the values will just be ignored).


swcurran (approved):
Tested in demo with mediation and with AATH.

Issue #1044, opened 2021-03-21
use and accept updated media types


troyronda:
AcaPy should be using the Aries RFC media types rather than `application/ssi-agent-wire`.

swcurran:
Agreed, and I believe we've added that.  But for backwards compatibility, both need to be accepted.

@andrewwhitehead can you provide a status on this?

swcurran:
A priority for accomplishing AIP 2.0 Interop with Aries Framework Go.

andrewwhitehead:
At the moment we don't support this, but for testing it could be enabled under the same flag that turns on the new message types for outgoing requests.

swcurran:
What is the work and who should do it?  Is this something you should do, or perhaps @shaangill025 ?

Presumably, in receiving messages, we could handle 0 or more values.  We can also send them out, but I'm guessing we don't know what those receiving our messages would do.  Correct?

andrewwhitehead:
On receiving we don't enforce the media type. I can do a quick pass today along with the did:peer changes for testing.

swcurran:
Completed

Issue #1042, opened 2021-03-21
https prefix for PIURI as the default


troyronda:
Please use the https prefix for PIURI as the AcaPy default. We should be demonstrating the https prefix as the outgoing PIURI in the test harness.

swcurran:
Completed

PR #2517, opened 2023-09-26
Change arg_parse to always set --emit-new-didcomm-prefix and --emit-new-didcomm-mime-type to true

swcurran:

Approach used was to accept the parameters, but whether set or not, set corresponding config value to True.

Left the config settings, but updated the help to say not setting the value is not possible.

Didn't change the tests, but added a comment about why the of the old setting is not particularly useful...


swcurran:
Fixes (finally!!) Issues #1042 and #1044.

Open to suggestions if we should be more aggressive at removing the setting where it is used.

usingtechnology (approved):
nice work!

dbluhm (approved):

PR #2520, opened 2023-09-27
Update shared components

andrewwhitehead:
Updates to:
- Indy-VDR 0.4 release (with did:indy support)
- Indy-Credx 1.1.0 minimum version
- Askar 0.3.0-dev.1


swcurran:
@andrewwhitehead — any thoughts on the failing Integration tests? I’ve not looked at the logs.

I did run the acapy AATH test suite and everything ran perfectly.

andrewwhitehead:
It looks like those test failures are highlighting the remaining places where store sessions are left open. I'll submit a separate PR with fixes for those, but I suspect that the default maximum number of connections has been reduced for Sqlite, based on the 'available parallelism' it can detect for the current environment. The integration tests could override that but we might want to increase the default maximum number of connections as well.

swcurran:
So this relates to that “define max connections based on available parallelism” — the thing that we saw a month or two back with tests on GHA failing, when working elsewhere?

Not sure what you mean here:

>  The integration tests could override that but we might want to increase the default maximum number of connections as well.

Where else do you mean?  Overriding it in the integration tests would get them to pass, but where else would we increase it?

andrewwhitehead:
We would increase the default that Askar's sqlite backend settles on, at the moment it can be as low as 2 if it thinks it's on a single-processor system (as I think it can appear on CI).

dbluhm:
Are we still planning to get this in for 0.11.0?

swcurran:
Looks like there was a breakthrough this morning on what was holding up getting the official releases for Askar and AnonCreds-RS, which in turn was holding up this. Let’s give it a bit more time.

dbluhm (approved):

PR #2521, opened 2023-09-28
Avoid multiple open wallet connections

andrewwhitehead:
Fixes issues with timeouts during integration tests when the maximum number of connections is low.


swcurran:
Looking good. Ran all the AATH acapy tests and all good.

dbluhm (approved):

PR #2523, opened 2023-10-02
Add /rotate to revocation api.

usingtechnology:
Added in `handle_full_registry` but seems to be some issues with how the counts work.
Cannot set the `next_index` to start at 0 because the credential data is invalid.
And because we don't track the index and current revocation id externally, we can't trigger the `handle_full_registry` until after a credential is created. So for every list/index we lose 1 slot. The only way I could successfully have registries rollover AND create the credentials successfully was triggering 1 slot before the max. cred. num.

### DIFF
--- a/aries_cloudagent/anoncreds/models/anoncreds_revocation.py
+++ b/aries_cloudagent/anoncreds/models/anoncreds_revocation.py
@@ -135,6 +135,8 @@ class RevRegDefState(BaseModel):
     STATE_FAILED = "failed"
     STATE_ACTION = "action"
     STATE_WAIT = "wait"
+    STATE_DECOMMISSIONED = "decommissioned"
+    STATE_FULL = "full"
 
     class Meta:
         """RevRegDefState metadata."""
@@ -179,6 +181,8 @@ class RevRegDefStateSchema(BaseModelSchema):
                 RevRegDefState.STATE_FAILED,
                 RevRegDefState.STATE_ACTION,
                 RevRegDefState.STATE_WAIT,
+                RevRegDefState.STATE_DECOMMISSIONED,
+                RevRegDefState.STATE_FULL,
             ]
         )
     )
--- a/aries_cloudagent/anoncreds/revocation.py
+++ b/aries_cloudagent/anoncreds/revocation.py
@@ -10,6 +10,7 @@ from pathlib import Path
 import time
 from typing import List, NamedTuple, Optional, Sequence, Tuple
 from urllib.parse import urlparse
+from uuid import uuid4
 
 from aries_askar.error import AskarError
 import base58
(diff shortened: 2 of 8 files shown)


usingtechnology:
I am adding more context for the issues I hit dealing with rev list indexes/cred revocation id

First, this is when I set the starting `"next_index"` to [zero](https://github.com/usingtechnology/aries-cloudagent-python/blob/eb962c0c43df3898fbb760e4ed3c761646d02ea6/aries_cloudagent/anoncreds/revocation.py#L421)...

This is the holder side exception:

```
Credential: state = credential-received, cred_ex_id = b98bbfcd-385c-4729-9950-8f52442e475f
2023-10-03 16:27:41,207 aries_cloudagent.protocols.issue_credential.v2_0.formats.indy.handler ERROR Error storing credential: None - Error processing received credential
2023-10-03 16:27:41,208 aries_cloudagent.messaging.base_handler ERROR Error storing issued credential
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/anoncreds/holder.py", line 198, in store_credential
    cred_recvd = await asyncio.get_event_loop().run_in_executor(
  File "/usr/local/lib/python3.9/asyncio/futures.py", line 284, in __await__
  File "/usr/local/lib/python3.9/asyncio/tasks.py", line 328, in __wakeup
Credential: state = abandoned, cred_ex_id = b98bbfcd-385c-4729-9950-8f52442e475f
Problem report message: Error processing received credential. Error: Invalid structure. Caused by: Issuer is sending incorrect data.
    future.result()
  File "/usr/local/lib/python3.9/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/usr/local/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
Credential: state = done, cred_ex_id = b98bbfcd-385c-4729-9950-8f52442e475f
  File "/home/vscode/.local/lib/python3.9/site-packages/anoncreds/types.py", line 292, in process
    bindings.process_credential(
  File "/home/vscode/.local/lib/python3.9/site-packages/anoncreds/bindings.py", line 715, in process_credential
    do_call(
  File "/home/vscode/.local/lib/python3.9/site-packages/anoncreds/bindings.py", line 504, in do_call
    raise get_current_error(True)
anoncreds.error.AnoncredsError: Error: Invalid structure
Caused by: Issuer is sending incorrect data


  File "/workspaces/aries-cloudagent-python/aries_cloudagent/protocols/issue_credential/v2_0/handlers/cred_issue_handler.py", line 71, in handle
    cred_ex_record = await cred_manager.store_credential(cred_ex_record)
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/protocols/issue_credential/v2_0/manager.py", line 623, in store_credential
    await cred_format.handler(self.profile).store_credential(
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/protocols/issue_credential/v2_0/formats/indy/handler.py", line 435, in store_credential
    raise e
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/protocols/issue_credential/v2_0/formats/indy/handler.py", line 415, in store_credential
    cred_id_stored = await holder.store_credential(
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/anoncreds/holder.py", line 207, in store_credential
    raise AnonCredsHolderError("Error processing received credential") from err
aries_cloudagent.anoncreds.holder.AnonCredsHolderError: Error processing received credential
```

dbluhm:
Quick ack from me; I'll take a closer look when I can!

usingtechnology:
Every rev list will be missing 1 slot; not a huge deal on larger sets.  

I am not quite sure where the logic needs to be updated; it's possible it needs to be changed in the `anoncreds` `Credential.create` method. It would be nice if the library knows that it is on the last index to publish an event or something. Or if it has a pre-check method that would indicate how much space is left in the list? 

It may be possible that it is a timing issue for the additional steps (upload tails file and register the revocation list), although I tried adding plenty of time (5 seconds) to see if that allowed those tasks to complete.

dbluhm:
Speaking from memory, I think I once asked @andrewwhitehead why the first index is unused. I remember there being a good reason but I don't recall the details unfortunately. For a time I thought that the generic interface would have to be 0-based and then we'd have to map it to Indy where it was strictly expected to be 1-based. I had logic that performed this mapping and then, while testing, I found it was actually causing issues. I believe the anoncreds library is expecting indexes to be 1-based for most operations but the full revocation status list is of course just a list making it subject to 0-based indexing for us in python.

andrewwhitehead:
There's no technical reason why the counting has to start at 1, but that's the convention right now. The entries in the tails file are multiplied by powers of the secret `gamma` value, and the first usable tails entry is the first power, which is I think why the numbering was started there.

swcurran:
I’m not clear what “it" is that is starting skipping 1.  Could someone clarify?

usingtechnology:
Near as I can tell, if I set everything to be zero-based, so pass 0 to the anoncreds library to Credential.create, the credential is created successfully but the HOLDER cannot store the credential.

```
anoncreds.error.AnoncredsError: Error: Invalid structure
Caused by: Issuer is sending incorrect data
```

So something... the wallet? the spec? does not like zero-based, so the first credential is created using `1`. But everything else is zero-based.

andrewwhitehead:
I added a couple changes here to verify the revocation index when a credential is created, as zero shouldn't be accepted: https://github.com/hyperledger/anoncreds-clsignatures-rs/pull/24

In indy-credx (using deltas), the indexes are always 1-based. For status lists, I'm not sure at the moment if the first index is skipped, or if it's interpreted as starting at 1.

PR #2527, opened 2023-10-03
chore: update pydid

dbluhm:
This update includes changes recently committed to the 0.10.X releases, correcting an issue that prevented mediators from handling `didcomm:transport/queue` as a service endpoint.

A small but notable difference between this and the 0.10.X releases: in changes queued for the "next" release, we added `peerdid`. `peerdid` also depends on pydid. I had to backport fixes to the pydid 0.3.x line to workaround conflicts caused by using 0.4.0 of pydid with peerdid. I'm working with the SICPA team to resolve this. In the meantime, 0.3.9 will work exactly as we need anyways.


dbluhm:
Gentle nudge on this one; this is ready for review @swcurran / @usingtechnology

swcurran (approved):

PR #2529, opened 2023-10-04
Feat: Support subwallet upgradation using the Upgrade command

### LINKED ISSUE #2507: Support updating sub-wallets [Multitenant env] with `Upgrade` command
The scope will include:
- add 2 startup arguments for upgrade command, `--all-subwallets` for upgrading all subwallets or `--subwallet-id` for specifying individual subwallet to upgrade. 
- Then get all WalletRecords [`find_all_records` or `retrieve_by_id` based upon the above option] to upgrade,  call `MultitenantManager -> get_wallet_profile` to get the wallet_profile  for each of them and apply the existing upgrade logic.

shaangill025:
- resolve #2507


dbluhm:
It's not something that was introduced by this PR but I think it's worth calling out nonetheless:

First, I think the whole `upgrade` method is due for a refactor. It's a beast of a method. When we recently worked through the upgrade process, working out how to potentially do an upgrade for a change we were working on, it took quite some time to parse through the code.

Second, consider this line:

https://github.com/hyperledger/aries-cloudagent-python/blob/9b16660508c3b1e9c70af46e38afbe51f8fcc6d1/aries_cloudagent/commands/upgrade.py#L395

This line is retrieving ALL records of the given type at once. This can result in out-of-memory issues if there are enough records in the wallet of that record type. We should probably be using a `BaseStorageSearch` to perform this record retrieval so that it can be done in batches: https://github.com/hyperledger/aries-cloudagent-python/blob/9b16660508c3b1e9c70af46e38afbe51f8fcc6d1/aries_cloudagent/storage/askar.py#L203-L235

shaangill025:
@dbluhm Re the need to refactor, I mentioned it to @swcurran and I think this can be taken up in the future (separate from this PR). Re using `BaseStorageSearch -> search_records`, I have replaced `query` in `upgrade.py` with it.

dbluhm (approved):

PR #2533, opened 2023-10-05
refactor: make ldp_vc logic reusable

### LINKED ISSUE #2468: JSON-LD Sign/Verify endpoints do things differently than ICv2/PPv2
The JSON-LD Sign and Verify endpoints provide a mechanism for doing LDP signatures on JSON-LD documents and verifying signed JSON-LD documents. Issue Credential v2 and Present Proof v2 provide the DIDcomm interface for doing the same when using the ld-proof-detail attachment format. The scope of these components is of course different; however, at some point in the call stack, they should probably be calling the same code for signing and verifying signatures. They do not.

This has the additional side effect that recent updates to add additional Signature Suite types to the JSON-LD signatures are not usable from the sign/verify endpoints.

These endpoints should be updated to use the same underlying calls as ICv2 and PPv2.

dbluhm:
This PR is a refactor of LDP VC handling within ICv2 and PPv2 to make it usable outside of those protocol's handlers. I believe this results in a cleaner implementation, separating the concerns of preparing a credential from the ICv2 protocol and its messages and handlers for those messages.

This refactor enables the other major addition in this PR, adding a couple of new endpoints for LDP VC issue and verify, `POST /vc/ldp/issue` and `POST /vc/ldp/verify`. These endpoints are similar to the `/jsonld/sign` and `/jsonld/verify` endpoints. However, the big difference is that the `/jsonld` endpoints did not use the same LDP signing methods used in ICv2 or PPv2. This has caused the two implementations to diverge more and more over time. With this PR, these endpoints are still available but are marked as deprecated. One other significant difference, the inputs for endpoints have changed to be more specific to signing and verifying LDP VCs and not just arbitrary JSON-LD documents.

Fixes #2468.

### DIFF
--- a/aries_cloudagent/config/default_context.py
+++ b/aries_cloudagent/config/default_context.py
@@ -141,6 +141,7 @@ class DefaultContextBuilder(ContextBuilder):
         plugin_registry.register_plugin("aries_cloudagent.messaging.jsonld")
         plugin_registry.register_plugin("aries_cloudagent.resolver")
         plugin_registry.register_plugin("aries_cloudagent.settings")
+        plugin_registry.register_plugin("aries_cloudagent.vc")
         plugin_registry.register_plugin("aries_cloudagent.wallet")
         if wallet_type == "askar-anoncreds":
             plugin_registry.register_plugin("aries_cloudagent.anoncreds")
--- a/aries_cloudagent/messaging/jsonld/routes.py
+++ b/aries_cloudagent/messaging/jsonld/routes.py
@@ -59,7 +59,11 @@ class SignResponseSchema(OpenAPISchema):
     error = fields.Str(required=False, metadata={"description": "Error text"})
 
 
-@docs(tags=["jsonld"], summary="Sign a JSON-LD structure and return it")
+@docs(
+    tags=["jsonld"],
+    summary="Sign a JSON-LD structure and return it",
+    deprecated=True,
+)
 @request_schema(SignRequestSchema())
 @response_schema(SignResponseSchema(), 200, description="")
 async def sign(request: web.BaseRequest):
@@ -119,7 +123,11 @@ class VerifyResponseSchema(OpenAPISchema):
     error = fields.Str(required=False, metadata={"description": "Error text"})
 
 
-@docs(tags=["jsonld"], summary="Verify a JSON-LD structure.")
+@docs(
+    tags=["jsonld"],
+    summary="Verify a JSON-LD structure.",
+    deprecated=True,
+)
 @request_schema(VerifyRequestSchema())
 @response_schema(VerifyResponseSchema(), 200, description="")
 async def verify(request: web.BaseRequest):
--- a/aries_cloudagent/protocols/issue_credential/v2_0/formats/ld_proof/handler.py
+++ b/aries_cloudagent/protocols/issue_credential/v2_0/formats/ld_proof/handler.py
@@ -2,7 +2,7 @@
 
 
 import logging
-from typing import Mapping, Optional
+from typing import Mapping
 
 from marshmallow import EXCLUDE, INCLUDE
 from pyld import jsonld
@@ -11,30 +11,12 @@ from pyld.jsonld import JsonLdProcessor
 from ......messaging.decorators.attach_decorator import AttachDecorator
 from ......storage.vc_holder.base import VCHolder
 from ......storage.vc_holder.vc_record import VCRecord
-from ......vc.ld_proofs import (
-    AuthenticationProofPurpose,
-    BbsBlsSignature2020,
-    CredentialIssuancePurpose,
-    DocumentLoader,
-    Ed25519Signature2018,
-    Ed25519Signature2020,
-    LinkedDataProof,
-    ProofPurpose,
-    WalletKeyPair,
-)
+from ......vc.ld_proofs import DocumentLoader
 from ......vc.ld_proofs.check import get_properties_without_context
-from ......vc.ld_proofs.constants import (
-    SECURITY_CONTEXT_BBS_URL,
-    SECURITY_CONTEXT_ED25519_2020_URL,
-)
 from ......vc.ld_proofs.error import LinkedDataProofException
-from ......vc.vc_ld import LDProof, VerifiableCredential, VerifiableCredentialSchema
-from ......vc.vc_ld import issue_vc as issue
-from ......vc.vc_ld import verify_credential
-from ......wallet.base import BaseWallet, DIDInfo
-from ......wallet.default_verification_key_strategy import BaseVerificationKeyStrategy
-from ......wallet.error import WalletNotFoundError
-from ......wallet.key_type import BLS12381G2, ED25519
+from ......vc.vc_ld import VerifiableCredential, VerifiableCredentialSchema
+from ......vc.vc_ld.manager import VcLdpManager, VcLdpManagerError
+from ......vc.vc_ld.models.options import LDProofVCOptions
 from ...message_types import (
     ATTACHMENT_FORMAT,
     CRED_20_ISSUE,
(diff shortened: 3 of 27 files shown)


dbluhm:
Some additional testing and fixes pending for this PR. I wanted to open this to see if there were any thoughts on this question:

Should I also move the holder related functions of storing LDP VCs to this interface?

swcurran:
Assuming that we will soon be supporting signing and verifying AnonCreds in W3C format proofs, is there anything else that needs to be done in preparation for that?  Will we be able to use the same interface to sign such credentials?  And ideally, sign a VC twice — once with AnonCreds, once with another key type?  AFAIK, the AnonCreds signature will be aligned with Linked Data Proofs.

Another thought. I believe the “LDP” term has been replaced with Data Integrity Proof…

I’ll edit this message — have to run right now...

dbluhm:
Good questions. I really enjoy how nice and simple everything is when it comes to VC formats :smile: 

Based on my understanding, AnonCreds in W3C format would be formatted as JSON-LD but would not be using Linked Data Proof signatures since they will of course use CL Signatures. So it would technically not fall in the category of ldp_vc. And to your point on Data Integrity Proofs replacing Linked Data Proofs, I believe you're correct. Based on casual observation, it seems the term `ldp_vc` will be sticking around for a while at least, given that the format is included in the OpenID4VCI spec as "VC secured using Data Integrity, using JSON-LD, with proof suite requiring Linked Data canonicalization" (which is hilariously verbose). That section clarifies:

> Note that Data Integrity used to be called Linked Data Proofs, hence "ldp" in the Credential format identifier

I haven't looked deeply at Data Integrity Proofs but I believe there would be changes required at layers lower than the ones I've touched in this PR. Superficially, it seems the biggest change would be how the signature is represented.

I think the interface I've added here would be beneficial to later adding DIP support, switching behavior by processing the `proof_type` option. Perhaps that means the manager itself should be renamed :thinking:

I don't think this extends to supporting AnonCreds in W3C format, though. AnonCreds would need a different set of options, at a minimum, which I think makes it not fit particularly well into this interface. I could be persuaded otherwise.

dbluhm:
Oh, I just read more of https://w3c.github.io/vc-data-integrity/; it seems that AnonCreds would actually be able to be considered a generic Data Integrity Proof so I think I'm wrong about what I just said.

andrewwhitehead:
Looks like a nice clean-up to me!

dbluhm:
Brief update: I'm working through the test failures and moving tests that used to apply to the ICv2 and PPv2 format handlers for ld_proof and  dif respectively to tests for the added manager!

dbluhm:
This is now ready for review! @swcurran I think this is a good change for JSON-LD creds but there's probably a conversation worth having about how this should interact with the `DataIntegrityProofs` in the future. This is a good stepping stone in that direction; it will be easier to make that switch when it's time.

swcurran:
@PatStLouis — this might be relevant to the work you are doing — being able to use the JSON-LD credentials independent of DIDComm protocols.  I think you mentioned that yesterday?  Please let me know. Your review of this PR would also be welcome!  :-)

PatStLouis:
@swcurran thank you for pointing this out to me. @dbluhm I will have a thorough look tomorrow but this seems like a great functionality to have. I'd like to have a discussion with you around the differences you highlighted between these new endpoints and the json-ld endpoints. I understand this will replace the existing json-ld endpoints?

2 comments/question I have

1. Have you considered using the endpoints `/credentials/issue` and `/credentials/verify` to align with the vc-api credential management spec? This would enable aca-py to be interoperable with w3c test-suites out of the box. They also have different endpoints for presentations, `/presentations/sign` and `/presentations/verify` which could be interesting to consider. I'm currently writing a PR on the vc-api spec around how error messages should be returned, I'd like to know your approach on this as well! I will base this on the VC data model spec problem details section.
[https://w3c-ccg.github.io/vc-api/#issue-credential](https://w3c-ccg.github.io/vc-api/#issue-credential)
[https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1338.html#problem-details](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1338.html#problem-details)

2. What are your thought's on credential storage? Currently the only way to store a json-ld credential is through the issue-credential v2 protocol. There is however some logic to query those credentials through the `/credentials/w3c` endpoint. Do you see adding another mechanism to store an already issued credential?

And you are correct, CL signatures are considered data integrity proofs. However, AnonCreds VC type will be difficult to be issued with an endpoint like this because the process for the credential request requires a back and forth between the issuer and the holder. There would need to be some reflections about how to provide the request to this endpoint and how it will look. Super interested in helping with this however.

PatStLouis:
Pulling in @wip-abramson since we recently had discussions around this for a project we are working on

dbluhm:
Hey @PatStLouis, thanks for your interest! I'm not necessarily opposed to implementing the VC-API; the interface is inspired by the VC-API, as was the LD-Proof-VC-Detail attachment format that this work kind of inherits from. There are subtle differences right now and it's been a minute since I did the bulk of this work. I'll take some time to compare and see if there's a clear reason for the differences.

PatStLouis:
Have a look here if you want an easy to run test-suite that would interact the same way. Just plug in the endpoints and see what happens: [https://github.com/credential-handler/verifiable-credential-playground-test-suite](https://github.com/credential-handler/verifiable-credential-playground-test-suite)

wip-abramson:
I agree with @PatStLouis.

I think we need an endpoint to store an issued JSON-LD VC. Or is it imagined that applications would handle storage of credentials themselves? 

I also think supporting AnonCreds issuance in this way will be hard. AnonCreds issuance requires interaction between issuer and holder passing cryptographic information that needs to be used in protocols other that json-ld sign. To implement this outside of DIDComm, I think we would need API calls to represent each of these interactions (leaving the communication out of band - potentially CHAPI or something else).

Really, AnonCreds jsonld issuance isn't JSON-LD sign at all (I don't think). Isn't it just JSONLD pack? Or W3C VC pack? Some transformation that takes a signed AnonCred and represents it in W3C VC format.

PatStLouis:
@wip-abramson You had some interesting points relating to didauth and credential storage, would you mind sharing these as well? This should be documented and discussed (its out of scope for this PR however discussing it doesn't hurt)

wip-abramson:
I'm not sure what I said more about credential storage.

But DIDAuth, is generally recommended when issuing a credential to a DID. First the issuer must request a DIDAuth from the holder for the DID they wish to include in the credential. See https://chapi.io/developers/wallets/#did-authentication-with-chapi

I actually think this would be possible with the endpoints proposed here. You would just JSONLD sign something like this:

```
const didAuthPresentation = {
    '@context': [
        'https://www.w3.org/2018/credentials/v1',
        'https://w3id.org/security/suites/ed25519-2020/v1'
    ],
    type: ['VerifiablePresentation'],
    holder: 'did:key:z6MkeprvBw4RFHJPQEmtioq4xRrN6Tk8EBSJ37eBCBQNHRjZ',
};
```

PatStLouis:
Bold of you to post javascript snippets in an aca-py thread. Thanks for the input!

PatStLouis:
@swcurran testing now

PatStLouis:
This looks great and works really well. I would have 2 comments:

Would it make sense to make the `proofType` optional? AFAIK in acapy the key type is selected upon wallet creation, so by providing a did:key wouldn't it be already bound to a specific proof type? If I'm wrong please enlighten me!

The created timestamp could be formatted to use the 'Z' alternative to specify UTC. Also according to the [XMLSCHEMA11-2 date-time specification](https://www.w3.org/TR/xmlschema11-2/#dateTime) the milliseconds should be omitted.

`2023-12-07T17:58:51.962131+00:00` -> `2023-12-07T17:58:51Z`

Question, when you verify a vp, do you also verify the vc signature it holds?

I like the included details upon verification, great context and useful.

Overall I like this alot in comparison to the json-ld endpoints

swcurran:
@dbluhm — let me know how you want to handle @PatStLouis ’s comments — include or merge as is.

Thanks @PatStLouis — good stuff!

PatStLouis:
@dbluhm you should come present this at the vc-api meeting next week, I know a few people who will be thrilled to see this alignment :)

dbluhm:
@PatStLouis Was it the proof created at timestamp specifically that you were referring to?

PatStLouis:
Sounds good, I was able to sign most of the credentials from [this list](https://github.com/credential-handler/vc-examples/tree/main/credentials)

Thumbs up from me!

swcurran (approved):
Approved based on evaluation of @PatStLouis . I’ll leave it to you @dbluhm to decide when to merge.

Awesome work developing and reviewing — thanks!

PR #2537, opened 2023-10-08
fix (backport): routing behind mediator 

dbluhm:


swcurran:
@codespree -- a "0.10.4-rc0" image has been created -- https://github.com/hyperledger/aries-cloudagent-python/pkgs/container/aries-cloudagent-python/135290717?tag=py3.9-0.10.4-rc0 -- are you able to give it a try?

@WadeBarnes -- if you can deploy this release in any BC Gov deployments, that would be great.

aries_cloudagent/resolver/default/legacy_peer.py
```diff
@@ -152,7 +163,7 @@ def didcomm_services_recip_keys_are_refs_routing_keys_are_did_key(
                     service["routingKeys"] = [
                         DIDKey.from_public_key_b58(key, ED25519).key_id
```
codespree, line 164: Without applying the transform in the first case, wouldn't it still be a key and not ref? As `DIDKey.from_public_key_b58(key, ED25519).key_id` would provide a `did:key:....` not a ref. In which case, it would still fail to give DIDCommService at the `pydid.deserialize_document()` stage? @dbluhm

dbluhm: `DIDKey.key_id` is a reference (perhaps not the best name for the property)

PR #2541, opened 2023-10-09
Anoncreds-rs pytest update

### LINKED ISSUE #2433: Update pytests that were "disabled" while updating endpoints.

usingtechnology:
Updates to pytests that were affected by anoncreds-rs updates.

Since we have not updated/implemented endorser in the anoncreds-rs branch, I have left endorser related tests as skipped. These are now marked:

`@pytest.mark.skip(reason="anoncreds-rs/endorser breaking change")`

closes #2433

### DIFF
--- a/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
+++ b/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
@@ -198,7 +198,7 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
             if not profile.settings.get_value("wallet.type"):
                 # TODO is this warning necessary?
                 reason += ": missing wallet-type?"
-            raise AnonCredsResolutionError(reason)
+            raise AnonCredsRegistrationError(reason)
 
         # Translate schema into format expected by Indy
         LOGGER.debug("Registering schema: %s", schema_id)
--- a/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
+++ b/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
@@ -198,7 +198,7 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
             if not profile.settings.get_value("wallet.type"):
                 # TODO is this warning necessary?
                 reason += ": missing wallet-type?"
-            raise AnonCredsRegistrationError(reason)
+            raise AnonCredsResolutionError(reason)
 
         # Translate schema into format expected by Indy
         LOGGER.debug("Registering schema: %s", schema_id)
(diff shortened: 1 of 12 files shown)


aries_cloudagent/anoncreds/default/legacy_indy/registry.py
```diff
@@ -198,7 +198,7 @@ async def register_schema(
             if not profile.settings.get_value("wallet.type"):
                 # TODO is this warning necessary?
                 reason += ": missing wallet-type?"
-            raise AnonCredsRegistrationError(reason)
+            raise AnonCredsResolutionError(reason)
```
dbluhm, line 201: It seems like this really ought to be a registration error rather than a resolution error; the operation attempted is a `register` rather than a `resolve` so, even in the event we were attempting to resolve something in order to register something immediately after, I think it is better for this to be a registration error from an interface perspective.

aries_cloudagent/anoncreds/issuer.py
```diff
@@ -220,7 +222,7 @@ async def create_and_register_schema(
             raise AnonCredsIssuerError(
                 "Schema already exists but was not in wallet; stored in wallet"
             ) from err
-        except AnoncredsError as err:
+        except (AnoncredsError, BaseAnonCredsError, LedgerError) as err:
```
dbluhm, line 225: `LedgerError` will only be raised by Indy components; `LedgerError` should be wrapped into a `BaseAnonCredsError` subtype rather than be handled in this abstract/general interface.

aries_cloudagent/anoncreds/issuer.py
```diff
@@ -335,7 +337,7 @@ async def create_and_register_credential_definition(
                 CredDef.from_native(cred_def),
                 options,
             )
-        except AnoncredsError as err:
+        except (AnoncredsError, BaseAnonCredsError, LedgerError) as err:
```
dbluhm, line 340: Same as previous comment.

aries_cloudagent/anoncreds/issuer.py
usingtechnology: Will do.

aries_cloudagent/anoncreds/default/legacy_indy/registry.py
usingtechnology: updated related code

aries_cloudagent/anoncreds/issuer.py
usingtechnology: updated related code

aries_cloudagent/anoncreds/issuer.py
usingtechnology: updated related code.

dbluhm (approved):

PR #2558, opened 2023-10-19
Fix: RevRegEntry Transaction Endorsement

### LINKED ISSUE #2441: Possible endorsement processing bug -- RevRegEntry transaction not being signed
A team has reported that there might be a problem with the RevRegEntry endorsement process -- the transaction is not being properly signed by the Endorser when requested by the Author.  The problem is only apparent on ledgers where the RevRegEntry transactions must be signed by an Endorser, and not all Indy ledgers require such a signature. The Sovrin MainNet does require it. It appears when the first RevRegEntry is processed (along with the RevRegDef), the transaction is signed (or perhaps it is part of the RevRegDef transaction?), but subsequent RevRegEntry transactions are not signed by the Endorser.

The first part of this task is to review the code and see if the endorser flow for a RevRegEntry is indeed processed in the same way as all other transactions. This may be through inspection or by observing the Author-Endorser interactions for publishing the RevRegs.

If the issue is confirmed, a fix would be appreciated.

This is an urgent issue for the team facing the problem and any help from the community would be appreciated.

shaangill025:
- resolve #2441


aries_cloudagent/revocation/manager.py
```diff
@@ -147,7 +151,27 @@ async def revoke_credential(
...
+                    await issuer_rr_upd.send_entry(self._profile)
+                else:
+                    async with self._profile.session() as session:
+                        try:
+                            connection_record = await ConnRecord.retrieve_by_id(
+                                session, connection_id
+                            )
+                        except StorageNotFoundError:
+                            raise RevocationManagerError(
+                                f"No connection record found for id: {connection_id}"
```
swcurran, line 164: Is this checking to see if there is an endorser connection?  Is so, the error message should be more explicit, I think.

shaangill025: Yes, to check that the `connection_id` if provided when calling `POST /revocation/revoke` exists. If no `connection_id` is provided and the profile is author role then connection_id will be retrieved from `endorser.endorser_alias` which will always exist. Maybe I can change it to `No endorser connection record .... `

esune (commented):
Do these changes also account for the scenario where the revocation entry is not published contextually to the call to the `/revocation/revoke` endpoint, but rather using the `/revocation/publish-revocations` endpoint to publish in bulk?

shaangill025:
> Do these changes also account for the scenario where the revocation entry is not published contextually to the call to the `/revocation/revoke` endpoint, but rather using the `/revocation/publish-revocations` endpoint to publish in bulk?

Thanks for pointing this out, now it does.

swcurran:
Not for this issue, but doesn’t the “revoke and publish” and the “publish” endpoints use a common “publish” method?  I would think they would...

shaangill025:
> but doesn’t the “revoke and publish” and the “publish” endpoints use a common “publish” method?

https://github.com/hyperledger/aries-cloudagent-python/blob/2c901a9f1b45e3cca54b3a9f737477ef91a31513/aries_cloudagent/revocation/manager.py#L150

https://github.com/hyperledger/aries-cloudagent-python/blob/2c901a9f1b45e3cca54b3a9f737477ef91a31513/aries_cloudagent/revocation/manager.py#L243

dbluhm:
Any word on when we can look at getting the failing tests resolved?

esune:
> Any word on when we can look at getting the failing tests resolved?

@shaangill025 is looking into them, I'm following-up today to see where we are at

andrewwhitehead:
It looks like the new unit tests were written before rebasing on some changes to the test framework. The fix seems to be just renaming 'async_mock' to 'mock' in test_manager.py.

andrewwhitehead:
In the integration test it looks like Acme receives a notification that their credential was revoked, but when checking the status it hasn't been updated yet. Could be a timing issue or an unpublished update?

swcurran:
Which is definitely related to the code being changed.

esune:
> In the integration test it looks like Acme receives a notification that their credential was revoked, but when checking the status it hasn't been updated yet. Could be a timing issue or an unpublished update?

That's what I was wondering. It looks like the workflow was re-run and now only 1 scenario is failing, which seems to be a good candidate for timing issues (but I admit I have not had a chance to look into the test code yet to be 100% sure).

esune (approved):
Changes look good to me. If @dbluhm can take a peek and confirm I did not miss anything I think it's good to go 👍🏻

dbluhm (dismissed):
I'm not thrilled about the continued propagation of `endorser_conn_id` and `write_ledger` through several interfaces but I think that's unavoidable until we have endorser support behind the ledger agnostic AnonCreds interface. LGTM.

dbluhm:
Do we want to try to get this out in a patch release separate from the AnonCreds-RS changes being made on main (in addition to merging this to main)?

esune:
> Do we want to try to get this out in a patch release separate from the AnonCreds-RS changes being made on main (in addition to merging this to main)?

Yes, I think this was the strategy we decided to go with.

dbluhm (approved):

PR #2559, opened 2023-10-19
refactor: replace multiformats library

### LINKED ISSUE #2501: Replace or fix multiformats library?
The multiformats library we're using has some issues that were reported back in May. The issue can be worked around by pinning typing-extensions to an older version but this workaround is causing some heartburn when it comes to solving dependency trees.

We should see if we can get this fixed or perhaps use another library.
Relevant issue: https://github.com/hashberg-io/multiformats/issues/10

dbluhm:
This replaces the multiformats library with a very basic implementation included directly within ACA-Py. Given that the multiformats library has gone stale and my PR has languished, I think this is a good alternative to using that library. The implementation is very simple and currently only supports the minimum currently required by ACA-Py. Expanding the implementation in the future should be trivial.

Fixes #2501


swcurran (approved):
LGTM

usingtechnology (approved):

PR #2561, opened 2023-10-24
refactor: use did-peer-2 instead of peerdid

dbluhm:
Speaking frankly, I felt that the peerdid library was a bit overly complicated. On top of that, peerdid is not up to date with the corrections proposed in this PR to the did:peer method specification:

https://github.com/decentralized-identity/peer-did-method-spec/pull/62

did-peer-2 is a minimal implementation of did:peer:2 and did:peer:3. It is up to date with the above PR. The corrections to the spec and the library lend themselves to a simpler implementation of the did:peer:2 and did:peer:3 resolvers. This refactor supports work to add did:peer:2 support to OOB + DID Exchange.

New in this PR is a cleanup mechanism for the did:peer:3 to did:peer:2 mapping records. When a connection is deleted, we check whether there are any potential mappings to remove.


dbluhm:
Integration tests failed. Looking at the logs, it seems like it was a JSON-LD context load failure: https://www.w3.org/2018/credentials/v1 failed to load. I think this is unrelated to my changes so I'm rerunning the action.

swcurran:
Looks good — running the AATH `acapy` runset...

swcurran (approved):

PR #2564, opened 2023-10-25
Playground needs optionally external network

usingtechnology:
Sigh... when documenting aries-mediator-service / redid-pq-demo and running through the instructions, I needed to make the playground network external. 

Unclear why it worked fine with the base mediator service and it's network but regardless, this is needed to allow the mediator service with redis demo to work.

Note that the `ELK_NETWORK_NAME` will always be an external network so adding that notation to always be `true`.


dbluhm (approved):

usingtechnology:
@dbluhm - Sorry about all these small fixes. Thanks for reviewing and merging them!

PR #2565, opened 2023-10-25
chore: dependency updates

dbluhm:
This PR is a routine dependency update with the exception of pytest-asyncio and pytest-cov, which were both updated to significantly newer versions.

I believe this update was postponed previously because of the flake8/ruff stuff.


swcurran (approved):

usingtechnology (approved):
LGTM - `./run_tests` ran successfully.

Issue #2567, opened 2023-10-25
Merging changes from anoncreds-rs branch


andrewwhitehead:
At the moment this branch has diverged significantly from the main branch, and there are likely too many conflicts to merge it directly, so we need to look at how to merge the changes that are there in separate phases. The branch also currently adds an `anoncreds` module instead of renaming/reusing classes that are present in the `indy` module, so it would be best to avoid code duplication while merging and ensure that we maintain test coverage.

For reference, there is a diff between the two branches here: https://github.com/hyperledger/aries-cloudagent-python/compare/anoncreds-rs (select the Files Changed tab)

Suggested steps are below, to be discussed and updated:

0. Rename the `indy` module to `anoncreds`, and rename `IndyHolder` et al to `AnonCredsHolder` et al.
 - Merge into main at the end of this work.
 - Indy SDK may still be supported, and CredX is still supported.

0a. Consider: Remove Indy SDK support?

1. Merge (from the anoncreds-rs branch via cherry picks or copy/paste/tweak as makes sense) the generic AnonCreds resolver/registry support
 - Code contained in `anoncreds/base.py` and `anoncreds/default`
 - Update `credential_definitions/routes` and `schemas/routes` to use the abstract registry class (this removes the custom handling around BaseMultitenantManager in those places)
    - per tenant ledger settings (copied into multiple places)
 - Update the `issue-credential` and `present-proof` indy handlers and to use the new resolver (see diff)

2. Merge the `AnonCredsHolder/Issuer/Verifier` classes, using `IndyHolder/Issuer/Verifier` as the base classes and supporting dependency injection
 - This may require adding new methods to the base classes
 - Tests should be gated by a new 'anoncreds-rs' flag defined in conftest.py
 - Initial merge may omit revocation support
 - Should include adding the `anoncreds` library to the images used for testing on CI

2.5. Update the routes
 - Create the new /anoncreds/schema, /anoncreds/creddef, /anoncreds/revocation routes
 - Change/stub the existing ones to point to the new implementations -
 - drop ones that we don't want anymore -- the ones that allow the controller to manage revocation, endorser
 - consider which to add back for "cleanup a mess" scenarios
    - draw on our experience -- not for backwards compatibility.

3. AnonCreds revocation support
 - Includes various changes in `protocols/present_proof/indy/pres_exch_handler.py` for example, to handle revocation status lists instead of deltas. Maybe this can be handled more generically in the `IndyVerifier` base class.
[...]

PR #2571, opened 2023-10-29
Anoncreds rs merge from main up to the latest

ianco:
Merge from main up to the point that poetry and ruff are added (commit 9fc598ffe2c406397b30484cb1aa37d6df8ff1cc - see branch https://github.com/ianco/aries-cloudagent-python/tree/ruff_updates_pre)

See comment in issue #2567

Also merged up to the latest code from main and did a bit of cleanup, there are still some test errors ...


ianco:
This PR has a few failing tests, but overall cleans up most of the differences between the main and anoncreds-rs branch (pulls all updates into the anoncreds-rs branch and updates the ruff formatting)

After this PR is merged there will be about ~100 files that differ between the 2 branches.

ianco:
Status - the `./scripts/run_tests` runs cleanly, however `./scripts/run_tests_indy` doesn't run at all.

The docker images builds and when the tests run it gets the following error:

`E   importlib.metadata.PackageNotFoundError: aries-cloudagent`

Not sure what's going on, I suspect the `poetry install` is installing code somewhere and then `poetry run` is looking for the code somewhere else ...

@dbluhm can you take a look?  I've compared the docker scripts in the various branches and can't tell what's going wrong ...

dbluhm:
@ianco looking at the GHA run of the tests, I didn't see the ImportError you mentioned; has this been resolved with your most recent push?

ianco:
@dbluhm  I get the error running locally.  If I run `./scripts/run_tests` everything runs fine but if I run `./scripts/run_tests_indy` then I get the error.  (Both run fine in both the `main` and the current `anoncreds-rs` branch).

The GHA just runs `poetry run pytest` it doesn't use either of these scripts.

usingtechnology:
i'm also getting it locally, seeing if I can figure it out. complicated dockerfile!

ianco:
OK it may be happening on `main` as well, I ran a quick test and just got the same error on `main`, I'm just doing a docker purge and re-testing again

ianco:
OK confirmed I get the same error in `main`.  Since the whole purpose was to build a dockerfile including indy-sdk and run the relevant tests - and we're ending support for indy-sdk - I suggest we ignore the error.

We'll need to review the tests to make sure we're running the full set of tests in GHA (and not just the non-indy-flagged tests).

usingtechnology:
@ianco 
in `Dockerfile.indy`, change line the `acapy-test` image (line 179). The change is to copy over the aries_cloudagent source and tell poetry to install it all. We need the source directory on the image for a "full" poetry install.

```
# ACA-Py Test
# Used to run ACA-Py unit tests with Indy
FROM indy-base as acapy-test

USER indy

RUN mkdir src test-reports

WORKDIR /home/indy/src

RUN mkdir -p test-reports && chown -R indy:indy test-reports && chmod -R ug+rw test-reports

ADD ./README.md pyproject.toml ./poetry.lock ./
COPY ./aries_cloudagent/ ./aries_cloudagent/

USER root
RUN pip install --no-cache-dir poetry
RUN poetry install --no-directory -E "askar bbs indy"

ADD --chown=indy:root . .
USER indy

ENTRYPOINT ["/bin/bash", "-c", "poetry run pytest \"$@\"", "--"]
```

andrewwhitehead:
@usingtechnology  Sounds like that fix might need to be added to the main branch?

andrewwhitehead (approved):

PR #2573, opened 2023-10-30
chore: point to official sd-jwt lib release

dbluhm:
This PR adjusts the version of sd-jwt pulled to be the official release now published to PyPI: https://pypi.org/project/sd-jwt/

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -938,19 +938,19 @@ test = ["pytest (>=6)"]
 
 [[package]]
 name = "filelock"
-version = "3.12.4"
+version = "3.13.1"
 description = "A platform independent file lock."
 optional = false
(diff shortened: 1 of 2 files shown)


dbluhm:
@swcurran I think it would be good to get this in for 0.11.0; assuming tests pass, I think it's a pretty innocuous change and removes the only "unofficial"/git dependency we have, which should make integrating with ACA-Py from plugins or elsewhere simpler.

swcurran:
Before or after 0.11.0-rc0?  Does it matter?

dbluhm:
Either way works for me!

swcurran (approved):
LGTM.  Should have merged before 0.11.0-rc0, as it failed to publish, likely because of this.  :-(

Issue #2549, opened 2023-10-17
Credential Request for a deleted message does not result in a Problem Report


swcurran:
Specific issue reported:

  1.  Start the Get Person Credential Process in the BC Wallet
  2.  Continue in the Wallet and BCSC app up to receiving the offer
  3.  Delete the credential exchange record using Postman
  4.  Accept the offer -- received the “This is taking longer than usual”.
  
Presumably, the "Request Credential" message from the wallet did not trigger a Problem Report message to the wallet. This is (likely) using the Issue Credential v1.0 protocol.

The most general issue is if/should ACA-Py should (automatically?) send a Problem Report to a connection that sends a message for  a non-existent protocol instance?  When the thread-id for a message is not found, what should be done?

Options include:

- send a problem report automatically, for all protocols.
- send a problem report or not based on a configuration setting.
- send a web hook to the controller and expose an endpoint such that the controller can send a problem report.
  - Does such an endpoint exist?

Interesting to know if the handling (whichever is chosen) requires an update per protocol, or can this be handled at a higher level?  My guess is that this would be a per protocol (and perhaps per message) change -- perhaps throwing an exception (e.g., "No protocol instance found") when the thread-id is not found, that can be in turn be caught, and a Problem Report sent or not.

Need to look at the Aries RFCs (notably, 0003 Protocols, 0005 DIDComm and 0008 Threads) to check for guidance there.

usingtechnology:
Just leaving a little update here on progress of the investigation.
I am using issue credential 2.0, I can go to 1.0 if needed... (and should to confirm behaviour). I am just running a multi-tenanted instance locally to do the investigation; allows me to breakpoint and debug.

Basic Steps:
- create issuer and holder tenants
- connect tenants
- issuer get a public did
- issuer create schema/cred def
- issuer make offer
- holder can find the offer
- issuer deletes their cred exchange record
- holder wants to accept the offer ()

In the issuer side of the code, we hit a logic loophole where we create a cred_ex record when one is not found. The comment is `holder sent this request free of any offer`. See [manager receive_request](https://github.com/hyperledger/aries-cloudagent-python/blob/f80639c5f139abc1dc42aadc68d67f2904449f71/aries_cloudagent/protocols/issue_credential/v2_0/manager.py#L425)

[...]

PR #2577, opened 2023-10-31
Send Problem report when CredEx not found

usingtechnology:
See #2549 

Address the situation when an issuer sends an offer and then deletes that offer (credential exchange) before the holder accepts it. When this is an `indy` credential exchange, a Problem Report is sent; when this is a `json-ld` exchange, the holder's accept actually becomes an unbound request for the credential and the flow carries on.

A discussion is warranted to see if we want to implement something more global? Seems like at any point in any protocol one side may not find their record and the other side should be notified of the abandonment. Maybe it is already done and this was just a gap...

Added BDD tests to address the deletion and followup acceptance for `indy` and `json-ld` and also the holder sending a request for credential to issuer. Again, `indy` is not allowed, `json-ld` can flow into a credential issuance.

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v2_0/manager.py
+++ b/aries_cloudagent/protocols/issue_credential/v2_0/manager.py
@@ -412,16 +412,16 @@ class V20CredManager:
         # the record by connection_id.
         connection_id = None if oob_record else connection_record.connection_id
 
-        format_handlers = []
-        format_handlers_without_offer = []
-        for format in cred_request_message.formats:
-            f = V20CredFormat.Format.get(format.format)
-            if f:
-                format_handlers.append(f)
-                if f.handler(self.profile).can_receive_request_without_offer():
-                    format_handlers_without_offer.append(f)
+        handlers = [
+            handler(self.profile)
+            for format in cred_request_message.formats
+            if (handler := V20CredFormat.Format.get(format.format))
+        ]
+        handlers_without_offer = [
+            handler for handler in handlers
+            if handler.can_receive_request_without_offer()
+        ]
 
-        handlers = format_handlers
         async with self._profile.session() as session:
             try:
                 cred_ex_record = await V20CredExRecord.retrieve_by_conn_and_thread(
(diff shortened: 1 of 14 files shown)


aries_cloudagent/protocols/issue_credential/v2_0/manager.py
```diff
@@ -412,6 +412,16 @@ async def receive_request(
...
         connection_id = None if oob_record else connection_record.connection_id
 
+        format_handlers = []
+        format_handlers_without_offer = []
+        for format in cred_request_message.formats:
+            f = V20CredFormat.Format.get(format.format)
+            if f:
+                format_handlers.append(f)
+                if f.handler(self.profile).can_receive_request_without_offer():
+                    format_handlers_without_offer.append(f)
```
dbluhm, line 422: Good case for a list comprehension:

```python
handlers = [
    handler(self.profile)
    for format in cred_request_message.formats
    if (handler := V20CredFormat.Format.get(format.format))
]
handlers_without_offer = [
    handler for handler in handlers
    if handler.can_receive_request_without_offer()
]
```

aries_cloudagent/protocols/issue_credential/v2_0/manager.py
```diff
@@ -420,31 +430,33 @@ async def receive_request(
...
-                    role=V20CredExRecord.ROLE_ISSUER,
-                    auto_remove=not self._profile.settings.get(
-                        "preserve_exchange_records"
-                    ),
-                    trace=(cred_request_message._trace is not None),
-                    auto_issue=self._profile.settings.get(
-                        "debug.auto_respond_credential_request"
-                    ),
-                )
+                if len(format_handlers_without_offer):
```
dbluhm, line 435: Nitpick but I'd say it's more idiomatic to just do `if format_handlers_without_offer:`

aries_cloudagent/protocols/issue_credential/v2_0/manager.py
usingtechnology: i like it; will do.

usingtechnology:
@dbluhm - just bumping this for another review. added your suggestions and then put a similar exception handling in the v1 code.

dbluhm (approved):
LGTM!

PR #2578, opened 2023-10-31
feat: emit did:peer:2 in DID Exchange

dbluhm:
This PR implements did:peer:2 support for DID Exchange. If the `--emit-did-peer-2` flag is set, ACA-Py will create and send a did:peer:2 DID instead of an unqualified DID and an attached Document in DID Exchange messages.

This PR also implements the `did_rotate~attach` from the [proposed DID Exchange v1.1](https://github.com/hyperledger/aries-rfcs/pull/795). It will include the `did_rotate~attach` attribute on responses when relevant and will expect to receive them on responses as well.

To achieve this, the following changes were made:
- Add `wallet.create_key`. `create_key` is just a more generic `create_signing_key`. Literally no difference in implementation; it just seemed an inappropriate name for how it was being used. I can be convinced that we should not do this.
- Add `wallet.store_did`. This enables us to more easily create DIDs outside of the wallet. The `wallet.create_local_did` method is starting to get a bit complicated. Rather than shoehorning all DID creation into the wallet interface, I think it makes sense to transition to a system where the wallet creates the primitives (such as keys) but then the creation of the DID itself is performed outside of the wallet. This was especially necessary because, in order to create a did:peer:2, we also need service info which can differ per connection (mediation used or not, for instance) and the wallet is simply not aware of the concept of services nor the context of the connection and associated mediation records. As part of an early pass, I even went as far as to introduce a DID Registration interface (which makes this attempt perhaps my third one in recent memory). I continue to struggle to create an interface that is generic enough to enable registering DIDs using plugged in registrars but still actually be useful. The requirements between methods varies so widely that the interface becomes nearly useless (at least in the context of ACA-Py). I'm finding it more reasonable to simply call a specific "registrar" directly. In these changes for did:peer:2, this is simply calling into the did-peer-2 library's `generate` method.
- Add `--emit-did-peer-2` flag and associated setting.
- Add support for easily creating an attachment with `text/string` mime type, as is used in the `did_rotate~attach`.
- Add the `PEER2` DID Method definition

Opening as a draft since, at the time of opening, this hasn't been thoroughly exercised by tests yet.

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -1168,6 +1168,12 @@ class ProtocolGroup(ArgumentGroup):
                 "using unencrypted rather than encrypted tags"
             ),
         )
+        parser.add_argument(
+            "--emit-did-peer-2",
+            action="store_true",
+            env_var="ACAPY_EMIT_DID_PEER_2",
+            help=("Emit did:peer:2 DIDs in DID Exchange Protocol"),
+        )
 
     def get_settings(self, args: Namespace) -> dict:
         """Get protocol settings."""
@@ -1234,6 +1240,10 @@ class ProtocolGroup(ArgumentGroup):
         if args.exch_use_unencrypted_tags:
             settings["exch_use_unencrypted_tags"] = True
             environ["EXCH_UNENCRYPTED_TAGS"] = "True"
+
+        if args.emit_did_peer_2:
+            settings["emit_did_peer_2"] = True
+
         return settings
 
 
--- a/aries_cloudagent/connections/base_manager.py
+++ b/aries_cloudagent/connections/base_manager.py
@@ -19,6 +19,7 @@ from pydid.verification_method import (
     Ed25519VerificationKey2018,
     Ed25519VerificationKey2020,
     JsonWebKey2020,
+    Multikey,
 )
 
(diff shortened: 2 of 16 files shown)


dbluhm:
@ianco I made a change to the integration tests to allow me to pass an arbitrary flag down through the layers to the start up of the agent. Thoughts on this approach? It seemed like it would be simpler to do this than to add flags all the way down the stack.

ianco:
> @ianco I made a change to the integration tests to allow me to pass an arbitrary flag down through the layers to the start up of the agent. Thoughts on this approach? It seemed like it would be simpler to do this than to add flags all the way down the stack.

Yes I think this is a good idea.  Will this work when running the alice/faber demo as well?

dbluhm:
I haven't gotten myself familiar enough with the demo yet to say for sure :sweat_smile: So that probably means it's not usable from the demo at this point.

ianco:
> I haven't gotten myself familiar enough with the demo yet to say for sure 😅 So that probably means it's not usable from the demo at this point.

No problem ping me when the PR is ready for review and I'll take a look ...

Jsyro:
Merged in https://github.com/hyperledger/aries-cloudagent-python/pull/2687.

PR #2580, opened 2023-10-31
Hardcode version due to poetry modifying version metadata containing `-`

### LINKED ISSUE #2579: ACA-Py Version has changed format from the last RC
I believe we've changed the format of the acapy version in the latest rc. I'm building agents from ACA-Py main every night in Aries Agent Test Harness and the version from `./bin/aca-py --version` is combing back as `0.11.0rc1` instead of the historical and expected `0.11.0-rc1`. There is some processing of the version in the AATH acapy backchannel that is used for version comparison that is failing because of this.
This issue started in the BC Wallet Test Pipeline on the evening of Oct 30. 

It is also likely that the `admin_url/status` api returns the illformed version format.

Gavinok:
This resolves #2579

The issue originates at commit 492091ec250e1ec5e66c792f67aebfa2ef177756 on September 6th

```diff
--- a/aries_cloudagent/version.py
+++ b/aries_cloudagent/version.py
@@ -1,4 +1,6 @@
 """Library version information."""
 
-__version__ = "0.10.1"
+from importlib import metadata
+
+__version__ = metadata.version("aries-cloudagent")
 RECORD_TYPE_ACAPY_VERSION = "acapy_version"
 --- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,12 @@
 [tool.poetry]
 name = "aries_cloudagent"
-version = "0.10.0-rc0"
-description = ""
+version = "0.10.1"
+description = "Hyperledger Aries Cloud Agent Python (ACA-Py) is a foundation for building decentralized identity applications and services running in non-mobile environments. "
 authors = ["Hyperledger Aries <aries@lists.hyperledger.org>"]
 license = "Apache-2.0"
 readme = "README.md"
 packages = [{include = "aries_cloudagent"}]
 classifiers = [
             "Programming Language :: Python :: 3",
             "License :: OSI Approved :: Apache Software License",
 	    "Operating System :: OS Independent",
```

Which syncs the version with the metadata version

This worked as expected until the latest rc where a `-` was introduced in commit 82385d8c18d9a2f9bac6aa00bfcd8f36d73e548c and 08d5a4be938527f8eb9d310e56cf0fead2d0bc63 which both occurred on october 30th 

```diff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,12 @@
 [tool.poetry]
 name = "aries_cloudagent"
-version = "0.11.0-rc0"
+version = "0.11.0-rc1"
 description = "Hyperledger Aries Cloud Agent Python (ACA-Py) is a foundation for building decentralized identity applications and services running in non-mobile environments. "
 authors = ["Hyperledger Aries <aries@lists.hyperledger.org>"]
 license = "Apache-2.0"
 readme = "README.md"
 packages = [{include = "aries_cloudagent"}]
 classifiers = [
             "Programming Language :: Python :: 3",
             "License :: OSI Approved :: Apache Software License",
 	    "Operating System :: OS Independent",
```

While this will not automatically track with the version in `pyproject.toml` it is the best known option for restoring the `-` in release candidate versions.

### DIFF
--- a/aries_cloudagent/version.py
+++ b/aries_cloudagent/version.py
@@ -2,5 +2,5 @@
 
 from importlib import metadata
 
-__version__ = metadata.version("aries-cloudagent")
+__version__ = "0.11.0-rc1"
 RECORD_TYPE_ACAPY_VERSION = "acapy_version"
--- a/aries_cloudagent/version.py
+++ b/aries_cloudagent/version.py
@@ -1,6 +1,4 @@
 """Library version information."""
 
-from importlib import metadata
-
 __version__ = "0.11.0-rc1"
 RECORD_TYPE_ACAPY_VERSION = "acapy_version"


swcurran:
Why does the `metadata.version` method strip off the “-“.  I’d rather not have yet another place to update the version for each release, so I like the use of the function to get it from the TOML file, but why does it strip off the “-“.  All of our RC releases have used “-“s, so I’d rather not drop that if we can avoid it.

Shouldn’t metadata.version retrieve and pass the version without change?

Gavinok:
@swcurran I think this comes down to how prerelease separators are normalized https://peps.python.org/pep-0440/#pre-release-separators

esune:
> @swcurran I think this comes down to how prerelease separators are normalized https://peps.python.org/pep-0440/#pre-release-separators

While I like the "look" of a `-` separated pre-release, I wonder if we should just follow the official normalization rules and fix this downstream instead rather than having custom handling.

Gavinok:
Would we be able to pull the non-normalized release version number? Looking into this now.

dbluhm:
While it may be uncomfortable, I'm in favor of following PEP 440 rules for release candidate versioning. I think that's most likely to cause the least long-term grief.

swcurran:
Sounds good to me.  So we drop the “-“  and make it “0.11.0rc1” in the release?  I just like the idea of being deliberate vs. adding yet another place where we’re tracking the version.

@WadeBarnes — do you see a problem with us doing that?

swcurran:
@WadeBarnes — or to ask more specifically — are we going to have a problem with the artifact publishing if we do that?  Anything we need to change?

Gavinok:
> Sounds good to me. So we drop the “-“ and make it “0.11.0rc1” in the release? I just like the idea of being deliberate vs. adding yet another place where we’re tracking the version.

Couldn't agree more. I personally was not in love with this solution. It was just the quickest non breaking solution.

WadeBarnes:
It does not look like dropping the `-` should affect anything.

PR #2600, opened 2023-11-08
Add ConnectionProblemReport handler

### LINKED ISSUE #2530: Unexpected state of (orphaned) connection when request is rejected
During the investigation of https://github.com/hyperledger/aries-cloudagent-python/issues/2528, connection requests from the issuer where rejected by the endorser or caused errors on the endorser side.  This was either not communicated back to the issuer or the issuer was not doing anything with the information - I was unable to confirm whether the expected problem report was generated and received by the issuer or not generated at all.

As a result the issuer (in this case) had a pending connection that was essentially orphaned. I would expect the request failure would be communicated so the connection record could be updated/deleted to indicate it was rejected.

### LINKED ISSUE #2599: ConnectionManagerError / ConnectionProblemReport logic clash
Related to #2530 - how to handle orphaned/abandoned connections.

ConnectionManagerError are raised and generally initialized with a descriptive text (ex. "Connection DID does not match DIDDoc id"). Further into the flow, the handlers catch ConnectionManagerErrors then raise ConnectionProblemReport. However, ConnectionProblemReports require very specific enums to set their `problem_code` property. If the ConnectionManagerError.error_code (the descriptive string) is used as the ConnectionProblemReport.problem_code the schema validation will fail.

So, long story short, when we think we are raising ConnectionProblemReports and sending them to be handled, they actually are not created/handled because they fail schema validation and are not created. We need to go through these and ensure the errors are passed properly and get handled as expected.

Another part of this is ConnectionProblemReports use AgentMessage as the base class while DIDExchangeProblemReport uses the ProblemReport as a base class. We should be consistent and update ConnectionProblemReport to follow DIDExchangeProblemReport "standard".

usingtechnology:
fixes #2530 
fixes #2599

Add new problem report handler for connection protocol.

Update the ConnectionManagerError and ConnectionProblemReport to match the same base classes and structure as found in the DIDX protocol. There were issues where the ConnectionManagerErrors were actually failing during marshalling causing problem reports to not get sent. This is addressed by updating the classes and error_codes.

There is still a significant logic gap that I will raise in another issue - ProblemReport delivery requires an existing connection which is means we can't deliver a problem report to indicate the issue is the connection protocol...  but leaving that discussion out of this PR.

### DIFF
--- a/aries_cloudagent/connections/models/conn_record.py
+++ b/aries_cloudagent/connections/models/conn_record.py
@@ -542,7 +542,7 @@ class ConnRecord(BaseRecord):
 
     async def abandon(self, session: ProfileSession, *, reason: Optional[str] = None):
         """Set state to abandoned."""
-        reason = reason or "Connectin abandoned"
+        reason = reason or "Connection abandoned"
         self.state = ConnRecord.State.ABANDONED.rfc160
         self.error_msg = reason
         await self.save(session, reason=reason)
--- a/aries_cloudagent/protocols/connections/v1_0/handlers/connection_invitation_handler.py
+++ b/aries_cloudagent/protocols/connections/v1_0/handlers/connection_invitation_handler.py
@@ -5,7 +5,6 @@ from .....messaging.base_handler import (
     BaseResponder,
     RequestContext,
 )
-
 from ..messages.connection_invitation import ConnectionInvitation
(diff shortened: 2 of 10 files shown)


usingtechnology:
makes sense to tackle #2599 in this PR. will be adding commits to address that issue.

dbluhm (approved):
LGTM!

PR #2611, opened 2023-11-17
feat: support resolving did:peer:1 received in did exchange

dbluhm:
This is an experimental implementation of supporting did:peer:1 resolution. This enables ACA-Py to correctly handle DID Exchange requests from AFJ (AFJ now fails to process our responses since we're not passing back a did:peer, still the unqualified values).

Notably, this does not support creating or sending did:peer:1, just storing and resolving.

Detailed changes:
- Storing a DID Document (using base connection manager) no longer first requires parsing as the (soon to be deprecated) DIDDoc class (not PyDID DIDDocument).
  - This alone removes a whole slue of issues that the [`LegacyCorrections`](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/resolver/default/legacy_peer.py#L27) class was written to deal with.
- Storing the key -> did -> connection mapping is now always delegated to `BaseConnectionManager.record_did`; this helps with consistently handling DIDs
- Fetching a DID Document with `BaseConnectionManager.fetch_did_document` no longer parses the response from the wallet using `DIDDoc`. This helps this call to be a primitive used by DID resolvers that rely on stored documents to resolve DIDs without the DIDDoc parsing logic getting in the way.
- Update the connection record to not require "Indy" shaped DIDs

Note to self, this resolver should have a clean up routine run when the associated connection record is deleted, similar to did:peer:2/3

### DIFF
--- a/aries_cloudagent/connections/base_manager.py
+++ b/aries_cloudagent/connections/base_manager.py
@@ -3,6 +3,7 @@
 For Connection, DIDExchange and OutOfBand Manager.
 """
 
+import json
 import logging
 from typing import List, Optional, Sequence, Text, Tuple, Union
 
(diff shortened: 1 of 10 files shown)


swcurran:
Adding @Jsyro to reviewer.  Can’t give a Maintainer approval, but I can translate… :-)

Jsyro:
Looks great, one question though. 

This maintains existing storage and fetching behaviour of the legacy unqualified DIDDocs, those documents are not going to be converted to did:peer:1?

swcurran:
> This maintains existing storage and fetching behaviour of the legacy unqualified DIDDocs, those documents are not going to be converted to did:peer:1?

No — existing documents need not be converted, since they are only ever used internally, never with external parties. At some point we might update them, although almost certainly not to did:peer:1.

PR #2630, opened 2023-11-25
Feature Suggestion: Include a Reason When Constraints Cannot Be Applied

Ennovate-com:
When the verifier agent is unable to apply a constraint filter, it would be helpful to have a reason why it failed to do so. For example, instead of:

```
Constraint specified for input_descriptor_123 does not apply to the enclosed credential in $.verifiableCredential[0]
```

... the message could include the reason that the constraint could not be applied, with the specific path or item that did not match. For example:

```
Constraint specified for input_descriptor_123 does not apply to the enclosed credential in $.verifiableCredential[0]
Reason: Credential is not applicable for field 11111111-2222-3333-4444-56789abcdef0 with paths ['$.credentialSubject.part1.part2']
```

This indicates that the field indicated in the message was not found in the Reveal Document or the full Credential.

or

```
Constraint specified for input_descriptor_123 does not apply to the enclosed credential in $.verifiableCredential[0]
Reason: No field in constraints for part2 under parent path "$.credentialSubject.part1"
```

This indicates that the "part2" property found in the Reveal Document under "$.credentialSubject.part1" was not found in the list of fields in the constraints.

### DIFF
--- a/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
+++ b/aries_cloudagent/protocols/present_proof/dif/pres_exch_handler.py
@@ -1450,7 +1450,7 @@ class DIFPresExchHandler:
             if is_limit_disclosure:
                 field = await self.get_updated_field(field, cred_dict)
             if not await self.filter_by_field(field, credential):
-                return f'Credential is not applicable for field "{field}"'
+                return f'Credential is not applicable for field {field.id} with paths {field.paths}'
             field_paths = field_paths + (
                 await self.restrict_field_paths_one_of_filter(
                     field_paths=field.paths, cred_dict=cred_dict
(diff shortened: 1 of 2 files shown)


dbluhm:
I like this suggestion; however, the implementation is a little less "Pythonic" than it ought to be. First, the nitpicky comment: variables should be snake_case (`failReason` is what I'm referring to specifically). And second, I think the fail reasons should not be reported as a string from the constraint application method. The semantics are a bit too loose, IMO. I would recommend reporting the fail reason in an exception that is caught by the caller.

dbluhm (approved):
Thanks for the contribution and for the prompt turn around on feedback!

Ennovate-com:
@dbluhm And, likewise, thanks for the quick feedback and merge. Much appreciated.

PR #2632, opened 2023-11-27
Integrate Anoncreds rs into credential and presentation endpoints

ianco:


dbluhm (approved):
LGTM! A lot of copy-pasted code, of course, but as we've discussed, this is a necessary step to keeping the AnonCreds RS changes isolated until ready for use. Especially since we anticipate deprecating and removing the original handlers in favor of the new anoncreds handlers, which will eliminate the duplication.

PR #2653, opened 2023-12-06
Connection and DIDX Problem Reports

### LINKED ISSUE #2598: DIDExchange handlers - error before sending Problem Reports
During the investigation on #2530, I was purposefully setting failure points to see how/if problem reports were sent and if the receiver of a problem report could then act on it.

There is a bug in [DIDXManager/BaseConnectionManager.diddoc_connection_targets](https://github.com/hyperledger/aries-cloudagent-python/blob/e13339650932f7b31f036366fdaa6e179dd760d1/aries_cloudagent/connections/base_manager.py#L585) that prevents the Problem Report from being sent. The crux is that it passes in an `AttachDecorator` not a `DIDDoc` as expected. So the `diddoc_connection_targets` method throws an exception and no Problem Report is delivered.

Think we need a DIDXManager implementation or another method to add to [response_handler.py](https://github.com/hyperledger/aries-cloudagent-python/blob/e13339650932f7b31f036366fdaa6e179dd760d1/aries_cloudagent/protocols/didexchange/v1_0/handlers/response_handler.py#L43)

### LINKED ISSUE #2650: Problem Report not delivered until connection is active
See #2599 and PR #2600.

Problem Reports require an active connection for delivery. However, if the issue is in the early stages of the connection/didexchange protocols, we have no way of successfully communicating that back to the other agent. We do have logic that catches exceptions and fires a ProblemReport in the request and response handlers but those reports will be swallowed and not delivered because the underlying connection is not active.

```
SKIP_ACTIVE_CONN_CHECK_MSG_TYPES = [
    "didexchange/1.0/request",
    "didexchange/1.0/response",
    "connections/1.0/invitation",
    "connections/1.0/request",
    "connections/1.0/response",
]

....

        if (
            message.connection_id
            and msg_type
            and msg_type not in SKIP_ACTIVE_CONN_CHECK_MSG_TYPES
            and not await super().conn_rec_active_state_check(
                profile=context.profile,
                connection_id=message.connection_id,
            )
        ):
            raise RuntimeError(
                f"Connection {message.connection_id} is not ready"
                " which is required for sending outbound"
                f" message {msg_id} of type {msg_type}."
            )
        return await self._send(context.profile, message, self._inbound_message)

```

Should we add problem report the skip types? Which would allow the message to be delivered? or add some other flag/logic that would allow problem reports in the skipped type protocol stages to be delivered? Like a problem report that occurs during the execution of a skipped type?

usingtechnology:
- add current message thread to problem report messages to identify which message/thread encountered the problem.
- problem report handlers will find connections based on thread/request id
- register didx problem report so handler will be active and deal with didx problem reports
- update tests to match new handler logic

fixes #2650 
fixes #2598

A lot of underlying issues uncovered (and now addressed) arising from #2530 investigation. 

Synopsis is the implemented logic for sending/handling connection (and didx) problem reports that require a fully working connection. This prevented us from notifying if the problem arose during the request or early response stages. That is addressed by allowing the problem report message types to skip the active check in the responder. Then the next issue was the agent receiving the problem report could not match up their connection, so the sender adds the failing message's thread to the problem report message before delivery, the receiver can then look up connections by a known thread id. Then the DIDX problem report code was always failing when trying to determine the messages targets, so that code was updated to handle diddoc attach conversion to diddoc and then resolve the targets. Then uncovered the fact that the DIDX problem reports (and handler) were not registered in the protocol message registry so those wouldn't get handled when we finally parsed the targets.

Sigh, lots of stuff that just never worked... 🤷 ...hoping that these fixes will be of use when there are issues in the early stages of connecting agents.

### DIFF
--- a/aries_cloudagent/messaging/responder.py
+++ b/aries_cloudagent/messaging/responder.py
@@ -5,26 +5,26 @@ in response to the message being handled.
 """
 import asyncio
 import json
-
 from abc import ABC, abstractmethod
-from typing import List, Sequence, Union, Optional, Tuple
+from typing import List, Optional, Sequence, Tuple, Union
 
 from ..cache.base import BaseCache
-from ..connections.models.connection_target import ConnectionTarget
 from ..connections.models.conn_record import ConnRecord
+from ..connections.models.connection_target import ConnectionTarget
 from ..core.error import BaseError
 from ..core.profile import Profile
 from ..transport.outbound.message import OutboundMessage
-
-from .base_message import BaseMessage
 from ..transport.outbound.status import OutboundSendStatus
+from .base_message import BaseMessage
 
 SKIP_ACTIVE_CONN_CHECK_MSG_TYPES = [
     "didexchange/1.0/request",
     "didexchange/1.0/response",
+    "didexchange/1.0/problem_report",
     "connections/1.0/invitation",
     "connections/1.0/request",
     "connections/1.0/response",
+    "connections/1.0/problem_report",
 ]
 
 
--- a/aries_cloudagent/protocols/connections/v1_0/handlers/connection_invitation_handler.py
+++ b/aries_cloudagent/protocols/connections/v1_0/handlers/connection_invitation_handler.py
@@ -31,5 +31,6 @@ class ConnectionInvitationHandler(BaseHandler):
                 ),
             }
         )
+        report.assign_thread_from(context.message)
         # client likely needs to be using direct responses to receive the problem report
         await responder.send_reply(report)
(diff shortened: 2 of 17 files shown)


dbluhm (approved):
This looks good! Nice work!

dbluhm:
Those DIDExchange Manager issues might have been me :sweat_smile:

usingtechnology:
> Those DIDExchange Manager issues might have been me 😅

Haha, no big deal, just makes me wonder about other little corners that haven't had issues arise yet. 😬

Issue #2669, opened 2023-12-12
[BUG] VC LDP credential issuance requires the credential.type array to have more than 1 item


PatStLouis:
The VC data model 1.1 spec states that the credential.type array can **optionally** have a second type:

`VerifiableCredential and, optionally, a more specific verifiable credential type.`
https://www.w3.org/TR/vc-data-model/#types

I would recommend to remove the validation on the `length == 1` for the credential.type as indicated in the specification.

Happy to provide a PR if this is approved.

swcurran:
PR welcome :-)!

PR #2670, opened 2023-12-13
Remove if condition which checks if the `credential.type` array is equal to 1

PatStLouis:
This PR addresses  #2669

There is currently a validation that prevents credential with only the base type from being signed. The data model spec states that additional credential types are optional. This also prevents conforming to some test-suites where [single credential type](https://github.com/w3c/vc-data-model-2.0-test-suite/blob/main/tests/input/credential-ok.json) credentials are used as test data.

This is a minor change with no impact.


PatStLouis:
@dbluhm I fixed the failing test, anything else I would need to do here?

swcurran (approved):

swcurran:
@PatStLouis -- the PR needs to be updated to the base branch before we can merge. Can you please make that change?

PatStLouis:
@swcurran done, just need to wait for the tests to complete

PR #2672, opened 2023-12-13
Anoncreds - Add unit testing

jamshale:
Adds unit tests for routes, holder, verifier, issuer.

Screenshot from 2023-12-13 10-50-52

There is still more unit testing to do. Mainly the `revocation.py` file. Also, there was a tricky part with scanning the wallet in holder and using `async for` I haven't figured out yet. Also there is some error handling I haven't covered. 

I tried to create objects with the `anoncreds` library as much as possible to catch problems in the future when using the library. When I couldn't manage to make that work I would mock the object a minimally as possible. I also had some trouble creating some DB/wallet `Entry` responses. To get around this i created some minimal class objects. These are used in holder and issuer files.

I made 2 small changes to src code. In verifier.py changed an error handling message and in holder.py added the mime-types attribute constant that seemed to be missing.


ianco (approved):
Looks good, very thorough!

PR #2676, opened 2023-12-15
Cache TAA by wallet name

jamshale:
The TAA acceptance was being cached at pool level. This caused problems in traction if the tenant changed pods and also TAA acceptance between different tenants in multi-tenant mode.

My understanding is limited in this area. 
I thought maybe I could cache at a wallet level, but I couldn't find a way to do this currently or any other examples in the code base. So, I inject the cache at profile level and then have a different cache key per wallet/tenant.

This will use more cache memory per tenant. Could possibly store the cache for less time.

### DIFF
--- a/aries_cloudagent/ledger/indy_vdr.py
+++ b/aries_cloudagent/ledger/indy_vdr.py
@@ -1,20 +1,19 @@
 """Indy-VDR ledger implementation."""
 
 import asyncio
-import json
 import hashlib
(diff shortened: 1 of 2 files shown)


esune (approved):
Looks good to me, thanks @jamshale!

swcurran (approved):
LGTM — and more importantly, @esune .

Issue #2678, opened 2023-12-17
[BUG] Improper Use of TimeZone-Naive datetime.utcnow()


Ennovate-com:
When `datetime.timestamp()` is called on a timezone-naive instance of `datetime`, it assumes the value is a local datetime. Therefore, the code `datetime.utcnow().timestamp()` should never be used, since the non-local timezone-naive `datetime.utcnow()` violates that assumption.

The "old" correct way to get the UTC timestamp is:
`datetime.now().timestamp()`

A better way of doing it is to avoid the naive datetime, using the timezone-aware version, formerly using:
`datetime.utcnow().replace(tzinfo=timezone.utc).timestamp()`
or now, because `datetime.utcnow()` has been deprecated since Python version 3.12, using:
`datetime.now(tz=timezone.utc).timestamp()`

A fairly easy way to see the effect is with the following code in an environment where the timezone offset is non-zero:

```
from datetime import datetime, timezone
now1 = datetime.now()
now2 = datetime.utcnow()
now3 = datetime.now(tz=timezone.utc)
now1
now1.timestamp()
now2
now2.timestamp() # WRONG!
now3
now3.timestamp()
```

This produces output like this:

```
>>> from datetime import datetime, timezone
>>> now1 = datetime.now()
>>> now2 = datetime.utcnow()
>>> now3 = datetime.now(tz=timezone.utc)
>>> now1
datetime.datetime(2023, 12, 16, 11, 50, 0, 979519)
>>> now1.timestamp()
1702749000.979519
>>> now2
datetime.datetime(2023, 12, 16, 17, 50, 0, 982995)
>>> now2.timestamp() # WRONG!
1702770600.982995
>>> now3
datetime.datetime(2023, 12, 16, 17, 50, 0, 983405, tzinfo=datetime.timezone.utc)
>>> now3.timestamp()
1702749000.983405
```

[...]

PR #2679, opened 2023-12-17
Fix: Change To Use Timezone Aware UTC datetime

Ennovate-com:
Fix issue #2678

Based on 0.11.0rc2 code because when I attempt to use the main branch as the base I am getting an error:

```
  File "/app/aries_cloudagent/anoncreds/__init__.py", line 3, in <module>
    from ..config.injection_context import InjectionContext
ImportError: attempted relative import beyond top-level package
```


swcurran:
@Ennovate-com — I’ve kicked off the failed integration tests again in case it might have been an intermittent failure. If not, can you please take a look at the failing test?

dbluhm (approved):

dbluhm:
If tests pass, I'll merge in main and enable auto-merging so we don't have to wait for the int tests again

Issue #2667, opened 2023-12-12
[Feature] Add `did:web` issuer support for ldp issuance/verification endpoints


PatStLouis:
The newly added endpoints to issue/verify json-ld vc's make use of the issuer field to find the signing key, limiting the issuance to `did:key`. In the event I have registered a `did:key` in my wallet and chose to host the verification method through a `did:web` identifier, I should be able to use this as my issuer id.

I would like to see a way for the agent to resolver a `did:web` document and fetch the verification keys from said document to decide which key to sign the payload with. Since acapy already supports a resolver endpoint this can enable such a feature.

The same would be made for the verification, I would like to refer to a `did:web` verification key which can be resolved to verify the payload.

The endpoints in question are `/vc/ldp/issue` and `/vc/ldp/verify`

Issue #2677, opened 2023-12-16
[BUG] `did:web` resolution incorrect service type validation


PatStLouis:
When resolving a `did:web` with the `/resolver/resolve/` endpoint the validation checks if the `service.type` is a string. According to the [did core spec](https://www.w3.org/TR/did-core/#services), this value can be a string or an array. 

> The value of the type property MUST be a [string](https://infra.spec.whatwg.org/#string) or a [set](https://infra.spec.whatwg.org/#ordered-set) of [strings](https://infra.spec.whatwg.org/#string). In order to maximize interoperability, the [service](https://www.w3.org/TR/did-core/#dfn-service) type and its associated properties SHOULD be registered in the DID Specification Registries [[DID-SPEC-REGISTRIES](https://www.w3.org/TR/did-core/#bib-did-spec-registries)].

Having an array for the `service.type` raises this error:
`
500: Response was incorrectly formatted. Failed to deserialize DIDDocumentRoot:. 1 validation error for ParsingModel[DIDDocumentRoot]. __root__ -> service -> 0 -> type. str type expected (type=type_error.str). [ErrorWrapper(exc=ValidationError(model='DIDDocumentRoot', errors=[{'loc': ('service', 0, 'type'), 'msg': 'str type expected', 'type': 'type_error.str'}]), loc=('__root__',))].
`

Changing the `service.type` to a string instead of an array resolves this error however both should be valid.

dbluhm:
Shoot. This is a PyDID issue. This has the potential to be a painful change though; several routines scan through services expecting to find a string.

Out of curiosity, do you have an example of this in use?

PatStLouis:
The traceability interoperability spec uses a list for the `service.type`. Aca-py isnt't able to process the document any further if the value is not a string, therefore can't fetch a `verificationKey` to verify a VC:
https://w3c-ccg.github.io/traceability-interop/draft/#service

PatStLouis:
Tagging this link to keep breadcrumbs of the work:
[https://github.com/Indicio-tech/pydid/issues/81](https://github.com/Indicio-tech/pydid/issues/81)

PatStLouis:
@dbluhm we fixed this in the pydid library and a new version was released. Is there anything we need to do on aca-py side to pull this new version or is it auto-magical?

dbluhm:
[...]

PR #2684, opened 2023-12-18
Add did web method type as a default option

PatStLouis:
Addresses issue #2667.

Inspired by the [did methods documentation](https://github.com/hyperledger/aries-cloudagent-python/blob/main/DIDMethods.md).


swcurran:
@PatStLouis — please check the Lint test failure.

@dbluhm — what do you think of this?  OK as a start?

PatStLouis:
I tested issuance/verification in both single tenant/multitenant modes.

I had a strange error with verification at first but I'm unable to reproduce it....verification works now using a `did:web` issuer...I'll keep an eye out, here's the error the verification endpoint was returning
`
('Could not expand input before framing.',)\nType: jsonld.FrameError\nCause: ('Tried to nullify a context with protected terms outside of a term definition.',)\nType: jsonld.SyntaxError\nCode: invalid context nullification  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 1032, in frame\n    expanded = self.expand(input_, options)\n  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 870, in expand\n    expanded = self._expand(active_ctx, None, document, options,\n  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 2329, in _expand\n    self._expand_object(\n  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 2728, in _expand_object\n    expanded_value = self._expand(\n  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 2233, in _expand\n    e = self._expand(\n  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 2324, in _expand\n    active_ctx = self._process_context(\n  File \"/home/aries/.local/lib/python3.9/site-packages/pyld/jsonld.py\", line 3076, in _process_context\n    raise JsonLdError(\n"
`

PatStLouis:
I think I found the issue, it seems to be related with the [traceability context url](https://w3id.org/traceability/v1). Trying to verify a vc issued by a `did:web` issuer who has this context included in it's did document seems to raise this strange error. Here's a vc to test (can be verified with other verification services such as the [univerifier](https://univerifier.io/)). Verifying the same vc without the traceability context returns true in aca-py.
```
{
        "@context": [
            "https://www.w3.org/2018/credentials/v1"
        ],
        "id": "urn:uuid:31288134-1c6a-42a2-aa37-c5c38dea09c0",
        "type": [
            "VerifiableCredential"
        ],
        "issuer": "did:web:PatStLouis.github.io",
        "issuanceDate": "2023-12-19T01:13:13.401Z",
        "credentialSubject": {
            "id": "did:example:123"
        },
        "proof": {
            "type": "Ed25519Signature2018",
            "proofPurpose": "assertionMethod",
            "verificationMethod": "did:web:PatStLouis.github.io#verkey",
            "created": "2023-12-19T01:13:13.401Z",
            "jws": "eyJhbGciOiAiRWREU0EiLCAiYjY0IjogZmFsc2UsICJjcml0IjogWyJiNjQiXX0..ahIwxPaTgLmUS4aS9nFApVQmx-uklAAR0I-W_-IKzhvMQFmCl0g7etQw3DG8oKqVs-0hlewuCtUxKZGO1GyyBw"
        }
    }
```

swcurran:
That is quite an `@context` file!!

ACA-Py is now pre-loading contexts — I wonder if that is the issue.  This was recently added in PR #2587 that is part of 0.11.0.  It might be worth trying 0.10.5 to see if the same issue occurs.  Given the pre-loading of context, I’m not sure how contexts that are not pre-loaded are handled.  I assume they are loaded dynamically, but that is not considered “best practice” in JSON-LD land, so perhaps not.  Which of course leads to the question, how does one add ones own contexts to an instance of ACA-Py?

PatStLouis:
I tried [adding and caching the context in the document loader](https://github.com/OpSecId/aries-cloudagent-python/blob/19e02bfaf4f6c5d2b2790edb783d0f009de546c7/aries_cloudagent/vc/ld_proofs/document_downloader.py#L45) as well and get the same result. Also putting the agent in debug mode doesn't log any errors so this seem unhandled from the `pyld` package. I'm going to try and find out where in the code this verification fails. Now the did core spec **recommends** having a context file for your services, its not required so this won't prevent any test cases AFAIK. The [issue raised](#2677) about the `service.type` being an array does get tested. I'll have a look at the `PyDID` library like @dbluhm suggested to understand the impact.

```
// Service array is used to look up traceability API service endpoint
pm.test("Response must include 'service' array", function() {
    const { service } = pm.response.json().didDocument;
    pm.expect(service).to.be.an('array').that.is.not.empty;
});
```

swcurran:
@PatStLouis -- can you please check on the conflicts?

PatStLouis:
fixed the conflict between the did:peer and did:web wallet methods

swcurran (approved):

PR #2687, opened 2023-12-19
Emit did:peer:2 for didexchange

Jsyro:
Rather than manage repo permissions. I'm reopening @dbluhm 's [PR](https://github.com/hyperledger/aries-cloudagent-python/pull/2578) here so I can verify and add/restore tests. 

- [x] Restore Unittests
- [x] Restore integration tests
- [x] Add new unit tests (check initiation with flag, check responding in kind without flag)
- [x] Add new integration tests (check connection with both flags set and one flag set)


dbluhm (approved):
Thanks!

PR #2688, opened 2023-12-21
Add unit tests for anoncreds revocation

jamshale:
Adds unit testing for the anoncreds revocation module.

Tried to use the anoncred module as much as possible but when I was having too much trouble I mocked calls that loaded/created objects or converted forms. 

Mocked the database/wallet Entry responses a lot with basic classes. Couldn't find a better way to do this or an examples in the repo.

Coverage is 92%. I tried not to make the tests too brittle but wanted to have good coverage for my own understanding. I'm hoping this will help others trying to understand the code along with having a good testing layout for future refactoring or features.


ianco (approved):
Very nice!!!

PR #2690, opened 2023-12-29
Improve api documentation and error handling

jamshale:
Add examples and descriptions to the swagger docs. Mostly used the indy examples.

There was some confusion with request params object attributes with camel vs snake casing. The anoncreds library does camel case like the spec decribes. Everywhere else has snake case. I changed everything so that only the inner anoncreds schema, cred_def and rev_reg_def objects are in camel case. I think this makes things a bit more clear but it is still a bit confusing when you are used to the indy api. 

I changed the POST rev_reg_def endpoint to have a wrapper like schema and cred_def with an options object where you can put the endorser connection id.

Changed the GET all creds defs return object to be in a wrapper like schemas.

Fixed a couple mistakes and improved the error handling. I think a majority of errors give information about the cause now. Can still get 500 errors with unexpected param values.


aries_cloudagent/anoncreds/base.py
```diff
@@ -103,7 +103,10 @@ def supported_identifiers_regex(self) -> Pattern:
 
     async def supports(self, identifier: str) -> bool:
         """Determine whether this registry supports the given identifier."""
-        return bool(self.supported_identifiers_regex.match(identifier))
+        try:
+            return bool(self.supported_identifiers_regex.match(identifier))
+        except TypeError:
```
dbluhm, line 108: Mostly curious, under what conditions were we receiving a TypeError?

aries_cloudagent/anoncreds/default/legacy_indy/registry.py
```diff
@@ -636,15 +636,23 @@ async def register_revocation_list(
             profile, rev_list, rev_reg_def.type, rev_reg_entry
         )
 
+        seq_no = None
+        try:
+            seq_no = rev_entry_res["result"]["txnMetadata"]["seqNo"]
```
dbluhm, line 641: Similar question, under what circumstances do we not see the seqNo in the return?

dbluhm (approved):

aries_cloudagent/anoncreds/default/legacy_indy/registry.py
jamshale: This one can happen when you try and publish a rev reg def that was already posted to the ledger. It would get a 500 KeyError because txnMetadata was missing. If the result is returned without the sequence number then it gets a 400 error with a duplicate entry response now.

aries_cloudagent/anoncreds/base.py
jamshale: Looking at this again I'm not sure how I got this one to happen. I'll look it again and can probably remove it.

jamshale:
I need to fix the revocation api integration tests to use the new payload type.

aries_cloudagent/anoncreds/models/anoncreds_revocation.py
```diff
@@ -116,13 +131,17 @@ class Meta:
     issuer_id = fields.Str(
         description="Issuer Identifier of the credential definition or schema",
         data_key="issuerId",
+        example=INDY_OR_KEY_DID_EXAMPLE,
```
dbluhm, line 134: I'm a little hesitant about the only examples showing as old Indy values. Is it possible to provide more than one example so we could do something like `did:example:123`?

Also, I believe most of these fields should be in the `metadata` kwarg now after updates to marshmallow, as a side note. That change probably belongs in a separate PR.

dbluhm (approved):
The snake_case vs camelCase changes have me a little nervous since I vaguely recall discussions about the names but I've been away from this long enough that I trust your judgement over my own on the details here. One small comment. Otherwise, LGTM!

jamshale:
> The snake_case vs camelCase changes have me a little nervous since I vaguely recall discussions about the names but I've been away from this long enough that I trust your judgement over my own on the details here. One small comment. Otherwise, LGTM!

I think the casing caused some confusion when this was getting implemented. The camel casing still applies for the anoncred objects so I'm pretty sure it's good. I basically just removed any discrepancies with search params and outer objects being mixed cases anymore.

ianco (approved):
Looks good!!!  Regarding @dbluhm 's comments (snake vs camel case, number of examples, etc) if there is anything we need to follow up later we should add as new issues.

aries_cloudagent/anoncreds/models/anoncreds_revocation.py
jamshale: I had thought about that. I was thinking when we add more methods we could figure something out for the other examples. For the metadata stuff I saw both methods used and wasn't sure the correct way.

PR #2692, opened 2024-01-02
did peer 4 resolution

Jsyro:
Using https://github.com/decentralized-identity/did-peer-4.

based on the behaviour expected by the specification. 

https://identity.foundation/peer-did-method-spec/#method-4-short-form-and-long-form


swcurran:
Looks good to me, but I defer to @dbluhm on this one.  We’ll want an “emit” capability — that will be a follow up PR?

Jsyro:
> Looks good to me, but I defer to @dbluhm on this one. We’ll want an “emit” capability — that will be a follow up PR?

Yes, I planned to add 'emit' functionality in future work when prioritized, but this a change that adds functionality and can be tested !

esune (commented):
Code changes look good to me, will leave the final review and approval/comments to @dbluhm

dbluhm (approved):

PR #2694, opened 2024-01-03
fix: save multi_use to the DB for OOB invitations

frostyfrog:
It was reported by a collegue that when using Out-of-band invitations with the connections handshake protocol, along with a public DID and connection re-use enabled, ACA-Py was only able to handle a single invitations. As if Connection Re-use wasn't working or the multi_use flag was ignored.

Turns out, we flat out weren't saving the multi_use flag to the database which is a big problem if you want to re-use the invitation. When DID-Exchanged is used, since DID-Exchange allows for a connection-less exchange of information. We only say the behaviour of the OOB Invitation getting deleted if we ran through the receive_reuse_message() method in the manager, which deletes the record if it's not set for multi_use.

### DIFF
--- a/aries_cloudagent/protocols/out_of_band/v1_0/models/oob_record.py
+++ b/aries_cloudagent/protocols/out_of_band/v1_0/models/oob_record.py
@@ -303,6 +303,6 @@ class OobRecordSchema(BaseExchangeSchema):
         required=False,
         metadata={
             "description": "Allow for multiple uses of the oobinvitation",
-            "example": True
+            "example": True,
         },
     )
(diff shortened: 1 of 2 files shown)


dbluhm:
The changes look good but it would be nice to see some tests for the record demonstrating that the value is definitely stored and recalled. Would also be valuable to include an integration test.

dbluhm (approved):
LGTM! Given that @frostyfrog and I are at the same org, I'll leave this to @swcurran or @ianco for final approval and merge.

ianco (approved):

PR #2696, opened 2024-01-04
Feature/emit did peer 4

Jsyro:
Open for comments on how to update the long form with the short form. The resolver handles its own mapping, however i'm storing a duplicate DIDInfo, one where the did is long form and another where it's short form (for the same verkey).

### DIFF
--- a/aries_cloudagent/config/argparse.py
+++ b/aries_cloudagent/config/argparse.py
@@ -2,7 +2,6 @@
 
 import abc
 import json
-
 from functools import reduce
 from itertools import chain
 from os import environ
@@ -10,15 +9,12 @@ from typing import Type
 
 import deepmerge
 import yaml
-
 from configargparse import ArgumentParser, Namespace, YAMLConfigFileParser
 
 from ..utils.tracing import trace_event
-
 from .error import ArgsParseError
-from .util import BoundedInt, ByteSize
-
 from .plugin_settings import PLUGIN_CONFIG_KEY
+from .util import BoundedInt, ByteSize
 
 CAT_PROVISION = "general"
 CAT_START = "start"
@@ -1177,6 +1173,13 @@ class ProtocolGroup(ArgumentGroup):
             help=("Emit did:peer:2 DIDs in DID Exchange Protocol"),
         )
 
+        parser.add_argument(
+            "--emit-did-peer-4",
+            action="store_true",
+            env_var="ACAPY_EMIT_DID_PEER_4",
+            help=("Emit did:peer:4 DIDs in DID Exchange Protocol"),
+        )
+
     def get_settings(self, args: Namespace) -> dict:
         """Get protocol settings."""
         settings = {}
@@ -1245,6 +1248,8 @@ class ProtocolGroup(ArgumentGroup):
 
         if args.emit_did_peer_2:
             settings["emit_did_peer_2"] = True
+        if args.emit_did_peer_4:
+            settings["emit_did_peer_4"] = True
 
         return settings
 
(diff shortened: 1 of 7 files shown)


demo/features/0160-connection.feature
```diff
@@ -16,6 +16,8 @@ Feature: RFC 0160 Aries agent connection functions
          | Acme_capabilities           | Acme_extra        | Bob_capabilities | Bob_extra        |
          | --public-did --did-exchange | --emit-did-peer-2 | --did-exchange   |--emit-did-peer-2 |
          | --public-did --did-exchange | --emit-did-peer-2 | --did-exchange   |                  |
+         | --public-did --did-exchange | --emit-did-peer-4 | --did-exchange   |--emit-did-peer-4 |
+         | --public-did --did-exchange | --emit-did-peer-4 | --did-exchange   |                  |
```
dbluhm, line 20: I think it'd be interesting to also test 2 -> 4 and 4 -> 2

aries_cloudagent/protocols/didexchange/v1_0/manager.py
```diff
@@ -822,6 +843,13 @@ async def accept_response(
             await self.record_did(response.did)
 
         conn_rec.their_did = their_did
+
+        # The long format I sent has been acknoledged, use short form now.
+        if LONG_PATTERN.match(conn_rec.my_did or ""):
+            conn_rec.my_did = await self.long_did_peer_4_to_short(conn_rec.my_did)
```
dbluhm, line 849: I think as a community we need to formalize this a bit more but I think this is a logical starting point.

dbluhm (approved):
A minor suggestion and a minor comment but I think this looks good!

swcurran:
Regards the issue of updating the long form to the short.

I would think that from the start on received DIDs, we want to store the short form on the connection ID, and when we need to resolve a long form, we convert it before looking it up.  Since it is impossible to look up a short form via its long form, I think we should always do that.

I guess on creation its trickier, since we need to send it to the other party as the long form. Do we ever need to use the long form again after we send it to the other party?  Especially if we assume the other party stores it as a short form (per comment above).

Also, do we store the DIDDoc?  If so, is there a DIDDoc to Long Form transformation that would provide a way to get the long form back if ever needed?

Jsyro:
Ready for final review. 

@swcurran we currently need the protocol manager class to manage the ins and outs of each protocols behaviour, but that's still going to have maintenance issues if there are interleaved blocks of code handling unqualified, did:peer:2/3 and did:peer:4. 

There is no code access at the point in time you are referring to, outbound queuing system uses the same values that we are manipulating. so doing any changes after we have gotten a response after sending the long form seems like the best and safest. Sending out the long form and then trying to change the agent's connection record may result in more convoluted behaviour than it's worth. 

This is definitely something to keep an eye on. 

We are not storing the did documents at all in storage. The short form did looks up the long form did, resolves that long form did into the document and then makes the appropriate substitutions to make it behave like the specification.

swcurran:
Got it — thanks.  Sounds good.

aries_cloudagent/config/argparse.py
```diff
@@ -1245,6 +1248,8 @@ def get_settings(self, args: Namespace) -> dict:
 
         if args.emit_did_peer_2:
             settings["emit_did_peer_2"] = True
+        if args.emit_did_peer_4:
+            settings["emit_did_peer_4"] = True
```
esune, line 1252: Wondering when one of these will become the new default, rather than having to specify the behaviour at startup (I am a bit out of the loop on this so I am likely missing some background info).

dbluhm (approved):
:+1:

PR #2704, opened 2024-01-10
fix: minor type hint corrections for VcLdpManager

dbluhm:
These are some really minor fixes just to make type checking happy and to clarify the use of some values in the VcLdpManager.


aries_cloudagent/vc/vc_ld/manager.py
```diff
@@ -311,16 +317,17 @@ async def _get_suite_for_credential(
...
 
-    async def _get_all_suites(self):
+    async def _get_all_proof_suites(self) -> List[LinkedDataProof]:
         """Get all supported suites for verifying presentation."""
-        suites = []
-        for suite, key_type in SIGNATURE_SUITE_KEY_TYPE_MAPPING.items():
-            suites.append(
-                suite(
+        return [
+            cast(
```
PatStLouis, line 323: This section could use a short comment to understand what is happening, there's a lot going on in the return statement.

I could suggest:
`# We iterate through the PROOF_KEY_TYPE_MAPPING dictionary to return an array of it's items as LinkedDataProof objects`

dbluhm:
> Looks good, I see a lot of attributes are marked as Optional, if they were required beforehand, what is the purpose of making them optional?

Good question; the type hints for the changed parameters all had a default value of `None`. This was intended to indicate that the value was a `str | None` or whatever other type. This was an old convention that ACA-Py followed but is no longer accepted by tools like pyright/pylance. The added `Optional` annotation aligns the type hint with the use of the field, updating it to match what modern type checking tools expect.

dbluhm:
> Otherwise I just want to point out that technically the term LinkedDataProof was renamed to DataIntegrityProof. Given the structure of the project this would be a pretty big change to implement however it's important to note that when we are referring to linked data proofs here we are in fact talking about data integrity proofs.

I think this is a good one for us to file an issue on but probably not something to be addressed by this PR directly, IMO.

PatStLouis:
I definitely don't think this PR should address this as this would entail a massive restructuring and might become a significant endeavor. Deciding if it's a worthwhile thing to address could be discussed in a separate issue. There's a distinction here between functional coding and proper term definition. AFAIK linked data proofs and data integrity proofs refer to the same thing/concept.

aries_cloudagent/vc/ld_proofs/suites/linked_data_signature.py
```diff
@@ -21,8 +21,8 @@ class LinkedDataSignature(LinkedDataProof, metaclass=ABCMeta):
     def __init__(
         self,
         *,
-        proof: dict = None,
-        verification_method: str = None,
+        proof: Optional[dict] = None,
+        verification_method: Optional[str] = None,
         date: Union[datetime, None] = None,
```
andrewwhitehead, line 26:
```suggestion
        date: Optional[datetime] = None,
```

swcurran (approved):

PR #2705, opened 2024-01-10
feat: inject profile

dbluhm:
This PR adds the Profile to the injection context. This enables creating a `ClassProvider` for a class that takes a profile in its init method:

```python
class ExamplePluggable:
    def __init__(self, profile: Profile):
        self.profile = profile
    def do_something(self):
        self.profile.notify("some:event", {"with": "data"})

async def setup(context: InjectionContext):
    context.injector.bind_provider(
        ExamplePluggable,
        ClassProvider(ExamplePluggable, ClassProvider.Inject(Profile))
    )

# ... Somewhere else ...
plugged_in = profile.inject(ExamplePluggable)
plugged_in.do_something()
```

It's a common pattern to create a class with a profile instance. This will make it easier to turn some of these components into pluggable components. For example, the `VcLdpManager`. At the moment, it's created directly where it's needed:

```python
manager = VcLdpManager(self.profile)
# do something with manager
```

With this change, I can create an (override-able) class provider for VcLdpManager, which changes the above into:

```python
manager = self.profile.inject(VcLdpManager)
# do something with manager
```

The only way this could be done previously was by binding providers in the Profile init method itself. This made it more difficult to make more components pluggable.

Side note: it's critical that the profile instance be injected as a weak reference or it will never be "finalized" (garbage collected) in multitenant contexts.

### DIFF
--- a/aries_cloudagent/config/base.py
+++ b/aries_cloudagent/config/base.py
@@ -158,5 +158,5 @@ class BaseInjector(ABC):
 class BaseProvider(ABC):
     """Base provider class."""
 
-    def provide(self, settings: BaseSettings, injector: BaseInjector):
+    def provide(self, settings: BaseSettings, injector: BaseInjector) -> Any:
         """Provide the object instance given a config and injector."""
--- a/aries_cloudagent/config/provider.py
+++ b/aries_cloudagent/config/provider.py
@@ -2,7 +2,7 @@
 
 import hashlib
 
-from typing import Sequence, Union
+from typing import Optional, Sequence, Union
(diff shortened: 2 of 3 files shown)


dbluhm:
Feedback on any unforeseen consequences of this are welcome; given that this pattern was kind of present in those profile init bind provider calls, I think it should be fine.

dbluhm:
Sorry, mixed up the assignee and reviewer fields... :sweat_smile:

jamshale (approved):
I can't comment on unforeseen consequences, but it looks good to me.

PR #2706, opened 2024-01-11
feat: make VcLdpManager pluggable

dbluhm:
As a follow up to #2705, this PR implements the example I gave of making `VcLdpManager` pluggable. This may be desirable to, for instance, change the issuance or verification behavior (e.g. using an external service to perform the signature or similar).

This is a bit "heavier" a change than what is proposed in #2348 in the sense that it would require slightly more effort to define the alternate DID lookup mechanism but I think this would still address the same problem it was seeking to solve. Given the straightforwardness of the change required in ACA-Py to support this, I am more in favor of it being handled this way.


ianco (approved):
LGTM

Issue #2693, opened 2024-01-03
[Question] Verifying presentations with the `/vc/ldp/verify` endpoint


PatStLouis:
The endpoint accepts a "vp" in the body, so I'm assuming it can verify presentations. However, when I provide a vp it returns the following errors:
```
{
  "vp": {
    "type": [
      "type must include VerifiableCredential"
    ],
    "issuer": [
      "Missing data for required field."
    ],
    "issuanceDate": [
      "Missing data for required field."
    ],
    "credentialSubject": [
      "Missing data for required field."
    ]
  }
}
```

It looks like its still expecting a vc, is this the intended behavior?

swcurran:
Can you add the VP you provided?  I’m guessing it did have those fields?

swcurran:
Note that a VP embeds an array of VCs in it, so it is not surprising that is expects a VC.  Might not be the issue, but hard to tell without seeing the data.

PatStLouis:
Here's a presentation. It can be verified with the uni resolver but aca-py returns the previous errors
```
{
    "@context": [
        "https://www.w3.org/2018/credentials/v1"
    ],
    "type": [
        "VerifiablePresentation"
    ],
    "holder": "did:web:trace.opsec.id:organization:test-20",
    "verifiableCredential": [
        {
            "@context": [
                "https://www.w3.org/2018/credentials/v1"
            ],
            "type": [
                "VerifiableCredential"
            ],
            "issuer": {
                "id": "did:web:trace.opsec.id:organization:test-20"
            },
            "issuanceDate": "2022-05-01T00:00:00Z",
            "credentialSubject": {
                "id": "did:key:123"
            },
            "id": "urn:uuid:cc173c16-aabb-11ee-a29f-0242ac1b0006",
            "proof": {
                "type": "Ed25519Signature2018",
                "verificationMethod": "did:web:trace.opsec.id:organization:test-20#verkey",
                "proofPurpose": "assertionMethod",
                "created": "2024-01-04T04:43:26Z",
                "jws": "eyJhbGciOiAiRWREU0EiLCAiYjY0IjogZmFsc2UsICJjcml0IjogWyJiNjQiXX0..XNFLfGs2kjPp78CyoVZ0_N8rwrWnB60AAtkZ85pvCKpfi0TUMArEkONROCYL638E0VJu9jSY3PVBHOJS2jg9CA"
            }
        }
    ],
    "proof": {
[...]

PR #2710, opened 2024-01-15
Enable presentation issuance/verification through vc-api endpoints

PatStLouis:
This PR enables presentation issuance and verification using the vc_ld manager and addresses #2693.

There is also 4 new routes included matching the vc-api specification. (minor changes from the `/vc/ldp` endpoints, mostly key names and defaulting to `Ed25519Signature2018` `proofType` for issuance).
@dbluhm could you help me by reviewing the changes I've made to the vc_ld manager and models? You will see I added an if condition in the `verify_presentation` section to enable `assertionMethod` purpose if no challenge is provided, while minimizing the impact on the existing code/workflows. I also improved the presentation models, based on the existing credential ones.

I have a test instance of these changes running [here](https://agent.vc.opsec.id/api/doc#/vc-api).


aries_cloudagent/vc/vc_ld/manager.py
```diff
@@ -345,18 +355,51 @@ async def issue(
...
+        )
+        return VerifiablePresentation.deserialize(vp)
+
     async def verify_presentation(
         self, vp: VerifiablePresentation, options: LDProofVCOptions
     ) -> PresentationVerificationResult:
         """Verify a VP with a Linked Data Proof."""
-        if not options.challenge:
+
+        if vp.proof.proof_purpose == "assertionMethod":
```
PatStLouis, line 385: @dbluhm here is the section I mentioned in the PR comment

PatStLouis:
The last commit added a method to store an issued VC

swcurran:
I'm trying to re-run the failed test, as it doesn't look like it is related to this PR.

You had flagged this is a [WIP], @PatStLouis , so I marked it as a "Draft" PR. Have you completed it at this point?

Thanks!

swcurran:
Ah...also there are conflicts with this PR to be resolved. Could you also look at that @PatStLouis ?  Thanks!

PatStLouis:
@swcurran yeah I noticed the tests were failing because it wasn't able to write a schema on the ledger which seemed odd, looks like it passed now. Let me fix the conflicts then I would consider this ready.

swcurran:
@PatStLouis — I assume this is now ready for review?  I’ve removed the “Draft” flag on it.  You still have “WIP” in the title.

dbluhm:
Thoughts on having the VC API routes be provided by plugin rather than be a built in, @PatStLouis @swcurran? That's what we did for the OpenID4VCI and these routes seem like an analogous chunk of functionality.

I think the plugin approach could be more ergonomic; it can be revised independent of ACA-Py's release cycle which could help it stay up to date with any spec changes or additions made in the VC API world. I think it could also help keep ACA-Py itself more maintainable in the long run so someone who isn't as well versed in the VC API details doesn't have to worry about making updates to that chunk of code.

Still reading into the details but I don't think I would suggest moving everything in this PR to a plugin, mostly just the strictly VC API portions.

But then again, maybe I inadvertently moved us in the direction of having them be included directly in ACA-Py when I added the `/vc` endpoints :sweat_smile: :man_shrugging:

PatStLouis:
Is there any plans for the `/vc` routes making it a bad idea to replace them with this set of vc-api endpoints? Sounds like they are looking to achieve the same goal and are already 95% similar, which is why I was happy to see that PR.

If namespacing is a concern, we could do:
`/vc-api/credentials/issue`
`/vc-api/credentials/verify`
`/vc-api/presentations/prove`
`/vc-api/presentations/verify`

The original ld-proof-vc-detail attachment format which inspired the request format of `credential/options` request body already references the vc-api (formaly known as the VC HTTP API) as seen in [RFC0593: JSON-LD Credential Attachment format for requesting and issuing credentials](https://github.com/hyperledger/aries-rfcs/tree/main/features/0593-json-ld-cred-attach#reference), below the options section.

These routes are not meant to create exchanges which is a significant difference with the oidc4vci.

After some reflections I believe these should be part of aca-py if we can have a consensus on replacing the `/vc` routes, otherwise it would be somewhat redundant. I don't see a big enough difference with the purpose/format of the vc-api routes and the current vc routes to warrant a plugin instead of a replacement/improvement.

I would also argue that they would actually be easier to maintain since there is a specification available.

The benefit of being compatible with the w3c test-suites is also interesting to get green checkmarks and claim interoperability out-of-the-box.

dbluhm:
I'm fine replacing the `/vc` routes since they've not made it into an official release yet. The only minor concern I have is with causing potential confusion with the existing `/credentials` endpoints which list credentials received by ACA-Py as a holder. So I think I'm in favor of namespacing the endpoints as you suggest.

aries_cloudagent/vc/vc_ld/manager.py
```diff
@@ -273,9 +274,17 @@ async def prepare_credential(
...
+        document: Union[VerifiableCredential, VerifiablePresentation],
+        options: LDProofVCOptions,
     ) -> LinkedDataProof:
-        issuer_id = credential.issuer_id
+        document_type = document.type[0]
+
+        if document_type == "VerifiableCredential":
+            issuer_id = document.issuer_id
+        if document_type == "VerifiablePresentation":
+            issuer_id = document.holder_id
```
dbluhm, line 286: Given that we have Python types defined for these objects, I would recommend checking the type with:

```python
if isinstance(document, VerifiableCredential):
```

aries_cloudagent/vc/vc_ld/models/credential.py
```diff
@@ -341,6 +349,13 @@ class Meta:
         metadata={"example": CREDENTIAL_SUBJECT_EXAMPLE},
     )
 
+    credential_status = DictOrDictListField(
+        required=False,
+        data_key="credentialStatus",
+        # validate=CREDENTIAL_STATUS_VALIDATE,
+        # metadata={"example": CREDENTIAL_STATUS_EXAMPLE},
```
dbluhm, line 356: General recommendation to not include commented out code. If these are intended to represent a future `TODO` item, I'd replace the commented lines with a `# TODO add validation and example`.

aries_cloudagent/vc/vc_ld/store.py
```diff
@@ -0,0 +1,42 @@
...
+from ...admin.request_context import AdminRequestContext
+from ...storage.vc_holder.base import VCHolder
+from ...storage.vc_holder.vc_record import VCRecord
+from ...vc.vc_ld import VerifiableCredential
+from ...vc.ld_proofs import DocumentLoader
+from pyld import jsonld
+from pyld.jsonld import JsonLdProcessor
+
+
+async def store_credential(
```
dbluhm, line 11: Should this belong as a method of the VcLdpManager?

aries_cloudagent/vc_api/examples.py
```diff
@@ -0,0 +1,189 @@
...
+    #         "example": [{
+    #             "verified": True,
+    #             "proof": {},
+    #             "purpose_result": {
+    #                 "valid": True,
+    #                 "controller": {}
+    #             }
+    #         }]
+    #     }
+    # )
```
dbluhm, line 84: As mentioned in a previous comment, strong recommendation not to include commented out code.

dbluhm (commented):
Overall, I think this looks good! We will need some unit tests to cover the additions to the VcLdpManager especially. I'm less concerned about tests for the routes themselves but some form of automated test for those would still be nice.

ianco:
> Overall, I think this looks good! We will need some unit tests to cover the additions to the VcLdpManager especially. I'm less concerned about tests for the routes themselves but some form of automated test for those would still be nice.

To echo @dbluhm 's comment, we need to make sure we have at least some basic unit tests for the new code.  Also we should have some integration tests (IMHO more important than the unit tests) and a document describing the functionality supported by the `/vc` endpoints.

ianco (changes requested):
See comments re tests and documentation

aries_cloudagent/vc_api/routes.py
```diff
@@ -0,0 +1,227 @@
...
+from .examples import (
+    IssueCredentialRequest,
+    IssueCredentialResponse,
+    VerifyCredentialRequest,
+    VerifyCredentialResponse,
+    ProvePresentationRequest,
+    ProvePresentationResponse,
+    VerifyPresentationRequest,
+    VerifyPresentationResponse,
+)
```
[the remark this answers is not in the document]
PatStLouis: I'm happy with using the current models in the vc module, these quick examples were more to have a usable swagger interface to demonstrate this easily. I would remove the examples file and use the existing vc models to avoid replicating code.

aries_cloudagent/vc/vc_ld/models/credential.py
PatStLouis: +1

aries_cloudagent/vc/vc_ld/store.py
PatStLouis: I think so, looking at the other actions such as verify/issue/etc.., they have their own files and a method calling it from the VcLdpManager, I would lean towards keeping this file and adding a method calling this function to keep the same model?

aries_cloudagent/vc_api/examples.py
PatStLouis: +1

PatStLouis:
@ianco I can add a readme file in the project for documentation. For the integration tests I will have to get more familiar with those.

@dbluhm could explain to me why a [challenge is required to verify a presentation](https://github.com/hyperledger/aries-cloudagent-python/blob/aa704a7775d7c374174ba99632829ac73d12791e/aries_cloudagent/vc/vc_ld/manager.py#L366C19-L366C19)? I ended up calling the verify function directly instead of the manager and provide an assertion purpose to bypass this.

ianco:
> @ianco I can add a readme file in the project for documentation. For the integration tests I will have to get more familiar with those.

It's very similar to the AATH tests, if you want an overview let me know and I can give you a quick walkthrough.

aries_cloudagent/vc_api/routes.py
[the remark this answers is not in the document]
dbluhm: Oh, I may have misinterpreted what these represented, woops. Sounds good.

aries_cloudagent/vc/vc_ld/manager.py
dbluhm: `_get_suite_for_document` seems fair to me!

aries_cloudagent/vc/vc_ld/store.py
dbluhm: Grabbed a few links for context/reference:
- Issue: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/vc_ld/issue.py
- Verify: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/vc_ld/verify.py
- Prove: https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/vc_ld/prove.py

I would categorize these as primitives that are stateless and encapsulate the inner workings of the LD Proofs package.

Storage, on the other hand is stateful, as it records the credential in the wallet. The manager already operates in this sphere (e.g. https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/vc_ld/manager.py#L92-L116) so I think I put my vote down for having storage of a credential be managed by the `VcLdpManager`.

aries_cloudagent/vc/vc_ld/store.py
```diff
@@ -0,0 +1,42 @@
...
+    credential: VerifiableCredential,
+    document_loader: DocumentLoader,
+    context: AdminRequestContext,
```
dbluhm, line 14: Another thing I caught just now, this should take a `Profile` instead of a context of any variety. If this method is moved to the manager, the profile will be provided in the class (`self.profile`). I think this is another good measure of stateful/stateless-ness; the issue, verify, prove modules do not take a context or profile anywhere.

aries_cloudagent/vc/vc_ld/store.py
PatStLouis: after looking at the code some more, this is also what I landed on

aries_cloudagent/vc/routes.py
```diff
@@ -49,6 +92,16 @@ async def ldp_issue(request: web.BaseRequest):
     context: AdminRequestContext = request["context"]
     body = await request.json()
     credential = VerifiableCredential.deserialize(body["credential"])
+    options = {} if "options" not in body else body["options"]
+    # Default to Ed25519Signature2018 if no proof type was provided
+    if "type" in options:
```
PatStLouis, line 97: While providing a proof type in the options field is not part of the vc-api spec, there is a convention of only using the key `type` instead of `proofType` when doing so. This bit of code will check both and prioritize `proofType` if its present as this is the current acapy field key which is used. It will also default to `Ed25519Signature2018` if nothing is provided. It will the pass whichever value comes out of it as `proofType` to the manager.

dbluhm: Might be good to leave this in a comment in the code but otherwise seems reasonable to me

PatStLouis:
I was having issues with the provided models from the `vc_ld` folder, so I created new schemas for the endpoint requests/responses which will be use to return specific validations and error codes. This is to leave the `vc` module classes and schemas intact. I added a `service` file in the vc-api folder which is responsible for interacting with the `VcLdpManager` and provide the expected `vc_ld` models. This service file contains a function for each route to which I pass the `web.request` object and process accordingly. The routes file will be used to process schema validations and error handling.

@amanji let's have a chat about how to handle requests and validations, you will find my models in the `/vc-api/models` directory

dbluhm (dismissed):

swcurran:
@PatStLouis — looks like this is near ready. Can you pleae update with the base branch so we can merge it?

PatStLouis:
@swcurran @dbluhm @ianco I'm happy with the state of this PR. I will make a subsequent PR with a postman collection in the demo folder, which will include documentation about these endpoints as well as simple test cases.

swcurran:
Sounds good.  The documentation is missing and that is a concern, so that needs to be added.

ianco:
> Sounds good. The documentation is missing and that is a concern, so that needs to be added.

Agree we need the documentation and we can deal with the rest in a subsequent PR.

PatStLouis:
Sounds good, I will add this tonight, I will put a small abstract on the main readme with a link to a page in the demo folder with more information!

PatStLouis:
@dbluhm I added a manger test for the storage function

swcurran:
@ianco — are your change requests ok now?

ianco (approved):

dbluhm:
@PatStLouis if you want to do one final merge of main, I've enabled auto merge so we can get this merged in next.

PR #2715, opened 2024-01-16
Anoncreds schema endorsement

jamshale:
Adds the flow for schema endorsement.

The endorsement flow is triggered by the agent having an author role or by manually setting the `create_transaction_for_endorser` option request param.

The anoncreds registry submits the transaction based on wallet type, and the post processing (storing the schema in the wallet) is triggered by the transaction manager using a new anoncreds schema event.

The endorsement integration tests are run with anoncreds wallets on the endorser and author roles. Added unit tests mostly on the registry register schema flow.

### DIFF
--- a/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
+++ b/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
@@ -3,11 +3,13 @@
 import json
 import logging
 import re
+import uuid
 from asyncio import shield
 from typing import List, Optional, Pattern, Sequence, Tuple
 
 from base58 import alphabet
 
+from ....anoncreds.default.legacy_indy.author import get_endorser_info
 from ....cache.base import BaseCache
 from ....config.injection_context import InjectionContext
 from ....core.profile import Profile
@@ -22,9 +24,20 @@ from ....ledger.multiple_ledger.ledger_requests_executor import (
     GET_CRED_DEF,
     IndyLedgerRequestsExecutor,
 )
+from ....messaging.responder import BaseResponder
 from ....multitenant.base import BaseMultitenantManager
-from ....revocation_anoncreds.models.issuer_cred_rev_record import IssuerCredRevRecord
+from ....protocols.endorse_transaction.v1_0.manager import (
+    TransactionManager,
+    TransactionManagerError,
+)
+from ....protocols.endorse_transaction.v1_0.util import is_author_role
+from ....revocation_anoncreds.models.issuer_cred_rev_record import (
+    IssuerCredRevRecord,
+)
 from ....revocation_anoncreds.recover import generate_ledger_rrrecovery_txn
+from ....storage.error import StorageError
+from ....utils import sentinel
+from ....wallet.did_info import DIDInfo
 from ...base import (
     AnonCredsObjectAlreadyExists,
     AnonCredsObjectNotFound,
@@ -194,11 +207,7 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
         # Assume endorser role on the network, no option for 3rd-party endorser
         ledger = profile.inject_or(BaseLedger)
         if not ledger:
-            reason = "No ledger available"
-            if not profile.settings.get_value("wallet.type"):
-                # TODO is this warning necessary?
-                reason += ": missing wallet-type?"
-            raise AnonCredsRegistrationError(reason)
+            raise AnonCredsRegistrationError("No ledger available")
 
         # Translate schema into format expected by Indy
         LOGGER.debug("Registering schema: %s", schema_id)
@@ -212,32 +221,90 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
         }
         LOGGER.debug("schema value: %s", indy_schema)
 
+        endorser_did = None
+        create_transaction = options.get("create_transaction_for_endorser", False)
+
+        if is_author_role(profile) or create_transaction:
+            endorser_did, endorser_connection_id = await get_endorser_info(
+                profile, options
+            )
+
+        write_ledger = (
+            True if endorser_did is None and not create_transaction else False
+        )
+
+        # Get either the transaction or the seq_no or the created schema
         async with ledger:
             try:
-                seq_no = await shield(
-                    ledger.send_schema_anoncreds(schema_id, indy_schema)
+                result = await shield(
+                    ledger.send_schema_anoncreds(
+                        schema_id,
+                        indy_schema,
+                        write_ledger=write_ledger,
+                        endorser_did=endorser_did,
+                    )
                 )
             except LedgerObjectAlreadyExistsError as err:
-                indy_schema = err.obj
-                schema = AnonCredsSchema(
-                    name=indy_schema["name"],
-                    version=indy_schema["version"],
-                    attr_names=indy_schema["attrNames"],
-                    issuer_id=indy_schema["id"].split(":")[0],
-                )
                 raise AnonCredsSchemaAlreadyExists(err.message, err.obj_id, schema)
             except (AnonCredsIssuerError, LedgerError) as err:
                 raise AnonCredsRegistrationError("Failed to register schema") from err
 
+        # Didn't need endorsement, so return schema result
+        if write_ledger:
+            return SchemaResult(
+                job_id=None,
+                schema_state=SchemaState(
+                    state=SchemaState.STATE_FINISHED,
+                    schema_id=schema_id,
+                    schema=schema,
+                ),
+                registration_metadata={},
+                schema_metadata={"seqNo": result},
+            )
+
+        # Need endorsement, so execute transaction flow
+        (schema_id, schema_def) = result
+
+        job_id = uuid.uuid4().hex
+        meta_data = {"context": {"job_id": job_id, "schema_id": schema_id}}
+
+        transaction_manager = TransactionManager(profile)
+        try:
+            transaction = await transaction_manager.create_record(
+                messages_attach=schema_def["signed_txn"],
+                connection_id=endorser_connection_id,
+                meta_data=meta_data,
+            )
+        except StorageError:
+            raise AnonCredsRegistrationError("Failed to store transaction record")
+
+        if profile.settings.get("endorser.auto_request"):
+            try:
+                (
+                    transaction,
+                    transaction_request,
+                ) = await transaction_manager.create_request(transaction=transaction)
+            except (StorageError, TransactionManagerError) as err:
+                raise AnonCredsRegistrationError(
+                    "Transaction manager failed to create request: " + err.roll_up
+                ) from err
+
+            responder = profile.inject(BaseResponder)
+            await responder.send(
+                message=transaction_request,
+                connection_id=endorser_connection_id,
+            )
+
         return SchemaResult(
-            job_id=None,
+            job_id=job_id,
             schema_state=SchemaState(
-                state=SchemaState.STATE_FINISHED,
+                state=SchemaState.STATE_TRANSACTION_REQUESTED,
                 schema_id=schema_id,
                 schema=schema,
             ),
-            registration_metadata={},
-            schema_metadata={"seqNo": seq_no},
+            registration_metadata={
+                "txn": transaction.serialize(),
+            },
         )
 
     async def get_credential_definition(
@@ -771,3 +838,31 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
                     applied_txn = ledger_response["result"]
 
         return (rev_reg_delta, recovery_txn, applied_txn)
+
+    async def txn_submit(
+        self,
+        profile: Profile,
+        ledger_transaction: str,
+        sign: bool = None,
+        taa_accept: bool = None,
+        sign_did: DIDInfo = sentinel,
+        write_ledger: bool = True,
+    ) -> str:
+        """Submit a transaction to the ledger."""
+        ledger = profile.inject(BaseLedger)
+
+        if not ledger:
+            raise LedgerError("No ledger available")
+
+        try:
+            return await shield(
+                ledger.txn_submit(
+                    ledger_transaction,
+                    sign=sign,
+                    taa_accept=taa_accept,
+                    sign_did=sign_did,
+                    write_ledger=write_ledger,
+                )
+            )
+        except LedgerError as err:
+            raise AnonCredsRegistrationError(err.roll_up) from err
--- a/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
+++ b/aries_cloudagent/anoncreds/default/legacy_indy/registry.py
@@ -298,7 +298,7 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
         return SchemaResult(
             job_id=job_id,
             schema_state=SchemaState(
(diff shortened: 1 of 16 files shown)


ianco:
Overall looks good.

I think we need to move where the ledger write actually happens, tho.  Right now the transaction manager writes to the ledger (https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py#L428) and then emits the event that we listen for (https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py#L526).

For anoncreds, I think the ledger write needs to move into our new (anoncreds) event handler method.

jamshale:
@ianco Ok, thanks. I will look into doing that, and add some unit tests and ping you when I'm finished.

ianco (dismissed):
looks good!

aries_cloudagent/anoncreds/default/did_web/registry.py
```diff
@@ -112,3 +112,7 @@ async def update_revocation_list(
     ) -> RevListResult:
         """Update a revocation list on the registry."""
         raise NotImplementedError()
+
+    async def txn_submit(self, profile: Profile, ledger_transaction: str) -> str:
+        """Submit a transaction to the ledger."""
+        raise NotImplementedError()
```
dbluhm, line 118: This addition is a red flag; `did:web` doesn't have a ledger or transactions.

dbluhm (changes requested):
Super quick comment, will come back with a more detailed review.

aries_cloudagent/anoncreds/default/did_web/registry.py
jamshale: Ok. Thanks for doing a review. This came from wanting the anoncreds registry wanting to do the transaction submission from Ian's comment. I was thinking it could just implement it to do nothing if that was the case for did:web. 

The most confusing thing for me still is what should happen in existing acapy and what should happen from the anoncreds registry.

dbluhm: The presence (or absence) of a ledger in use by anoncreds should be hidden behind the anoncreds interface and be a detail of a specific registry implementation. I'm between meetings so I haven't had a chance to look closer yet at the goal of the method to make alternate recommendations but I should be able to look closer later!

ianco: The alternative is to move the ledger write from within the transaction manager (here: https://github.com/hyperledger/aries-cloudagent-python/pull/2715/files#diff-7e8dcd5a6024ec4d34242c09d77451ed6c5309435710e4278eb82ebbfce65c2eR424) to the event listener in the anoncreds issuer (here: https://github.com/hyperledger/aries-cloudagent-python/pull/2715/files#diff-fad60a24ebdabddf55dab0b471caef11c0ab82243fef6e0f1df2f9f250e6acb6R680)

So ... within the transaction manager, it just needs to know enough to *not* try to write a ledger transaction for an `askar-anoncreds` wallet, and then the anoncreds issuer listener would need to do that.

ianco: Giving this some more thought, I think it's fair for us to define a general ledger API which includes functionality (like writing a ledger transaction) that isn't supported by all ledgers.  And then the implementation would just throw a "not implemented exception" as @jamshale has implemented here.

If we enforce an API including only those functions implemented by *all* ledgers then we're potentially limiting the functionality available to aca-py.  I.e. what if we get a new type of ledger later on that supported even *less* functionality?  Do we have to eliminate the API functions that this new ledger doesn't support?

dbluhm: I took a closer look at this. The `txn_submit` is called from the endorser protocol and that's the only spot (that I identified quickly at least). My read on this is that this just shouldn't be an operation you access through the AnonCreds Interface. The limited interface isn't about holding back ACA-Py from additional features, it's about keeping the role of the AnonCreds registry clear and straightforward. If it's obvious that all AnonCreds implementations will require endorsement, then I think we can justify making changes to the interface to better support that. But for the moment, it's an Indy specific concern. I don't see a problem with calling up an instance of the Indy registry directly (instead of through the generic interface) and call additional functions on it that would otherwise be unavailable on the generic interface.

dbluhm: Having the Endorser protocol call methods on the Indy registry directly I think is a more accurate representation of the dependency anyways since it is currently built explicitly for use by Indy networks.

jamshale: I think I understand. I was thinking you needed to access the indy_legacy registry through the interface. I'll try and implement it the way you descibed and then we can discuss again.

jamshale:
The latest commit removes the extending of the registry interface, and instead implements the txn_submit method from within the legacy_indy registry. Author triggered txn_submits in the indy ledger and endorser manager complete_transaction create an instance of the LegacyIndyRegistry to submit the transactions. 

This will still end up submitting the transaction from the indy ledger but will make sure that ledger is injected from the anoncreds registry.

One drawback is that because the transaction manager and ledger are imported from the ledger routes module on initialization, importing the LegacyIndyRegistry from the top level caused a circular import exception. To get around this the Registry is only imported when needed during anoncred transaction submit operations.

dbluhm (changes requested):
Apologies for taking so long to look deeper on this PR; I feel bad about making some comments early, requesting changes, and then requesting some more after you made those changes... But I think we're not quite aligned on vision for the endorser support yet.

Happy to discuss these points more.

dbluhm:
I fear I'm coming across as harsh but I think there were some mistakes made in the first iteration of endorser support in ACA-Py that really need to be corrected. As it stands, I think we're inheriting a bit too much of those mistakes in this iteration.

ianco (dismissed):
Not completely happy with how the endorsed transaction is getting saved ... I don't like the (circular) dependency between the transaction manager and Indy registry, and would prefer the ledger transaction gets saved in the event handler somehow (limits the specific ledger dependencies) but let's accept this PR and then re-visit in the next endorser PR.

aries_cloudagent/anoncreds/routes.py
```diff
@@ -193,9 +211,17 @@ async def schemas_post(request: web.BaseRequest):
     issuer = AnonCredsIssuer(context.profile)
     try:
         result = await issuer.create_and_register_schema(
-            issuer_id, name, version, attr_names, options=options
+            issuer_id,
+            name,
+            version,
+            attr_names,
+            options,
+            request["outbound_message_router"],
```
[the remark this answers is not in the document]
jamshale: Ok. Yes, I was wondering if there was another way to get the responder. I had looked but hadn't found it. I will clean that up.

aries_cloudagent/anoncreds/default/legacy_indy/registry.py
```diff
@@ -211,33 +219,82 @@ async def register_schema(
...
+                    "Transaction manager failed to create request: " + err.roll_up
+                ) from err
+
+            await outbound_handler(
+                transaction_request, connection_id=endorser_connection_id
+            )
+        return {
+            "sent": {"schema_id": schema_id, "schema": schema_def},
+            "txn": transaction.serialize(),
+        }
```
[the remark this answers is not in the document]
jamshale: Ok. This was an area of confusion for me. I had been somewhat copying what was originally done with endorsement which has a much different response object. I'll look at this again and how I can convert it to a SchemaResult object.

aries_cloudagent/anoncreds/issuer.py
```diff
@@ -201,17 +202,22 @@ async def create_and_register_schema(
...
                 options,
+                outbound_handler,
             )
 
-            await self._store_schema(schema_result)
+            # Endorsement transaction - don't store in wallet
+            if not isinstance(schema_result, SchemaResult):
+                return schema_result
+
+            await self.store_schema(schema_result)
```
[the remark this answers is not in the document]
jamshale: Ok, I had started doing this with the cred def stuff and have a better idea what you were doing here.

jamshale:
> I fear I'm coming across as harsh but I think there were some mistakes made in the first iteration of endorser support in ACA-Py that really need to be corrected. As it stands, I think we're inheriting a bit too much of those mistakes in this iteration.

I think that's fair. A lot of this is new to me so I appreciate the feedback. Some of the things you mentioned I had been confused about. I don't think it should take to long.

jamshale:
I think I'm getting a grip on the old implementation and how we want it to change with anoncreds. This should be way better now.

One thing I noticed is that the `GET /schema/{id}` endpoint doesn't return the state, so you can't see it's finished though the api even though the record is updated. Not sure if that should be change as another ticket. 

I added a state for transaction_requested and I used a random uuid for the job-id. Wasn't sure if this was exactly what we wanted to do.

ianco:
> One thing I noticed is that the `GET /schema/{id}` endpoint doesn't return the state, so you can't see it's finished though the api even though the record is updated. Not sure if that should be change as another ticket.

Isn't there a ledger transaction id that isn't available until the schema is written to the ledger?

jamshale:
> > One thing I noticed is that the `GET /schema/{id}` endpoint doesn't return the state, so you can't see it's finished though the api even though the record is updated. Not sure if that should be change as another ticket.
> 
> Isn't there a ledger transaction id that isn't available until the schema is written to the ledger?

That should still be there, yes. This is just for the anoncreds Result object. It has a state field I was setting to `transaction_requested` and then `finished` when it is written to the ledger. You can't ever see the `finished` state from the current GET api. Only the inner anoncreds object.

jamshale:
POST /schema

```json
{
  "job_id": "db16fd5f8c094d62ba3d1722dacb2467",
  "schema_state": {
    "state": "transaction_requested",
    "schema_id": "WgWxqztrNooG92RXvxSTWv:2:author-schema:1.1",
    "schema": {
      "issuerId": "WgWxqztrNooG92RXvxSTWv",
      "attrNames": [
        "score"
      ],
      "name": "author-schema",
      "version": "1.1"
    }
  },
  "registration_metadata": {
    "txn": {
      "state": "request_sent",
      "created_at": "2024-01-30T00:08:21.960157Z",
      "updated_at": "2024-01-30T00:08:21.968250Z",
      "trace": false,
      "transaction_id": "7e664eb2-2ed0-460e-8411-b6858aeb2fa2",
      "_type": "https://didcomm.org/sign-attachment/1.0/signature-request",
      "signature_request": [
        {
          "context": "did:sov",
          "method": "add-signature",
          "signature_type": "default",
          "signer_goal_code": "aries.transaction.endorse",
          "author_goal_code": "aries.transaction.ledger.write"
        }
      ],
      "signature_response": [],
      "timing": {
        "expires_time": null
      },
      "formats": [
        {
          "attach_id": "4cef3a8f-b910-4794-9326-6f0a638d683c",
          "format": "dif/endorse-transaction/request@v1.0"
        }
      ],
      "messages_attach": [
        {
          "@id": "4cef3a8f-b910-4794-9326-6f0a638d683c",
          "mime-type": "application/json",
          "data": {
            "json": "{\"endorser\": \"6bX4vukUZxxGKm7f5bBkmN\", \"identifier\": \"FB5yHWKaZk59hiKqjJKEHs\", \"operation\": {\"data\": {\"attr_names\": [\"score\"], \"name\": \"author-schema\", \"version\": \"1.1\"}, \"type\": \"101\"}, \"protocolVersion\": 2, \"reqId\": 1706573301954441235, \"signature\": \"5ihP9oXmgg6WRRMmNjHiiiP6W3r3S5xVfJFCru3BUzHSqiV6gkmRhvo3UgDzsTmQHU2Z4zp91hwU3cKHtLsG9xWA\"}"
          }
        }
      ],
      "meta_data": {
        "context": {
          "job_id": "db16fd5f8c094d62ba3d1722dacb2467",
          "schema_id": "WgWxqztrNooG92RXvxSTWv:2:author-schema:1.1"
        }
      },
      "connection_id": "2e64c27e-5c35-4090-87ef-a48e4db3f2ba"
    }
  }
}
```

GET /schems/{id}

```json
  "schema": {
    "issuerId": "FB5yHWKaZk59hiKqjJKEHs",
    "attrNames": [
      "score"
    ],
    "name": "author-schema",
    "version": "1.1"
  },
  "schema_id": "FB5yHWKaZk59hiKqjJKEHs:2:author-schema:1.1",
  "resolution_metadata": {
    "ledger_id": null
  },
  "schema_metadata": {
    "seqNo": 1234
  }
}
```

aries_cloudagent/anoncreds/routes.py
```diff
@@ -126,11 +127,24 @@ class SchemaPostOptionSchema(OpenAPISchema):
...
 
+    create_transaction_for_endorser = fields.Bool(
+        description="""
+            Create transaction for endorser (optional, default false). 
+            Use this for agents who don't specify an author role but want to 
+            create a transaction for an endorser to sign.""",
+        required=False,
+        example=False,
+    )
+
```
dbluhm, line 147: I think this is a good use of options. However, I wonder if the first option might be better handled by having an established default endorser (kind of like how we set our public DID and that persists across calls). And I'm honestly not sure what the second one means; is this perhaps a case where we could simplify the interface by sticking to a reasonable default? In what circumstances would we set this value to True?

aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py
```diff
@@ -797,7 +807,14 @@ async def endorsed_txn_post_processing(
...
             # Notify schema ledger write event
-            await notify_schema_event(self._profile, schema_id, meta_data)
+            if self._profile.settings.get("wallet.type") == "askar-anoncreds":
+                event_bus = self.profile.inject(EventBus)
+                await event_bus.notify(
+                    self._profile,
+                    SchemaRegistrationFinishedEvent.with_payload(meta_data),
+                )
+            else:
+                await notify_schema_event(self._profile, schema_id, meta_data)
```
dbluhm, line 817: What do you think of just calling `AnonCredsIssuer.finish_schema` directly?

dbluhm (dismissed):
A few comments but I think this is still a great first set of changes. I'm okay with my comments being ignored in the interest of unblocking this from moving forward!

aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py
jamshale: I think that's a good idea. I was just following the existing pattern here but that would be simpler.

jamshale:
> A few comments but I think this is still a great first set of changes. I'm okay with my comments being ignored in the interest of unblocking this from moving forward!

I'll update it. I'd rather address them here before the next PR which will be a lot bigger.

aries_cloudagent/anoncreds/default/legacy_indy/registry.py
```diff
@@ -212,32 +221,90 @@ async def register_schema(
...
+                message=transaction_request,
+                connection_id=endorser_connection_id,
+            )
+
         return SchemaResult(
-            job_id=None,
+            job_id=job_id,
             schema_state=SchemaState(
-                state=SchemaState.STATE_FINISHED,
+                state=SchemaState.STATE_TRANSACTION_REQUESTED,
```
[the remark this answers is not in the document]
jamshale: :+1: I had second guessed adding a more specific state here. Wait makes sense.

jamshale:
@dbluhm Addressed the comments now.

dbluhm (approved):

PR #2716, opened 2024-01-17
Fix incorrect Sphinx search library version reference

swcurran:
I did test it locally, but I think my process is not picking things like this up.


swcurran:
Figured out how I should have tested it...and verified it failed previously, fixed now.

WadeBarnes (approved):

PR #2719, opened 2024-01-17
Upgrade anoncreds to 0.2.0.dev7

jamshale:
Used an exact version of the dev pre-release. Figured is would be changed anyway when this becomes non pre-release. 

Fixed some calls to the anoncreds library. Required fetching some additional objects from the wallet. Fixed the unit tests.

These would have been quicker and easier to fix if the unit tests caught them. For some of the unit tests I was able to interact with anoncreds and create objects with it, but for some objects like creating credentials and revocation registries I was mocking the objects. These ended up being the broken calls so only the integration tests caught them.

Would be nice if the unit tests could create all the anoncreds objects instead of mocking them. Will try and do this more when I see the opportunity. 

Changed the devcontainer poetry version to match the version used to upgrade.


swcurran (dismissed):

jamshale:
More things to fix here then I thought there would be. Quite a few changes to anoncreds object create methods. Some of the complex objects I wasn't able to unit test well, like creating credentials and some revocation objects. These are the failing ones.

swcurran:
Sorry this is a bigger than expected task, but I think it is the priority, so this is good for you to work through.

It wasn’t until I saw your PR that I realized we weren’t on the `0.2.0` path.  Getting on that is necessary.

jamshale:
This should be good now. All the integration tests pass. Updated the description comment.

dbluhm (approved):

ianco:
> Would be nice if the unit tests could create all the anoncreds objects instead of mocking them. Will try and do this more when I see the opportunity.

Yes this would be nice.  Prob best to deal with this strategically since the integration tests test the full stack top-to-bottom.

ianco (approved):

PR #2723, opened 2024-01-17
Remove exception on connectionless presentation problem report handler

loneil:
This is to support being able to get callbacks to [vc-authn-oidc](https://github.com/bcgov/vc-authn-oidc) when a user declines the presentation request sent to their agent.

Presently ACA-PY is throwing an exception if the problem report does not contain a connection record.
Remove that exception and allow the recieve_problem_report to continue (and supply None for the connection ID in that case).


jamshale (approved):
lgtm

PR #2737, opened 2024-01-26
bump pydid to v 0.4.3

PatStLouis:
Addresses #2677

@dbluhm I ran a `poetry update pydid` after changing the version in the `pyproject.toml` file. Seems like it did the trick.


PatStLouis:
strange, when running the black code formatter locally it detects no changes...I'm using version 23.7.0

PatStLouis:
@swcurran I updated to the new black formatter version (24.1) used by the tests, it was released yesterday and seems to behave differently. This 2 line PR turned into a 302 lines PR...

dbluhm:
That's unfortunate -- @PatStLouis no worries on the formatting changes. I think your separation into two different commits is a good compromise. Thanks!

It's probably a good idea for us to pin the version of black used in the actions so we can keep it in sync with what we have down in the dev dependencies. But we don't need to worry about that in this PR

dbluhm (approved):

dbluhm:
I'll wait for the int tests to run to completion but this looks good to me!

dbluhm:
Test failed but looks like a fluke. Rerunning.

PatStLouis:
the error seems to have to do with the [anoncreds revocation registry not being available](https://github.com/hyperledger/aries-cloudagent-python/actions/runs/7673810864/job/20918164545?pr=2737#step:3:14719)
`aries_cloudagent.anoncreds.revocation.AnonCredsRevocationError: No active registry`

The linting did change the indy/web/legacy registries files however I never heard of code linting breaking code. This would be wild.

dbluhm:
@ianco or @jamshale any hints what might be going wrong? I agree with Patrick that it seems unlikely that these changes would impact the anoncreds tests.

jamshale:
> @ianco or @jamshale any hints what might be going wrong? I agree with Patrick that it seems unlikely that these changes would impact the anoncreds tests.

The anoncreds integration tests are broken for anoncreds-dev8. They should get fixed with dev9.

I don't think we need to block this for them.

ianco:
I think this is related to the last anoncreds-rs update, we can check again when we get the anoncreds-rs `dev9` release

dbluhm:
Merging this as is then; thanks @ianco and @jamshale! And thanks for your patience @PatStLouis!

Issue #2743, opened 2024-01-29
[BUG] PR ready, traceability context framing issue with pyld


PatStLouis:
There is an ongoing issue with the w3c-ccg traceability context which raises an error when trying to expand a did document containing a specific combination of fields. The issue is raised [here](https://github.com/w3c-ccg/traceability-interop/issues/613) for reference.

The process of verifying a VC in aca-py currently frames the document at 2 moments:
- when fetching the [verification method](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/ld_proofs/suites/linked_data_proof.py#L145)
- when validating the [controller](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/ld_proofs/purposes/controller_proof_purpose.py#L57)

I will submit a PR with a workaround and explanation shortly and would like some input. I welcome any better solution...

swcurran:
@PatStLouis — great if you could make the ACA-Py maintainers meeting on Tuesday (2024.01.30) 9:00 Pacific / 18:00 Central Europe.  Meeting agenda: https://wiki.hyperledger.org/display/ARIES/2024-01-30+Aries+Cloud+Agent+-+Python+Maintainers+Meeting

PatStLouis:
closing as this has been addressed by #2795

PR #2744, opened 2024-01-29
Address traceability context bug #2743

PatStLouis:
This PR is to address #2743

There is an odd bug that's been ongoing for some time where when trying to expand a did document with the `pyld` library it fails.

Since this happen in a very specific condition, this PR looks to implement a workaround.

When we need to verify a json-ld VC where the issuer/verificationMethod is a `did:web:` scheme, we first manually resolve the did document with the `requests` module, then extract the traceability context form the document if present, then frame the document directly....

Its not pretty, but its the best solution I could come up with that has minimal impact on the existing code. If anyone has a better solution, I'm happy to change it!

@swcurran @dbluhm this is the last critical PR I would like to see in the 0.12 release

### DIFF
--- a/JsonLdCredentials.md
+++ b/JsonLdCredentials.md
@@ -16,6 +16,7 @@ By design Hyperledger Aries is credential format agnostic. This means you can us
 - [Issuing Credentials](#issuing-credentials)
 - [Retrieving Issued Credentials](#retrieving-issued-credentials)
 - [Present Proof](#present-proof)
+- [VC-API Endpoints](#vc-api)
 
 ## General Concept
 
@@ -210,3 +211,18 @@ Call the `/credentials/w3c` endpoint to retrieve all JSON-LD credentials in your
 ## Present Proof
 
 > ⚠️ TODO: https://github.com/hyperledger/aries-cloudagent-python/pull/1125
+
+## VC-API
+
+In order to support these functionalities outside of the respective DIDComm protocols, a set of endpoints conforming to the [vc-api](https://w3c-ccg.github.io/vc-api) specification are available. These endpoints should be used by a controller when building an identity platform.
+
+These endpoints include:
+- `GET /vc/credentials` -> returns a list of all stored json-ld credentials
+- `GET /vc/credentials/{id}` -> returns a json-ld credential based on it's ID
+- `POST /vc/credentials/issue` -> signs a credential
+- `POST /vc/credentials/verify` -> verifies a credential
+- `POST /vc/credentials/store` -> stores an issued credential
+- `POST /vc/presentations/prove` -> proves a presentation
+- `POST /vc/presentations/verify` -> verifies a presentation
+
+To learn more about using these endpoints, please refer to the available [postman collection](./demo/AriesPostmanDemo.md#experimenting-with-the-vc-api-endpoints).
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ ACA-Py supports a Transaction Endorsement protocol, for agents that don't have w
 
 ACA-Py supports deployments in scaled environments such as in Kubernetes environments where ACA-Py and its storage components can be horizontally scaled as needed to handle the load.
 
+### VC-API Endpoints
+
+A set of endpoints conforming to the vc-api specification are included to manage w3c credentials and presentations. They are documented [here](JsonLdCredentials.md#vc-api) and a postman demo is available [here](./demo/AriesPostmanDemo.md#vc-api).
+
 ## Example Uses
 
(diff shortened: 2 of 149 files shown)


dbluhm:
Not pretty indeed :sweat_smile: I'm quite hesitant on this one. Could you provide some more details on how pyld fails when resolving the traceability context? Has the issue been raised with the authors of that library?

PatStLouis:
So I've raised an issue on the [pyld repo](https://github.com/digitalbazaar/pyld/issues/188) and the traceability [interop repo](https://github.com/w3c-ccg/traceability-interop/issues/613). The `pyld` maintainers never acknowledged the issue and the traceability folks claimed the issue was most likely in the traceability context. There was a [fix suggested](https://github.com/w3c-ccg/traceability-vocab/pull/923) which didn't address the issue. The traceability folks are willing to help however its been difficult to get hold of them. 

For a minimal example, here's a code snippet. Its the specific combination of the traceability context, the traceability service endpoint and type. Remove any one of those 3 and the error isn't raised. 
```
from pyld import jsonld

document = {
  "@context": [
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/traceability/v1"
  ],
  "service": [
    {
      "id": "did:web:example.com#traceability-api",
      "type": ["TraceabilityAPI"],
      "serviceEndpoint": "https://example.com"
    }
  ]
}

jsonld.expand(document)
```

The issue is with the expanding part, which is a substep of the framing. The error returned is:
```
pyld.jsonld.JsonLdError: ('Tried to nullify a context with protected terms outside of a term definition.',)
Type: jsonld.SyntaxError
Code: invalid context nullification
```

I'm really trying to find a solution here since aca-py currently won't be able to verify `TraceableCredentials` because of this bug. 

There's a traceability meeting tomorrow I will try to get more info but it's been over a month I noticed this issue. It bothers me that nobody else encountered it because as far as I can tell acapy is doing things the right way and not being able to expand a linked data document sounds like a significant issue.

dbluhm:
If I stumbled on the pyld repo and saw that it hadn't seen a commit in 4 years and had 20 open PRs and 50+ issues, I would assume the project was dead. That doesn't seem far from the truth, frankly. A quick scan for an alternative does not yield much in python. So that's too bad lol... Even if we did poke at the library and figure out what needs to be fixed, it seems like it would not help since PRs aren't getting merged.

dbluhm:
I did a bit of `pdb`ing on the example you provided and from a very uninformed perspective, I thin it has to do with the empty `@context` in the `TraceabilityAPI` section of the context

dbluhm:
Can we call step 5.1.1 to the attention of the traceability folks in this section of the json-ld-api spec? https://w3c.github.io/json-ld-api/#algorithm

dbluhm:
Ah, you know what, I think I see the issue:

https://github.com/digitalbazaar/pyld/blob/316fbc2c9e25b3cf718b4ee189012a64b91f17e7/lib/pyld/jsonld.py#L3073

`if not ctx:` should be `if ctx is None:`. `ctx` will be falsey when given an empty dictionary. The algorithm states that the error condition should only be met when the ctx is explicitly a null or, in python, a None.

dbluhm:
I still don't understand the function of the empty context, though. If that's truly necessary according to the traceability context authors then we could try patching the library or putting a PR out for it on the off chance it gets merged (and released....). But if it's not strictly necessary, maybe it could be dropped in a revision of the context.

PatStLouis:
@dbluhm I will try the proposed edit for the pyld library. The maintainers are in the vc-api call I will point this out to them this afternoon and we can get a PR in if it solves the issue. Do you want to author it since you found the solution?

I will also touch on this at the traceability call this afternoon and bring up the point you mentioned.

dbluhm:
> @dbluhm I will try the proposed edit for the pyld library. The maintainers are in the vc-api call I will point this out to them this afternoon and we can get a PR in if it solves the issue. Do you want to author it since you found the solution?
> 
> I will also touch on this at the traceability call this afternoon and bring up the point you mentioned.

Feel free to open the PR! I have no preference to being the author of the solution :slightly_smiling_face:

PatStLouis:
@dbluhm see this [PR](https://github.com/digitalbazaar/pyld/pull/189) for the solution

swcurran:
@PatStLouis — can we close this with the new #2795 change?

swcurran:
Sorry for the extra question — I had been looking at the open PRs in prep for the release candidate and didn’t see your comment on #2795.  Doh!

PatStLouis:
@swcurran Yes when #2795 is merged we won't need this PR as this was a less desirable workaround. The other PR will address the core issue directly.

PR #2755, opened 2024-02-02
Remove asynctest dependency and fix "coroutine not awaited" warnings

ff137:
As initially done by @dbluhm in #2566, this drops the asynctest dependency, as it's massively out of date and generates hundreds of warnings in the tests - making it harder to improve on other warning logs.

It seems the dependency was subtly reintroduced in #2596. This PR attempts to refactor tests to only use unittest, as they fully support async tests since python 3.8 and external deps for async tests shouldn't be needed.

Edit: Summary of changes:

:heavy_minus_sign: Drops dependencies: `asynctest` (deprecated), and `async-case` (unused)

:recycle: Refactors tests to use unittest mocking

:white_check_mark:  Includes fixes for 3 tests that were silently not implemented properly (`coroutine not awaited`)

:construction_worker: Added a condition to tests workflow that it will fail if "coroutine not awaited" appears in the warning logs. (Here's a [sample result](https://github.com/hyperledger/aries-cloudagent-python/actions/runs/7756059087/job/21152653635?pr=2755) of a workflow failing, where all tests passed, but the warning appears in logs)

### DIFF
--- a/aries_cloudagent/protocols/didexchange/v1_0/tests/test_manager.py
+++ b/aries_cloudagent/protocols/didexchange/v1_0/tests/test_manager.py
@@ -2036,7 +2036,7 @@ class TestDidExchangeManager(IsolatedAsyncioTestCase, TestConfig):
                 await self.manager.accept_response(mock_response, receipt)
 
     async def test_accept_response_find_by_thread_id_no_did_doc_attached(self):
-        mock_response = mock.AsyncMock()
+        mock_response = mock.MagicMock()
         mock_response._thread = mock.MagicMock()
         mock_response.did = TestConfig.test_target_did
         mock_response.did_doc_attach = None


ff137:
Funny enough, the tests actually report _more_ warnings now ... 
Doing a diffcheck I see that the "deprecated @coroutine method" warnings are indeed gone, but without asynctest there are now warnings displayed for modules in `aries_cloudagent/anoncreds` which wasn't displaying previously. Maybe asynctest wasn't logging things properly. 
They're just Marshmallow warnings. I'll see if I can patch some of those in this PR too

ff137:
Quite a few different cases of deprecations warnings to fix, so I'll try tackle that in #2756

ff137:
@dbluhm just a heads up that some of the "`coroutine 'AsyncMockMixin._execute_mock_call' was never awaited` warnings appear to have been introduced in this PR: https://github.com/hyperledger/aries-cloudagent-python/pull/2687

I could fix the one case, in `test_accept_response_find_by_thread_id_no_did_doc_attached`. 
`mock_response = mock.MagicMock()` was changed to an AsyncMock, which let the test pass but didn't do things properly. 

Currently, test_create_request_emit_did_peer_2 and _4 aren't being mocked properly. Seems to be the only other "coroutine never awaited" problem. I'll see if I can figure those out too

ff137:
I figured it's a good idea to modify the tests workflow to always fail if a coroutine is not awaited ... because it indicates tests weren't set up properly, and it will avoid "silently failing" tests from merging in the future. So, I added that to the tests workflow:
```yml
      - name: Tests
        run: |
          poetry run pytest 2>&1 | tee pytest.log
          if grep -Eq "RuntimeWarning: coroutine .* was never awaited" pytest.log; then
            echo "Failure: Detected unawaited coroutine warning in pytest output."
            exit 1
          fi
```

Suggestions are welcome

.github/workflows/tests.yml
```diff
@@ -27,4 +27,9 @@ jobs:
           pip install poetry
           poetry install -E "askar bbs"
       - name: Tests
-        run: poetry run pytest
+        run: |
+          poetry run pytest 2>&1 | tee pytest.log
+          if grep -Eq "RuntimeWarning: coroutine .* was never awaited" pytest.log; then
+            echo "Failure: Detected unawaited coroutine warning in pytest output."
+            exit 1
+          fi
```
dbluhm, line 35: I think it is wise to fail tests on coroutine never awaited errors. I have an inkling that pytest may support this directly? Perhaps it just has a "fail on warning" and "don't fail on warning" and doesn't allow granular failures. Will piping to `tee` preserve the exit code for the pytest invocation? In other words, will the action still fail when there's a test failure and not just on a warning we don't want to see?

dbluhm (commented):
Nice! The AnonCreds code was on the back of my mind when I did the first round of async test removal but I'd since forgotten it needed attention after merging. Good catch on the did:peer tests as well! Quick comment on the test workflow adjustments.

.github/workflows/tests.yml
ff137: Ah yes, I thought the same to config pytest to fail on warning, but afaik it only lets you specify the kind of warning (RuntimeWarning / Deprecation warning) as a catch all. So to fail on specific warnings can be solved with this pipe to file and grep approach.

I also wondered if the pytest exit code is preserved -- thought the `2>&1` logic might accommodate that, but not sure. I'll test and refactor as needed 👌

ff137: @dbluhm  Indeed! It wasn't preserving the exit code ... doh. (Thanks GPT)
Luckily one can use `${PIPESTATUS[0]}`, because PIPESTATUS preserves the exit code of the most recent pipeline.
(Also thanks GPT) :smile: 

Testing fail/success cases now

ff137:
Testing new workflow changes:
- :white_check_mark: should fail if any test fails: 
  - https://github.com/hyperledger/aries-cloudagent-python/actions/runs/7757812743/job/21158132015
- :white_check_mark: should fail when all tests pass but a coroutine was not awaited: 
  - https://github.com/hyperledger/aries-cloudagent-python/actions/runs/7757905821/job/21158429703
- :white_check_mark: should succeed when all tests pass and no coroutine warnings: 
  - https://github.com/hyperledger/aries-cloudagent-python/actions/runs/7757963935/job/21158616988

Checks out :+1:

dbluhm (approved):
Thanks again and nice work!

PR #2775, opened 2024-02-12
Implement B006 rule

jamshale:


aries_cloudagent/multitenant/askar_profile_manager.py
```diff
@@ -60,6 +61,8 @@ async def get_wallet_profile(
             Profile: Profile for the wallet record
 
         """
+        extra_settings = extra_settings or {}
+
```
jamshale, line 65: The `extra_settings` dict is overridden later in this function. I didn't want to change anything not related to the task but it seems like this might be a mistake.

aries_cloudagent/anoncreds/routes.py
```diff
@@ -228,10 +228,10 @@ async def schema_get(request: web.BaseRequest):
     try:
         schema = await anoncreds_registry.get_schema(context.profile, schema_id)
         return web.json_response(schema.serialize())
-    except AnonCredsObjectNotFound:
-        raise web.HTTPNotFound(reason=f"Schema not found: {schema_id}")
+    except AnonCredsObjectNotFound as e:
```
jamshale, line 231: I had all of the `B` linting rules enabled for a bit and it flagged this.

swcurran (approved):

PR #2782, opened 2024-02-13
Anoncreds revoke and publish-revocations endorsement

jamshale:
Opening for visibility. I'm still going to manually test more but the Integration tests are passing for endorsement and publishing revocations.

One area of confusion for me is what happens with the local wallet verse the ledger. Write now an endorsement transaction is created when the `update_revocation_list` is called with changes. Then the local wallet is updated. At the same time the endorsement manager receives the `114` event and notifies the anoncreds `RevocationListFinished` event handler. It checks if the wallet contains a record for the list via `rev_reg_def_id` and if it doesn't it creates the record. If it does then it assumes the record has been updated already and does nothing. 

I was trying some other stuff like deleting the record and creating a new one with a `wait` state but couldn't decide if this was necessary, or something we wanted to do.

I deleted the `/anoncreds/revoke` and `/anoncreds/publish-revocation` endpoints and instead just have the `/revocation/revoke` and `/revocation/publish-revocation`. I thought this was appropriate because the revocation endpoint are all loaded for anoncreds specifically and the rest of the anoncreds endpoints are to do with creating objects. I changed the params for requesting a transaction manually to use the body options and updated the integration tests.

### DIFF
--- a/aries_cloudagent/anoncreds/revocation.py
+++ b/aries_cloudagent/anoncreds/revocation.py
@@ -483,7 +483,7 @@ class AnonCredsRevocation:
                     value_json={
                         "rev_list": rev_list.serialize(),
                         "pending": None,
-                        # TODO THIS IS A HACK; this fixes ACA-Py expecting 1-based indexes
+                        # TODO THIS IS A HACK; this fixes ACA-Py expecting 1-based indexes  # noqa: E501
                         "next_index": 1,
                     },
                     tags={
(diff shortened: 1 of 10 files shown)


ianco:
> I deleted the `/anoncreds/revoke` and `/anoncreds/publish-revocation` endpoints and instead just have the `/revocation/revoke` and `/revocation/publish-revocation`. I thought this was appropriate because the revocation endpoint are all loaded for anoncreds specifically and the rest of the anoncreds endpoints are to do with creating objects. I changed the params for requesting a transaction manually to use the body options and updated the integration tests.

I think this happened when I did the original merge of the `anoncreds-rs` branch.  You're correct we should just have the endpoints in one place and it makes sense to be under the `/revocation` section.

ianco:
Overall looks good, is it ready to go or are you still working on it?

jamshale:
My manual testing and integration testing has been good, so I think it's ready.

The last integration test run failed but I'm pretty sure it was a random unrelated fail.

ianco (approved):

Issue #2724, opened 2024-01-18
Enable the "emit-did-peer..." startup options to be tenant specific


swcurran:
We would like to see the new ability to emit DID Peer connections to tenant specific, so that we can experiment with enabling connection reuse.

swcurran:
@esune , @loneil — what is the effort to do this?  Are there instructions on how this is done?

esune:
[This PR](https://github.com/hyperledger/aries-cloudagent-python/pull/2233) has the changes that relate to per-tenant settings.
At minimum, we would extend the settings to be included/supported in the list that currently holds which settings can be "tenantized". We could also look at cleaning-up a bit that list (which is actually currently two lists) so that more settings are supported out-of-the-box without changes (probably a slightly bigger change).

swcurran:
So the steps are:

- update the Multitenancy.md documentation to add the item tot he list
- update the admin/routes.py to add the item to the two lists

Sound right?

esune:
Yes, that sounds about right if we just want to add this flag to the list without extra refactoring.

Issue #2648, opened 2023-12-04
Feature request: per-tenant setting for `preserve-exchange-records`


ff137:
Is it possible to add support for the `preserve-exchange-records` startup environment setting to be configurable at a per tenant level? It would be great if some holders/issuers can be onboarded with a wallet setting that automatically configures the desired behavior for preserving/deleting cred/pres exchange records, for that tenant in particular. If this is too much of a refactoring job, then that's np, we can get by with the `auto_remove` param that has been added to the issue/present methods.

Issue #2572, opened 2023-10-30
Per Tenant setting: requests-through-public-did


thiagoromanos:
Since I can define per tenant the setting `public_invites`, it would be great that I could also set per tenant the `requests-through-public-did` (in a case where I have a tenant that wants to receive those public did requests, but all other tenants don't).

swcurran:
Thanks — this is correct. @esune — is there a pattern/approach for making configuration parameters tenant-specific vs. global?

esune:
We have implemented an endpoint in a Traction plugin to facilitate setting these at runtime: see [this](https://github.com/bcgov/traction/blob/e0648f1c6210f5d29688e0427a5f8208c0a4d4bc/plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py#L363). I believe that anything listed in [this map](https://github.com/shaangill025/aries-cloudagent-python/blob/d705ca2fe9132fc5bdc07e427e10f74201bbfac2/aries_cloudagent/multitenant/admin/routes.py#L45-L62) is supported: we started with a subset of all the options that were available as there are several and we were not sure whether this would be useful for us only or others. I think if there are other flags that would be good to have set at a tenant level they just need to be added to that list.

c.c.: @shaangill025 for more insight.

thiagoromanos:
> We have implemented an endpoint in a Traction plugin to facilitate setting these at runtime: see [this](https://github.com/bcgov/traction/blob/e0648f1c6210f5d29688e0427a5f8208c0a4d4bc/plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py#L363). I believe that anything listed in [this map](https://github.com/shaangill025/aries-cloudagent-python/blob/d705ca2fe9132fc5bdc07e427e10f74201bbfac2/aries_cloudagent/multitenant/admin/routes.py#L45-L62) is supported: we started with a subset of all the options that were available as there are several and we were not sure whether this would be useful for us only or others. I think if there are other flags that would be good to have set at a tenant level they just need to be added to that list.
> 
> c.c.: @shaangill025 for more insight.

@esune , is there a plan to add this `requests-through-public-did` flag to that per-tenant list? Don't think that would be a breaking change to aca-py, since the global flag is there since 0.9.0.

esune:
> 
[...]

PR #2790, opened 2024-02-16
feature/per tenant settings

amanji:
- Adds `--emit-did-peer-{2 | 4}` per tenant setting
- Adds `--preserve-exchange-records` per tenant setting
- Adds  `--requests-through-public-did` per tenant setting

Resolves: #2724 
Resolves: #2648
Resolves: #2572

### DIFF
--- a/aries_cloudagent/multitenant/admin/routes.py
+++ b/aries_cloudagent/multitenant/admin/routes.py
@@ -42,6 +42,7 @@ ACAPY_LIFECYCLE_CONFIG_FLAG_MAP = {
     "ACAPY_EMIT_DID_PEER_2": "emit_did_peer_2",
     "ACAPY_EMIT_DID_PEER_4": "emit_did_peer_4",
     "ACAPY_PRESERVE_EXCHANGE_RECORDS": "preserve_exchange_records",
+    "ACAPY_REQUESTS_THROUGH_PUBLIC_DID": "requests_through_public_did",
 }
 
 ACAPY_LIFECYCLE_CONFIG_FLAG_ARGS_MAP = {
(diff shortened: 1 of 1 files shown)


swcurran:
Did any other work get done on this that would be worth documenting?  Other than simply adding the lines, was anything else done that might help future per tenant setting additions?

@loneil — does simply adding these items result in them being exposed/updateable in Traction, or is there another PR needed there to add these three settings to expose them in the Traction UI?

Thanks!

swcurran (approved):

loneil:
They will need a UI addition to add the fields on the settings page is all

amanji:
Should I do a quick PR to add them in traction?

amanji:
> Did any other work get done on this that would be worth documenting?  Other than simply adding the lines, was anything else done that might help future per tenant setting additions?
> 
> 
> 
> @loneil — does simply adding these items result in them being exposed/updateable in Traction, or is there another PR needed there to add these three settings to expose them in the Traction UI?
> 
> 
> 
> Thanks!

Not for this PR. It would be worthwhile exploring refactoring code to wholesale include parameters and create an exclusion list for those that shouldn't be tenant specific. I'd argue that a refactor of the argparse file should be explored to come up with consistent parameter names, etc.

loneil:
> Should I do a quick PR to add them in Traction?

This would be in an RC right? Maybe would have to just wait until the release that includes it is rolled out anyways?

jamshale:
The failing integration test seems to be a fragile test. The same test failed for my recent draft PR.

amanji:
> > Should I do a quick PR to add them in Traction?
> 
> 
> 
> This would be in an RC right? Maybe would have to just wait until the release that includes it is rolled out anyways?

Yeah we can file an issue for now and I'll add it as soon as a new version is released

PR #2795, opened 2024-02-16
Bump pyld version to 2.0.4

PatStLouis:
This is to address a bug we have found in the pyld library which was incorrectly asserting context elements as null.

Addresses issue #2743 and will close PR #2744.

For background information you can read [this closed issue on the pyld repo](https://github.com/digitalbazaar/pyld/pull/189).

@swcurran I tested these changes and was able to verify traceable credentials with the proper did document content.


swcurran:
I’ll watch for the tests, and we’ll include this in the pending release candidate.

swcurran (approved):

jamshale:
> I’ll watch for the tests, and we’ll include this in the pending release candidate.

A couple of the integration tests have been failing a lot today. I'm not sure why, as they had been passing consistently. Might be a resources and timing issue. I ran the failing test locally on main and it passed. Pretty sure we can ignore this fail if we want.

PatStLouis:
From the available logs the step that failed seems due to a [cred_def being on the ledger but not in the agent's wallet](https://github.com/hyperledger/aries-cloudagent-python/actions/runs/7937287801/job/21674211438?pr=2795#step:3:7355). Sounds like a false positive, do you have the ability to rerun the tests @jamshale ? We had occurrences before where the tests sometimes fail for some arbitrary reason, probably resource or timing issue like you mentioned.

PatStLouis:
I triggered a re-run be updating the branch, lets see

swcurran:
Add I merged another PR, so you are out of date again.  Update and I will bypass the protections..,

Issue #2792, opened 2024-02-16
Askar-Anoncreds wallet does not work with multitenancy/askar-profile


ianco:
For multitenancy, Askar supports two strategies for wallet storage:

- database-per-wallet - each wallet is stored in a separate database (this is the default)
- multi-wallet-per-database - multiple wallets are stored in a single database, using askar profiles

The second scenario is configured by providing the following options to aca-py:

```
--wallet-type askar
--multitenancy-config wallet_type=askar-profile
```

When using `--wallet-type askar-anoncreds`, aca-py throwns an error when trying to access the tenant wallet, one example is (from the faber demo):

```
Faber      |   File "/home/aries/aries_cloudagent/anoncreds/routes.py", line 196, in schemas_post
Faber      |     result = await issuer.create_and_register_schema(
Faber      |   File "/home/aries/aries_cloudagent/anoncreds/issuer.py", line 185, in create_and_register_schema
Faber      |     async with self.profile.session() as session:
Faber      |   File "/home/aries/aries_cloudagent/anoncreds/issuer.py", line 101, in profile
Faber      |     raise ValueError("AnonCreds interface requires AskarAnoncreds")
Faber      | ValueError: AnonCreds interface requires AskarAnoncreds
```

Looks like either a new profile is required under multitenancy, or the existing AskarProfile needs to be modified to support the `askar-anoncreds` scenario.

swcurran:
Could this be related to #2789 and what it "undid" #2705?  Perhaps try it again now that #2789 has been merged and the change reverted?

jamshale:
I don't think anoncreds fully supports multitenancy at this point. This is something i'm currently working on. 

Although `create_and_register_schema` was one endpoint that was supposed to be supported with default config. So this sounds like a separate issue other then just "not implemented yet". I can consider this use case when implementing multitenancy in general for the other endpoints.

ianco:
> I don't think anoncreds fully supports multitenancy at this point. This is something i'm currently working on.

I think it's best to wait on the wallet upgrade (for multitenancy scenarios) until we're confident all the anoncreds/multitenancy is working correctly.  We could proceed tho with the "single wallet" upgrade scenario (which is another ticket).

Issue #2786, opened 2024-02-14
Multitenancy - Subwallet - Author endorsement settings and automation


jamshale:
When creating a subwallet for an author agent some of the extra settings like `endorser-public-did` and `endorser-invitation`, etc... aren't available, and none of the auto setup with the endorser happens. Some of the author settings are available but not the settings for endorser connection and endorser info steps.

Having the ability to automate this process through configuration is very handy for controllers and should be a supported feature imho. It helps simplify tenant setup and prevents mistakes.

PR #2801, opened 2024-02-20
Add anoncreds multitenant endorsement integration tests

jamshale:
So, I had a misunderstanding. Endorsement for anoncreds in multitenancy is currently working when the admin agent is askar-anoncreds wallet type and the tenants are created with askar-anoncreds wallet type.

This PR will just add integration tests for this scenario.

I think it's important to have the feature https://github.com/hyperledger/aries-cloudagent-python/issues/2786 available for multitenancy going forward as this is where some of my confusion came from. When I first thought this didn't work, the author tenant wasn't setup completely. (It's pretty much finished other than some unit testing problems)

The other issue somewhat related to this issue is https://github.com/hyperledger/aries-cloudagent-python/issues/2792 but work can be done off of this issue as it is more relevant.

I'm thinking about how this is going to work with the admin agent on `askar` and a tenant is on `askar-anoncreds` and vice versa, and how this is going to work with the upgrade script etc. Currently this doesn't work because the wallet type of the admin agent is what decides what endpoints are available.

I don't think these questions are related to this ticket though so the tests should be enough to close this issue.

### DIFF
--- a/demo/features/0586-sign-transaction.feature
+++ b/demo/features/0586-sign-transaction.feature
@@ -36,13 +36,11 @@ Feature: RFC 0586 Aries sign (endorse) transactions functions
       @WalletType_Askar_AnonCreds @GHA
       Examples:
          | Acme_capabilities         | Bob_capabilities          | Schema_name    |
-         | --wallet-type askar-anoncreds | --wallet-type askar-anoncreds   | anoncreds-testing |
          | --wallet-type askar-anoncreds |                                 | driverslicense |
          |                               | --wallet-type askar-anoncreds   | anoncreds-testing |
-         |                               | --wallet-type askar-anoncreds --multitenant  | anoncreds-testing |
 
 
-   @T001.1-RFC0586 @GHA
+   @T001.1-RFC0586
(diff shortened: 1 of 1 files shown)


ianco:
> I'm thinking about how this is going to work with the admin agent on `askar` and a tenant is on `askar-anoncreds` and vice versa, and how this is going to work with the upgrade script etc. Currently this doesn't work because the wallet type of the admin agent is what decides what endpoints are available.

Per @swcurran 's comments about the upgrade process (askar to askar-anoncreds) I think we'll need to have a consistent set of API's available for the two wallet types.  Not sure why it's not working but endorsement definitely "should" work between the askar and askar-anoncreds wallet types.

There are other issues with how the endorser/author connection is setup but we have a couple of issues open on this topic.

ianco (dismissed):

jamshale:
Not sure how I did this after testing locally but I spelled `multitenant` wrong as the option and this caused the tests to hang.

I also slightly reduced the amount of endorsement test for `GHA`. They are getting pretty long and the few I removed are covered by the other tests.

ianco (approved):

PR #2804, opened 2024-02-21
Update publish-docs to operate on main and on branches prefixed with docs-v

swcurran:

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -2,13 +2,13 @@ name: publish-docs
 
 on:
   push:
-    # Publish `main` as latest
+    # Publish `main` as latest, and when pushes are done to branches with "v-doc" prefix
     branches:
       - main
+      - docs-v*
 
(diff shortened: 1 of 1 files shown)


swcurran:
@WadeBarnes -- check it out.  Should:

- publish docs for any created branches or tags ("on", "create")
- publish docs for pushes to branches that start with "docs-v" or main
- if the pushed or created tag / branch is not "main" or has a "v-docs" prefix, exit happily.
- otherwise, normal flow for mike (versioning support) and mkdocs

WadeBarnes:
Looks like it should work.  Could add the `- docs-v*` filter to the `on create`?

swcurran:
I had just added the GHA extension to VS-Code when I did this. Turns out that `create` doesn’t support `branches`.  I think early versions might have (found a reference on the internet that it did), but the extension flagged it as an error and when I checked the GHA docs, it doesn’t take any parameters.

WadeBarnes (approved):

PR #2816, opened 2024-02-27
feat: did-rotate

amanji:
Taken from #2494 (c/o @dbluhm)

This PR is an early implementation of the [RFC 0794 DID Rotate protocol](https://github.com/hyperledger/aries-rfcs/tree/main/features/0794-did-rotate). This was an exercise to feel out any potential issues with the protocol as defined.

This is not yet a complete implementation. Notably missing right now is a way to actually trigger a DID rotation through an Admin API call. I also need to finish wiring up the handlers to the manager. And who could forget about tests. 🙂

Overall, the protocol is solid. I'll report my thoughts on the protocol on the PR to the Aries RFCs.

An ACA-Py specific concern I discovered in this process that probably deserves some discussion: invalidating the connection target cache when the DID Rotation is committed. As currently structured, we never really expect the connection targets for a connection to change after the protocol establishing it concludes. DID Rotation shakes that expectation up.

Right now, in my changes, I've added a clear_connection_targets_cache helper method to partially address this. It's partial because it will only work for ACA-Py when run as a single instance with the default in memory cache or if you're using Indicio's redis cache plugin for whole cluster shared caching.

I think it would be better for ACA-Py to use a value including both DIDs of the connection in the cache key as a more complete solution. The problem with this and why I've not done it on a first pass is that there would need to be some restructuring of how we use the Responder classes so that the send methods expect a whole ConnRecord and not just a connection_id (or individually a their_did and my_did, I suppose) so we can extract the DIDs from it. 9 times out of 10, I would say that connection record is already available to the caller of the send method and my hypothesis is that in the remaining 1 time out of 10, it will not dramatically impact performance of critical operations (it is not common for us to know the connection_id and not also have a connection record -- maybe only a few places in the Admin API might be like this).

I'll do some more investigation on these points. Might be good to talk implications at the next ACA-PUG call.

cc @swcurran @TelegramSam
FYI @Jsyro this has ties back to updating the DIDs associated with connections that we've talked about recently.


amanji:
Tagging @dbluhm for review, please.

aries_cloudagent/protocols/did_rotate/v1_0/manager.py
```diff
@@ -0,0 +1,244 @@
...
+        except DIDMethodNotSupported:
+            raise UnsupportedDIDMethodError(RotateProblemReport.method_unsupported(did))
+        except DIDNotFound:
+            raise UnresolvableDIDError(RotateProblemReport.unresolvable(did))
+
+        try:
+            await conn_mgr.resolve_didcomm_services(did)
+        except BaseConnectionManagerError:
+            # TODO Make this reportable?
+            raise DIDRotateManagerError("Unable to resolve DIDComm services for DID")
```
dbluhm, line 244: I think we should make this a reportable error

aries_cloudagent/protocols/did_rotate/v1_0/manager.py
```diff
@@ -0,0 +1,244 @@
...
+        record.state = RotateRecord.STATE_ACK_RECEIVED
+        if not record.new_did:
+            raise ValueError("No new DID stored in record")
+
+        conn.my_did = record.new_did
+        async with self.profile.session() as session:
+            # Don't emit a connection event for this change
+            # Controllers should listen for the rotate event instead
+            await conn.save(session, reason="My DID rotated", event=False)
+            await record.save(session, reason="Received rotate ack")
```
dbluhm, line 191: Should we also delete the record?

aries_cloudagent/protocols/did_rotate/v1_0/routes.py
```diff
@@ -0,0 +1,121 @@
...
+
+    context: AdminRequestContext = request["context"]
+
+    profile = context.profile
+    did_rotate_mgr = DIDRotateManager(profile)
+
+    connection_id = request.match_info["conn_id"]
+
+    body = await request.json()
+    to_did = body["to_did"]
```
dbluhm, line 62: This implies that controller must first create a new DID to rotate the connection to. For pairwise connections (likely just about every single connection), we'd want to generate a new DID for every rotation. If you're rotating in bulk, that's double the number of requests... If rotating in bulk is needed, perhaps it would be best for there to be a separate Admin API for that.

The other issue is I don't believe there's currently a way to create a did:peer:2 or 4 through the Admin API. Creation of did:peer:2/4 is only done in DID Exchange at the moment. It shouldn't be too crazy to add to the `/wallet/did/create` route, I think. If we don't implement that in this PR, I think we'd want it before a release because the DID rotate protocol would be basically useless for the most common case without it.

pyproject.toml
```diff
@@ -60,6 +60,7 @@ ursa-bbs-signatures= { version = "~1.0.1", optional = true }
 
 # indy
 python3-indy= { version = "^1.11.1", optional = true }
+lxml = "^5.1.0"
```
dbluhm, line 63: What is this used for?

dbluhm (commented):
Looks pretty good! Some comments below. A general comment: we need integration tests for this protocol.

aries_cloudagent/protocols/did_rotate/v1_0/routes.py
amanji: I'd have to think about how to implement a bulk rotate in a nice interface given how the RFC is written. To be fair I have implemented the `/rotate` route directly based on how the message is described [here](https://github.com/hyperledger/aries-rfcs/tree/fe79f41e18786b628a91d65c1e13cfb824d2e16f/features/0794-did-rotate#rotate), not really thinking about how the DID would be generated in the first place. I think it makes sense to include a route (or update a current one) to accommodate the creation of did:peer:2/4 which would be fed into the `/rotate` endpoint. If we want to then follow up with a route to create and rotate to a new DID in one step, we can add a convenience handler for that.

Am I following you correctly?

aries_cloudagent/protocols/did_rotate/v1_0/manager.py
amanji: I think all three errors get reported back the to the sender at line: https://github.com/petridishdev/aries-cloudagent-python/blob/db05f1e072272f937fd71d886e56179fd136b2fe/aries_cloudagent/protocols/did_rotate/v1_0/manager.py#L123

Correct me if I'm wrong here

aries_cloudagent/protocols/did_rotate/v1_0/routes.py
dbluhm: Yep, that's about what I had in mind. Though, to clarify, I don't think it's necessary to create that bulk endpoint now. I think the minimum adjustment required to merge this PR is the did:peer:2/4 creation.

aries_cloudagent/protocols/did_rotate/v1_0/manager.py
dbluhm: I lean towards considering it overkill. The record itself doesn't store much and I can't think of a reason why the controller would want to try to query them later (nor are we exposing that ability anyways).

aries_cloudagent/protocols/did_rotate/v1_0/manager.py
amanji: Actually that one error extends from a non-reported exception. I can update that.

dbluhm: The `DIDRotateManagerError` is not a subclass of `ReportableDIDRotateError` so it's not captured in that `try/except` block. It would bubble up to the controller through the Admin API but the sender would not be notified. We would need to add a method like `method_unsupported` on the `RotateProblemReport` message to make this a reportable error.

pyproject.toml
amanji: Removed

aries_cloudagent/protocols/did_rotate/v1_0/manager.py
amanji: Replaced the `save` call with a `delete_record` method invocation.

aries_cloudagent/protocols/did_rotate/v1_0/routes.py
amanji: This is now implemented in the /wallet/did POST endpoint. Tests have been updated.

aries_cloudagent/connections/base_manager.py
```diff
@@ -739,6 +739,21 @@ async def get_connection_targets(
...
+    async def clear_connection_targets_cache(self, connection_id: str):
+        """Clear the connection targets cache for a given connection ID.
+
+        Historically, connections have not been updatable after the protocol
+        completes. However, with DID Rotation, we need to be able to update
+        the connection targets and clear the cache of targets.
+        """
+        # TODO it would be better to include the DIDs of the connection in the
+        # target cache key This solution only works when using whole cluster
+        # caching or have only a single instance with local caching
```
dbluhm, line 751: I think we should discuss this point; if we don't make this change now, there can be unexpected behavior in clusters. Additionally, once the rotation is completed, there is no way to handle any potentially out of order messages. If we did make this change, the cached value for the previous DIDs would be around until the cache expires them. There would still be issues in clustered environments without whole cluster caching, though.

amanji: TBH I'm not entirely familiar with how the different caching mechanisms work.

jamshale:
Not sure why this PR is getting so many random integration test fails. Not the same as the ones as I fixed recently. Might try an look at these other ones at some point. Re-running again.

aries_cloudagent/connections/base_manager.py
dbluhm: Yeah, I'm feeling like the right option right now is probably to log an issue to address handling potentially out of order messages received after a rotation. It's an unlikely edge case and I think it will be more valuable to keep this PR moving for now.

dbluhm (approved):

amanji:
@jamshale Yeah the tests seem so finicky.

jamshale:
> @jamshale Yeah the tests seem so finicky.

They seem to get worse sometimes and are better sometimes. The last fail was the one I thought I had fixed. Hadn't seen it in a while. Not sure if adding more retries might help or not.

Issue #2807, opened 2024-02-21
Anoncreds - support for anoncreds and askar wallets concurrently


ianco:
Support starting aca-py with both the credx and anoncreds libraries loaded, and support (in multitenancy) tenants with both askar and askar-anoncreds wallet types concurrently.  (This is necessary to allow tenants in a multi- environment to upgrade their sub-wallets from askar to askar-anoncreds at their convenience.  (Note that the upgrade function will be documented in a separate issue.)

When starting up aca-py, load both the credx and anoncreds libraries, and enable the API endpoints for both libraries (schema, credential definition and revocation).  Ensure that the endpoints are under different url's.  (Schema and cred def are currently under different url's - the "new" API's are under `/anoncreds`.  For revocation, the new endpoints will need to be moved to the `/anoncreds` url, alongside the schema and cred def.

In each API endpoint that is duplicated, add a check to each endpoint to make sure it is being called with the correct wallet type (for example for the new anoncreds-related API's, if the wallet type is `askar` then return an error code.)

When in multi-tenancy mode, create new tenant wallets using the overall aca-py's configured wallet type.  (For example if aca-py is started with `--wallet-type askar` then created tenant wallets with `askar` type, likewist with `askar-anoncreds`.)

When running in multi-tenant mode, ensure that the tenant wallet's wallet type is in scope for all processing related to that tenant (i.e. Admin API calls and processing for received agent-to-agent messages).

On aca-py startup, verify that the configured `--wallet-type` matches the actual database.  If the wallet is `askar` and aca-py is started with `--wallet-type askar-anoncreds`, aca-py should shut down with an error message.  Likewise if the wallet is `askar-anoncreds` and aca-py is started with `--wallet-type askar`.  If there is a database upgrade in progress (askar to askar-anoncreds, see issue #2808) then aca-py should allow startup but the only endpoint available will be the upgrade endpoint.  (Note that for a multi-tenant scenario, this just applies to the base wallet.)

WadeBarnes:
My vote is for a versioned API approach.  IMO this will enable the API to remain consistent from a endpoint path perspective between versions where appropriate (such as with the schema, cred-def, and revocation endpoints) which will in-turn make migration from one API version to another easier for developers; consistent APIs with different parameters/functionality.

I think this will also enable:
- Setting a default API, in the case an API version is not explicitly defined by the caller.  The default can be moved up as support for deprecated APIs is removed.
[...]

PR #2822, opened 2024-03-05
Anoncreds - support for anoncreds and askar wallets concurrently

jamshale:
Implements https://github.com/hyperledger/aries-cloudagent-python/issues/2807.

1. Enables schema, cred_def and revocation endpoints for both askar and askar-anoncreds in multitenant mode. All anoncreds endpoints are prefixed with `/anoncreds/`. For some api marshmallow schema objects they were the same name as the existing  objects which displayed a warning. I simply appended Anoncreds to the end of these objects that had conflicts.
3. Returns 403 errors when using the wrong endpoints with message. I thought about trying to do something fancier like a middleware, but decided because it was only a small subset that it would be easier to just add a method to the top of the endpoints. In anoncreds some of the endpoints already had a profile type check and through a ValueError. I did handle this, but for consistency and to prevent unrelated errors I still added the profile check method to the top of the method.
4. Prevent subwallets from being created that are a different type then the base wallet. For example, in the case of having the base wallet askar and trying to create a askar-anoncreds wallet it will throw a 403 and have a message. If the wallet type is empty is defaults to the base wallet.
5. Prevents wallets from being started as an askar wallet and then trying to change the config to askar-anoncreds by using a similar flow as versioning. If the `wallet-type` record is empty and there is a version record (existing wallet) it will assume the wallet is askar. I'm still unsure if anything needs to happen with `indy` wallets. I'm having trouble installing the library to test.
6. Updates integration tests and runs a few of the tests that touch most of the effected endpoints in multitenancy mode.
7. Updates the faber agent in the demos.

### DIFF
--- a/aries_cloudagent/anoncreds/default/legacy_indy/tests/test_registry.py
+++ b/aries_cloudagent/anoncreds/default/legacy_indy/tests/test_registry.py
@@ -86,7 +86,7 @@ mock_schema = AnonCredsSchema(
 class TestLegacyIndyRegistry(IsolatedAsyncioTestCase):
     async def asyncSetUp(self):
         self.profile = InMemoryProfile.test_profile(
-            settings={"wallet-type": "askar-anoncreds"},
+            settings={"wallet.type": "askar-anoncreds"},
             profile_class=AskarAnoncredsProfile,
         )
         self.registry = test_module.LegacyIndyRegistry()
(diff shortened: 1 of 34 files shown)


ianco:
> 4\. I'm still unsure if anything needs to happen with `indy` wallets. I'm having trouble installing the library to test.

I don't think we need to do anything with Indy wallets, other than document that if someone is still running an Indy wallet they need to upgrade to Askar before upgrading ton AnonCreds.  For multitenancy there is no "run Indy and Aksar in parallel" so they will need to upgrade *all* Indy tenants to Askar.

ianco (approved):

Issue #2703, opened 2024-01-09
Spike: Investigate/document how to use OOB Connection Reuse with did:peer


swcurran:
Now that we have `did:peer:2/4` support in ACA-Py, we *should* be able to support connection reuse with:

- OOB multi-use invitations
- using `did:peer:2/4`

This task is to investigate this use of connections to determine how a Bifold Wallet might be able to use this.

swcurran:
@ianco, could you please take a look at this one?  I think it should be easy to try out with the demo and the Swagger API.

PatStLouis:
I can look if IDLab can provide a web tool to conveniently and repeatedly generate an OOB credential offer QR code. Would this be useful @ianco ?

ianco:
Just started to look at this today.  I'm assuming this will use the `/out-of-band/create-invitation` endpoint?

How do you setup the payload to specify `did:peer:2/4` (or is it an internal coding change)?

```
{
  "accept": [
    "didcomm/aip1",
    "didcomm/aip2;env=rfc19"
  ],
  "alias": "Barry",
  "attachments": [
    {
      "id": "attachment-0",
      "type": "present-proof"
    }
  ],
  "goal": "To issue a Faber College Graduate credential",
  "goal_code": "issue-vc",
  "handshake_protocols": [
    "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0"
  ],
  "mediation_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "metadata": {},
  "my_label": "Invitation to Barry",
  "protocol_version": "1.1",
  "use_public_did": false
}
```

swcurran:
There is a on option for triggering emitting a DID peer 2 in these scenarios `—emit-did-peer-2`. Integration test using did-peer-2 can be found here: https://github.com/hyperledger/aries-cloudagent-python/blob/9031a4a59505719c6446a55dfc0437e5a6f40e30/demo/features/0160-connection.feature#L15

I think that is what you need to use.  The trick I’m wondering about is being able to establish a connection, and then being being able to use `reuse` when establishing a second connection (so you only have one connection…).

swcurran:
My hope is that it works out of the box, and we just need to describe how to do it.  But if it doesn’t work out of the box, what does happen and how do we fix it?

ianco:
See PR #2713  ...  I created an integration test to issue a credential and send a proof request and it fails :-(

[...]

PR #2823, opened 2024-03-05
Support connection re-use for did:peer:2/4

ianco:
See issue https://github.com/hyperledger/aries-cloudagent-python/issues/2703

This PR is READY FOR REVIEW!, reuse working for public DID and did:peer:2/4.

- adds additional meta-data to the wallet DID, to flag a did:peer for re-using in invitations
- add the DID meta-data to the admin API `GET /wallet/did` response
- add a flag to the OOB create invitation to request a unique did (default is to re-use the invitation DID in the wallet)
- reuse works for did:peer:2 and did:peer:4, and also public DID
- added separate flags to the demo for:
	- `--public-did-connections` - use the inviter's public DID in invitations, and allow use of implicit invitations
	- `--reuse-connections` - support connection re-use (invitee will reuse an existing connection if it uses the same DID as in the new invitation)
	- `--multi-use-invitations` - inviter will issue multi-use invitations
	- note that the `--reuse-connections` flag needs to be enabled in both Faber and Alice for the demo to enable connection reuse

Also added a few new integration tests with the `--connection-reuse` flag.


swcurran:
My thoughts on the questions:

> reuse works for did:peer:2 and did:peer:4, and also public DID
should this be generic? and how to determine if the DID can be used for connection reuse? should it work for ALL DID methods?

The requirement for reuse in the spec. is that the `service` item in the invitation be a DID -- not a JSON structure such as happens with an unqualified DID (what we've mostly used to now) or a did:peer:1. As we are dropping support for either of those, likely any DID is valid -- since the DID is resolved.  `did:peer:2/4` are magic in that by "resolving" them, the DIDDoc (the data that would be in the JSON structure) is decoded.  Note that `did:key` is not permitted as it does not have an endpoint to which messages can be sent.

Does that help?

> adds additional meta-data to the wallet DID, to flag a did:peer for re-using in invitations

Perhaps this should also allow flagging ledger-based (published/public) DIDs?  E.g. any DID that supports "resuse"?

> add the DID meta-data to the admin API GET /wallet/did response

Sounds necessary to me.

amanji:
Is there any way we can pull out the create methods for did:peer:2/4?. It would be useful to have for integration testing, but I can also patch in the change from this PR.

ianco:
> > adds additional meta-data to the wallet DID, to flag a did:peer for re-using in invitations
> 
> Perhaps this should also allow flagging ledger-based (published/public) DIDs? E.g. any DID that supports "resuse"?

It's already used for flagging public DID's, so I'm just piggybacking on an existing feature.

ianco:
> The requirement for reuse in the spec. is that the `service` item in the invitation be a DID -- not a JSON structure such as happens with an unqualified DID

OK this is clear, I'll update to check for a DID, and remove the specific `did_peer` checks that I added

ianco:
For `did:peer:4` should the connection use the long form or short form of the did?

```
      "my_did": "did:peer:4zQmbCiTt4dS2XZz9BMGYpZTKH92mYzq9Q1cavQPtfrFEfiV",
      "their_did": "did:peer:4zQmUpGK93RFrakSGmJbwL9QJXqs4uC2S8NQeSyzcomCqKzz",
```

vs ...

```
    "my_did": "did:peer:4zQmbCiTt4dS2XZz9BMGYpZTKH92mYzq9Q1cavQPtfrFEfiV:zCo9LoqasV3ZFnxnYNDU8yhBL9F6kGgrTrneiFJHM5Z8L7jw9Gj3H61gCqSKjiMxqWzztJykZfSAM5AXaeyD6wtFwjgyoPMZ6EYSoJ2HwFRd4oafTXtCvGmTC8E4FpjcFaQgnbe1uVbtRmZ1q96mPeMwsDszPiKuyH5D9rHULHkETFwd9hjPKSroX9XVoNxRaprNVr9JLWproTuDzCFwrRYou24dwVAUSsVtnM1ti2PJuZk3D5sdYqDeL8y4yw7WwVvzqd9jykuCt9nQyPBE2DE6saLSAyJRY3tYin2kcdeLCgWJfmaubDVqaEKQnFLZFJodbuts52PHhKGKcgoVXu4nKowE6zedeVtTmZEjn4USXg3oePgnQ47wDbFC5Z1nQgkZG3o7HtWWxmpYZQGqxxZxNDYckfLVxyAdNGZcXN2uHd6i788bc5B5tEEmBptAr9BeaggjobsZanaMDxNHkMLTHbVpn199pNwYXLwpZR8mdB4WMk99FQHyziWHbBLLUihEJqJzsdX88v8",
      "their_did": "did:peer:4zQmUwLj6dLCg2Sv9YirbFkzwFaMyRvmoqSn5Pge9fdP9ATi:zCo9LoqasV3ZFnxnYNDU8yhBL9F6kGgrTrneiFJHM5Z8L7jw9Gj3H61gCqSKjiMxqWzztJykZfSAM5AXaeyD6wtFwjgyoPMZ6EYSoJ2HwFRd4oafTXtCvGmTC8E4FpjcFaQgnbe1uVbtRmZ1q96mPeMwsDszPiKuyH5D9rHULHkETFwd9hjPKSroX9XVoNxRaprNVr9JLWproTuDzCFwrRYou24dwVAUSsVtnM1qintrVJ618X7sPxk13tCXhXHsMUHxK7PeL8sZG8dUFpc75k5waWuCTWK7gQhpvBGoijh4JCZMbqhFWGqUEwRZ3UbgC5f2DWM1MDJFbVtWmgkWtjtiEcJ1366gsLPRpuLYjPfAbMGRD6fXhFLaAmAxJjkEKpR4SvHUucdzv1LYeUZT6dMxf15dY5of2ca8YYSeGdpWYxrfpeMDkkfCSqXbTvXd8opaDEWjdmYMUnQVDBiWiW7C1cuPs73ewcCaeJP11yUbgD24qNtZZFLziaz3XoGRc5N1UfJ6HmCUKTi",
```

swcurran:
@dbluhm and @TelegramSam — can you please weigh in on this one to help Ian complete the implementation?  Thanks!

dbluhm:
> For `did:peer:4` should the connection use the long form or short form of the did?

I'd say short form unless there's a strong reason not to (there isn't an obvious reason that comes to mind for me).

dbluhm:
10 minute scan didn't reveal any red flags for me. I'll have to set aside some time to look deeper.

ianco:
Integration tests have failed twice in a row, so I ran them locally and everything passed.

aries_cloudagent/wallet/askar.py
```diff
@@ -348,12 +357,21 @@ async def get_local_did_for_verkey(self, verkey: str) -> DIDInfo:
...
-            return self._load_did_entry(dids[0])
+            ret_did = dids[0]
+            ret_did_info = ret_did.value_json
+            if len(dids) > 1 and ret_did_info["did"].startswith("did:peer:4"):
+                # if it is a peer:did:4 make sure we are using the short version
+                other_did = dids[1]  # assume only 2
+                other_did_info = other_did.value_json
+                if len(other_did_info["did"]) < len(ret_did_info["did"]):
+                    ret_did = other_did
+                    ret_did_info = ret_did.value_json
```
amanji, line 373: Shouldn't this be `ret_did_info = other_did.value_json`?

amanji (commented):
Looks good so far. Leaving comments as I see things.

aries_cloudagent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -279,6 +284,99 @@ async def create_invitation(
                     routing_keys=[],
                 ).serialize()
 
+        elif did_peer_4 or did_peer_2:
```
amanji, line 287: This function is getting quite long and is difficult to track logic. I would suggest at some point we refactor this out into smaller and more digestible blocks.

amanji (commented):
Small comment. Otherwise it LGTM!

aries_cloudagent/wallet/askar.py
ianco: Could be either I guess, it's the same value

aries_cloudagent/protocols/out_of_band/v1_0/manager.py
ianco: Yes I agree, I think that's a task for the next time a DID method needs to be supported for connection reuse

I chatted with Stephen about this general topic and we agreed to put it off.  The next DID(s) - that we know of - that need to support connection reuse are all public, so we should be good for now ...

aries_cloudagent/wallet/askar.py
amanji: Could we just keep it the same to alleviate confusion? I know it's a bit nitpicky but from an outside perspective I would assume there's a reason they're not referencing the same object.

ianco: Sure I'll fix it

ianco: Fixed :-P

amanji (approved):

PR #2759, opened 2024-02-05
feat: add new format and implement VCDICredFormatHandler (Draft)

tra371:
- add `VC_DI` in `CredFormat.Format`
- lay out `VCDICredFormatHandler`
- add new schema classes for cred offer and request
- implement CredExRecordVCDI
- fix schemas on cred_abstract and cred_request, improve on vc_di support on alice-faber demo

next step
- fix unittests issues

PR #2830, opened 2024-03-08
feat: add new format and implement VCDICredFormatHandler

gvelez17:
- add VC_DI in CredFormat.Format
- lay out VCDICredFormatHandler
- add new schema classes for cred offer and request
- implement CredExRecordVCDI
- fix schemas on cred_abstract and cred_request, improve on vc_di support on alice-faber demo

next step

fix unittests issues

Note: this is a rebased version of @tra371 's pr #2759


aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
```diff
@@ -43,18 +43,22 @@
     CRED_20_PROPOSAL: {
         V20CredFormat.Format.INDY.api: "hlindy/cred-filter@v2.0",
         V20CredFormat.Format.LD_PROOF.api: "aries/ld-proof-vc-detail@v1.0",
+        V20CredFormat.Format.VC_DI.api: "hlindy/cred-filter@v2.0",
```
gvelez17, line 46: @tra371 @sarthakvijayvergiya @supersonicwisd1 is this right or should this be didcomm ^

aries_cloudagent/protocols/issue_credential/v2_0/models/detail/vc_di.py
```diff
@@ -0,0 +1,135 @@
...
+            tag_filter={"cred_ex_id": cred_ex_id},
+        )
+
+    def __eq__(self, other: Any) -> bool:
+        """Comparison between records."""
+        return super().__eq__(other)
+
+
+class V20CredExRecordVCDISchema(BaseRecordSchema):
+    """Credential exchange indy detail record detail schema."""
```
gvelez17, line 88: should this comment say "indy" or should it say "vc_di" ?

aries_cloudagent/protocols/issue_credential/v2_0/routes.py
```diff
@@ -241,9 +291,11 @@ class V20IssueCredSchemaCore(AdminAPIMessageTracingSchema):
     def validate(self, data, **kwargs):
         """Make sure preview is present when indy format is present."""
```
gvelez17, line 292: update comment too pls

demo/features/steps/0454-present-proof.py
```diff
@@ -1,24 +1,13 @@
...
     async_sleep,
 )
-from runners.agent_container import AgentContainer
 from runners.support.agent import (
-    CRED_FORMAT_INDY,
-    CRED_FORMAT_JSON_LD,
-    DID_METHOD_SOV,
-    DID_METHOD_KEY,
-    KEY_TYPE_ED255,
-    KEY_TYPE_BLS,
```
gvelez17, line 21: huh is this just because we auto-cleaned unused imports? not sure why else we are making this change

demo/runners/agent_container.py
```diff
@@ -972,6 +971,25 @@ async def issue_credential(
 
             return cred_exchange
 
+        elif self.cred_type == CRED_FORMAT_VC_DI:
+            cred_preview = {
```
gvelez17, line 975: not required, but we could refactor with section above, just set some variable and then have common code block

demo/runners/faber.py
```diff
@@ -314,6 +340,70 @@ def generate_proof_request_web_request(
...
+                if revocation:
+                    vc_di_proof_request["non_revoked"] = {"to": int(time.time())}
+
+                proof_request_web_request = {
+                    "presentation_request": {"vc_di": vc_di_proof_request},
+                    "trace": exchange_tracing,
+                }
+                if not connectionless:
+                    proof_request_web_request["connection_id"] = self.connection_id
+                return proof_request_web_request
```
gvelez17, line 405: this is good, do we need to add any additional tests to demonstrate that the vc is a valid verifiable credential?  or is that out of scope of our changes?

demo/runners/performance.py
```diff
@@ -760,6 +764,13 @@ async def check_received_pings(agent, issue_count, pb):
             "('debug', 'info', 'warning', 'error', 'critical')"
         ),
     )
+    parser.add_argument(
+        "--cred-type",
+        type=str,
+        metavar="<cred-type>",
+        default=None,
+        help=("Specifyng the credential type"),
```
gvelez17, line 772: good idea to have this simple flag.  is there a default credential type that is used if the flag is not present?  should we use that as the default for the flag value to give the user a clue about that behavior

demo/runners/support/utils.py
```diff
@@ -115,7 +115,7 @@ def output_reader(handle, callback, *args, **kwargs):
             break
         try:
             run_in_terminal(functools.partial(callback, line, *args))
-        except AssertionError as e:
+        except AssertionError:
```
gvelez17, line 118: why remove as e?  the error object is useful right, maybe this was some artifact ?

gvelez17:
Overall this looks good to me, but do we also want to add some readme/markdown about what VC_DI is and how to get the verifiable credential out or how it can interoperate with other wallets?  we could also do that separtely from the repo or later, but i'd like to make sure new folks have an idea why, when and how to use them?

gvelez17:
@dhh1128 would greatly appreciate (and recognize) your guideance here to get this over the line nicely, if you have any suggestions.

aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
ianco: This is the issue I mentioned on Discord.  Technically the credential proposal is out of scope, so theres no "didcomm" version of the proposal, however when aca-py issues a credential it creates a "dummy" proposal on-the-fly and attaches it to the credential exchange record.  It then uses this to determine the format of the credential offer, issue, etc.  So even though you specify format "vc_di" the exchange ends up using Indy format.

You can see this if you run the demo:

```
./run_demo faber --wallet-type askar-anoncreds --cred-type vc_di  --events
```

... and ...

```
./run_demo alice --wallet-type askar-anoncreds  --events
```

Then connect the agents and issue a credential, and you can see the format in the resulting credential offer:

```
    "cred_offer": {
...
        "formats": [
            {
                "attach_id": "indy",
                "format": "hlindy/cred-abstract@v2.0"
            }
        ],
...
```

There are two possible fixes for this behaviour:

1. in the V20CredManager.create_offer() method, use the originally requested format, not the format that is attached to the proposal

or

2. Fake out a vc_di format proposal (that looks exactly like the Indy proposal, but with a vc_di format type) so that the create_offer() method picks up the correct type

I think #2 is probably a more elegant solution.  (I know the vc_di proposal was out of scope, but we didn't realize the detail of how it's used in the credential offer flow ...)

Let me know if you have any questions.  I can throw together a quick fix for #2 if you like ...

ianco:
> @ianco do we care that we don't have codecov? should we worry if we actually covered all the new stuff?

Personally I'm not too concerned about the level of code coverage as long as we have good integration tests.  However we do need a minimal set of unit tests to confirm that the code works.  I did a quick review of the code and it doesn't look like there are any new tests (?) in this PR yet.

Regarding the broken tests, many are throwing a "StopAsyncIteration" error - for example:

```
./scripts/run_tests aries_cloudagent/protocols/issue_credential/v2_0/tests/test_routes.py::TestV20CredRoutes::test_credential_exchange_issue
```

... raises this error.  I'm not sure the cause (it's raised during a "mock" call) but I suggest we focus on getting the demo working (see my comments about the credential proposal format), and once we confirm the demo is working properly we can revisit the tests.

aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
gvelez17: I agree, I'd love for us to do #2, or it doesn't feel like this is "real"

And def would love to see any thoughts/fix you have for it @ianco !

gvelez17:
> > @ianco do we care that we don't have codecov? should we worry if we actually covered all the new stuff?
> 
> Personally I'm not too concerned about the level of code coverage as long as we have good integration tests. However we do need a minimal set of unit tests to confirm that the code works.

mmm yeah usually I am lazy and start by looking at the codecov report, to make sure those unit tests are added.  @supersonicwisd1 do you want to try to just add some tests, or maybe run a quick local codecov and see where we are at - we need to make sure new code has tests

aries_cloudagent/protocols/issue_credential/v2_0/models/detail/vc_di.py
supersonicwisd1: Yes it should. It would be updated. Thank for catching it

aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
supersonicwisd1: I believe the #2 solution is preferred

ianco: I tried changing the format to:

```
    CRED_20_PROPOSAL: {
        ...,
        V20CredFormat.Format.VC_DI.api: "didcomm/w3c-di-vc-filter@v0.1",
    },
```

(I didn't make any other changes)

... and then I got an error from Alice when receiving the credential:

```
Alice      | 2024-03-08 22:34:36,122 aries_cloudagent.messaging.models.base ERROR V20CredOffer message validation error:
Alice      | Traceback (most recent call last):
Alice      |   File "/home/aries/aries_cloudagent/messaging/models/base.py", line 199, in deserialize
Alice      |     schema.loads(obj) if isinstance(obj, str) else schema.load(obj),
Alice      |   File "/home/aries/.local/lib/python3.9/site-packages/marshmallow/schema.py", line 723, in load
Alice      |     return self._do_load(
Alice      |   File "/home/aries/.local/lib/python3.9/site-packages/marshmallow/schema.py", line 910, in _do_load
Alice      |     raise exc
Alice      | marshmallow.exceptions.ValidationError: {'cred_def_id': ['Unknown field.'], 'nonce': ['Unknown field.'], 'schema_id': ['Unknown field.'], 'key_correctness_proof': ['Unknown field.']}
Alice      | 2024-03-08 22:34:36,123 aries_cloudagent.core.dispatcher ERROR Message parsing failed: Error deserializing message: V20CredOffer schema validation failed, sending problem report
```

With a bit of digging, it looks like the credential offer attachment is wrong, it is being generated as:

```
{
    "schema_id": "DqRJRAnD1es61Xvx4Q99k:2:degree schema:85.58.17",
    "cred_def_id": "DqRJRAnD1es61Xvx4Q99k:3:CL:12:faber.agent.degree_schema",
    "key_correctness_proof":
    {
        ... etc ...
    },
    "nonce": "1032768035276427881713135"
}
```

However the VCDI schema (defined here: https://github.com/hyperledger/aries-cloudagent-python/pull/2830/files#diff-f4aec3b9b074e0ceff1202bfc8f5dd81a387c14d116915fbb178dab705119176R325) doesn't match.

demo/runners/support/utils.py
supersonicwisd1: The exception is not returning the error. There is just a pass keyword after

ianco:
> We did pass the integration tests which run alice-faber so we should be close now...just need the real presentation format maybe

There aren't any integration tests for the new VC_DI format tho ...

aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
supersonicwisd1: It was the format suggest here that was followed 
https://github.com/hyperledger/aries-rfcs/pull/809/files#diff-40b1f86053dd6f0b34250d5be1319d3a0662b96a5a121957fe4dc8cceaa9cbc8R25-R69

ianco: I think that's the correct format but that's not what the code is producing ...

The schema is defined here, and that's what the received message is getting validated against:  https://github.com/hyperledger/aries-cloudagent-python/pull/2830/files#diff-f4aec3b9b074e0ceff1202bfc8f5dd81a387c14d116915fbb178dab705119176R325

The cred offer is getting created here, and it looks like it's just reusing the Indy offer right now:  https://github.com/hyperledger/aries-cloudagent-python/pull/2830/files#diff-f7d9c8d89f2442a02ac7a5924be4046e195d411791e4cbfd0fd9e7e2ecb987adR211

aries_cloudagent/protocols/issue_credential/v2_0/formats/vc_di/handler.py
```diff
@@ -0,0 +1,458 @@
...
+        attached_credential = cred_ex_record.cred_offer.attachment(
+            VCDICredFormatHandler.format
+        )
+        detail_credential = VCDICredAbstract.deserialize(attached_credential)
+        binding_proof = cred_ex_record.cred_request.attachment(
+            VCDICredFormatHandler.format
+        )
+        detail_proof = VCDICredRequest.deserialize(binding_proof)
+        manager = self.profile.inject(VcLdpManager)
+        # TODO - implement a separate create_credential for vcdi
```
dhh1128, line 349: Does this TODO need to be addressed before the PR is merged?

demo/features/steps/0453-issue-credential.py
```diff
@@ -1,7 +1,5 @@
 from behave import given, when, then
 import json
-from time import sleep
```
dhh1128, line 3: I'm a little bit puzzled about the fact that this demo got simpler by deleting stuff. Is it because we wrote a new demo that shows vc_di plus the stuff that's been deleted here? I would have expected this demo to get more complex as vc_di was added to the demo as well...

demo/features/steps/0454-present-proof.py
dhh1128: I have a similar reaction to Golda. Why is this stuff no longer needed?

demo/playground/examples/tests/test_mediator_ping_agents.py
```diff
@@ -99,23 +99,23 @@ def initialize_mediation(agent: Agent, invitation):
 
 @pytest.fixture(scope="session")
 def faber_mediator(faber, mediation_invite):
-    logger.info(f"faber_mediator...")
+    logger.info("faber_mediator...")
```
dhh1128, line 102: Although these string constants do not need the f"..." prefix, I think the author of this code used that convention to be consistent, knowing that very often, log statements would end up having parameters inserted: f"...{param}...". By putting the prefix on the string, it is guaranteed to be safe to insert the parameter. Without that prefix, every insert of a parameter requires the developer to analyze whether the prefix is needed. If the developer makes a mistake, the line has a bug, but the bug isn't detected unless the logged event actually happens, leaving the code a bit vulnerable.

For that reason, I'm not sure it's such a great idea to remove the prefix. But I don't feel super strongly about it. I'm just mentioning it as something to think about.

demo/runners/support/agent.py
```diff
@@ -732,7 +731,7 @@ async def register_or_switch_wallet(
         public_did=False,
         webhook_port: int = None,
         mediator_agent=None,
-        cred_type: str = CRED_FORMAT_INDY,
+        cred_type: str = CRED_FORMAT_INDY or CRED_FORMAT_VC_DI,
```
dhh1128, line 734: Same as previous. Isn't this a bug?

dhh1128 (commented):
I left several minor comments, but mostly, this looks like a great early version of a feature to me.

demo/runners/support/agent.py
```diff
@@ -672,11 +673,9 @@ async def register_did(
         did: str = None,
         verkey: str = None,
         role: str = "TRUST_ANCHOR",
-        cred_type: str = CRED_FORMAT_INDY,
+        cred_type: str = CRED_FORMAT_INDY or CRED_FORMAT_VC_DI,
```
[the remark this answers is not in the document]
ianco: I think @dhh1128 is correct this is a bug

demo/runners/support/agent.py
ianco: I think @dhh1128 is correct this is a bug

demo/features/steps/0453-issue-credential.py
ianco: I suspect the deletions are a side effect of a formatting tool, I've noticed that occasionally there are code changes due to code formatting standards changing.

gvelez17: yeah i think its automatically pruning unused imports, at least that is what i suspect

gvelez17: verified, at least with spot check.  resolving.

demo/playground/examples/tests/test_mediator_ping_agents.py
gvelez17: agree, restored.

aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
gvelez17: mmm are we sure its getting created here, i can't see where this is called.  maybe its the schema_id if it is called here? is it name magic?

aries_cloudagent/protocols/issue_credential/v2_0/formats/vc_di/handler.py
```diff
@@ -0,0 +1,458 @@
...
+        issuer = AnonCredsIssuer(self.profile)
+        ledger = self.profile.inject(BaseLedger)
+        cache = self.profile.inject_or(BaseCache)
+
+        cred_def_id = await issuer.match_created_credential_definitions(
+            **cred_proposal_message.attachment(VCDICredFormatHandler.format)
+        )
+
+        async def _create():
+            # TODO - implement a separate create_credential_offer for vcdi
```
gvelez17, line 212: @ianco any clue here quite welcome - do we really need a separate create_credential_offer?

Or can we get away with some kind of custom `issuer` that contains the right schema_id?

aries_cloudagent/protocols/issue_credential/v2_0/formats/vc_di/handler.py
gvelez17: yes we are trying to see how exactly to do it...agree this is the key place.  @dhh1128 or @ianco any suggestion on the right method very welcome.

aries_cloudagent/anoncreds/issuer.py
```diff
@@ -632,3 +632,62 @@ async def create_credential(
             raise AnonCredsIssuerError("Error creating credential") from err
 
         return credential.to_json()
+
+    async def create_credential_vc_di(
```
gvelez17, line 636: is this called?  @ianco i am not sure if there is name magic, if not i don't see where this is called - perhaps if we call it from the appropriate place - 
@sarthakvijayvergiya i hadn't thought about that, can we maybe see if we can just call this function to create the credential??

aries_cloudagent/protocols/issue_credential/v2_0/formats/vc_di/handler.py
gvelez17: @tra371 @sarthakvijayvergiya  is it possible that here we should call create_credential_vc_di from https://github.com/hyperledger/aries-cloudagent-python/blob/22d11e2c0084ae8ced2d920714e248abd7ff4f0a/aries_cloudagent/anoncreds/issuer.py#L636 ?

gvelez17: but that creates the credential itself, so maybe that is the attachment?

demo/runners/support/agent.py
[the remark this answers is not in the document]
gvelez17: well it will just be CRED_FORMAT_INDY, unless that is falsy, in which case it will be CRED_FORMAT_VC_DI
but that is also not what we want

```>>> False or 'hi'
'hi'
```
but yeah its still a bug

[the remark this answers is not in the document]
gvelez17: @tra371 @supersonicwisd1 lets fix this today and see what we want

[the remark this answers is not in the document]
gvelez17: its not a major bug - its just a default - its just confusing.  we probably should leave the original default

demo/runners/support/agent.py
gvelez17: it is a confusing default and a no-op change.  lets leave it as it was, with INDY being the default (which it still is, we are just confusing people)

aries_cloudagent/protocols/issue_credential/v2_0/message_types.py
gvelez17: ah this was a dupe function i have deleted it

aries_cloudagent/anoncreds/issuer.py
gvelez17: have deleted this, it was WIP

ianco:
> As currently structured, this strictly supports only AnonCreds format, right? In the future, what will be required to add support for issuing, say, DIP-VC or JWT-VC? Is there anything we can or should do now to make that easier for us later?

Keeping in mind that this initial work is being done under a Code With Us (https://marketplace.digital.gov.bc.ca/opportunities/code-with-us/7afcbd7c-2bbc-41ed-bf27-b6ba6e2903c5) we need to be mindful of scope creep.  @swcurran  what do you think?  My preference would be to keep the initial use case as simple as possible.

gvelez17:
> > As currently structured, this strictly supports only AnonCreds format, right? In the future, what will be required to add support for issuing, say, DIP-VC or JWT-VC? Is there anything we can or should do now to make that easier for us later?
> 
> Keeping in mind that this initial work is being done under a Code With Us (https://marketplace.digital.gov.bc.ca/opportunities/code-with-us/7afcbd7c-2bbc-41ed-bf27-b6ba6e2903c5) we need to be mindful of scope creep. @swcurran what do you think? My preference would be to keep the initial use case as simple as possible.

We're more than happy to structure for future extensibility, not concerned with scope creep if we can possibly complete after the formal approvals (once we get to there!)

do you guys have specific suggestions about structuring this for extensibility?

swcurran:
Agree that we will need future extensibility with this, so if there are some things that are obvious and easy, let’s do them. But highest priority is delivery the AnonCreds in W3C VCDM, so that’s the current focus.  I can’t answer Golda’s question about specifics to do, but if others have suggestions, please add them.  Thanks!!!

swcurran:
I should add that the goal would be to transition from the existing 0592 and 0593 Credential Attachment RFCs for both legacy AnonCreds and JSON-LD VCs (and 0771 AnonCreds) to the VC-DI attachments.  It will take some time, but that is the goal.

demo/runners/performance.py
ianco: The credential type is defaulted when creating the agent:

`cred_type: str = CRED_FORMAT_INDY,`

... it would probably be more clearer to assign the default in the input argument tho ...

gvelez17:
> , DIP-VC or JWT-VC

hey @dbluhm we are definitely interested in this after we get the first stage working, if we can plan a sync to understand your thoughts on it.  @tra371 @clehner

dbluhm:
@gvelez17 I was hoping you might have recommendations for where the right points in the implementation would be to extend later to support other credential formats and signature types.

gvelez17:
> @gvelez17 I was hoping you might have recommendations for where the right points in the implementation would be to extend later to support other credential formats and signature types.

hey Daniel is this still relevent based on the conversation in Anoncreds working group today that V2 will always default to a W3C compliant format?  Are we moving more towards a consensus format rather than ability to plug in many?  Happy to still look at this however will give some suggestions

dbluhm:
> > @gvelez17 I was hoping you might have recommendations for where the right points in the implementation would be to extend later to support other credential formats and signature types.
> 
> hey Daniel is this still relevent based on the conversation in Anoncreds working group today that V2 will always default to a W3C compliant format? Are we moving more towards a consensus format rather than ability to plug in many? Happy to still look at this however will give some suggestions

Yes; the VCDI credential attachment format is explicitly capable of issuing credentials with multiple cryptosuites, of which, AnonCreds is just one variety. We will want to be able to support signing credentials with other cryptosuites besides just AnonCreds.

sarthakvijayvergiya:
New PR:
https://github.com/hyperledger/aries-cloudagent-python/pull/2861

gvelez17:
Closed in favor of https://github.com/hyperledger/aries-cloudagent-python/pull/2861

PR #2835, opened 2024-03-13
feat: external suite provider interface

dbluhm:
This PR adds a pluggable component enabling usage of an externally provided signature suite. This is a refinement on previous attempts to make this aspect of ACA-Py more flexible; namely, #2706 which was later reverted due to issues with including profile in the context and multi-tenancy. This approach minimizes the "surface area" of the interface that the plugin must fulfill (at least when compared to what was proposed in #2706).

This is an advanced use case requiring some fairly deep knowledge of the management of DIDs and VCs. I have an example implementation of using an external KMS for signing JSON-LD creds that can be found here: https://github.com/dbluhm/acapy-ld-signer

### DIFF
--- a/aries_cloudagent/vc/vc_ld/external_suite.py
+++ b/aries_cloudagent/vc/vc_ld/external_suite.py
@@ -1,0 +1,44 @@
+"""Plugin hook for providing an external signature suite implementation.
+
+This enables greater control over where JSON-LD credentials are signed without
+requiring knowledge of the complexities of the JSON-LD/VC-LDP subsystem.
+"""
+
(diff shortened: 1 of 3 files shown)


swcurran:
Haven’t looked too carefully at this, but I think at minimum some sort of document is needed (likely in docs/features) that covers this and at least a start at how to use it.

ianco:
The update looks pretty straightforward but agree with @swcurran we need a doc describing the feature

dbluhm:
Basic docs, with a link to an example project, added!

swcurran (approved):

Issue #2838, opened 2024-03-15
Sending a Cred-Def supporting revocation to the ledger while ACA-Py has no tails-server configured


JoLuPuma:
Hello,
Sending a Cred-Def supporting revocation to the ledger while ACA-Py has no tails-server configured has displayed an error as expected: `aries_cloudagent.revocation.error.RevocationError: tails_server_base_url not configured`
But it didn't stop ACA-py to publish the [cred-def to the ledger](http://test.bcovrin.vonx.io:3707/tx/BCOVRIN_TEST/domain/467305). 

When I issued a credential using the cred-def in question, the holder (in this case the BC Wallet) get the credential-offer and he is even able to see the credential-preview, but after accepting the credential it never arrives. As expected too, car there is no Revocation registry created in the ledger, ACA-Py is unable to send the credential : 
`aries_cloudagent.protocols.issue_credential.v1_0.manager.CredentialManagerError: Cred def id 2VtgXP3ChNVSqR1vKjbUbE:3:CL:464042:jplc has no active revocation registry`

I want to know if this is a known behavior of ACA-Py or can this be considered as a bug?
Is it possible to make aca-py to verify if all the configuration for revocation has been set up before publishing to the ledger a Cred-Def supporting revocation?

Thanks

ianco:
Sounds like a bug to me.  If aca-py is creating a revocable credential definition (or credential, or revoking a credential) it should probably check that a tails server is configured.

jamshale:
I've seen this before, but thought this was by design, as you could later configure the tails server, and then create the revocation registries. I don't think this is good though, as you can get into the case described here, without knowing what happened unless you look at the logs. So, I agree with what Ian says. It should check the configuration and prevent the cred def from being created in the first place.

swcurran:
When we first implemented revocation and left it up to the Controller to do all the work to manage revocation, then what you are saying is “reasonable”.  But leaving to up to the Controller to manage revocation itself is “unreasonable” — way too hard. So we ONLY want to support ACA-Py handling all of the revocation details, so the if the CredDef says `revocation: true`, then ACA-Py should (MUST) be checking everything up front from creating the CredDef through the creation of the initial two RevRegs, and the publication of the tails file.  Ideally, its all or nothing — or as close as we can come to that.

PR #2849, opened 2024-03-21
Prevent revocable cred def being created without tails server

jamshale:
Addresses https://github.com/hyperledger/aries-cloudagent-python/issues/2838.

Simply returns an error response when creating a revocable cred def without a tails file configured.


swcurran (approved):

jamshale:
Going to merge. The one integration test fail was the endorsement promote public did one. Going to add more retries to it so it fails less often when ledger is slow.

PR #2854, opened 2024-03-22
Increase promote did retries

jamshale:
Doh! I think I must have accidentally deleted the wait in between retries somehow and not noticed.

Hopefully this should prevent the errors during promote did.

### DIFF
--- a/demo/features/steps/0586-sign-transaction.py
+++ b/demo/features/steps/0586-sign-transaction.py
@@ -40,8 +40,9 @@ def step_impl(context, agent_name, did_role):
     )
 
     # make the new did the wallet's public did
-    retries = 3
+    retries = 5
     for retry in range(retries):
+        async_sleep(1.0)
         published_did = agent_container_POST(
             agent["agent"],
             "/wallet/did/public",


swcurran (approved):

swcurran:
Test fail is super interesting — a CredDef that is attempting to be published fails because it already exists on the ledger.  I checked the ledger and `N1LoA9rMwyskHEW9csYpQJ:3:CL:360946:Faber.agent.Schema_DriversLicense_v2` was created on Feb. 16 — http://test.bcovrin.vonx.io:3707/tx/BCOVRIN_TEST/domain/360950.

How does that happen?  That means that the DID being used in the test was the same as being used on Feb. 16.

🤯

jamshale:
> Test fail is super interesting — a CredDef that is attempting to be published fails because it already exists on the ledger. I checked the ledger and `N1LoA9rMwyskHEW9csYpQJ:3:CL:360946:Faber.agent.Schema_DriversLicense_v2` was created on Feb. 16 — http://test.bcovrin.vonx.io:3707/tx/BCOVRIN_TEST/domain/360950.
> 
> How does that happen? That means that the DID being used in the test was the same as being used on Feb. 16.
> 
> 🤯

That's the other fail I've seen. Not as often. 

Feb 16. That doesn't make any sense???

swcurran:
The only thing I can think of is that the same seed for the DID was generated on both test runs.  Is it possible we’re not being random enough when coming up with the seed?

jamshale:
I just had another fail like this `69HTqHL7PjCS8K4Bp44g46:3:CL:293648:Faber.agent.Schema_DriversLicense_v2` from Jan 29. Seems like the tests must be recreating the same dids :/

PR #2860, opened 2024-03-28
Feature: use decorators for admin api authentication

### LINKED ISSUE #2318: Multitenancy: do not require main wallet api-key when invoking subwallet functions
When running ACA-Py in multitenancy mode, requests directed to an API controlling a subwallet function still require the main wallet `x-api-key` to be provided in the request headers for the request to succeed.

While there are ways to "mask" this from a user by proxying the request, it would be nice if the main wallet `x-api-key` header was only required for operations related to tenant provisioning and management, and not for requests directed to a subwallet.

esune:
Resolves #2318 

Opening draft PR early for feedback while I work through the changes across the code (and fixing/updating/adding tests as needed).

The PR removes the authentication middleware and the logic they deal with, and implements two decorators:

- `admin_authentication`: to be used for routes that should ONLY be invoked by an administrator, such as the multitenancy endpoints, the server endpoints and so on, independently of the mode the agent is running as.
- `tenant_authentication`: to be used to require authentication by either providing a tenant token (multi-tenant mode) or a valid api-key (single-tenant mode).

Both decorators account for unauthenticated `options` requests as well as insecure mode. Insecure paths will just not be decorated. Middleware code - currently commented-out - will be removed.

I think the bit of refactoring required for this to work (including plugins once released) is well worth the flexibility - looking for early feedback especially from @dbluhm, @ianco, @jamshale


ianco:
Looks good upon first glance.

esune:
> Looks good upon first glance.

Thanks @ianco . I've pushed updates to all the route handlers, now working on cleaning-up and fixing/adding tests and validating things really work as expected.

dbluhm (commented):
I like the flexibility and straightforward-ness of the approach. I like a nice function decorator lol. I always tend toward caution with them though; they can have some gotchas. Where these are being called by the aiohttp router, I think the gotcha potential should be minimal. LGTM so far.

aries_cloudagent/admin/server.py
```diff
@@ -306,87 +298,8 @@ async def make_application(self) -> web.Application:
         # we check here.
         assert self.admin_insecure_mode ^ bool(self.admin_api_key)
 
-        def is_unprotected_path(path: str):
```
amanji, line 309: Nice cleanup of this code and other spots where the route is being hardcoded!

aries_cloudagent/admin/decorators/auth.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @functools.wraps(handler)
+    async def wrapper(request):
+        context: AdminRequestContext = request["context"]
+        profile = context.profile
+        header_admin_api_key = request.headers.get("x-api-key")
+        valid_key = general_utils.const_compare(
+            profile.settings.get("admin.admin_api_key"), header_admin_api_key
+        )
+        insecure_mode = bool(profile.settings.get("admin.admin_insecure_mode"))
```
amanji, line 27: Is it better to just shortcut out of the function if `insecure_mode` is set, right at the start of the function?

aries_cloudagent/admin/decorators/auth.py
```diff
@@ -0,0 +1,80 @@
...
+    @functools.wraps(handler)
+    async def wrapper(request):
+        context: AdminRequestContext = request["context"]
+        profile = context.profile
+        authorization_header = request.headers.get("Authorization")
+        header_admin_api_key = request.headers.get("x-api-key")
+        valid_key = general_utils.const_compare(
+            profile.settings.get("admin.admin_api_key"), header_admin_api_key
+        )
+        insecure_mode = bool(profile.settings.get("admin.admin_insecure_mode"))
```
amanji, line 62: Same comment as above. Is it better to just shortcut out of the method if `insecure_node` is set or if the `OPTIONS` method is being used?

amanji (commented):
So far, this all LGTM! Really liking how clean those decorators are.

aries_cloudagent/admin/decorators/auth.py
esune: The condition would still need to be checked, so I preferred having only two exit points rather than a bigger if/else block. I can get those two extracted to a first condition if it seems better, but overall they all follow the same logic.

esune:
Fixed the route tests, still need to tackle removing the now failing middleware tests and add tests for the decorator functions.
Unsure why the formatter check is failing as it looks (to me) like the change it suggests is the same as the change that is already there?

esune:
Rebased code off the latest main. I am currently fighting with failing unit tests: when calling endpoints in the `server` API namespace, I get the following error and have been so far unable to figure out why it is being triggered even after stepping through the code execution:
```log
2024-04-08 18:21:29,365 aiohttp.server ERROR Error handling request
  File "/home/vscode/.local/lib/python3.9/site-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
  File "/home/vscode/.local/lib/python3.9/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/home/vscode/.local/lib/python3.9/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/admin/server.py", line 180, in ready_middleware
    return await handler(request)
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/admin/server.py", line 217, in debug_middleware
    return await handler(request)
  File "/workspaces/aries-cloudagent-python/aries_cloudagent/admin/server.py", line 362, in setup_context
    return await task
  File "/usr/local/lib/python3.9/asyncio/futures.py", line 284, in __await__
  File "/usr/local/lib/python3.9/asyncio/tasks.py", line 328, in __wakeup
    future.result()
  File "/usr/local/lib/python3.9/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/usr/local/lib/python3.9/asyncio/tasks.py", line 256, in __step
    result = coro.send(None)
  File "/usr/local/lib/python3.9/site-packages/aiohttp_apispec/middlewares.py", line 45, in validation_middleware
    return await handler(request)
TypeError: admin_auth() takes 1 positional argument but 2 were given
```

An easy way to reproduce is to start the agent and call any of the endpoints protected with `admin-authentication` in the `server` namespace. Please note that calling similarly set-up endpoints in the `multitenancy` namespace does work as expected.
Any input is appreciated as this is now blocking progress (want to fix this scenario before creating new tests for tyhe decorators).

esune:
All tests should be fixed now. I updated/synced the version of Black as three different versions were specified between `pyproject.toml`, `.pre-commit-config.yaml` and `blackformat.yml` and ran the formatter on the project, however it still raises potential changes - if anyone has an idea of why/how to address this please let me know.

jamshale:
> All tests should be fixed now. I updated/synced the version of Black as three different versions were specified between `pyproject.toml`, `.pre-commit-config.yaml` and `blackformat.yml` and ran the formatter on the project, however it still raises potential changes - if anyone has an idea of why/how to address this please let me know.

I haven't looked into the pre-commit or black format files. I use the `ruff` vscode extension to format my files and it seems to work well with the project. If I do get a fail from black on the PR I use the `Run and Debug` black or ruff process from the devcontainer and it re-formats the code correctly for me.

esune:
This is - finally - ready for review.

WadeBarnes (approved):

jamshale:
There's one failing unit test because of a ruff formatting error. `aries_cloudagent/config/argparse.py:650:91: E501 Line too long`.
You can just add `  # noqa: E501` to the end of that line.

poetry.lock
```diff
@@ -1,88 +1,88 @@
-# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
jamshale, line 1: I don't know what the policy is with package upgrades on this repo. But, I think it probably better to only upgrade the package that you changed in the pyproject.toml file. All the tests are passing so it's probably ok, but it's not really related to the task.

jamshale:
Looks good and I will approve. I just had one comment about how there is a upgrade on all packages. I think it's better to only upgrade the packages related to the PR and do a general upgrade in a separate PR. In the case we do find a problem with an upgrade it will be easier to isolate.

poetry.lock
esune: I re-generated the lockfile after updating the version of black: the dependency and the workflows were mismatched, so I updated the version and re-locked the dependencies so it should be only that. I can revert that change if preferable.

jamshale (approved):

Issue #2855, opened 2024-03-22
Integration tests creating existing DID's


jamshale:
The integration tests are recreating the same did's they previously created causing fails because the schema/cred_def's they are trying to create already exist on the ledger. Should be able to make the seed/did creation more random to prevent this from happening.

PatStLouis:
have we found in the code where the seed is generated? I was thinking in [here](https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/run_bdd) but I couldn't see it

ianco:
The aca-py seed is created here:  https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/runners/support/agent.py#L211

It's kind of hokey and not really robust enough for running integration tests, where we may create hundreds of DID's in a single run of the tests.  (It was initially implemented for a simple demo but should be updated ...)

PatStLouis:
I can suggest to switch to python [secrets.token_urlsafe](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe) to create a random seed with large entropy instead of randint.

ianco:
> I can suggest to switch to python [secrets.token_urlsafe](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe) to create a random seed with large entropy instead of randint.

I'll take a look, thanks for the suggestion.

PatStLouis:
@ianco looks like @dbluhm is addressing this in another pr with a similar approach, using `token_hex` instead. Less entropy but good enough!

Issue #2857, opened 2024-03-26
Updates to enable the backwords compatibility and transitioning to did:peer:2/4 and all qualified DIDs


swcurran:
In looking at the approach that has evolved in release 0.12.0 for handling the transition to the use of qualified DIDs only, we've decided on the following changes to the current implementation:

- Remove the new configuration parameters that were added as part of 0.12.0 `--emit-did-peer-2` and `--emit-did-peer-4`.
  - Remove them as well from the "per-tenant" settings list of parameters.
  - We've decide that finer grained control is needed at the point of creating invitations and DID Exchange requests is more useful than configuration settings.
- A `use_did` parameter will be added to the `outofband/create_invitation` that takes either
   - a DID Method (options: `did:peer:2` `did:peer:4`) so that a DID of that type is created (if necessary), and used in the invitation. If a DID of the type has to be created, it is flagged as the "invitation" DID and used in all future invitations so that connection reuse is the default behaviour.
      - This is the recommend approach, and we further recommend using `did:peer:4`.
      - Some care is needed when the DID method is called with one type than another. I would suggest each time a different DID Method is called (not the current default) a new DID be created and flagged for reuse in future invitations.
   - or
   - a complete DID, which will be used for the invitation being established.  This supports the edge case of an entity wanting to use a new DID for every connection. It is the responsibility of the controller to create the DID before passing it in.
   - If not provided, the 0.11.0 behaviour of an unqualified DID is used.
     - We expect this behaviour will change in a later release to be that a `did:peer:4` is created and (re)used.
- A `use_did` parameter will be added to the `didexchange/create_request`. `didexchange/accept_invitation` that takes either
   - a DID Method (options: `did:peer:2` `did:peer:4`) so that a DID of that type is created for use in the connection being established. This is the recommend approach, and we further recommend using `did:peer:4`
   - or
   - a complete DID, which will be used for the connection being established.  This supports the edge case of an entity wanting to use the same DID for all connections.
   - If not provided, the 0.11.0 behaviour of an unqualified DID is created if DID Exchange 1.0 is used, and a DID Peer 4 is used if DID Exchange 1.1 is used.
     - We expect this behaviour will change in a later release to be that a `did:peer:4` is created.
- When `auto-accept` is used with DID Exchange, the an unqualified DID is created if DID Exchange 1.0 is being used, and a DID Peer 4 is used if DID Exchange 1.1 is used.

[...]

PR #2862, opened 2024-04-01
refactor: introduce use_did and use_did_method

dbluhm:
This implements #2857. With one modification: rather than having a single parameter `use_did`, I found that the processing became significantly simpler by having a `use_did` and `use_did_method` parameter. Hopefully that's an acceptable trade off.

WIP: Currently, just the updates to the did exchange routes have been made. Working through the OOB route updates now. Wanted to open early for feedback.


amanji (commented):
LGTM so far. Some nice cleanup

dbluhm:
Quick update, I took a stab at wrangling the logic of the `create_invitation` method on the OOB manager by introducing a separate class for the purpose. Felt like overkill but the more I leaned into the approach, the clearer to me that the logic is rather complex and having it be split up this way might help. Interested in feedback if anyone has any.

dbluhm:
https://github.com/hyperledger/aries-cloudagent-python/actions/runs/8542753211/job/23405071552?pr=2862#step:3:31413

I think this test failure is unrelated so I'll focus on fixing unit tests first and removing the emit flags officially.

swcurran:
Yes — we are getting this error too often. Pretty sure it is that the DID Seed is not random enough — we keep reusing DIDs used in earlier tests...

dbluhm:
Quick note that with the removal of the `--emit-did-peer-4` and `--emit-did-peer-2` flags, I believe that would mean that the AATH tests recently added for qualified DIDs might need to be updated since those tests depend on using these flags.

dbluhm:
I was seeing test failures that I was not certain were related to my changes here so I went ahead and fixed #2855 in 62744c1

dbluhm:
Interop testing is looking pretty good. I have some small issues to sort out with backwards compat still (my respond-in-kind for the message version is a little wonky) but should be wrapping up soon.

dbluhm:
This is ready for review!

ianco (dismissed):
LGTM.  Are there any docs we need to update?

dbluhm:
Good point! I'll see if I can toss in a quick doc update somewhere

dbluhm:
As I was writing the docs, I realized there is one case I overlooked: when the inviter creates an invite with didexchage/1.1 only and the invitee supports 1.1, it is still sending an unqualified DID in a 1.1 message if the invitee accepts the invitation without specifying a `use_did_method`. I believe it would be preferable for the invitee to automatically prefer using `did:peer:4` in this case. I'll go ahead and make that tweak but open to thoughts if there are any.

jamshale (dismissed):
LGTM!

dbluhm:
@ianco @jamshale I made a couple of fixes and added a new doc outlining use of qualified DIDs in ACA-Py, with details on OOB and DID Exchange

jamshale (approved):

ianco (approved):

swcurran:
@dbluhm — are you happy that this is ready to merge?  If so — we can push 0.12.0-rc3 out.

dbluhm:
Yep, ready! I was half keeping an eye on the integration tests just to be sure but my testing indicates that it should be fine

PR #2866, opened 2024-04-02
Fix run_tests script

ianco:
Fix a change that I think @dbluhm did awhile back, which is now causing issues when trying to run tests locally.

@dbluhm  what was the reason for the fix?  This PR removes your change but maybe there is an alternative that addresses your reasons without breaking local tests ...


dbluhm:
Ah, shoot. Sorry about that. This PR fixed a minor annoyance in my environment and I incorrectly assumed that it would work in more "traditional" docker environments.

dbluhm (approved):

esune:
Thanks @ianco !

PR #2867, opened 2024-04-02
Add missing verificaiton method option

PatStLouis:
I replaced the custom `IssuanceOptionsSchema` schema on the vc/issue endpoint with the previous `LDProofOptionsSchema`. The idea will be to revisit the idea of having custom options field for different endpoints in a post 1.0 version. The verificationMethod option is necessary when issuing using unconventional did methods, such as did:web where the wallet can't correlate a verificationMethod with a specific did

### DIFF
--- a/aries_cloudagent/vc/routes.py
+++ b/aries_cloudagent/vc/routes.py
@@ -186,7 +186,7 @@ async def prove_presentation_route(request: web.BaseRequest):
         options = {} if "options" not in body else body["options"]
 
         # We derive the proofType from the holder DID if not provided in options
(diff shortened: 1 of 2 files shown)


PatStLouis:
@dbluhm @jamshale this should be a fairly straight forward review. For a refresher, on the vc-api issuance endpoints, if the controller doesn't provide a `proofType`, it will default to `Ed25519Signature2020` or `BbsBlsSignature2020` depending on the key method of the did.

dbluhm (approved):

PR #2892, opened 2024-04-13
feat: drop indy sdk

### LINKED ISSUE #2865: Officially remove Indy SDK support
Clarification: Indy SDK != Indy Network support. Using the Askar wallet type, Indy Network support is retained through the Indy VDR library and AnonCreds support through Indy CredX and, soon, AnonCreds RS.

The Indy SDK has been deprecated since 0.9.0. Support should be officially removed.

dbluhm:
Resolves #2865

This PR removes all components and tests referencing the Indy SDK library. This was accomplished by working through each module depending on the library, removing it, and then adjusting the places that referenced that module to be updated to point to replacing components.

It should be noted that the code base was already in a pretty good state; the Indy SDK dependencies were pretty neatly isolated. The "worst" of these updates was the odd test here and there that was directly using Indy SDK pieces when they should have been using things like `InMemoryProfile.test_profile()` or similar.

We have already ceased running tests for Indy specific components automatically in GHA. Because of this, I'm confident the removal of Indy specific tests should not cause a gap we've not already filled. However, I think this will technically result in less overall code coverage of the tests, if you were considering both the indy and non-indy tests.


dbluhm:
I'm not sure about my changes in `aries_cloudagent/ledger/multiple_ledger/manager_provider.py`

jamshale (dismissed):
This looks good to me and will approve. Should probably get another opinion with more historical experience, though.

ianco (dismissed):

dbluhm:
@jamshale @ianco merged in main and resolved conflicts in the poetry lock. This will need a fresh review but then should be ready to merge. Thanks!

jamshale (approved):

PR #2898, opened 2024-04-17
:art: fix typos

ff137:
The most minor of quality of life improvements.

The motivation was to fix typos that appear in the openapi spec, because typos in the model descriptions would propagate to autogenerated clients. 

Fixing those, I would 'find all' to fix all instances in the codebase. So at first it was just to fix typos in model descriptions, and that expanded to more parts. 
When I started correcting double spaces, that's when I knew I'd gone too far.

Using the Code Spell Checker vscode extension made this pretty straightforward. Can recommend.

No external model names or anything changed, so no breaking changes as a result of corrections. Just internal variable names, comments, and some readme improvements.

Edit: I also modified model descriptions to no longer contain \n characters or excessive whitespace, and committed the updated openapi specs, to reflect the changes there as well.

Edit2: Went a bit above and beyond to correct typos in more modules. Lots of files changed, I know... but a scanning review should just take a couple minutes!


dbluhm (approved):

dbluhm:
Looks like a regen of the openapi and swagger json files is necessary after merging #2899 :slightly_smiling_face:

aries_cloudagent/resolver/default/peer3.py
```diff
@@ -64,7 +64,7 @@ async def _resolve(
         return doc
 
     async def create_and_store(self, profile: Profile, peer2: str):
-        """Injest did:peer:2 create did:peer:3 and store document."""
+        """Inject did:peer:2 create did:peer:3 and store document."""
```
dbluhm, line 67:
```suggestion
        """Ingest did:peer:2 create did:peer:3 and store document."""
```
This one might actually be meant to be "ingest," from the context; i.e. taking in a did:peer:2, creating a did:peer:3, and storing a document.

dbluhm (approved):

aries_cloudagent/resolver/default/peer3.py
ff137: Ingest was my first guess 😄 but then I doubted myself when I noticed that "inject" is also just 1 character away. Plus, inject is less gastrointestinal.

If ingest makes more sense, I'll amend that 👌

The other funny one was "encluding", which was the hardest to correct 😂 both "including" and "excluding" are just a 1 character diff. So that's another one I'd appreciate confirmation that I guessed right 😄

dbluhm: I think it's probably clear enough as it is :slightly_smiling_face: The wording in general on this one is a bit odd

PR #2903, opened 2024-04-18
fix Faber demo to use oob with aip10 to support connection reuse

ianco:


demo/runners/agent_container.py
dbluhm, line 1: Maybe a nitpick but did we want to keep the `print` statements in this file?

dbluhm (approved):

Issue #2901, opened 2024-04-18
`PublishRevocations` model replaced with `PublishRevocationsSchemaAnoncreds`


ff137:
Good day

I notice from working with the openapi spec, in 0.12.0 there is a new model: `PublishRevocationsSchemaAnoncreds`.
It looks like this (from running openapi-generator):
```py
class PublishRevocationsSchemaAnoncreds(BaseModel):
    rrid2crid: Optional[Dict[str, List[Annotated[str, Field(strict=True)]]]] = Field(
        default=None, description="Credential revocation ids by revocation registry id"
    )
```

This new model is now used to replace the previous `PublishRevocations` models in the Revocation API:
- The following methods: `clear_pending_revocations` and `publish_revocations` 
- they now return `PublishRevocationsSchemaAnoncreds` instead of the old `PublishRevocations`
- Additionally, the `TxnOrPublishRevocationsResult` model, which was previously using `PublishRevocations`, is now also using `PublishRevocationsSchemaAnoncreds`, which seems to be an unnecessary change.

Now, the confusion set in when I noticed the Anoncreds Revocation API is using the old `PublishRevocations`, instead of the new `PublishRevocationsSchemaAnoncreds`. The naming suggests that this is an oversight.

The old model was modified to now take extra `options`, which the `PublishRevocationsSchemaAnoncreds` does not.
```py
class PublishRevocations(BaseModel):
    options: Optional[PublishRevocationsOptions] = None
    rrid2crid: Optional[Dict[str, List[Annotated[str, Field(strict=True)]]]] = Field(
        default=None, description="Credential revocation ids by revocation registry id"
    )
```

So, my guess is that `PublishRevocationsSchemaAnoncreds` is the model that should be used in the new AnoncredsRevocationApi, and that this is the model that should have the extra `options` field, not the old one. 

All in all, this isn't problematic for the internals of ACA-Py, because the models are identical, it's just a difference in names, with the extra options field. But, the model names propagate to the openapi spec, which is causing some confusion!

If my observations are correct that this is an oversight, then do let me know and I can contribute a quick renaming fix.

ff137:
Additionally, I notice that the old Revocation API also replaces `CredRevRecordResult` with `CredRevRecordResultSchemaAnoncreds`, in `get_revocation_status`. I assume this is also an oversight and that the Revocation API should use the same models as previous; that only the AnoncredsRevocationApi should be using these new models instead.

[...]

PR #2909, opened 2024-04-23
Fix api schema mixup in revocation routes

jamshale:
Fixes https://github.com/hyperledger/aries-cloudagent-python/issues/2901.

I somehow mixed up the schema naming between anoncreds revocation and existing revocation modules in a couple places. Shouldn't cause problems but still needs to be fixed.

Also a couple schemas existed in anoncreds that weren't used (copy and paste error)

##### ***Note***: 
This does remind me of something. The clear pending revocations method doesn't exist in the anoncreds implementation currently. Not sure how important that is and if it should be implemented for the upcoming 1.0.0 release ?

### DIFF
--- a/aries_cloudagent/revocation/routes.py
+++ b/aries_cloudagent/revocation/routes.py
@@ -270,7 +270,7 @@ class RevokeRequestSchema(CredRevRecordQueryStringSchema):
     )
 
 
-class PublishRevocationsSchemaAnoncreds(OpenAPISchema):
+class PublishRevocationsSchema(OpenAPISchema):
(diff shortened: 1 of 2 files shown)


dbluhm (approved):

ianco (approved):

jamshale:
Hmm. Got 1 fail `aries_cloudagent.anoncreds.revocation.AnonCredsRevocationError: Revocation registry not found`. Pretty sure this has happened before when the test von network got out of sync. I did pass before main got merged in.

Going to run it again. This error could be handled better. Shouldn't be a 500.

Issue #2913, opened 2024-04-24
Anoncreds library still referencing dev11 pre-release


loneil:
ACA-Py 0.12.0 is still using `anoncreds= { version = "0.2.0.dev11", optional = true, allow-prereleases = true }` as a dependency but anoncreds back in Feb looks to have a real release that ACA-Py could be referencing instead that came out shortly after that one.
https://pypi.org/project/anoncreds/#history

Doesn't look like much difference between 0.2.0dev11 and 0.2.0 proper.

dbluhm:
Fixed in #2917

PR #2914, opened 2024-04-24
Update AnonCreds delendency to 0.2.0

swcurran:
Per #2913 -- trying to update to AnonCreds 0.2.0 final.

Copied what was done in #2758 -- update the pyproject.toml file and poetry.lock.

There are a massive number of changes in the poetry.lock -- not sure about that.  Please confirm that the poetry.lock update is correct before approving. Definitely looks questionable to me...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ did-peer-4 = "^0.1.4"
 aries-askar= { version = "~0.3.0", optional = true }
 indy-credx= { version = "~1.1.1", optional = true }
 indy-vdr= { version = "~0.4.0", optional = true }
-anoncreds= { version = "0.2.0", optional = true, allow-prereleases = true }
+anoncreds= { version = "0.2.0", optional = true }
 
 # bbs
 ursa-bbs-signatures= { version = "~1.0.1", optional = true }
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -53,7 +53,7 @@ did-peer-4 = "^0.1.4"
 aries-askar= { version = "~0.3.0", optional = true }
 indy-credx= { version = "~1.1.1", optional = true }
 indy-vdr= { version = "~0.4.0", optional = true }
-anoncreds= { version = "0.2.0.dev11", optional = true, allow-prereleases = true }
+anoncreds= { version = "0.2.0", optional = true, allow-prereleases = true }
```
loneil, line 56: Can (someone correct me if I'm wrong...) I think remove the `allow-prereleases = true` part since 0.2.0 is a release. (I did that when testing plugins in Traction at least and it worked for me)

swcurran: Updated.  I debated doing that when changing and just left, but now removed.

loneil:
Lock file changes look like they make sense to me after a `poetry update`.

Some things (example `aiohttp="~3.9.4"`) are from compatibility set to major version, so would pull in 3.9.5 on an update

Some things are sub-dependencies in a version range that would allow an update
Ex:
image

jamshale:
The pyproject.toml looks right now.

I think you probably updated all the packages with `poetry install` or similar. Poetry is a little bit weird like this and will always try and upgrade anything it can with install.

To update just anoncreds you could do `poetry add anoncreds` or `poetry lock --no-update`. I think both should work.

swcurran:
I’ll close this and try again.

PR #2916, opened 2024-04-25
fix: rev notifications on publish pending

### LINKED ISSUE #2915: Regression(0.12.X): Revocation Notifications not getting emitted
It looks like #2558 introduced a bug where revocation notifications are no longer getting emitted when using the `/revocation/publish-revocations` endpoint. It seems to be fine when revocation is immediately published on revoke but if the issuer desires to batch them together and publish all pending revocations at once, no notifications are emitted to the holder connection.

This seems to be an issue in both 0.12.0 and 0.12.1rc0. Not sure why we didn't catch this one sooner :thinking:

dbluhm:
Fixes #2915.

Ensure revocation notifications are emitted when publishing batched pending revocations when _not_ using an external endorser to publish the transaction.

### DIFF
--- a/aries_cloudagent/revocation/manager.py
+++ b/aries_cloudagent/revocation/manager.py
@@ -256,6 +256,7 @@ class RevocationManager:
             if limit_crids:
                 crids = crids.intersection(limit_crids)
             if crids:
+                crids = list(crids)
                 (delta_json, failed_crids) = await issuer.revoke_credentials(
                     issuer_rr_rec.cred_def_id,
                     issuer_rr_rec.revoc_reg_id,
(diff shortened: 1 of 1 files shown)


aries_cloudagent/revocation/manager.py
```diff
@@ -294,6 +295,9 @@ async def publish_pending_revocations(
                         )
                     else:
                         rev_entry_resp = await issuer_rr_upd.send_entry(self._profile)
```
jamshale, line 297: Before I removed this it had a `if not connection_id:` statement on this. I think that's partly what confused me initially before I got more familiar with the flow.

jamshale (approved):

PR #2870, opened 2024-04-03
refactor: logging configs setup

amanji:
This PR adds some cleanup to the logging setup to make it easier to understand. Redundant tests have been removed.

PR #2918, opened 2024-04-25
fix: fixes a regression that requires a log file in multi-tenant mode

amanji:
This fixes a regression introduced by #2870 that required a log file to be present on file when using the default config setup for multi-tenant mode. The log file is only required if the user specifically sets the `--log-file` startup param.

### DIFF
--- a/aries_cloudagent/config/default_multitenant_logging_config.ini
+++ b/aries_cloudagent/config/default_multitenant_logging_config.ini
@@ -21,7 +21,7 @@ args = (sys.stderr,)
 class = logging.handlers.TimedRotatingFileMultiProcessHandler
 level = DEBUG
 formatter = formatter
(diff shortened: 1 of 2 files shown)


jamshale:
I don't see anything wrong with the changes made, but several integration tests are failing related to the logging config.

swcurran:
@amanji — are you on this?

amanji:
@swcurran Yes. Just running local tests

jamshale (approved):
Honestly, I'm still a bit confused by all this logging files and different configurations. I'm going to approve it though as you have detailed comments and looks like you've got a solid understanding.

amanji:
It has been just as perplexing to me having gone through the code a few times. There are likely some historical design decisions that I'm not aware of. I do think this can be distilled further down but probably best left for another refactor.

PR #2919, opened 2024-04-26
:arrow_up: Upgrade pydid (pydantic v2)

ff137:
Testing pydid upgrade to pydantic v2: https://github.com/Indicio-tech/pydid/pull/72

Edit: Great! All checks have passed. Marking as draft until pydid update is released


dbluhm:
@ff137 I think this should be ready for review now; would you be able to regenerate the lock file?

ff137:
@dbluhm Nice! I missed that 0.5.0 was released. I'll rebase 👍 

One thing to keep in mind is that the acapy-plugins will need to also be upgrade to pydantic v2, to ensure compatibility.

dbluhm:
@jamshale @amanji heads up on @ff137's point. Do you anticipate any challenges with the plugins you maintain?

I think there would be a conflict with the pickup plugin we maintain. We haven't done major updates to that plugin in a while though so maybe it's just time to retire it.

amanji:
I don't think it would affect DRPC

ff137:
FWIW the redis-events plugin uses pydantic v1, and we did an upgrade for that when testing this release. So I can contribute pydantic V2 for that one plugin. Can help with the others too, because I've memorised a lot of the migration steps :-)

jamshale:
Yes. I think some plugins will need to be updated. I know the oid4vci plugin has a pydantic dependency. 

We have the release management system merged for the plugin repo just now, so the upgrades could be prepared and then merged after the aca-py release.

jamshale:
This is why the release management should help. If we don't update a plugin and retire it. It is now tagged with 0.12.1 (All plugins are currently passing). It can just be left behind unless we chose to upgrade it. No need to remove it or anything.

dbluhm (approved):

PR #2926, opened 2024-05-01
0.12.1

swcurran:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,16 @@
 # Aries Cloud Agent Python Changelog
 
-## 0.12.1rc1
+## 0.12.1
 
 ### April 26, 2024
 
-Release 0.12.1rc1 is a small patch to cleanup some edge case issues in the handling of Out of Band invitations, revocation notification webhooks, and connection querying uncovered after the 0.12.0 release. Fixes and improvements were also made to the generation of ACA-Py's OpenAPI specifications.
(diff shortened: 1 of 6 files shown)


swcurran:
@jamshale -- could you please look at the integration test error? Presumably intermittent, but not one I recognize.

dbluhm (approved):
Testing with example scenarios in minimal example repo all look good: https://github.com/Indicio-tech/acapy-minimal-example/pull/88

Tests against Credo (run locally) also passed using the 0.12.1rc1 image.

LGTM! Assuming the integration test failure is indeed just an intermittent issue.

jamshale:
The integration test got a duplicate record error setting up endorser roles. I've never seen this before. I'd like to have a look at this closer and figure out if it's a blocking issue or not.

jamshale:
I'm going to look into that error from the integration tests and make a ticket. I'm sure it's intermittent and already existed, but still, it shouldn't be able to happen. 

Doesn't affect the release.

jamshale:
FYI on the failed integration test. Seems like, somehow the test created a race condition, by duplicating a /set-endorser-role request, with these lines.
```
7hAcme.agent | Updating endorser role for connection: d6283a4d-aa71-49b7-9e80-aa58f69cd8f7, TRANSACTION_ENDORSER
7hAcme.agent | Updating endorser role for connection: d6283a4d-aa71-49b7-9e80-aa58f69cd8f7, TRANSACTION_ENDORSER
```
creating a duplicate record. I tried to recreate the problem but haven't been able to. Because I've never seen this error before and seems very unlikely to happen on this endpoint in practice, I'd say we don't have to worry about it.

PR #2927, opened 2024-05-01
Update README.md

KPCOFGS:
Changed all relative paths to absolute paths for PyPI project description compatibility


swcurran:
Thanks for this!  We publish documents in a bunch of other places that I worry about, but hadn’t been thinking about PyPi.  Good changes!

swcurran (approved):

PR #2942, opened 2024-05-09
:construction_worker: Add pip to dependabot config

ff137:
Configures pip dependency management with dependabot

I've specified:
```yml
    schedule:
      interval: daily
    update-types:
      - "minor"
      - "major"
    commit-message:
      prefix: "chore(deps): "
    labels:
      - "dependencies"
      - "python"
```

Let me know if there are any desired changes.

`update-types` specifies to ignore patch releases, so it doesn't spam too much (and because ~ or ^ versioning already would include patch releases when lock file is updated)
Then, `commit-message: prefix` and `labels` is to match the existing dependabot PR style (which I assume is currently created manually)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
- # For details on how this file works refer to:
- #   - https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+# For details on how this file works refer to:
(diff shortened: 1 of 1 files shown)


WadeBarnes:
Hi @ff137, We have one of our team members working on updating the dependabot configurations too.  See his preliminary work for aca-py here; https://github.com/rajpalc7/aries-cloudagent-python/pull/1/files

ff137:
@WadeBarnes good stuff! Thanks, that one's much more comprehensive 👍

PR #2943, opened 2024-05-10
fix(interop): overly strict validation

dbluhm:
This change relaxes some overly strict validation that was preventing presenting proof using LDP-VC from Credo 0.5.X based agents.


jamshale (approved):

dbluhm:
cc @PatStLouis

dbluhm:
@PatStLouis The `id` properties changed are in the presentation submission, which is outlined in the presentation exchange specification as follows:

> The presentation_submission object MUST contain an id property. The value of this property MUST be a unique identifier, such as a [UUID](https://tools.ietf.org/html/rfc4122).
> The presentation_submission object MUST contain a definition_id property. The value of this property MUST be the id value of a valid [Presentation Definition](https://identity.foundation/presentation-exchange/#term:presentation-definition).

I think the only change made to the VC itself is loosening the requirements for the `LinkedDataProof` schema `domain` property

TimoGlastra (approved):

PatStLouis:
@dbluhm thanks for the clarification, I misread that this applied to the presentation itself!

Issue #2934, opened 2024-05-07
`Anoncreds` Revoking one credential of many of the same type fails proof


jamshale:
With anoncreds specifically. If `faber` issues multiple credentials to `alice` of the same type and then revokes any of the credentials, the next proof request will fail verification.

Steps (using demo):
- start faber `./run_demo run faber --wallet-type askar-anoncreds --revocation`
- start alice `./run_demo run alice --wallet-type askar-anoncreds`
- connect
- issue 3 credentials to alice
- revoke credential 2
- request proof
- result will be `verified = false`

[Screencast from 2024-05-07 09:39:59 AM.webm](https://github.com/hyperledger/aries-cloudagent-python/assets/31809382/29afa09f-16f4-4903-902c-1ac5860c2eb1)

swcurran:
Sounds like this is either a bug in ACA-Py’s selection of a default credential to use when there are multiple — should choose the most recent non-revoked one, or in Alice picking the wrong one to use from a set.

swcurran:
That said, I don’t know how “default” is defined by ACA-Py.  Presumably it is the first in the array of candidate VCs that could be used to satisfy the request.

PR #2946, opened 2024-05-13
Anoncreds - Send full registry list when getting revocation states

jamshale:
See https://github.com/hyperledger/aries-cloudagent-python/issues/2934.

So this ended up being a one line fix. The anoncreds legacy_indy implementation for the holder, was getting the revocation list (registry) from the ledger. Then it was removing index 0, which doesn't actually represent a credential but is expected by anoncreds-rs when getting the revocation state. See https://github.com/hyperledger/anoncreds-rs/issues/336.

This is a bit weird and will need to be remembered when other implementations get added.

I have tested with the demo and manually. I still want to make an integration test for this scenario.


swcurran (approved):
Looks good to me.  Crazy about this fix and the “1” start. I wonder how that decision got made, many years ago...

ianco (approved):

PR #2947, opened 2024-05-14
DOC: Verifiable Credential Data Integrity (VC-DI) Credentials in Aries Cloud Agent Python (ACA-Py)

kenechukwu-orjiene:


dbluhm:
@jamshale given that this is just a doc update, I'm guessing the integration test failure is a fluke? Any insights?

jamshale:
> @jamshale given that this is just a doc update, I'm guessing the integration test failure is a fluke? Any insights?

Yes. This is the one I've seen twice now. Somehow the integration test creates a race condition when setting the endorser's role. Very random.

docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,91 @@
...
+```jsonc
+{
+  "method": "key",
+  "options": {
+    "key_type": "ed25519"
+  }
+}
+```
+
+### Issue a Credential
```
ianco, line 89: I think this section needs a bit more description, an example of ow to issue the credential and a high-level description of how the process works.

(Likewise for the present proof, which I assume will come after the implementation is complete.)

ianco (changes requested):
Overall this is pretty high level, but I'll defer to @swcurran as to how much documentation is required.

I think there is more info required tho on the implementation, and a reference in the demo docs for how to select this credential type in the demo.

docs/features/W3cCredentials.md
kenechukwu-orjiene: Noted. I should include that soon

supersonicwisd1:
> Overall this is pretty high level, but I'll defer to @swcurran as to how much documentation is required.
> 
> I think there is more info required tho on the implementation, and a reference in the demo docs for how to select this credential type in the demo.

Yes, I should add the select between indy and vc_di soon. Thank for the review.

PR #2948, opened 2024-05-14
Example integration test issuing 2 credentials under the same schema

ianco:


jamshale (approved):
This was pretty much exactly what I was trying to do. Thanks for the example!

swcurran:
When we added this test, can we delete one that just does a single issueance, since this one covers that?  Nice to keep the time to run the suite shorter…. :-).  Not sure it is worth such an effort to eliminate duplicates, but the tests take a long time to run these days.

ianco:
> When we added this test, can we delete one that just does a single issueance, since this one covers that? Nice to keep the time to run the suite shorter…. :-). Not sure it is worth such an effort to eliminate duplicates, but the tests take a long time to run these days.

The run for each PR includes all tests tagged as `@GHA`, so we can control the scope of the PR tests with tags, rather than deleting tests.

For the new test I added I didn't include it in the `@GHA` scope, since it was just a demo for @jamshale .  I agree we should review the scope of the `@GHA` tests and make sure we don't have too much duplication.

swcurran:
Sounds great. Agreed!

PR #2950, opened 2024-05-14
CodeCov report on PR's

jamshale:
This will bring back `codecov` reporting on PR's and add comments like:

image

The warning only happened because I forced pushed.

The readme badge should also start working again.

There is a bit of setup for the repo admin. I'm hoping the covecov account that was used a year ago can be used again. The repo admin has to setup a token. It's really easy. The instructions are here https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage.

Note: The unit tests for this PR won't pass until the token is available. 
image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,7 @@
 comment:
-  layout: "header, diff"
+  layout: " diff, flags, files"
   behavior: default
-  require_changes: no
+  require_changes: false
+  require_base: false
+  require_head: true
+  hide_project_coverage: false
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -14,3 +14,4 @@ jobs:
     with:
       python-version: "3.9"
       os: "ubuntu-latest"
+    secrets: inherit
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -14,4 +14,3 @@ jobs:
     with:
       python-version: "3.9"
       os: "ubuntu-latest"
-    secrets: inherit
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,9 @@ on:
       os:
         required: true
         type: string
+    secrets:
+      CODECOV_TOKEN:
+        required: true
 
 jobs:
   tests:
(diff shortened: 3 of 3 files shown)


jamshale:
I'm not sure who might have access to the old codecov account for this repo?

WadeBarnes:
Seeing if I do.  @ryjones should

WadeBarnes:
We haven't used it in over a year:
image

WadeBarnes:
Looking at updating the token.

WadeBarnes:
Done, `CODECOV_TOKEN` token added.

WadeBarnes:
@jamshale, You might need to push an update to get the new secret to be picket up.  It does not exist in the context of the existing action runs, so simply re-running them has no effect.

WadeBarnes:
Doh, I know what the issue is.  PRs don't have access to the GitHub secrets (at least by default).

jamshale:
> Doh, I know what the issue is. PRs don't have access to the GitHub secrets (at least by default).

Ok. Ya I'm trying to figure out what to do. Didn't anticipate this problem.

jamshale:
Hmm. I tried tokenless upload with v3, which is supposed to work with open source repos. However, I got a 429 error due to rate limiting. I seems like a prevalent problem and the codecov team has an open issue they are working on.

https://github.com/codecov/feedback/issues/301

I'm not sure if it's possible to get the token working from a forked repo PR. I might try and run it a few more times and see if it works a decent amount of the time. Otherwise I'm not sure what to do, other than wait for them to fix it or possibly switch code coverage reporting providers.

jamshale:
I've researched the codecov github action repo quite a bit now and I think it's basically broken for our use case. So many people are reporting the same problems and I have no idea when they are going to fix it.

I'm going to figure out an alternative to switch to. Probably `coveralls`. Although i was having some trouble before with getting the commenting working.

WadeBarnes:
@jamshale, How about setting up ConarCloud to handle the code coverage as well?  It's already running and reporting on the PRs.  Additional setup details here; https://docs.sonarsource.com/sonarcloud/enriching/test-coverage/overview/

jamshale:
@WadeBarnes I'm still going to look at alternatives because I'm not a big fan of how codecov github actions work for our use case. However, their team did say that we need to install the app to our repo to get our own bandwidth for uploads, and I never mentioned that. https://github.com/codecov/codecov-action/issues/1431#issuecomment-2116233477.

Not sure if you want to try this of just continue finding a better option?

WadeBarnes:
@ryjones, Any opinions?

ryjones:
I think SonarCloud, if it handles what you need, is better.

PR #2955, opened 2024-05-16
Add OpenSSF Scorecard GHA - weekly

swcurran:
Copies the Fabric version of this GHA, with the only changes being removal of the Fabric copyright comment and the change of the day of the week the job runs.

Might be nice to run immediately and see where we are.

Once we have the scorecard, we can investigate what other updates we want to do to improve the score.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  workflow_dispatch:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '17 21 * * 4'
+  push:
+    branches: [ "main" ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@0864cf19026789058feabb7e87baa5f140aac736 # v2.3.1
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          # (Optional) "write" PAT token. Uncomment the `repo_token` line below if:
+          # - you want to enable the Branch-Protection check on a *public* repository, or
+          # - you are installing Scorecard on a *private* repository
+          # To create the PAT, follow the steps in https://github.com/ossf/scorecard-action?tab=readme-ov-file#authentication-with-fine-grained-pat-optional.
+          # repo_token: ${{ secrets.SCORECARD_TOKEN }}
+
+          # Public repositories:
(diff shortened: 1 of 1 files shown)


WadeBarnes (changes requested):
I'd like to see the actions updated to use the version tags rather than the commit hashes.  It will be easier for Dependabot to monitor for updates.

Otherwise LGTM

swcurran:
Checked the OpenSSF Repo and they are doing that as well — https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml

I’d say we leave it as this for now.

What would we change it to if we did change it?

WadeBarnes:
> Checked the OpenSSF Repo and they are doing that as well — https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml
> 
> I’d say we leave it as this for now.
> 
> What would we change it to if we did change it?

Example:
```
uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
=>
```
uses: actions/checkout@v4
```

swcurran:
Updated 3 out of 4. Didn’t know what to do with the other one…suggestions?

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,76 @@
...
+          #   - See https://github.com/ossf/scorecard-action#publishing-results.
+          # For private repositories:
+          #   - `publish_results` will always be set to `false`, regardless
+          #     of the value entered here.
+          publish_results: true
+
+      # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
+      # format to the repository Actions tab.
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@97a0fba1372883ab732affbe8f94b823f91727db # v3.pre.node20
```
WadeBarnes, line 65:
```suggestion
        uses: actions/upload-artifact@v4 # was v3.pre.node20 - 97a0fba1372883ab732affbe8f94b823f91727db
```

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      # Upload the results to GitHub's code scanning dashboard (optional).
+      # Commenting out will disable upload of results to your repo's Code Scanning dashboard
+      - name: "Upload to code-scanning"
+        uses: github/codeql-action/upload-sarif@3 # was v3.24.9 - 1b1aada464948af03b950897e5eb522f92603cc2
```
WadeBarnes, line 74:
```suggestion
        uses: github/codeql-action/upload-sarif@v3 # was v3.24.9 - 1b1aada464948af03b950897e5eb522f92603cc2
```

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,76 @@
...
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@4 # was v4.1.1 - b4ffde65f46336ab88eb53be808477a3936bae11
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@2 # was v2.3.1 - 0864cf19026789058feabb7e87baa5f140aac736
```
WadeBarnes, line 43:
```suggestion
        uses: ossf/scorecard-action@v2.3.3 # was v2.3.1 - 0864cf19026789058feabb7e87baa5f140aac736
```

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,76 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@4 # was v4.1.1 - b4ffde65f46336ab88eb53be808477a3936bae11
```
WadeBarnes, line 38:
```suggestion
        uses: actions/checkout@v4 # was v4.1.1 - b4ffde65f46336ab88eb53be808477a3936bae11
```

WadeBarnes (changes requested):

WadeBarnes:
The versions need to be specified as valid tags.

.github/workflows/scorecard.yml
WadeBarnes: This project only defines explicit tags.  For example the latest version tagged is `v2.3.3`, but unlike other projects they do not define a `v2` tag pointing to the same commit.

WadeBarnes: See [ossf/scorecard-action/tags]()https://github.com/ossf/scorecard-action/tags

swcurran:
Changes made -- should be good now. I hope.

Back to you @WadeBarnes !

WadeBarnes (approved):

PR #2959, opened 2024-05-21
DIDComm V2 Initial Implementation

TheTechmage:
The purpose of this PR is to add basic DIDComm V2 support to ACA-Py. It is our intention to add support gradually with small PRs, rather than one massive PR. To that end, at present, here's what we've added:

- Added `--experimental-didcomm-v2` flag to enable the DIDComm V2 code
- Added the ability to decrypt/unpack V2 messages
- When a message is received, we respond back with a DIDComm V1 problem report (there-by, providing a foundation to rely upon when adding protocols) and pack it the same way that we'd pack a DIDComm V2 message


dbluhm:
@frostyfrog Some minor conflicts in this PR after #2892 was merged

dbluhm:
Why not return a DIDComm v2 problem report?

aries_cloudagent/core/dispatcher.py
```diff
@@ -108,12 +108,56 @@ def queue_message(
...
+        # responder.connection_id = connection and connection.connection_id
+        error_result = ProblemReport(
+            description={
+                "en": str("No Handlers Found"),
+                "code": "message-parse-failure",
+            }
+        )
+        if inbound_message.receipt.thread_id:
+            error_result.assign_thread_id(inbound_message.receipt.thread_id)
+        logging.getLogger(__name__).debug("CONSTRUCTED V2 DISPATCHER RESPONSE")
```
dbluhm, line 156: Should use `self.logger`

aries_cloudagent/core/dispatcher.py
```diff
@@ -108,12 +108,56 @@ def queue_message(
...
+        )
+
+        context.injector.bind_instance(BaseResponder, responder)
+        # responder.connection_id = connection and connection.connection_id
+        error_result = ProblemReport(
+            description={
+                "en": str("No Handlers Found"),
+                "code": "message-parse-failure",
+            }
+        )
```
dbluhm, line 153: I think we should prepare a minimum viable DCv2 problem report here; I understand the motivation for reusing the v1 problem report but I think, for the sake of "correctness," it would be valuable to make sure a DIDComm v2 agent sending a message to ACA-Py is able to expect a DIDComm v2 message in return.

aries_cloudagent/connections/base_manager.py
dbluhm, line 1: Calling out a limitation: experimental v2 support will only work using did:peer:2 right now.

Should we implement it for did:peer:4 as well at this stage?

aries_cloudagent/messaging/v2_agent_message.py
dbluhm, line 1: Is this used? Perhaps we omit it for now until we have something more complete?

aries_cloudagent/transport/inbound/session.py
```diff
@@ -203,6 +203,7 @@ def process_inbound(self, message: InboundMessage):
     async def parse_inbound(self, payload_enc: Union[str, bytes]) -> InboundMessage:
         """Convert a message payload and to an inbound message."""
         session = await self.profile.session()
+        LOGGER.debug("INBOUND SESSION")
```
dbluhm, line 206: Should probably delete

aries_cloudagent/wallet/base.py
```diff
@@ -55,6 +56,34 @@ async def create_key(
...
+    async def get_key_by_kid(self, kid: str) -> KeyInfo:
+        """Fetch a key by looking up its kid.
+
+        Args:
+            kid: the key identifier
+
+        Returns:
+            The key identified by kid
+
+        """
```
dbluhm, line 85: Calling out a fundamental paradigm shift starting here for ACA-Py: to work with DCv2, these changes start using full DID URL references to verification methods to identify keys instead of the base58 encoding of the public key.

And a limitation: There is only one kid associated with a key. Does it make sense for a key to be looked up by multiple different key ids?

aries_cloudagent/transport/pack_format.py
```diff
@@ -218,3 +300,106 @@ def get_recipient_keys(self, message_body: Union[str, bytes]) -> List[str]:
...
+            except CryptoServiceError:
+                LOGGER.debug("Message unpack failed, falling back to JSON")
+            else:
+                # Set message_dict to be the dictionary that we unpacked
+                message_dict = message_unpack.message
+
+        if message_unpack.sender_kid:
+            receipt.sender_verkey = message_unpack.sender_kid.split("#")[0]
+            receipt.recipient_verkey = message_unpack.recipient_kid.split("#")[0]
+            pass
```
dbluhm, line 356: The `pass` is not needed

aries_cloudagent/transport/pack_format.py
```diff
@@ -218,3 +300,106 @@ def get_recipient_keys(self, message_body: Union[str, bytes]) -> List[str]:
...
+                message_unpack = await messaging.unpack(message_json)
+            except CryptoServiceError:
+                LOGGER.debug("Message unpack failed, falling back to JSON")
+            else:
+                # Set message_dict to be the dictionary that we unpacked
+                message_dict = message_unpack.message
+
+        if message_unpack.sender_kid:
+            receipt.sender_verkey = message_unpack.sender_kid.split("#")[0]
+            receipt.recipient_verkey = message_unpack.recipient_kid.split("#")[0]
```
dbluhm, line 355: Calling out another start of a paradigm shift: sender and receiver are identified by DIDs instead of by base58 encoded public keys.

Should the sender/receiver be a kid rather than a DID? Technically, as seen in the unpack result metadata, it is fundamentally a kid that the message is sent to and received by. Perhaps it would be better to prefer keeping that specificity around until it is no longer needed.

aries_cloudagent/transport/pack_format.py
```diff
@@ -218,3 +300,106 @@ def get_recipient_keys(self, message_body: Union[str, bytes]) -> List[str]:
...
+        Raises:
+            MessageEncodeError: If the message could not be encoded
+
+        """
+        messaging = session.inject(DIDCommMessaging)
+
+        if sender_key and recipient_keys:
+            message = await messaging.pack(
+                message=message_json,
+                to=recipient_keys[0],
```
dbluhm, line 399: Should the `pack` method of DMP that we're using here support multiple recipients?

dbluhm (commented):
Some of these comments are requested changes. Others are notes for other reviewers to pay attention to. Others still are open questions. Probably not all of them need to be answered before I would feel good about merging this PR.

That being said, I had a hand in these changes so I defer to other ACA-Py contributors for a full review.

Nice work to this point, @frostyfrog and @mepeltier!

aries_cloudagent/messaging/v2_agent_message.py
mepeltier: That is not used just yet; I think I accidentally included it in one of my commits 😅. I should be more careful with my `git add .`'s lol

aries_cloudagent/connections/base_manager.py
mepeltier: Given how early we are in the process, I don't think its _necessary_, but at the same time, it might be worth getting it in before our changes are more solidified. It'll also probably make it easier to switch off of did:peer:2, which is a good thing imo

aries_cloudagent/core/dispatcher.py
mepeltier: To be honest, that log can probably just get removed. At the very least it should be adjusted to be more consistent with other logs in ACA-Py. I'm pretty sure I put it in there in my early testing to see what parts of the code I was hitting

frostyfrog: Yep, this one can be removed

aries_cloudagent/transport/pack_format.py
frostyfrog: Correct. It was added as I was typing out the code and testing to prevent the code from failing. It can be removed 

One of the downsides of Python is the need for pass in empty blocks xP

aries_cloudagent/wallet/base.py
dbluhm: Now that I think on it, the way we implemented the interface, it doesn't actually preclude the ability to associated a key with multiple kids. But the implementation of the interface does impose this limitation.

aries_cloudagent/transport/pack_format.py
frostyfrog: It probably should. Another problem that I ran into is that it implicitly forward wraps the message, even if we are responding on the same connection. Something I planned to look at later

aries_cloudagent/connections/base_manager.py
frostyfrog: I do believe that it's not limited to `did:peer:2`. Other `did:peer` methods should be supported as well. Once I have something better to key off of for outbound, things like `did:web` should work (in theory). Though you are right in that all `did` methods, aside from `did:peer:2`, are currently untested

dbluhm:
Most of the feedback so far has been from those who have already had a hand in the PR; it would be great to get more feedback from outside that group.

The ability to assign a kid to a key has been identified as something we would like to see before we lose some of the old JSON-LD endpoints. This PR adds that ability incidentally as this same support is required for efficient key lookup when using kids in DIDComm v2. If we end up having issues that prevent this PR from going forward, we should consider breaking that component out.

cc @PatStLouis

jamshale:
I can't add any good comments about the approach other than it like it's been done well as a non intrusive way to add the changes and the code generally looks really good. This isn't a strong area for me.

We will need to fix the unit testing and add some coverage. Also, a couple integration tests would be good to have setup before we got this merged into v1.0.0. Could probably be done as another task if there's been good manual testing.

frostyfrog:
> I can't add any good comments about the approach other than it like it's been done well as a non intrusive way to add the changes and the code generally looks really good. This isn't a strong area for me.
Thank you for your feedback! I was worried that we weren't getting any eyes on the changes

> We will need to fix the unit testing and add some coverage. Also, a couple integration tests would be good to have setup before we got this merged into v1.0.0. Could probably be done as another task if there's been good manual testing.

In regards to testing, I used the docker-compose within [this repo](https://github.com/frostyfrog/didcomm-v2-test-util) of mine. It uses the same underlying `didcomm-messaging` library to send and receive DIDComm V2 messages and has been used in conjunction with the [DIDComm Demo](https://demo.didcomm.org) hosted by the DIF. This has allowed us to quickly iterate upon changes and verify whether or not they work. I'm not quite sure how ACA-Py's unit tests are structured at the affected layers of the stack, but the [test script](https://github.com/frostyfrog/didcomm-v2-test-util/blob/main/src/__main__.py) itself is small and easy enough to follow. I hope that clears up the concerns.

The new code should be gated under the experimental flag and the changes were designed to allow for a series of many PRs, rather than one large PR with everything done all at once. This being the first PR of many.

jamshale:
> The new code should be gated under the experimental flag and the changes were designed to allow for a series of many PRs, rather than one large PR with everything done all at once. This being the first PR of many.

Sounds good. We'll need to at least get the unit tests passing and then we can consider getting this merged as a experimental feature.

jamshale:
I think this is getting pretty close. I'm not opposed to merging it with the unit testing and integration testing coming later.

Sonarcloud flagged a few minor things that should be addressed. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=2959&id=hyperledger_aries-cloudagent-python

jamshale:
> It feels odd to me to define a constant for error text. I'm willing to be convinced otherwise, but it feels like it's not really an issue?

I'm not really concerned about these. We might have to change sonarcloud settings to get this passing. We haven't really done this anywhere in the project. However, when we have the same error message in many places we should have a constant for it. It's just good practice to not need to find and replace all if we ever wanted to change it.

PatStLouis:
I finally got to give this a look.

@frostyfrog when building the image and running with the flag `--experimental-didcomm-v2` I get the error `ModuleNotFoundError: No module named 'didcomm_messaging'`. I can see it's in the poetry lock file so it should be installed, any idea why I get this behavior? Am I missing something?

To build the image I checked out the `feat/didcommv2-proof-of-concept` branch.

Full error track:
```
agent-1  | 2024-06-09 16:38:14,894 aries_cloudagent.commands.start ERROR Exception during startup:
agent-1  | Traceback (most recent call last):
agent-1  |   File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/commands/start.py", line 72, in init
agent-1  |     await startup
agent-1  |   File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/commands/start.py", line 28, in start_app
agent-1  |     await conductor.setup()
agent-1  |   File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/core/conductor.py", line 127, in setup
agent-1  |     context = await self.context_builder.build_context()
agent-1  |   File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/config/default_context.py", line 69, in build_context
agent-1  |     from didcomm_messaging import (
agent-1  | ModuleNotFoundError: No module named 'didcomm_messaging'
```

frostyfrog:
Did you install it with the didcommv2 extra (similar to installing with Askari)? This is how it's done for my test docker container: https://github.com/frostyfrog/didcomm-v2-test-util/blob/main/Dockerfile.acapy

mepeltier:
Yeah, you'll need to run something like `poetry install --all-extras` or `poetry install --extras "didcommv2 askar"` (I think the `askar` extra is needed for this branch, but I'm not actually 100% sure if that's the case)

frostyfrog:
Wait, @mepeltier, wasn't your custom import error supposed to trigger to talk about the extra, or am I thinking of something else?

PatStLouis:
@frostyfrog I'm using the [dockerfile from the project](https://github.com/Indicio-tech/aries-cloudagent-python/blob/feat/didcommv2-proof-of-concept/docker/Dockerfile), I'll give these solutions a try.

frostyfrog:
@PatStLouis looking at that Dockerfile, it looks like there's a build argument called: `ARG acapy_reqs=[askar,bbs]`. If this is changed to include there DIDComm extra (I think it was called `didcommv2`. It was mentioned earlier), then you should be golden!

frostyfrog:
It was discussed in Today's ACA-Pug call to add the didcommv2 extra to the docker file, so I'm doing that now.

dbluhm:
@mepeltier is in the process of putting together tests to pass the code coverage threshold SonarCloud is requiring.

jamshale:
FYI. I'm pretty sure we can actually merge this with the sonarcloud check failing. So if we think there is adequate test coverage below 80% we should be able to go ahead.

dbluhm (approved):
I'm happy with the state of these changes. More work to come!

dbluhm:
Gentle nudge for a review from @jamshale (so we don't have to keep merging upstream and resolving poetry lock conflicts :sweat_smile:). I think the 61% coverage is adequate for the nature of these changes. Uncovered code is mostly configuration only used when the experimental flag is enabled.

jamshale:
Day off in Canada. Will prioritize this for tomorrow morning.

jamshale (dismissed):
I haven't studied what was needed for the didcommv2 messaging, but the code looks well tested, formatted and commented :+1: 

I'm going to approve and merge because this is an experimental feature and @dbluhm approved it

TheTechmage:
The conflicts are always occurring with the poetry lock. Let me re-lock it to "clear" the conflict (it's been happening for a while)

TheTechmage:
Or, Micah will beat me to the punch! xD Should be good to go

jamshale (approved):

PR #2964, opened 2024-05-22
Sonarcloud with code coverage

jamshale:
This will add code coverage and duplication reporting to the sonarcloud comment and have details in the sonarcloud UI.

image

The admin of the sonarcloud account need to do 2 things for the project. 

1. Turn off automatic analysis. The upload needs to happen after the unit tests complete and doing it automatically isn't supported yet.
image

image

The github action will initialize the analysis now and should look like the image.

2. Create a token in my account --> security, and add it to the repo's secrets called `SONAR_TOKEN`

image

image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,0 @@
-comment:
-  layout: "header, diff"
-  behavior: default
-  require_changes: no
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -11,6 +11,7 @@ jobs:
   tests:
     name: Tests
     uses: ./.github/workflows/tests.yml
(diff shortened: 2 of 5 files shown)


WadeBarnes:
@jamshale, Is the `SONAR_TOKEN` referenced in the workflow that will be running for a PR?  If so are we going to run into the same issue with the secrets not being available for PRs from external forks?

jamshale:
> @jamshale, Is the `SONAR_TOKEN` referenced in the workflow that will be running for a PR? If so are we going to run into the same issue with the secrets not being available for PRs from external forks?

Ugh. Of course. That was stupid. I'm looking at a solution on their forum. I'll see if I can get it working.

PR #2966, opened 2024-05-22
Sonarcloud with code coverage

jamshale:
This is an attempt at getting code coverage working with sonarcloud with github actions from forked repos. It works by doing the testing and then uploading the results and the PR number from the pull request. The sonar workflow then fires after the testing workflow has finished where it should now have access to the token. It will download the test results and checkout the pull request branch from the forked repo to do the scan.

I can't be sure everything going to work yet because this still requires the token in secrets and for automatic analysis to be disabled. See https://github.com/hyperledger/aries-cloudagent-python/pull/2964.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,0 @@
-comment:
-  layout: "header, diff"
-  behavior: default
-  require_changes: no
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,54 @@
+name: Sonar Scan and Coverage
+
+on:
+  workflow_run:
+    workflows: [ PR Tests ]
+    types: 
+      - completed
+
+jobs:
+  SonarCloud:
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success'
+    steps:
+      - uses: actions/checkout@v3
+        with:
+            fetch-depth: 0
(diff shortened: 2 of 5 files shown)


WadeBarnes:
There is a `SONAR_TOKEN` registered already.  Let me know if it works.
image

jamshale:
> There is a `SONAR_TOKEN` registered already. Let me know if it works. 

The other thing that needs to happen is to disable the automatic analysis. We won't get any sonarcloud scanning after it's disabled until this gets merged. But we can just turn it back on.

WadeBarnes:
> > There is a `SONAR_TOKEN` registered already. Let me know if it works.
> 
> The other thing that needs to happen is to disable the automatic analysis. We won't get any sonarcloud scanning after it's disabled until this gets merged. But we can just turn it back on.

Done

jamshale:
This is a bit tricky. I don't think the `sonar` workflow will exist until after it's merged :/

I wonder if I can try and fork my fork to test this better before trying to merge it.

We can probably re-enable automatic analysis when I try and think of something.

WadeBarnes:
I've re-enabled the automated analysis for now.

jamshale:
> I've re-enabled the automated analysis for now.

Thanks. I'm using 2 accounts to test this. Will reopen when I feel confident about it.

PR #2989, opened 2024-05-27
Enable `no-transport` mode as startup parameter

PatStLouis:
This PR adds a `--no-transport` startup parameter for making the agent start without requiring the `--enpoint`, `--inbound-transport` and `--outbound-transport` to be set.

The idea was shared during an aca-pug meeting and there was no objection. This can greatly simplify deployment for controllers that do no require didcomm messaging.

Currently it just overrides what is required for startup.

@dbluhm @swcurran @ianco


PatStLouis:
I will have to review this looks like some other parameters got mixed up in my PR

PR #2991, opened 2024-05-27
Add `proof_key` issuance option

PatStLouis:
Following the discussions in PR #2348, here's my suggestion to add an optional field to specify a verkey to use when querying the wallet for the `didInfo` used in issuing credentials.

This would satisfy the requirements raised by @Jsyro and some of the work we are doing around binding `did:web` to a `did:indy`.

Please review and share some toughts @dbluhm @swcurran @jamshale @ianco @TimoGlastra .

This will also preserve this enabling feature from the deprecated `/jsonld/sign` endpoint, giving controllers flexibility while using other did methods.

### DIFF
--- a/aries_cloudagent/messaging/valid.py
+++ b/aries_cloudagent/messaging/valid.py
@@ -17,6 +17,7 @@ from .util import epoch_to_str
 B58 = alphabet if isinstance(alphabet, str) else alphabet.decode("ascii")
 
 EXAMPLE_TIMESTAMP = 1640995199  # 2021-12-31 23:59:59Z
+VERKEY_EXAMPLE = "3Dn1SJNPaCXcvvJvSbsFWP2xaCjMom3can8CQNhWrTRx"
 
 
 class StrOrDictField(Field):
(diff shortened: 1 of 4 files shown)


dbluhm:
As discussed in the ACA-PUG call today, we're looking at adding similar functionality by making it possible to tag keys with their verification method IDs. This was something we needed in #2959 as well. We will work with @PatStLouis to make sure the required interface changes can make it into ACA-Py regardless of the ultimate fate of #2959 (if needed, we can break them out of that PR and open it separately).

dbluhm:
@PatStLouis the DCv2 PR was just merged; I think there are some minor updates needed to make it so we can associate multiple kids with a key but otherwise the changes made their should support this use case. I propose we close this PR and create an issue to track that.

dbluhm:
Issue opened; @PatStLouis if there's good reason to have this PR open, feel free to reopen.

PatStLouis:
@dbluhm that sounds good to me, thanks!

Issue #2962, opened 2024-05-21
🐛 presenting a revocable credential, when verifier requested "non_revoked" before rev registry was created, leads to TypeError


ff137:
Scenario:
- a verifier posts an IndyProofRequest to a holder using `/present-proof-2.0/send-request`
  - the verifier's proof request specifies: `non_revoked: to: <any timestamp from distant past>`
- the prover responds to this presentation (by posting to `/present-proof-2.0/records/{pres_ex_id}/send-presentation`)
  - the prover presents a cred_id pointing to a newly minted (non-revoked) revocable credential.
- through some mightily difficult-to-read code in `IndyPresExchHandler.return_presentation`, since `non_revoked.to` was included in the request, we end up calling the "revocation registry delta" (in this case from the indy_vdr ledger).
- the response we get from the ledger does not contain some expected keys, and leads to a TypeError:
```sh
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/asyncio/tasks.py", line 256, in __step
    result = coro.send(None)
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/present_proof/v2_0/routes.py", line 1199, in present_proof_send_presentation
    pres_ex_record, pres_message = await pres_manager.create_pres(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/present_proof/v2_0/manager.py", line 277, in create_pres
    pres_tuple = await pres_exch_format.handler(self._profile).create_pres(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/present_proof/v2_0/formats/indy/handler.py", line 162, in create_pres
    indy_proof = await indy_handler.return_presentation(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/present_proof/indy/pres_exch_handler.py", line 155, in return_presentation
    (delta, delta_timestamp) = await ledger.get_revoc_reg_delta(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/ledger/indy_vdr.py", line 1056, in get_revoc_reg_delta
	"accum": response_value["accum_to"]["value"]["accum"],
TypeError: 'NoneType' object is not subscriptable
```

Edit: Occurs in 0.11 as well as latest 0.12.1

[...]

PR #2995, opened 2024-05-27
Fix issue with requested to revoke before registry creation

jamshale:
This fixes https://github.com/hyperledger/aries-cloudagent-python/issues/2962 by fetching the registry list start time and then using it for the revoke_to value and recalculating the delta, when the `accum_to` value is missing from the ledger response. This means that if the presentation request has a revoke_to value before the registry was created that the holder will use the creation time to calculate the delta.

There is some conversation about the desired behaviour in the ticket. I'm not certain this is the way it should be fixed, but my limited understand and testing appears like it is valid.

### DIFF
--- a/aries_cloudagent/ledger/indy_vdr.py
+++ b/aries_cloudagent/ledger/indy_vdr.py
@@ -1092,6 +1092,29 @@ class IndyVdrLedger(BaseLedger):
             ) from err
 
         response_value = response["data"]["value"]
+        accum_to = response_value.get("accum_to")
+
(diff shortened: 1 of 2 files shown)


swcurran:
@cvarjao — this might be the source of the issue with some Lawyers being unable to produce a valid revocation proof. If this same issue is happening in Credo, the result would be the “accum_to” being “NONE”, which IIRC was the error that was coming up.  Further, if the lawyer got a new credential, their existing credential would be revoked, triggering an update of the RevReg, which might in turn would make the problem go away.

dbluhm (approved):

Issue #2952, opened 2024-05-15
Manage integration test workflows


jamshale:
The integration tests on PR currently take a very long time (~ 1hr). We should have a better idea about what tests have have overall, which ones need to be run on PR's and which ones instead, could possibly be run on a schedule and on releases. This would save resources and time for developers, while still catching the same amount of problems.

1. Map out the tests and get feedback on the tests required for PR's.
2. Adjust workflows.

PR #2996, opened 2024-05-27
Manage integration tests with GitHub Actions (#2952)

jamshale:


ianco (approved):
Looks good to me.  I agree we can probably scope down the `@PR` tests even further but this is a good start, gives us the framework, and we can always move tags around on these tests later on ...

jamshale:
@ianco I rebased to fix a merge conflict and need re-approval, please.

ianco (approved):

Issue #2777, opened 2024-02-13
"Cannot re-enter scope: profile:" errors in plugins using ACA-Py 0.12.0rc0 (from injection_context.py)


loneil:
I've run into an issue in a couple different contexts when testing out using the 0.12.0rc0 ACA-Py image and libraries.

I'm wondering if there's some relation to the changes made here? https://github.com/hyperledger/aries-cloudagent-python/pull/2705
Either with multitenancy? Or plugin compatibility? I'm not really sure the root cause...

**1. Building and testing multitenant_provider plugin**
This is using the plugin from https://github.com/hyperledger/aries-acapy-plugins and the devcontainer and startup/testing  instructions there.
Update pyproject aca-py version to 0.12.0rc0 and build in devcontainer.
Unit tests pass
All integration tests failing with 
```
Failed: Failed to create wallet: b'400: Cannot re-enter scope: profile:multitenant_sub_wallet.'
```
1 example:
![image](https://github.com/hyperledger/aries-cloudagent-python/assets/17445138/7abae6e2-5c01-4b31-a7d6-e632f1eb26b3)

**2. Updating Traction to 0.12.0rc and running it**
Get errors when trying to do anything that creates or accesses a wallet in 
Traction Innkeeper plugin
Multitenant provider plugin

Creating wallet on fresh start (Innkeeper plugin):
```
2024-02-09 13:44:34 Message: "Error creating wallet ('traction_innkeeper')."
2024-02-09 13:44:34 Arguments: (InjectionContextError('Cannot re-enter scope: profile:askar-wallet'),)
```

Accessing a wallet (Multitenant provider):
```
raise InjectionContextError("Cannot re-enter scope: {}".format(scope_name))
aries_cloudagent.config.injection_context.InjectionContextError: Cannot re-enter scope: profile:multitenant_sub_wallet
```

Drilling in more and I find the root cause in 
```
2024-02-09 13:44:34   File "/home/aries/.venv/lib/python3.9/site-packages/aries_cloudagent/core/profile.py", line 40, in __init__
2024-02-09 13:44:34     self._context = context.start_scope(scope)
2024-02-09 13:44:34   File "/home/aries/.venv/lib/python3.9/site-packages/aries_cloudagent/config/injection_context.py", line 82, in start_scope
2024-02-09 13:44:34     raise InjectionContextError("Cannot re-enter scope: {}".format(scope_name))
```

[...]

PR #2997, opened 2024-05-29
feat: inject profile and session

dbluhm:
This is yet another attempt at #2705 which was reverted in #2789 because of issues reported in #2777. This time, I have explicitly tested using the integration tests in the Multitenant provider plugin to ensure the same issue does not arise.

The same benefits mentioned in #2705 of course apply here. This makes it easier to provide pluggable components that depend on a Profile without the interface of those components accepting a profile in every method.

The primary difference between this PR and #2705 is that this ensures a unique scope based on wallet id of multi-tenant subwallets.

I have also tacked on injecting the session. This enables binding dependencies that require a session from plugins and other components. For instance, the DIDComm v2 work in #2959 could use this to bind a provider for the `SecretsAdapter` needed to provide access to Askar to the pack and unpack methods.

### DIFF
--- a/aries_cloudagent/admin/request_context.py
+++ b/aries_cloudagent/admin/request_context.py
@@ -21,13 +21,13 @@ class AdminRequestContext:
         self,
         profile: Profile,
         *,
-        context: InjectionContext = None,
-        settings: Mapping[str, object] = None,
-        root_profile: Profile = None,
-        metadata: dict = None
+        context: Optional[InjectionContext] = None,
+        settings: Optional[Mapping[str, object]] = None,
+        root_profile: Optional[Profile] = None,
+        metadata: Optional[dict] = None
     ):
         """Initialize an instance of AdminRequestContext."""
-        self._context = (context or profile.context).start_scope("admin", settings)
+        self._context = (context or profile.context).start_scope(settings)
         self._profile = profile
         self._root_profile = root_profile
         self._metadata = metadata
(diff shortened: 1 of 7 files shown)


jamshale (dismissed):
Looks good to me. I'd like to look at it closer for my own understanding, but multitenancy was the only problem last time. 

This should help with unit testing, too?

dbluhm:
> This should help with unit testing, too?

@jamshale I hadn't thought about that but this could be used to inject test/mock dependencies in unit tests. We had strategies to do this already though so I'm not sure it changes too much. If we took advantage of this in more places, mocking dependencies would be simpler but that would require transitioning more places to using the injected profile.

dbluhm:
I've still got issues with entering a scope that isn't unique in the integration tests... I will take a deeper look. I'm beginning to question the value of the named scopes. ContextVars, for instance, don't name the contexts and require them to be globally unique. I'll look into that angle.

dbluhm:
I removed named scopes entirely. They served no significant purpose in practice and they were causing all sorts of trouble.

jamshale (approved):
I'm a bit confused why there was named scopes to begin with? If they didn't really do anything. But, I don't know the history of the project to have any opinion on it. It looks like a good change and that you know what you're doing, so I'm going to approve.

dbluhm:
@jamshale Yeah, the tests implied that there was some perceived use to the named scopes at some point in the past but it looks like in practice they were just never really used.

You could "recall" a particular scope by name and retrieve values out of it. But this would only ever be possible from a parent scope and it's just not common practice to, for example, obtain the bound values of a session from the profile scope that spawned it. That would probably be a bad idea, in fact lol.

In practice, the lifetimes of the InjectionContexts and references to the contexts are, I think, appropriately narrow and naturally fall out of (lexical) scope at the right times. So we don't need to access the scopes by name since we're using the references to them as bound to a profile or session, etc.

PR #3010, opened 2024-06-03
feat: soft binding for plugin flexibility

dbluhm:
This PR introduces a way to bind providers in profiles in such a way that they may be overridden by a plugin.

I have a fairly pointed use case: my goal is to be able to provide an alternate implementation of VC Holder. To see an example of this in use, check out this demo I put together: https://github.com/dbluhm/acapy-vc-holder. This demo uses an external KMS, delegating secure storage and retrieval of received LDP-VCs to an external component.

The "soft-binding" functionality has so far only been applied to the binding of a VC Holder instance.

An alternative to this approach would be to move the VC Holder provider binding code outside of the profile. This is possible now that the profile itself is bound to the context after #2997. I think this would create a pattern where we have dynamic providers based on wallet type. I don't think that's necessarily a bad thing but it moves logic specific to a wallet type out of the Profile definition for the wallet type. This seemed less than desirable. However, making a special case binding mechanism also feels a bit less than desirable, though it was straightforward to implement.

I'm open to feedback on this one.


jamshale:
FYI. The code coverage check is 80%. I think this is the default from sonarcloud. We can adjust it if we feel it's too high.

ianco (commented):
Looks good to me.

jamshale (approved):

PR #3033, opened 2024-06-14
:sparkles: Add pagination support for listing Connection, Cred Ex, and Pres Ex records

ff137:
Marking as draft to first confirm functionality works as expected

Edit: Success! I've written a local test to assert that limit/offset does indeed work as expected, and that unique records are returned across pages. 
___

This adds `limit` and `offset` query parameters to:
- listing connection records (in v1 connections API)
- listing credential exchange records (in v1 and v2 issue_credentials API)
- listing presentation exchange records (in v1 and v2 present_proof API)

As with #3000: the default behavior is now changed to no longer attempt to fetch all records for these endpoints. A default page size of 100 is used, with a maximum allowable page size of 10'000.

As you can see, adding the functionality is as simple as extending `PaginatedQuerySchema` for the query parameter schemas, and then to read limit and offset, and passing it to the Records.query method.

There are probably more endpoints where this functionality can be added, so I'm open for suggestions to apply the same elsewhere, but it should be straight forward enough for other contributors to add too.

These are just what I believe to be the essential endpoints, such that issuers/verifiers listing connections or exchange records won't try to read potentially millions of records.

### DIFF
--- a/aries_cloudagent/messaging/models/paginated_query.py
+++ b/aries_cloudagent/messaging/models/paginated_query.py
@@ -1,5 +1,8 @@
 """Class for paginated query parameters."""
 
+from typing import Tuple
+
+from aiohttp.web import BaseRequest
 from marshmallow import fields
 
 from ...messaging.models.openapi import OpenAPISchema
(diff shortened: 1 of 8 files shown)


ff137:
The PR Tests passed, but something went wrong with artifact upload. Will empty commit to re-run

jamshale (approved):
This might possibly cause issues with existing UI implementations like https://github.com/bcgov/traction. Should make sure there's something about it in the release notes.

jamshale:
I guess sonarcloud is complaining because the `int(request.query.get("limit", DEFAULT_PAGE_SIZE))` could be a helper function. I don't have a problem with it the way it is. 

We could see if we can merge it without changing sonarcloud settings.

jamshale:
I tried to update the pr branch with main. Hopefully didn't mess anything up.

ff137:
@jamshale my fork had diverged with extra commits. Managed to resolve it now. And I was gonna post a comment about how deduplication is sometimes just tedious, but then I figured to just do it 😆 so included a helper method

jamshale (approved):
It is a bit nicer with the helper method :+1:

Issue #3030, opened 2024-06-12
Startup can fail with certain multi-tenancy subwallets [Multi-tenancy]


jamshale:
There was an issue reported here https://github.com/hyperledger/aries-cloudagent-python/pull/2922#pullrequestreview-2112805554 with the anoncreds upgrade feature. For some reason it is unable to query the subwallet and/or create to subwallet profile.

It needs to handle this scenario correctly. 

I have tried to reproduce with sqlite DB and normal sub-wallet and base-wallet setting but have been unable to reproduce the issue so far.

ff137:
Thank you! I'll try modify provision=True and test with that... Will report back tomorrow 👌

ff137:
I've tested just with the simple change setting provision to True, but we still get the same error:
```sh
Traceback (most recent call last):
aries_askar.error.AskarError: Backend error
Caused by: error returned from database: Request returned an error: database "a8f2692b4e4b4312a9ae696d76643362" does not exist.
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/commands/start.py", line 72, in init
    await startup
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/commands/start.py", line 29, in start_app
    await conductor.start()
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/core/conductor.py", line 522, in start
    await self.check_for_wallet_upgrades_in_progress()
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/core/conductor.py", line 846, in check_for_wallet_upgrades_in_progress
    subwallet_profiles = await get_subwallet_profiles_from_storage(
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/utils/profiles.py", line 54, in get_subwallet_profiles_from_storage
    await MultitenantManager(root_profile).get_wallet_profile(
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/multitenant/manager.py", line 85, in get_wallet_profile
    profile, _ = await wallet_config(context, provision=provision)
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/config/wallet.py", line 51, in wallet_config
    profile = await mgr.provision(context, profile_cfg)
File "/usr/local/lib/python3.9/site-packages/aries_cloudagent/askar/profile.py", line 293, in provision
    opened = await store_config.open_store(provision=True)
[...]

PR #3048, opened 2024-06-19
:adhesive_bandage: add exception handling to wallet-upgrade check

ff137:
Simple fix for non-critical issue.

Currently, if something goes wrong with the `check_for_wallet_upgrades_in_progress` method, then startup will be aborted. This PR just adds a try-except block, such that the exception is logged and startup can proceed.

Relates to #3030

### DIFF
--- a/aries_cloudagent/core/conductor.py
+++ b/aries_cloudagent/core/conductor.py
@@ -519,7 +519,12 @@ class Conductor:
             except Exception:
                 LOGGER.exception("Error accepting mediation invitation")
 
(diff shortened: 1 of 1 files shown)


jamshale (approved):

jamshale:
I'm still trying to wrap my head around how this problem is happening. But, this is an easy fix to get around the problem. Very unlikely the problem and this functionality would conflict. Especially for this upgrade which should happen in a few seconds maximum.

PR #3049, opened 2024-06-19
Handle failed tails server issuance [Anoncreds]

jamshale:
- Handles the scenario where the tails server isn't available during revocation rotation.
- Fixes a bug with the openapi verse expected value for `max_cred_num`
- Refactors the create_credential function
- Removes some TODO's. Most with documentation that I think is good enough.

Keeps the revocation list in a failed state if the revocation registry has never been uploaded to the tails server. During issuance it will check for a failed state and try to upload and finish the revocation list. Issuance will not occur until a successful tails server and connection occurs.

Thought about keeping the revocation registry definition itself in a failed state but seemed to make sense to have it this way so the registry will still handle the full registry scenario.


ianco (approved):
Excellent!

Issue #3020, opened 2024-06-06
Interoperability 0.12.1 and credo-ts 0.5.3 - OOB connections


jamshale:
It's been reported that credo-ts version 0.5.3 expects the `did-rotate` attachment to be present oob connections. I'm not sure if credo-ts should not throw an error when the `did_rotate~attach` attachment is missing or acapy should have the `did_rotate~attach`.

We need to either let credo-ts know to fix it on their side or fix it on our side because currently the versions don't work for oob connections.

See https://discord.com/channels/905194001349627914/1244928900341960746 for more context.

kukgini:
I see that `did_rotate` decorator (`did_rotate~attach`) is optional in the RFC.
https://github.com/hyperledger/aries-rfcs/blob/main/features/0023-did-exchange/README.md

TimoGlastra:
Maybe there's a bug, but it should only require the `did_rotate` if the did is diferent than `did:peer:1`. So if the did in the response is `did:peer:1` we try to extract it from the `did_doc~attach`, and otherwise we verify the signature on the `did_rotate~attach`.

but it is correct that we REQUIRE either one to be present to verify the signature.

Which did method are you using?

kukgini:
@jamshale could you confirm which `did:peer` is used? If `did:peer:2` or more is used then `aca-py` should attach did_rotate as @TimoGlastra said.

jamshale:
@amanji @dbluhm Do you have any insight into this issue? I don't really have an interop setup atm, to look into this better. I'm not entirely sure what settings is causing the problem, but it's been reported by two separate people.

kukgini:
@TimoGlastra I captured aca-py log of `didexchange` response. It seem's like `did:peer:4` is used and there is `did_rotate~attach`. Note that this response is for two aca-py agent.

`{
  "@type": "https://didcomm.org/didexchange/1.1/response",
  "@id": "94c5b4ba-4f9d-42e5-a8e4-c5b604c3a210",
  "~thread": {
    "thid": "bd1ca149-7573-42a5-beed-86c47bac81ea",
    "pthid": "d8446259-64f8-4dfd-99c0-31c1bff589b4"
  },
[...]

PR #3050, opened 2024-06-19
fix: respond to did:peer:1 with did:peer:4

dbluhm:
This adjusts the default did method ACA-Py will respond to did:peer:1 with to did:peer:4. This should help the situation reported in #3020.

Marked as draft until I have a chance to test against Credo.


dbluhm:
I tested against Credo and added a quick unit test. All looked good and worked as expected. This should resolve #3020.

jamshale (approved):

PR #3056, opened 2024-06-24
Add vcdm 2.0 issuance/verification

PatStLouis:
Initial attempt at adding VCDM  2.0 issuance with `ED25519Signature2020` proof type.

Will add a couple test cases.

What are opinions regarding if the credential models/schemas should be bundled into the same or have 2 distinct models?

I went with the latter.


jamshale:
@PatStLouis Is this going to be rebased and worked on? Or can we close the PR?

PatStLouis:
@jamshale I'm closing thiss, I will open a new PR with the latest work. This is outdated.

PR #3057, opened 2024-06-24
fix: didexchange manager not checking the did-rotate content correctly

gmulhearn-anonyome:
minor error i noticed while browsing acapy code whilst working on aries-vcx: https://github.com/hyperledger/aries-vcx/issues/1226.

if statement was not correcting the correct data, resulting in it always being `true`


jamshale (approved):

dbluhm (approved):
Good catch

PR #3067, opened 2024-06-25
Upgrade python to version 3.12

jamshale:
Changes all usage of python 3.9 to 3.12. There was almost no changes to code needed.

The only changes were in the demo and the `wait_for` function call had a change in arguments. Also one unit test assertion, and a timing or race condition issue with one integration test. I believe might be caused by speed improvements with python. A small sleep statement seemed to solve it. 

The acapy docker images for 3.12 obviously aren't available yet and all of the demos that use docker images should be updated later after release.


jamshale:
I think we can ignore the security issue reported. It's to do with running the demo image as root user.

ianco (commented):
Looks good.  I like that there are no (few) code changes required,

ianco (approved):

ianco (approved):

PR #3068, opened 2024-06-25
For proof problem handler, allow no connection record (OOB cases), prevent unhandled exception

loneil:
Details in https://github.com/hyperledger/aries-cloudagent-python/issues/3069

Just add a conditional to handle the case where there's no connection record. Already did this for v1.0 here https://github.com/hyperledger/aries-cloudagent-python/pull/2723 but missed this was done in a different place for 2.0


jamshale (approved):
lgtm :+1: 

Could have a unit test just for completeness if wanting to be thorough.

PR #3072, opened 2024-06-28
Rule D417

jamshale:
Updates function documentation to the appropriate level needed for rule D417 https://docs.astral.sh/ruff/rules/undocumented-param/. Helps developers understand and more quickly use functions if these are accurate.

Mostly used co-pilot to fill in wrong or missing arguments. Reviewed most of it and it looked like it was doing a good job.

### DIFF
--- a/aries_cloudagent/indy/credx/holder.py
+++ b/aries_cloudagent/indy/credx/holder.py
@@ -379,8 +379,15 @@ class IndyCredxHolder(IndyHolder):
         """Check ledger for revocation status of credential by cred id.
 
         Args:
-            credential_id: Credential id to check
+            ledger (BaseLedger): The ledger to check for revocation status.
+            credential_id (str): The ID of the credential to check.
+            fro (int, optional): The starting sequence number of the revocation registry
+                delta. Defaults to None.
+            to (int, optional): The ending sequence number of the revocation registry
+                delta. Defaults to None.
 
+        Returns:
+            bool: True if the credential is revoked, False otherwise.
         """
         cred = await self._get_credential(credential_id)
         rev_reg_id = cred.rev_reg_id
@@ -537,17 +544,24 @@ class IndyCredxHolder(IndyHolder):
     ) -> str:
         """Create current revocation state for a received credential.
 
+        This method creates the current revocation state for a received credential.
+        It takes the credential revocation ID, revocation registry definition,
+        revocation delta, delta timestamp, and tails file path as input parameters.
+
         Args:
-            cred_rev_id: credential revocation id in revocation registry
-            rev_reg_def: revocation registry definition
-            rev_reg_delta: revocation delta
-            timestamp: delta timestamp
+            cred_rev_id (str): The credential revocation ID in the revocation registry.
+            rev_reg_def (dict): The revocation registry definition.
+            rev_reg_delta (dict): The revocation delta.
+            timestamp (int): The delta timestamp.
+            tails_file_path (str): The path to the tails file.
 
         Returns:
-            the revocation state
+            str: The revocation state.
+
+        Raises:
+            IndyHolderError: If there is an error creating the revocation state.
 
         """
-
         try:
             rev_state = await asyncio.get_event_loop().run_in_executor(
                 None,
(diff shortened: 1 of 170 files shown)


aries_cloudagent/indy/credx/holder.py
```diff
@@ -379,8 +379,15 @@ async def credential_revoked(
         """Check ledger for revocation status of credential by cred id.
 
         Args:
-            credential_id: Credential id to check
+            ledger (BaseLedger): The ledger to check for revocation status.
+            credential_id (str): The ID of the credential to check.
+            fro (int, optional): The starting sequence number of the revocation registry
```
swcurran, line 384: Do we care this should be “from”?

aries_cloudagent/indy/holder.py
```diff
@@ -42,8 +42,15 @@ async def credential_revoked(
         """Check ledger for revocation status of credential by cred id.
 
         Args:
-            credential_id: Credential id to check
+            ledger (BaseLedger): The ledger to check for revocation status.
+            credential_id (str): The ID of the credential to check.
+            fro (int, optional): The starting time of the revocation status check range.
```
swcurran, line 47: And again

aries_cloudagent/revocation/indy.py
```diff
@@ -208,7 +208,14 @@ async def get_issuer_rev_reg_delta(
         """Check ledger for revocation status for a given revocation registry.
 
         Args:
-            rev_reg_id: ID of the revocation registry
+            rev_reg_id (str): ID of the revocation registry
+            fro (int, optional): The sequence number to start from (exclusive).
```
swcurran, line 212: And again

swcurran:
Way cool — that’s wild. And the config change means these will be kept up to date?

@jamshale — would you mind look at the warnings we are getting when we generate the read the docs?  To see them:

- go into the `docs` folder
- Run the two instructions in the `UpdateRTD.md` file
  - The sphinx command to copy the latest set of files (e.g. `rm -rf generated; sphinx-apidoc….`)
  - The `sphinx-build` command

No worries if it is not obvious what is needed.  Obvs, you need to install sphinx (or use docker) to run the commands.

FYI — I’m using version `sphinx-build 5.3.0`.

aries_cloudagent/indy/credx/holder.py
jamshale: Ya. I'll change this here. It was like this before, but no point in truncating a word this short.

jamshale:
Yes. I'll will at least fail the `ruff` lint check that is done with the unit tests and force developers to keep the function descriptions up to date. I think we should fully switch to ruff for linting, because right now `ruff` and `black` are both tools that can do the same thing. Ruff is just more popular now and faster. Not important to switch immediately, though. 

I'll have a look at the doc warnings :eyes:

dbluhm:
Full agreement from me on switching to ruff for formatting too; it is noticeably faster than black

aries_cloudagent/indy/credx/holder.py
jamshale: So, this was actually done on purpose because `from` is a reserved keyword in python. I'm going to change these arguments to `timestamp_from` and `timestamp_to` instead.

jamshale:
@swcurran Did you used to get any warnings or errors when running `sphinx-build`? I fixed one set of warnings related to these changes but am getting other warnings for unrelated documentation. 

I can continue looking into them but might create another task for it. I'm having quite a bit of trouble trying to resolve them effectively.

jamshale:
I'm slowly figuring out how to fix the sphinx errors and warnings. I'll include the changes in this PR once I resolve them all.

swcurran:
Should you push that to another PR?  Don’t want to take too much of your time on it.

jamshale:
We can merge this the way it is. They warnings and errors only effect the individual docs. The rest will still be generated. 

I think I should be done soon though. Fixed most of them.

jamshale:
This is updated and should be good to go now.

swcurran (approved):

PR #3080, opened 2024-07-03
Switch from black to ruff

jamshale:
Completely removes black from the project in favour of ruff for linting and formatting.

Adjust the decontainer and the github actions workflow.


jamshale:
Also, I forgot to mention that the ruff linting based on the rules in `pyproject.toml` weren't actually being checked by the github action pipelines. This fixes that issues now and makes sure all linting and formatting is successful when opening a PR.

ianco (approved):

Issue #3065, opened 2024-06-25
Combination of terse web hooks without --preserve-exchange-records makes completing a present proof 2.0 impossible 


Gavinok:
As the verifier it is impossible to complete a present proof 2.0 transaction without the `--preserve-exchange-records` flag.

When the webhook confirming a proof request has been completed acapy will already delete the record before the controller can retrieve it.

swcurran:
To confirm — the issue is not that ACA-Py has already deleted the record before the controller can retrieve it — that is expected/intended behaviour.  The controller shouldn’t be trying to retrieve the record after the end of the protocol.  Rather the issue is that the “terse" webhook that was added at some point into the code base (for good reason — it carries too much information) does not include the required information for the verifier — the presentation — that is needed by the controller.

Do I have that right?

jamshale:
Can we replicate this with the demo? If so what are the steps?

I removed the `preserve-exchange-records` flag from the faber and alice demo and could still complete the proof despite the presentation record getting deleted by both agents.

swcurran:
You just need to check the webhook to see what data comes back in it.  If it is insufficient, it is a problem.  The timing issue is not the problem.  The controller/verifier should NOT be trying to retrieve the record via the API when that flag is not set.

Gavinok:
The issue is that in our case we don't even seem to manage to verify the presentation before it is deleted. Her are some of our logs

```
controller-1     | 2024-06-25T19:34:52.470174Z [info     ] GOT A PRESENTATION, TIME TO VERIFY [api.routers.acapy_handler]
controller-1     | 2024-06-25T19:34:52.470306Z [debug    ] >>> verify_presentation        [api.core.acapy.client]
controller-1     | Starting new HTTP connection (1): aca-py:8077
aca-py-1         | 2024-06-25 19:34:52,475 aries_cloudagent.admin.server ERROR Handler error with exception: Record not found: pres_ex_v20/714760ae-3c25-40b3-8ae9-a6413fc9c431.
aca-py-1         | 
aca-py-1         | =================
aca-py-1         | Traceback (most recent call last):
aca-py-1         |   File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/present_proof/v2_0/routes.py", line 1270, in present_proof_verify_presentation
aca-py-1         |     pres_ex_record = await V20PresExRecord.retrieve_by_id(session, pres_ex_id)
[...]

PR #3081, opened 2024-07-03
Add by_format to terse webhook for presentations

ianco:
Fix for issue #3065

Note that this is for the 2.0 presentation protocol only, do we need to apply a similar fix for 1.0?

This adds the `by_format` attribute for all webhooks, not just for status `done`, however the demo requires this to work with terse webhooks (and no additional calls to the api endpoints)

Added `by_format` for issue cred as well - was failing in the demo for json-ld :-(

Makes the terse webhooks slightly less terse, but is required if we want the webhook to contain "enough" info for the controller to respond to.

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v2_0/messages/cred_ex_record_webhook.py
+++ b/aries_cloudagent/protocols/issue_credential/v2_0/messages/cred_ex_record_webhook.py
@@ -23,6 +23,7 @@ class V20CredExRecordWebhook:
         "credential_definition_id",
         "schema_id",
         "credential_id",
+        "by_format",
         "trace",
         "public_did",
         "cred_id_stored",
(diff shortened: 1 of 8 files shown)


jamshale:
> @jamshale do you know if it's possible to ignore the SonarCloud error? It doesn't like that we don't explicitly set the user in the demo dockerfile ....

Sonarcloud passing isn't actually required to merge. So we can just ignore this and approve/merge as normal. I don't think there's an easy way to ignore it through the sonarcloud UI.

swcurran:
Is it easy to fix?  Obvs if not, we can leave it — just wondering.

ianco:
> Is it easy to fix? Obvs if not, we can leave it — just wondering.

Prob not difficult to fix but may be time consuming to figure it out.  The dockerfies are all slightly different so we'd need to get them all lined up with each other or figure it out for each one.  I suggest we just log a ticket for now ...

ianco:
The integration test failures are related to json-ld, I'll need to look into it ...

jamshale (approved):
:+1:

Gavinok (approved):
Thanks for getting to this so quickly

PR #3082, opened 2024-07-03
:bug: fix storage record pagination with post-filter query params

ff137:
Related to: https://github.com/hyperledger/aries-cloudagent-python/issues/3001#issuecomment-2206110085

Marking as draft to first validate things work as expected (testing here: https://github.com/didx-xyz/aries-cloudapi-python/pull/850)

PR #3083, opened 2024-07-04
:art: improve record querying logic

ff137:
Minor amendments that I wanted to include before #3082 was merged.

Expands error handling for the unlikely case where json.loads fails, and improves readability / reduces complexity of nested if conditions.


jamshale (approved):
:+1: Much more readable

PR #3092, opened 2024-07-08
1.0.0rc4

swcurran:
The release is (I think) ready to go.  Please review the changelog entry and propose any changes to it -- items that should be highlighted, and breaking changes.

I noticed that a number of the dockerfiles reference a container image for what I assume is the latest release -- `3.9-01.12.1`.  When do those get updated following a release?  I assume they don't happen as part of a release -- or do they?  Should I update those in this PR?


jamshale:
I dont think those docker images will be available until after the release. We will have to have a task after the release is complete.

ianco:
We haven't dropped the `1.0` protocols yet for issue cred and present proof but I think we should include these in "breakign changes" anyways, as VERY STRONG recommendations to upgrade or else.

jamshale (approved):
Looks good to me :+1:

ianco (approved):
Looks good to me.  I added a comment about the deprecation of the 1.0 endpoints but it's just a suggestion.

jamshale:
The integration tests should do a fuller suite of tests for a release that could take 70 minutes +. The same set has been passing nightly pretty consistently, but there is a higher chance of intermittent fails.

dbluhm (approved):

PR #3095, opened 2024-07-10
Check connection is ready in all connection required handlers

jamshale:
This is for https://github.com/hyperledger/aries-cloudagent-python/security/advisories/GHSA-cxwq-rwgr-qp5m.

I'm pretty sure this fixes the main issues reported with basic_messages and handshake_reuse protocols. I also applied the connection ready check to all other protocols that were missing and that I think need an established connection.

I didn't apply it to any problem report protocols because I don't think a connection is required in the cases. I'm not entirely sure if the problem report protocols could still be exploited in this way. Maybe an nefarious actor could send problem reports before the connection was established and cause issues?

The details of all these protocols isn't a strong point for me. I did my best to remediate the problem but I good review by someone more knowledgeable would be appreciated.

### DIFF
--- a/aries_cloudagent/protocols/actionmenu/v1_0/handlers/menu_handler.py
+++ b/aries_cloudagent/protocols/actionmenu/v1_0/handlers/menu_handler.py
@@ -3,9 +3,9 @@
 from .....messaging.base_handler import (
     BaseHandler,
     BaseResponder,
+    HandlerException,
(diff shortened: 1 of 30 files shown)


jamshale:
The code coverage wants me to tests all the exception scenarios.

dbluhm (approved):

PR #3105, opened 2024-07-17
Make single wallet config more explicit

jamshale:
This changes some naming and configuration to make the multitenant single wallet vs multi wallet settings and usage more clear. Askar wallets can use either the basic manager `MultitenantManager` which is used by default and creates a DB/wallet for each tenant. Or, it can use the `AskarProfileMultitenantManager` to create subwallets in a single DB/wallet. This has been renamed to `SingleWalletAskarMultitenantManager` for clarity, and the config changed to `multitenancy-config: '{"wallet_type": "single-wallet-askar"}'` from `multitenancy-config: '{"wallet_type": "askar-profile"}'`

Also fixed an issue with the upgrade check during startup that was using the base manager before initialization and creating subwallet db's on restart when in single wallet mode. I'm looking into this more because the `BaseMultitenantManager` is injected in many other places. Need to make sure they are also using the correct manager via the `MultitenantManagerProvider`.

Added some documentation in `Multitenancy.md`

***Note:*** the multitenant_provider plugin needs to be update because it inherits this class. Already done in my forked repo.

### DIFF
--- a/.vscode-sample/multitenant-admin.yml
+++ b/.vscode-sample/multitenant-admin.yml
@@ -8,10 +8,11 @@ inbound-transport:
 
 outbound-transport: http
 
-wallet-type: askar-anoncreds
+wallet-type: askar
 wallet-storage-type: default
 wallet-name: multitenant-admin-wallet
 wallet-key: multitenant-admin-wallet-key
+multitenancy-config: '{"wallet_type": "single-wallet-askar"}'
 
 admin-insecure-mode: true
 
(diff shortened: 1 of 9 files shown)


ianco (approved):

ff137:
NB: renaming `{"wallet_type": "askar-profile"}'` to `'{"wallet_type": "single-wallet-askar"}'` is not listed in the rc5 / rc6 changelogs :-) should probably be noted as a breaking change.

jamshale:
Yes, it should. I thought it had been listed. It is a critical change for multitenant single wallet deployments. @swcurran

PR #3106, opened 2024-07-17
Fix the check for vc_di proof

ianco:
Fix the error that is getting raised in AATH.  (Double checked and the AATH tests will now pass.)


jamshale (approved):
:+1:

PR #3110, opened 2024-07-19
DOC: Verifiable Credential Data Integrity (VC-DI) Credentials in Aries Cloud Agent Python (ACA-Py) #2947

kenechukwu-orjiene:


docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,633 @@
...
+    "jws": "eyJhbGciOiJFZERTQSJ9..."
+  }
+}
+```
+</details>
+
+2. **Select Credential type**
+The ability to choose the credential type (indy, vc_di) to be issued. The credential type is used to determine the schema for the credential data.
+
+The format to change credential can be seen in the ["Demo Instruction"](https://github.com/sarthakvijayvergiya/aries-cloudagent-python/blob/whatscookin/feat/vc-di-proof/docs/demo/README.md)
```
PatStLouis, line 139: This link should point to the aca-py github page, could this be a relative path?

docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,633 @@
...
+  "auto_remove": false,
+  "comment": "Issuing a test credential",
+  "credential_preview": {
+    "@type": "https://didcomm.org/issue-credential/2.0/credential-preview",
+    "attributes": [
+      {"name": "name", "value": "John Doe"}
+    ]
+  },
+  "trace": false
+}
```
PatStLouis, line 159: don't we need to include a filter for issue-credential 2.0?

docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,633 @@
...
+      "credentialSubject": {
+        "id": "did:key:z6Mkh......",
+        "name": "John Doe"
+      },
+      "proof": {
+        "type": "Ed25519Signature2020",
+        "created": "2023-01-01T00:00:00Z",
+        "proofPurpose": "assertionMethod",
+        "verificationMethod": "did:key:z6MkqG......#z6MkqG......",
+        "jws": "eyJhbGciOiJFZERTQSJ9..."
```
PatStLouis, line 209: Same as above, `jws` is an invalid signature format here, it needs to be `proofValue` with a multibase representation of the signature.

docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,633 @@
...
+  "credentialSubject": {
+    "id": "did:key:z6Mkh......",
+    "name": "John Doe"
+  },
+  "proof": {
+    "type": "Ed25519Signature2020",
+    "created": "2023-01-01T00:00:00Z",
+    "proofPurpose": "assertionMethod",
+    "verificationMethod": "did:key:z6MkqG......#z6MkqG......",
+    "jws": "eyJhbGciOiJFZERTQSJ9..."
```
PatStLouis, line 130: jws is not used in data integrity proofs, please use `proofValue` with a multibase representation of the proof:
https://w3c.github.io/vc-di-eddsa/#example-signed-credential-1

docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,633 @@
...
+Ensure that every property key in the document is mappable to an IRI. This requires either the property key to be an IRI by default or to have the shorthand property mapped in the `@context` of the document.
+
+```json
+{
+    "@context": [
+        "https://www.w3.org/2018/credentials/v1",
+        "https://w3id.org/security/data-integrity/v2",
+        {
+            "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#"
+        }
```
PatStLouis, line 67: I would include note about avoiding `@vocab` outside of development and/or examples, and implementers should define their own terms in a context file they provide.

docs/features/W3cCredentials.md
```diff
@@ -0,0 +1,633 @@
...
+        "jws": "eyJhbGciOiJFZERTQSJ9..."
+      }
+    }
+  ]
+}
+```
+</details>
+
+2. **Send the Verification Request:**
+Use the `/verifier-credential/2.0/verify` endpoint.
```
PatStLouis, line 218: I don't think this endpoint exists.

docs/features/W3cCredentials.md
PatStLouis: For this specific example, { "name": "https://schema.org/name" } instead of {"@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#"} would also be a better solution to defining the attribute.

docs/features/W3cCredentials.md
kenechukwu-orjiene: Thanks, this have been updated

docs/features/W3cCredentials.md
kenechukwu-orjiene: Great one picking this out. I have included filter

docs/features/W3cCredentials.md
kenechukwu-orjiene: do you mean something like this
```json
..
"@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/security/data-integrity/v2",
    "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#"
    ]
 ..
 ```

swcurran:
Please update this PR to the latest so we can merge this.  Thanks.

docs/features/W3cCredentials.md
PatStLouis: no, leave as is this will work, I would just put a note since using `@vocab` is considered a bad practice outside of demonstrations and development. Implementers are encouraged to write their own context or use community available contexts. The existing w3c documentation outlines this here:
https://github.com/hyperledger/aries-cloudagent-python/blob/main/docs/features/JsonLdCredentials.md#json-ld-context

swcurran (approved):

docs/features/W3cCredentials.md
gvelez17: @kenechukwu-orjiene ^^

PR #3112, opened 2024-07-21
fix: multiuse invites with did peer 4

### LINKED ISSUE #3111: Multiuse invitations and Qualified DIDs
Affected versions: 0.12.1, 1.0.0rc4, nightly-2024-07-19

There seems to be an issue in determining the correct inbound connection when using multi-use invitations and didexchange 1.1 with qualified DIDs.

Reproducible example can be found here: https://github.com/Indicio-tech/acapy-minimal-example/blob/test/multiuse-oob/examples/multiuse-oob/example.py

Small amount of context: ACA-Py uses a connection record to represent multi-use invitations.

If a connection is made with a multi-use invitation and then a message is sent to the resulting connection, the response to that message will be handled as though it had been sent by the multi-use invitation connection. In the example, this is demonstrated by sending a feature discovery message; ACA-Py's handler for the disclose message checks the `connection_ready` condition on the context and finds it as false, which is accurate because the multi-use invitation conn record is still in state `invitation`. The connection record as extracted from the logs of the call to the DiscloseHandler:

```python
ConnRecord
(_id='1cd87e77-9a4a-49c7-93d0-3d972a0f1849', _last_state='invitation', _new_with_id=False, state='invitation', created_at='2024-07-19T16:27:10.277265Z', updated_at='2024-07-19T16:27:10.27726
5Z', my_did=None, their_did=None, their_label=None, their_role='invitee', invitation_key='De2YsJc9m3Lr9XeErdbjjMCSGFk9YwRBmw9V8q1Md3nd', invitation_msg_id='d39fca16-eaaa-40e7-b447-b56d8c085d
12', request_id=None, error_msg=None, inbound_connection_id=None, accept='manual', invitation_mode='multi', alias=None, their_public_did=None, connection_protocol='didexchange/1.1')
```

dbluhm:
This corrects an issue where did:peer:4 connection records were failing to be resolved on inbound message, resulting in the multiuse invitation that created the connection being resolved instead.

Fixes #3111.


jamshale (approved):
Manually tested with the scenario that was failing for me and this fixed it :+1:

swcurran:
OK to merge with the SonarCloud state? @dbluhm what do you think?

dbluhm:
I'm okay with it being merged as is but I am also happy to get a quick unit test written. Would not take a lot to cover the changes since it wasn't a lot of changes.

jamshale:
Sometimes reformatted code will make sonarcloud report areas that didn't have any logic changes. I think that happened a bit here. Could think about a couple tests for the new code.

dbluhm:
Sonar cloud is still unhappy but I'd say the new unit tests + integration tests will be adequate to cover these changes.

jamshale (approved):

dbluhm:
@jamshale https://github.com/hyperledger/aries-cloudagent-python/actions/runs/10045120793/job/27761704780?pr=3112#step:4:18060
Odds of this being a false-negative? Seems unrelated to my changes at first glance at least

jamshale:
> @jamshale https://github.com/hyperledger/aries-cloudagent-python/actions/runs/10045120793/job/27761704780?pr=3112#step:4:18060 Odds of this being a false-negative? Seems unrelated to my changes at first glance at least

This test does have false negatives quite often. I think it's probably the case as other tests would have failed as well. Let's see if it passes a new run.

jamshale:
@dbluhm This test is actually failing locally. I think it must be related to the changes. I'm not sure why the `retrieve_by_tag_filter` query would have been affected for this test.

dbluhm:
Interesting -- I'll dig in! Thanks for checking locally

swcurran:
@dbluhm — sorry to bug you, but another failed test in the PR Tests.  Looks to be in the same area as the most recent commit.

dbluhm:
> @dbluhm — sorry to bug you, but another failed test in the PR Tests. Looks to be in the same area as the most recent commit.

Yeah, pushed while my tests were running locally still :slightly_smiling_face: PR tests should be good. I think int tests are too but they take a while to run locally even when narrowed to just the endorser scenarios.

aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py
```diff
@@ -772,20 +767,17 @@ async def set_transaction_my_job(self, record: ConnRecord, transaction_my_job: s
         return tx_job_to_send
 
     async def set_transaction_their_job(
-        self, tx_job_received: TransactionJobToSend, receipt: MessageReceipt
+        self, tx_job_received: TransactionJobToSend, connection: ConnRecord
```
jamshale, line 770: I was looking at this for a bit and didn't understand why it used MessageReceipt here?

jamshale (approved):

aries_cloudagent/protocols/endorse_transaction/v1_0/manager.py
dbluhm: Indeed; my impression is that it was just a question of the implementer being unfamiliar with the request context or something along those lines.

Issue #3094, opened 2024-07-10
Receiving unauthenticated DIDComm messages from a connection in the invitation state


jamshale:
See https://github.com/hyperledger/aries-cloudagent-python/security/advisories/GHSA-cxwq-rwgr-qp5m

PR #3120, opened 2024-07-23
Apply security patch 0.11.x

jamshale:
Apply security patch to branch 0.11.x
- #3094 

Integration tests currently need to be ran manually on patch releases. I ran all the @GHA tag tests against a local von network and tails server and all 4 features passed.

### DIFF
--- a/aries_cloudagent/protocols/actionmenu/v1_0/handlers/menu_handler.py
+++ b/aries_cloudagent/protocols/actionmenu/v1_0/handlers/menu_handler.py
@@ -3,9 +3,9 @@
 from .....messaging.base_handler import (
     BaseHandler,
     BaseResponder,
+    HandlerException,
     RequestContext,
 )
-
 from ..messages.menu import Menu
 from ..util import save_connection_menu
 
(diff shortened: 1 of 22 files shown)


swcurran:
Running AATH tests…so far so good.

swcurran:
Very weird. With these changes, all but a set of 7 peer-did tests fail. When I run with 0.11.x branch in ACA-Py a whole bunch more tests fail — coordinate mediation, oob, issue credential and one present proof.  Running again, in case it was a glitch, but local tests are usually pretty reliable…

jamshale:
Interesting. I've only run the integration tests for acapy for this patch. The AATH acapy runset all passed for me for the 0.12.x patch.

dbluhm:
Most of the peer did work wasn't merged until 0.12.0, right? We should expect AATH peer did tests to fail, wouldn't we?

jamshale:
I think that aligns with what tests failed. I kind of expected some failing tests with this branch.

dbluhm (approved):

swcurran:
Ran the AATH `acapy` runset with this PR, 0.11.x/3.12 and 0.11.x/3.9.  All were consistent, with the same set of tests failing that would be expected — things added after 0.11.x.  However, as noted above, one of the two 0.11.x/3.12 tests had some additional failures — but worked the next time.

I think we’re good with this, and we can go ahead.

Next up will be a PR to publish the patch release.  Stay tuned!

PR #3122, opened 2024-07-24
0.11.2

swcurran:


swcurran:
@jamshale @dbluhm -- do we need an rc release, or should I update this to the actual release. Happy to do either -- this PR or after a bit of testing, we do a second PR for the actual release.

jamshale (approved):

Issue #2999, opened 2024-05-31
When running version 0.12.0, an error occurs if --wallet-key flag is missing


kukgini:
I'm tying to upgrade aca-py from 0.8.2 to 0.12.1

Before release 0.12.0, the --wallet-key parameter was not required on askar wallet type (as far as i know, it is required only on indy wallet type). However,  it is required on version 0.12.0. 

I see that the change was made by Ian Contanzo in 12/5/23 but not described in Breaking Changes. 
(commit hash: bfa0a5ec91ad18f2e9cddc3bdf876908dc271cc0)

I didn't set the parameter before, but instead set the --wallet-key-derivation-method to "kdf:argon2i:mod". So, I guess it was set to DEFAULT_KEY

kukgini:
However, I realized that my service was set DEFAULT_KEY which is empty string because I didn't set --wallet-key flag.

IMHO, The reason why DEFAULT_KEY cannot be used when --wallet-key-derivation-method is RAW is because the wallet key must be 32 bits. And if it is set to Argon KDF, even if it is an empty string, it can be used as a wallet key because it is converted to 32bit. So, It seems like the restriction is technical reason.

However, In any cases, I think It is nessicery to set --wallet-key for persistance storage for security reason. So, basically I think your (@ianco) change is correct. The reason that the original source code requires a wallet-key only for indy is probably to distinguish it from in-memory. So, The mistake have been made when askar was firstly added. 

The problem is that it seems like rekey feature is blocked, I can't change my current key which is empty string to another one.

ianco:
I think the `--rekey` parameter needs to support the use case where the existing key is blank (the DEFAULT_KEY).

As to whether `--wallet-key` must be provided for askar wallets I thought it was a requirement, I didn't realize there was a DEFAULT_KEY value available.

I'm going to tag @swcurran  and @andrewwhitehead  on this issue for their input.  At the very least we need to support the 0.8.2 to 0.12.1 upgrade

swcurran:
I don’t quite get the nuances/interactions of the options, but from the sounds of it we need to do three things:

1. Make the `—rekey` update as per @kukgini / @ianco agreement.
2. Update the breaking changes documentation in the release notes, and perhaps call out the issue further in the https://aca-py.org/latest/deploying/UpgradingACA-Py/ documentation.
3. Get a 0.12.2 release out there that is a patch on 0.12.1 (not based on latest main).

Does that make sense?

PR #3125, opened 2024-07-25
Add rekey feature with blank key support

jamshale:
See #2999.

This adds the support to change wallet keys and also allows any wallets created with a blank key before it was required in 0.12.0 to change their key and unblock the upgrade. 

This allow an agent stuck in a state where they used a blank key to configure a wallet-key that won't be used, but will be replaced by the `rekey` variable. After using rekey they can then use that key in the wallet-key config.

### DIFF
--- a/aries_cloudagent/askar/store.py
+++ b/aries_cloudagent/askar/store.py
@@ -36,20 +36,17 @@ class AskarStoreConfig:
             config = {}
         self.auto_recreate = config.get("auto_recreate", False)
         self.auto_remove = config.get("auto_remove", False)
+
         self.key = config.get("key", self.DEFAULT_KEY)
         self.key_derivation_method = (
             config.get("key_derivation_method") or self.DEFAULT_KEY_DERIVATION
         )
 
-        if (
-            self.key_derivation_method.lower() == self.KEY_DERIVATION_RAW.lower()
-            and self.key == ""
-        ):
-            raise ProfileError(
-                f"With key derivation method '{self.KEY_DERIVATION_RAW}',"
-                "key should also be provided"
-            )
         self.rekey = config.get("rekey")
+        self.rekey_derivation_method = (
+            config.get("rekey_derivation_method") or self.DEFAULT_KEY_DERIVATION
+        )
+
         self.name = config.get("name") or Profile.DEFAULT_NAME
         self.in_memory = self.name == ":memory:"
         if self.in_memory:
@@ -131,6 +128,20 @@ class AskarStoreConfig:
                 )
             raise ProfileError("Error removing store") from err
 
+    def _handle_open_error(self, err: AskarError, retry=False):
+        if err.code == AskarErrorCode.DUPLICATE:
+            raise ProfileDuplicateError(
+                f"Duplicate store '{self.name}'",
+            )
+        if err.code == AskarErrorCode.NOT_FOUND:
+            raise ProfileNotFoundError(
+                f"Store '{self.name}' not found",
+            )
+        if retry and self.rekey:
+            return
+
+        raise ProfileError("Error opening store") from err
+
     async def open_store(self, provision: bool = False) -> "AskarOpenStore":
         """Open a store, removing and/or creating it if so configured.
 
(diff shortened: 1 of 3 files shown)


aries_cloudagent/askar/store.py
```diff
@@ -156,16 +168,26 @@ async def open_store(self, provision: bool = False) -> "AskarOpenStore":
                     self.key_derivation_method,
                     self.key,
                 )
+                if self.rekey:
+                    await Store.rekey(store, self.key_derivation_method, self.rekey)
```
esune, line 172: Is the key derivation method something that we would want to expose, or using the currently "saved" one is the only feasible option?

jamshale: This is from the config value. Currently it uses the same value. Meaning you need to use the same method for the rekey as the original key method. I saw a config that doesn't exist commented out that would allow the rekey to use a different method.

jamshale: It it isn't provided it defaults to `kdf:argon2i:mod`

kukgini: Perhaps we could consider adding wallet.rekey_derivation_method to allow for simultaneous changes to the KDF.

jamshale: Yes. I think that's a good idea.

jamshale:
rekey derivation method support has been added.

esune (dismissed):
I think this looks good 👍🏻

aries_cloudagent/config/argparse.py
```diff
@@ -1638,6 +1638,17 @@ def add_arguments(self, parser: ArgumentParser):
...
         )
+        parser.add_argument(
+            "--wallet-rekey-derivation-method",
+            type=str,
+            metavar="<rekey-derivation-method>",
+            env_var="ACAPY_WALLET_REKEY_DERIVATION_METHOD",
+            help=(
+                "Specifies the key derivation method used for the replacement"
+                "rekey encryption. If RAW key derivation method is used, also"
+                "--wallet-key parameter is expected."
```
kukgini, line 1649: I think it should be --wallet-rekey instead of --wallet-key

aries_cloudagent/config/argparse.py
```diff
@@ -1694,6 +1705,10 @@ def get_settings(self, args: Namespace) -> dict:
             settings["wallet.type"] = args.wallet_type
         if args.wallet_key_derivation_method:
             settings["wallet.key_derivation_method"] = args.wallet_key_derivation_method
+        if args.wallet_key_derivation_method:
+            settings["wallet.rekey_derivation_method"] = (
+                args.wallet_rekey_derivation_method
```
kukgini, line 1710: Why does it use the `wallet_rekey_derivation_method` value when `wallet_key_derivation_method` is declared?

aries_cloudagent/askar/store.py
```diff
@@ -49,9 +49,10 @@ def __init__(self, config: dict = None):
                 f"With key derivation method '{self.KEY_DERIVATION_RAW}',"
                 "key should also be provided"
             )
-        # self.rekey = config.get("rekey")
-        # self.rekey_derivation_method = config.get("rekey_derivation_method")
-
+        self.rekey = config.get("rekey")
+        self.rekey_derivation_method = (
+            config.get("rekey_derivation_method") or self.DEFAULT_KEY_DERIVATION
```
kukgini, line 54: How about implement the logic to set the default value when there is no setting in argparse.py?

aries_cloudagent/config/argparse.py
jamshale: Yes. I think this was a careless copy/paste error. Good spot. The brackets are just for the line length.

aries_cloudagent/config/argparse.py
jamshale: Yes. I copied and pasted from wallet-key and didn't go over it properly.

aries_cloudagent/askar/store.py
jamshale: I think that's a good idea, but all the other configs used in the store use this pattern of loading with defaults here. I don't want to change it in this PR.

aries_cloudagent/askar/store.py
```diff
@@ -36,21 +36,16 @@ def __init__(self, config: dict = None):
             config = {}
         self.auto_recreate = config.get("auto_recreate", False)
         self.auto_remove = config.get("auto_remove", False)
+
         self.key = config.get("key", self.DEFAULT_KEY)
         self.key_derivation_method = (
             config.get("key_derivation_method") or self.DEFAULT_KEY_DERIVATION
         )
 
-        if (
```
jamshale, line 44: Removed this, as key is now required for all methods.

ianco (approved):
Looks good to me

PR #3133, opened 2024-07-29
Adds documentation site docs for releases 0.11.0, 0.11.1, 0.11.2, and eliminates docs for RCs

swcurran:
This PR adds the static site documentation pages for releases 0.11.0, 0.11.1 and 0.11.2, and deletes those for RC releases that have been supplemented by final releases.  By adding in the 0.11.x releases, we now have all of the LTS release documentation published from this ACA-Py repo, vs. from the aries-acapy-docs repo.  The RC release removals is just a bit of house cleaning.

As of today, I've switched the https:aca-py.org site to use gh-pages from the ACA-Py site.  In doing that, we lose access to all the docs before 0.12.0, hence this PR to add the 0.11.x releases.  We lose the docs from prior to 0.11.0, but so be it -- I think that is OK -- although we could also move the rest of the releases, just as I did the 0.11.x ones.

The only challenge will be that if we do another 0.11.x release, I'll have to generate the docs from the aries-acapy-docs repo, and then transfer them from there to here.  Not too difficult -- just an extra step.  Documentation from now on will come out of this repo.


swcurran:
The description of this change was updated after submitting this PR -- so check it out ^^^^^ above.

jamshale (approved):

jamshale:
So many files :open_mouth:

PR #3148, opened 2024-08-02
jq-1.6 is already in the source container

rngadam:


rngadam:
Looks like I was wrong, jq is not in the source container:

````
docker run --entrypoint /usr/bin/jq  ghcr.io/hyperledger/aries-cloudagent-python:py3.9-0.12.1 
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/usr/bin/jq": stat /usr/bin/jq: no such file or directory: unknown.
````

PR #3164, opened 2024-08-14
Update `TxnOrPublishRevocationsResultSchema`

cl0ete:
Update `TxnOrPublishRevocationsResultSchema` to accommodate changes made in PR #3107

Tweak return values from `publish_revocations` to match response model.

Ran the script `generate-open-api-spec` to update `openapi.json` and `swagger.json` files


swcurran:
Tested that the OpenAPI and SwaggerAPI files match the generated.  I don’t know enough about the functionality to determine if it is right — adding reviewers.  For reviewers — is this OK to go into 1.0.0 or should we hold it until that is complete?

jamshale (approved):
Thanks! This looks good.

jamshale:
@swcurran This should be safe to include in 1.0.0. 

I didn't think it was important enough to need to get in, but we could include it if it doesn't hold anything up.

PR #3166, opened 2024-08-15
👷 fix lint workflow and 🎨 apply ruff linting

ff137:
Just upgrades the ruff pre-commit hook, and applies the `pre-commit run --all-files` command, as introduced in #3080

edit: also fixes the lint workflow to do a formatting check

Because, when I just ran ruff format, it rekt my uncommitted work. :smile: (Now I can run format without 328 files changed)


.github/workflows/format.yml
```diff
@@ -16,3 +16,6 @@ jobs:
           python-version: "3.12"
       - name: Ruff Format and Lint Check
         uses: chartboost/ruff-action@v1
+        with:
+          version: 0.5.7
+          args: "format --check"
```
ff137, line 21: @jamshale PS: I think this action was just doing `ruff check`, since it didn't include `args: "format --check"`. 
https://github.com/ChartBoost/ruff-action?tab=readme-ov-file#use-ruff-format

ff137:
Confirmed that now the lint workflow fails with faulty line:
https://github.com/hyperledger/aries-cloudagent-python/actions/runs/10402565667/job/28807290328?pr=3166

.github/workflows/format.yml
jamshale: This checks linting and formatting now? I thought I had tested this but might have just been checking linting.

jamshale:
Thanks! I think this formatting is cleaner. I don't think it's necessary to squeeze into v1.0.0 but will approve and merge shortly.

jamshale (approved):

ff137:
It sounded like a 1.0.0 release is imminent. Is there any reason _not_ to include format changes? 😄

jamshale:
It is imminent :smile: I do think this is safe enough to include. I'm going to merge and let @swcurran decide on whether to include it in the tagged release.

PatStLouis:
I'm also having linting issues with the current main branch using black, we need to make sure the tests pick this up. We had occurrences in the past where a black version upgrade introduced a significant amount of changes. This might be related to the recent black version released august 4th which bumped the version from 24.4 to 24.8.

https://pypi.org/project/black/24.8.0/

PatStLouis:
IMO linting should have no impact on the code

jamshale:
@PatStLouis we don't have black as a formatter for acapy anymore. You should be using ruff instead.

PatStLouis:
Oh that explains a lot, missed that one. I just switched and it works pretty good! Thanks for the heads up.

PR #3169, opened 2024-08-15
:art: organize imports

ff137:
Now's probably as good a time as any to organize imports for the codebase!

At present, the Ruff formatter doesn't sort imports (they say it is planned): https://docs.astral.sh/ruff/formatter/#sorting-imports

But, they allow to sort imports and format with:
```sh
ruff check --select I --fix
ruff format
```

This is mostly identical to [`isort`](https://github.com/PyCQA/isort), but `isort` includes some additional improvements in structure.

So, I applied `isort . --profile black`, followed by `ruff check --select I --fix` and `ruff format`.

Tada. :tada: One bulk change to neaten up imports across the board.

Thought this is maybe worth squeezing in before the official release. Can come later if it's just a spanner in the works.

I don't see any need to include a rule like this to the lint workflow. That can probably be added eventually when ruff supports organizing imports in one command.


jamshale (approved):
Looks good to me :+1:

ff137:
Been itching to contribute this for a while. I reflexively sort imports, and always have to withhold here cuz it changes too much 😄

PR #3181, opened 2024-08-19
Add di support w/ eddsa-jcs-2022

PatStLouis:
This PR looks to bring Data Integrity to aca-py over arbitrary json (NOT VC).

It adds a eddsa-jcs-2022 cryptosuite to the vc plugin and 2 routes to the wallet plugin (similar to the jwt sign/verify features).

It also moves the request/response schemas to the models directory of the wallet plugin.


PatStLouis:
I'm having issue with the poetry lock file after adding the `canonicaljson` package, I don't recall having issues previously when updating the lock file. I'll keep investigating but any pointers would be welcomed.

aries_cloudagent/wallet/models/web_requests.py
dbluhm, line 1: Nitpick but I think I'd rather have these be closer to the request handlers that use them rather than being in a separate file.

aries_cloudagent/vc/vc_di/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,119 @@
...
+            secured_document = document.copy()
+            secured_document["proof"] = existing_proof
+            secured_document["proof"].append(proof)
+
+            return secured_document
+        except Exception:
+            raise DataIntegrityProofException()
+
+    async def verify_proof(self, unsecured_document, proof):
+        # https://www.w3.org/TR/vc-data-integrity/#verify-proof
```
dbluhm, line 82: Nitpick: why not put these links (which are very helpful) into the docstring? This would ensure that they are rendered into the documentation.

aries_cloudagent/vc/vc_di/cryptosuites/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .eddsa_jcs_2022 import EddsaJcs2022
+
+CONTEXTS = {"data-integrity-v2": "https://w3id.org/security/data-integrity/v2"}
```
dbluhm, line 3: When would we look up this context by the string `data-integrity-v2`? I get the feeling it might be simpler to just have `DIV2_CONTEXT = "..."`.

dbluhm (commented):
A few nitpicks. Looks good overall to me!

aries_cloudagent/vc/vc_di/cryptosuites/eddsa_jcs_2022.py
PatStLouis: will do, thanks for the tip (I think ruff was giving me warning when I was doing so but I'll revisit it)

aries_cloudagent/vc/vc_di/cryptosuites/__init__.py
PatStLouis: I agree, not sure why I settled for this.

I would like to create a file similar to [this file](https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/vc/ld_proofs/constants.py), but focused on contexts with their digests. 

I'm just not convinced by the current structure of the vc directory. I think we should think about the purpose of this directory and re-organize how files are structured. I would like to create a linked-data sub directory where the document loaders and context uri/files live, and manage all linked-data related operations in there, providing some functions to detect dropped terms, etc...

Then just focus on data integrity proofs...

aries_cloudagent/wallet/models/web_requests.py
PatStLouis: My plan was to put the cryptosuite there and later on build issuing VC with it by adding a layer of data model validation. When you say these, what exactly do you refer to?

PatStLouis:
@dbluhm this is ready for a final review, I've addressed most of the comments and added 4 unit tests. Do you know how to make SonarCloud happy?

PatStLouis:
@jamshale can I get a pair of eyes on this pr please?

jamshale:
I think this looks good 👍 

Was the did_key manager stuff needed for this PR or was it an accident?

aries_cloudagent/vc/vc_di/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .proof import DIProof
+from .proof_options import DIProofOptions, DIProofOptionsSchema
+
+__all__ = ["DIProof", "DIProofSchema", "DIProofOptions", "DIProofOptionsSchema"]
```
jamshale, line 4: You need to import `DIProofSchema` to fix the issue reported by sonarcloud.

PatStLouis:
> I think this looks good 👍
> 
> Was the did_key manager stuff needed for this PR or was it an accident?

accident, let me fix this

PatStLouis:
closing as the features are supported by #3261

Issue #3182, opened 2024-08-19
ptvsd debugging option is deprecated (replaced by debugpy)


rngadam:
https://github.com/microsoft/ptvsd

> 🛑 ᴘᴛᴠsᴅ ɪs ᴅᴇᴘʀᴇᴄᴀᴛᴇᴅ 🛑
> A replacement implementation of DAP for Python is [debugpy](https://pypi.org/project/debugpy/), which is based on the development branch (5.x) of ptvsd. All future development is going to be happening in debugpy; there will be no further commits to or releases of ptvsd.
> 
> If you're using the [Python extension for Visual Studio Code](https://github.com/microsoft/vscode-python), you should be switched to debugpy automatically, so long as you're using an up-to-date version of the extension. Please file any bug reports directly in the [debugpy repository](https://github.com/microsoft/debugpy).

PR #3183, opened 2024-08-19
issue #3182: replace deprecated ptvsd debugger by debugpy

rngadam:
issue #3182 

keeping ptvsd environment variables for backward compatibility

### DIFF
--- a/aries_cloudagent/__main__.py
+++ b/aries_cloudagent/__main__.py
@@ -20,16 +20,16 @@ def init_debug(args):
 
     # --debug to use microsoft's visual studio remote debugger
     if ENABLE_PTVSD or "--debug" in args:
+        PTVSD_HOST = os.getenv("PTVSD_HOST", "localhost")
+        PTVSD_PORT = os.getenv("PTVSD_PORT", 5678)
         try:
-            import ptvsd
+            import debugpy
 
-            ptvsd.enable_attach()
-            print("ptvsd is running")
+            debugpy.listen((PTVSD_HOST, PTVSD_PORT))
             print("=== Waiting for debugger to attach ===")
-            # To pause execution until the debugger is attached:
-            ptvsd.wait_for_attach()
+            debugpy.wait_for_client()
         except ImportError:
-            print("ptvsd library was not found")
+            print("debugpy library was not found")
 
     if ENABLE_PYDEVD_PYCHARM or "--debug-pycharm" in args:
         try:
--- a/aries_cloudagent/__main__.py
+++ b/aries_cloudagent/__main__.py
@@ -20,13 +20,13 @@ def init_debug(args):
 
     # --debug to use microsoft's visual studio remote debugger
     if ENABLE_PTVSD or "--debug" in args:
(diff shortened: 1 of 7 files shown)


swcurran:
Thanks for the update.  Can you please search the documentation (*.md files) for references to ptvsd, and update changes in those as needed.  Also, are there any places in the dockerfiles that need to be updated?  Might not, but I have a faint memory there were some things there.

rngadam:
> Thanks for the update. Can you please search the documentation (*.md files) for references to ptvsd, and update changes in those as needed. Also, are there any places in the dockerfiles that need to be updated? Might not, but I have a faint memory there were some things there.

ENABLE_PTVSD, PTVSD_HOST, PTVSD_PORT should keep working as is, but with debugpy as the underlying implementation. Is that ok? If that's acceptable, then we don't need to change wherever those environment variables are used. Otherwise yes, there are a couple places where we'd need to rename PTVSD for DEBUGPY.

There are two places I ought to update in any cases:

* docs/features/DevReadMe.md (link to debugpy instead of ptvsd)
* conftest.py (the test won't work here)

swcurran:
I think it is OK to keep the existing config parameters, although in retrospect it is unfortunate that they weren’t just “DEBUG” in the first place.  Who knew the implementation would change :-).  Suggest at least updating the config parameters information to note that debugpy is being used instead.

OK — one more question.  Is it easy in argparse to have two names for the same config?  If that were possible we could have both the PTVSD* and DEBUG* for the config parameters.  Not a big deal...

Thanks!

rngadam:
> I think it is OK to keep the existing config parameters, although in retrospect it is unfortunate that they weren’t just “DEBUG” in the first place. Who knew the implementation would change :-). Suggest at least updating the config parameters information to note that debugpy is being used instead.

both PTVSD and debugpy are implementation of DAP. The reason it is not debug I think is because there is an alternative to debug with Pycharm too.

> OK — one more question. Is it easy in argparse to have two names for the same config? If that were possible we could have both the PTVSD* and DEBUG* for the config parameters. Not a big deal...

I added DAP_HOST and DAP_PORT as alternates environment variables. Note that this is done in __main__, not argparse before anything else starts.

Updated doc and conftest.py (although I'm not sure how to test the latter?).

rngadam:
Also, seems like the reason for ENABLE_PTVSD was to get around the fact that some of the modes of the agent would reject the --debug flag. Now that it is in its own group Debugger, all modes can be supported.

swcurran:
Thanks — I think that looks good, but leave to the dev maintainers to weigh in — @jamshale @dbluhm @amanji @ianco .

Please address the file conflict — required before merging — which looks like is also causing the synk issue.

rngadam:
> Please address the file conflict — required before merging — which looks like is also causing the synk issue.

done. BTW seems like poetry content-hash merge conflict is a long-standing issue:

https://github.com/python-poetry/poetry/issues/496

jamshale (approved):
I don't develop or debug with this method so i'm not really sure how to test it properly. I did run the example command and the debugger attached.

It looks tested and that you know what you're doing so I'll approve.

PR #3184, opened 2024-08-19
refactor: remove connection protocol

dbluhm:
This PR Removes the deprecated connections protocol.

In addition to the message handlers and routes, this also means:
- CLI flags relating to connections protocol removed.
- References to specific messages in base connection manager were removed; there will need to be some working around this in the plugin to continue to support the protocol but I think it's good and necessary to at least rip it out for now here.
- Introduction protocol used to use Connections invitations. I did a very rough-and-ready conversion to OOB invites.
- Connections management routes (e.g. querying connections) were moved to `aries_cloudagent.connections.routes` (from `aries_cloudagent.protocols.connections.v1_0.routes`).

I am expecting this to not be fully functional yet so opening as a draft for now.


PatStLouis:
@dbluhm it was mentioned on the call this was moved to a plugin. Is the code to enable using that plugin included in this pr?

swcurran:
It’s in the plugin repo: https://github.com/hyperledger/aries-acapy-plugins/pull/925

PatStLouis:
@swcurran thank you, I was referring towards the "code required to make this plugin work" which was highlighted in the meeting. 

The functionality lost/broken features [outlined here](https://github.com/hyperledger/aries-acapy-plugins/pull/925#issuecomment-2297477787), will this require code from within the aca-py code base or it can all be managed in the external plugin?

swcurran:
Doh…sorry about that. Good point.  And per my comment in the meeting — is it viable to simply document how to use the plugin?  I suppose not being able to use the artifacts directly (e.g., having to change the Python code) might be a bit too painful.

dbluhm:
With the most recent updates on this branch and to the corresponding plugin, I have successfully completed a connection using the plugged in connection protocol. There are some failing tests I will address but I will go ahead and mark this as ready for review.

jamshale (commented):
This looks great! I think this deprecation support path is really going to help manage complexity.

I'm just wondering if maybe a scenario test that installs the plugin and creates a v1 connection might be a good idea.

I don't see any reason not to approve this. We'll need good docs for when it gets released. Guaranteed people are going to upgrade without realizing this change has happened.

swcurran:
Cool.  Let’s include in our discussion tomorrow the version number if we remove this.  1.1.0?  

Should the SonarCloud analysis be addressed before merging?

dbluhm:
> I'm just wondering if maybe a scenario test that installs the plugin and creates a v1 connection might be a good idea.

We can do that though this is basically what's happening in the integration test of the plugin itself. So I'm not sure how valuable it is for it to be tested in both places.

> We'll need good docs for when it gets released.

Indeed, I have not adequately addressed this need yet.

> Should the SonarCloud analysis be addressed before merging?

I am not too concerned personally. The most complaints are coming from code that was just moved but detected as new in the connection routes (moved from the protocols.connections.routes to connections.routes).

dbluhm:
This PR has been refreshed after growing stale. It is ready for review and merge whenever we're ready with it.

swcurran:
@jamshale @loneil and other @openwallet-foundation/acapy-admins and @openwallet-foundation/acapy-committers -- please review and let's get this merged before it goes stale again.  I love the "new code not tested" since this is a PR about removal... I assume we can ignore that, but would like a dev to review/comment.

Thanks

jamshale (approved):
I can't find anything wrong with this. Looks good to me 👍 . Happy to get all this complexity removed from the core.

The most important thing is the plugin works...

acapy_agent/protocols/connections/v1_0/tests/test_routes.py
```diff
@@ -1,828 +0,0 @@
...
-        }
-        self.request = mock.MagicMock(
-            app={},
-            match_info={},
-            query={},
-            __getitem__=lambda _, k: self.request_dict[k],
-            headers={"x-api-key": "secret-key"},
-        )
-
-    async def test_connections_list(self):
```
ff137, line 36: Some of these tests should not be deleted, because they are still kept in the connections routes module. That's why SonarCloud reports the uncovered lines. Can use [their report](https://sonarcloud.io/component_measures?id=openwallet-foundation_acapy&pullRequest=3184&metric=new_coverage&selected=openwallet-foundation_acapy%3Aacapy_agent%2Fconnections%2Froutes.py) to check which tests should remain

ff137 (changes requested):

swcurran (approved):

acapy_agent/protocols/connections/v1_0/tests/test_routes.py
ff137: I still think it's a good idea to re-add those unit tests :eyes:

swcurran:
Sorry — I junped the gun on this one. I also should have waited until the connections plugin was in place.  I think there will be some falling out with this one before it is complete.

ianco:
> Sorry — I junped the gun on this one. I also should have waited until the connections plugin was in place. I think there will be some falling out with this one before it is complete.

Also this is now breaking some OATH tests tat are still using connection protocol ...

swcurran:
Yes — expected. Part of the fix/cleanup will be adding the plugin.  We’ll just live with it for now — I’ll try to deal with it later.  @ianco — much more concerned  about getting the intermittent fix resolved in OATH and integration tests here.

acapy_agent/protocols/connections/v1_0/tests/test_routes.py
dbluhm: Ah, good catch, I intended to move these tests along with the routes.py file to the core connections package. I will open a PR restoring these tests (the ones that are still relevant).

PR #3194, opened 2024-08-22
Change integration testing

jamshale:
Changes the automated integration testing so that there is 3 different types of tests.
1. ***BDD Interop tests (AATH)***- During a PR the workflow will clone AATH and then locate the forked repo location. Adjust the target for the tests to the forked repo and branch and run the acapy <-> acapy tests. These take about 43 mins. A little bit longer than the existing integration tests. These tests are more thorough and we want to encourage coverage here for the interop profile. The biggest problem here is that tests will need to be updated in AATH if they fail from aca-py, but I think that this shouldn't happen often as the interop protocols should be stable.
3. ***Local BDD Tests*** - These are the current integration tests, but they have been scoped down considerably so they don't cover the same things as the AATH tests. They will still be available for those who like the Behave/Gherkin and demo agent style of tests. These only take about 6 minutes now and only include the endorsement tests and the anoncreds upgrade. We could still make it so all the tests run on releases. I want to make sure we aren't losing coverage removing all these tests, but I don't think we are.
4. ***Scenario testing*** - These tests will use the acapy-controller (minimal-example) library created by Indicio. The examples have been copied from their repo as base tests. I thought about doing something trickier by cloning the repo but don't think it's necessary. If Indicio wants to copy the tests from acapy or vice versa then we can just copy and paste them. These tests only take about 3 minutes right now.

The scenario testing could probably be extended to have for e2e tests like what happens in the aries-acapy-tools repo. I imagine being able to test things like shut downs and upgrades and changing configurations from this setup.

***TODO:***
 - Adjust the AATH tags and run a smaller subset on PR's
 - What to do for nightly/release/manual workflow runs? Currently same as existing.
 - Tests are all pointing to http://test.bcovrin.vonx.io/ ledger and tails server. Maybe change to `dev` so we can wipe the env more often?


jamshale:
Going to try and fix these dockerfile sonarcloud warnings.

docs/testing/BDDTests.md
```diff
@@ -28,7 +28,7 @@ cd aries-cloudagent-python/demo
 
 Note that an Indy ledger and tails server are both required (these can also be specified using environment variables).
 
-Note also that some tests require a ledger with TAA enabled, how to run these tests will be described later.
+Note also that some tests require a ledger with TAA enabled, how to run tEhese tests will be described later.
```
esune, line 31: Nit.: typo

docs/testing/IntegrationTests.md
```diff
@@ -0,0 +1,25 @@
...
+3. Scenario testing.
+
+## Interop profile (AATH) BDD tests
+
+Interoperability is extremely important in the aries community. When implementing or changing features that are included in the [aries interop profile](https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0302-aries-interop-profile/README.md) the developer should try to add tests to this test suite. 
+
+These tests are contained in a separate repo [AATH](https://github.com/hyperledger/aries-agent-test-harness). They use the gherkin syntax and a http back channel. Changes to the tests need to be added and merged into this repo before they will be reflected in the automatic testing workflows. There has been a lot of work to make developing and debugging tests easier. See (AATH Dev Containers)[https://github.com/hyperledger/aries-agent-test-harness/blob/main/AATH_DEV_CONTAINERS.md#dev-containers-in-aath]
+
+The tests will then be ran for PR's and scheduled workflows for ACA-Py <--> ACA-Py agents. These tests are important because having them allows the AATH project to more easily test credo-ts <--> ACA-Py scenarios and ensure interoperability with mobile agents using interacting with python agents.
```
esune, line 14: Did you mean to only use one of `using interacting with`?

scenarios/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.10
```
esune, line 1: Should we use Python 3.12 instead, since it is what the latest release is based on?

scenarios/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "scenarios"
+version = "0.1.0"
+description = ""
+authors = []
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.10"
```
esune, line 9: Same as for `scenarios/Dockerfile`, might want to use Python 3.12

.github/workflows/bdd-integration-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: acapy-integration-tests
+name: BDD Integration Tests
```
esune, line 1: Might want to scope test execution to `paths` that contain code and/or test definitions, so that tests are not executed when no code changes are submitted. Same for other workflows.

esune (commented):
Left a couple minor comments, but overall looks good to me.
Still wondering about executing ALL interop tests every time a PR gets opened/updated/reopened as it is potentially a big overhead and use of executor time. We could *maybe* create a matrix mapping code paths with test tags and dynamically specify which sets to execute (e.g.: only issue-credential and not connections) if nothing has changed, but this would definitely be for a future improvement 🤔

scenarios/Dockerfile
jamshale: This is because the library acapy-minimal-example library targets 3.10. This is an issue with using an external library for this. I believe it would be pretty straightforward to get a PR for the indicio library to upgrade to 3.12 and then we could upgrade it here.

We are also targeting the `main` branch of the indicio github. Targeting a release tag would be a better strategy.

.github/workflows/bdd-integration-tests.yml
jamshale: Ya. That is a good idea. I could add that to this PR.

jamshale:
> Left a couple minor comments, but overall looks good to me. Still wondering about executing ALL interop tests every time a PR gets opened/updated/reopened as it is potentially a big overhead and use of executor time. We could _maybe_ create a matrix mapping code paths with test tags and dynamically specify which sets to execute (e.g.: only issue-credential and not connections) if nothing has changed, but this would definitely be for a future improvement 🤔

I do think that's a good idea and I could do it in an upcoming PR. I think we should only run a subset that only includes most of the happy path flows in AATH and the new scenarios. They will catch a vast majority of mistakes. Then for nightly and release runs we could do a more thorough testing suite on off hours.

ianco (dismissed):

jamshale:
@esune I fixed the spelling mistakes and change the acapy-controller dependency in scenarios to target a release.

I'm going to create 3 tickets for additional work.
1. Scoping the PR tests from AATH down
2. Only run integration tests when there's source code or testing changes.
3. Rework the nightly and release testing with scenarios and AATH testing.

esune (approved):

Issue #3202, opened 2024-08-28
Noisy logs


rngadam:
At the agent start, the stdout of the agent shows multiple banner messages and output from print statements.

Having noisy logs makes it more difficult to identify quickly actual problems. 

Expect to:

*  be able to suppress those banner messages 
* not have print statements (should be logging)

## list of print statements

https://gist.github.com/rngadam/3cccd0f515091544f745d681d8119ae4

WadeBarnes:
The deprecation banner and notices should only be displayed when you are using deprecated features.   Suppressing these messages defeats the purpose of the messages - to let you know you are using deprecated features and you should migrate to alternate (replacement) features.  So, the best practice way to get rid of the messages is to migrate away from the deprecated features you're using.

I do agree the messages should be logged rather than being printed.

rngadam:
> The deprecation banner and notices should only be displayed when you are using deprecated features. Suppressing these messages defeats the purpose of the messages - to let you know you are using deprecated features and you should migrate to alternate (replacement) features. So, the best practice way to get rid of the messages is to migrate away from the deprecated features you're using.

I expected that too, but if you look at the code the banner is printed everytime (there's no conditional):

from here:

https://github.com/hyperledger/aries-cloudagent-python/blob/90895bbc3f2bf203bbfe35f0464a9ea1f37ebeaa/aries_cloudagent/core/conductor.py#L321-L338

the only condition for printing the deprecation notice is having transports configured (aka transport is not disabled)

https://github.com/hyperledger/aries-cloudagent-python/blob/90895bbc3f2bf203bbfe35f0464a9ea1f37ebeaa/aries_cloudagent/config/logging.py#L389-L429

It's giving a deprecation notice whatever config and even though #2368 is not done yet. Correct me if I'm wrong, but there seems to be no way to "correctly" configure the agent to get rid of the deprecation notice.

rngadam:
> Regarding the standard ACA-Py startup banner. I think this is a very important piece of diagnostic information that should not be suppressed. It provides you with a summary of the agent configuration and version that can help you identify issues right off the bat.

Perhaps in regular usage but it interferes with testing (I'm forking 100 agents from the same parent process, creating a storm of output) and creates the habit of eventually ignoring output that could/should be important.

WadeBarnes:
[...]

PR #3203, opened 2024-08-28
introduce ACAPY_NO_BANNER--no-banner and replace print statements by logging in plugins

rngadam:
see issue #3202

### DIFF
--- a/aries_cloudagent/anoncreds/default/did_indy/registry.py
+++ b/aries_cloudagent/anoncreds/default/did_indy/registry.py
@@ -41,7 +41,7 @@ class DIDIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
 
     async def setup(self, context: InjectionContext):
         """Setup."""
-        print("Successfully registered DIDIndyRegistry")
+        LOGGER.info("Successfully registered DIDIndyRegistry")
 
     async def get_schema(self, profile: Profile, schema_id: str) -> GetSchemaResult:
         """Get a schema from the registry."""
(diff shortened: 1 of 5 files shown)


WadeBarnes:
See my comments here; https://github.com/hyperledger/aries-cloudagent-python/issues/3202#issuecomment-2315548060.  I don't think suppressing deprecation notices is a good idea, it defeats their purpose.

WadeBarnes (changes requested):
See my comments here; https://github.com/hyperledger/aries-cloudagent-python/issues/3202#issuecomment-2315548060. I don't think suppressing deprecation notices is a good idea, it defeats their purpose.

rngadam:
> See my comments here; [#3202 (comment)](https://github.com/hyperledger/aries-cloudagent-python/issues/3202#issuecomment-2315548060). I don't think suppressing deprecation notices is a good idea, it defeats their purpose.
> 
> Regarding the standard ACA-Py startup banner. I think this is a very important piece of diagnostic information that should not be suppressed. It provides you with a summary of the agent configuration and version that can help you identify issues right off the bat.

Outputting the banner is the default. ACAPY_NO_BANNER allows to suppress the banner output in situation where the information in diagnostic is already known to the user.

swcurran:
I've added this to the ACA-Pug meeting coming up this Tuesday, Sept. 3 at 8:00 Pacific / 17:00 Central Europe.  @rngadam -- would you be able to make that?  Details are here: https://wiki.hyperledger.org/display/ARIES/2024-09-03+Aries+Cloud+Agent+-+Python+Users+Group+Community+Meeting

jamshale:
Closing this as won't do. Feel free to re-open it if this is still something you really think is needed.

PR #3226, opened 2024-09-11
fix: anoncreds revocation notification when revoking

thiagoromanos:
When revoking a credential using anoncreds, if you set `publish: true` and `notify:true`, it wasn't sending the notification to the holder, because the publish part was happening **_before_** the notification "enqueue" part. Changing the order of things fixes the problem.

### DIFF
--- a/aries_cloudagent/revocation_anoncreds/manager.py
+++ b/aries_cloudagent/revocation_anoncreds/manager.py
@@ -145,6 +145,18 @@ class RevocationManager:
                 f"No revocation registry record found for id: {rev_reg_id}"
             )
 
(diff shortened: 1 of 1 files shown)


jamshale (approved):
I'm not sure how this got messed up because it's in the correct order for non-anoncreds, and looks like it's always been this way.

A side note. This scenario should get a new scenario integration test. I'm going to create a ticket for it. This should be tested with automated testing.

Thanks for fixing it. Your fix makes sense.

thiagoromanos:
@jamshale , can you merge this? I don't have permission on this repo

PR #3230, opened 2024-09-12
Only run integration tests on correct file changes

jamshale:
Only runs the integration tests when there is changes that should be tested.

Scenario Tests:
- scenarios or aries-cloudagent or poetry.lock directories change

BDD Integration Tests:
- demo or aries-cloudagent or poetry.lock directories change

BDD Interop Tests:
- aries-cloudagent or poetry.lock directory change

This is to speed things up and not put as much load on the remote test resources. Possibly, we could also stop running the unit tests for non-code changes. Although then the sonarcloud scan wouldn't run on it.

### DIFF
--- a/.github/workflows/bdd-integration-tests.yml
+++ b/.github/workflows/bdd-integration-tests.yml
@@ -28,14 +28,29 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Check changed files
+        id: check-changed-files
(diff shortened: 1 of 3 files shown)


jamshale:
Ugh. Forgot that all these tests should run if any libraries change.

jamshale:
Do not merge. Testing poetry.lock changes.

ianco (approved):
LGTM

PR #3232, opened 2024-09-13
Add anoncreds profile basic scenario test

jamshale:
Adds a basic test for the anoncreds profile and endpoints. Creates schemas, cred defs and revocation objects, issuance and presentation between alice and bob (both anoncreds profiles)

I'd like to do more testing, including a migration, but this at least tests the happy path including revocation notification.


jamshale:
The sonarcloud stuff can be ignored imo.

ianco (approved):

swcurran (approved):

PR #3237, opened 2024-09-16
did:tdw resolver

jamshale:
- Uses the trustdidweb-py library to implement the resolver. 
- Plugs the resolve endpoint and library into the base_resolver class for caching and future other resolution.


aries_cloudagent/wallet/did_method.py
```diff
@@ -90,6 +90,13 @@ def holder_defined_did(self) -> HolderDefinedDid:
     holder_defined_did=HolderDefinedDid.NO,
 )
 
+TDW = DIDMethod(
+    name="tdw",
+    key_types=[ED25519, X25519],
```
jamshale, line 95: unsure about this...

jamshale:
There's going to be an update to the spec so I'm not going to work on this until it gets approved and the library gets updated.

swcurran:
I don’t think the update to the spec will impact what is in this PR. AFAIK, it is making a call to a library. The call will not be impacted by the spec change, which is just some details around the DID Log format.

OK to leave if you think best — just noting that the spec change should not impact this.

jamshale:
OK. Sounds good. I can continue on this. I'm kind of blocked by bot getting it published. There's a pr in trustdidweb-py repo to do it but hasn't got completed yet.

aries_cloudagent/did/tdw/routes.py
```diff
@@ -0,0 +1,84 @@
...
+            "example": "did:tdw:QmRHpkso6BM1jRvGee5VoiS3a6wUfPAfuPnyGRBFo4rL6k:example.com",  # noqa: E501
+            "description": "Did TDW to resolve",
+        },
+    )
+
+
+class DidTdwResolveResponse(OpenAPISchema):
+    """Response schema for create DID endpoint."""
+
+    did_doc = fields.Dict(
```
jamshale, line 30: I'm thinking about just returning the diddoc dict here instead of inside this key.

swcurran:
Just as a matter of interest — if the DID to be resolved was a DID URL, could/should this code return the object that is referenced instead of the DIDDoc?  For example if what was to be resolved was `did:tdw:1234567:example.com/whois`, the resulting object would be a Verifiable Presentation.  Or maybe `did:tdw:1234567:example.com/OCA/ocabundle.json`?

jamshale:
Maybe that does make sense. After verifying and resolving the did it could get the response from the endpoint and return that. The way it is now, that responsibility would fall on the controller.

swcurran:
Perhaps not needed yet, but I think we want that in the future.  It gives the controller the choice, without having to get its hands dirty with the rules of DIDs and DID Resolution. Either it gets a valid object back that it can use, or it gets an error that something is broken with its request.

dbluhm:
> Perhaps not needed yet, but I think we want that in the future. It gives the controller the choice, without having to get its hands dirty with the rules of DIDs and DID Resolution.

I am all for continuing to push for controller simplicity and keeping the controllers concerns focused on business logic and not DID and Key minutia (except in circumstances where business requirements dictate being picky about those things).

I think there's a place for a `dereference` endpoint alongside the `resolve` endpoint. Having it be separate would help the caller be certain they won't unintentionally get a did document when they expected a verification method or vice versa.

It is notable however that I don't think there's generally a reason for the controller to resolve/dereference DIDs and DID URLs directly. The only time I've done this from a controller was when working around bugs or features not yet supported in ACA-Py. Which is perhaps a fair reason to justify their existence alone but, all going well, I think it would generally be unnecessary to call them.

aries_cloudagent/did/tdw/routes.py
dbluhm, line 1: Do we need a separate endpoint for this? The `/resolver/resolve` endpoint is already designed to hook into the DID Resolver

aries_cloudagent/did/tdw/tdw_manager.py
dbluhm, line 1: I'm a bit skeptical of the `DidManager` interface at the moment. I think it could work out but with the focus on DID Resolution in this PR, it's hard to say whether we've got the right interface or not. What do you think of omitting the interface until we have more context to inform how the interface should look?

aries_cloudagent/did/tdw/tests/test_routes.py
```diff
@@ -0,0 +1,66 @@
...
+mock_response = {
+    "@context": [
+        "https://www.w3.org/ns/did/v1",
+        "https://w3id.org/security/multikey/v1",
+    ],
+    "id": "did:tdw:Qma6mc1qZw3NqxwX6SB5GPQYzP4pGN2nXD15Jwi4bcDBKu:domain.example",  # noqa: E501
+}
+
+
+class TestAnoncredsRoutes(IsolatedAsyncioTestCase):
```
dbluhm, line 18: TestAnonCredsRoutes I assume is just a typo?

aries_cloudagent/did/tdw/tdw_manager.py
jamshale: Ya sounds good. It's not really needed at this point.

aries_cloudagent/did/tdw/tests/test_routes.py
jamshale: ugh. copy/paste error.

aries_cloudagent/did/tdw/routes.py
jamshale: I was thinking we were separating the resolve endpoint like `create, update, delete`. But looking at https://hackmd.io/06YnFvnVQLyP_BzGQjEdIg?view again, I see you left resolve out. I guess we'll just use that generic endpoint then. I had noticed that endpoint and wasn't sure If I should just use that or not.

PatStLouis:
+1 for what @dbluhm is suggesting, we should have a resolve endpoint that will return a did document and use a de reference endpoint for any other resources represented by a did url.

@jamshale are you planning to do any verification as part of the resolving?

[Did log processing](https://identity.foundation/trustdidweb/#read-resolve) is part of the resolving section in the spec.

jamshale:
@PatStLouis The library does the verification. I haven't gone over the spec with the code in detail, but basic testing seemed to be correct.

PatStLouis:
Sounds good, I wasn't aware how much of the process the library was covering. This is great!

acapy_agent/resolver/did_resolver.py
```diff
@@ -56,7 +56,9 @@ async def _resolve(
         if isinstance(did, DID):
             did = str(did)
         else:
-            DID.validate(did)
+            if not DID.is_valid(did) and not DIDUrl.is_valid(did):
```
jamshale, line 59: This allows for the DIDUrl to be resolved. I admittedly don't have a lot of knowledge here.

acapy_agent/resolver/routes.py
```diff
@@ -23,7 +24,10 @@ class W3cDID(validate.Regexp):
     """Validate value against w3c DID."""
 
     EXAMPLE = "did:ted:WgWxqztrNooG92RXvxSTWv"
-    PATTERN = DID_PATTERN
+    # Did or DidUrl regex
+    PATTERN = re.compile(
+        "^did:([a-z0-9]+):((?:[a-zA-Z0-9._%-]*:)*[a-zA-Z0-9._%-]+)(#\w+)?$"
+    )
```
jamshale, line 30: Changes the pattern to allow did url pattern. Couldn't figure out how to do it with the pydid library.

dbluhm: I'm confused why we're doing this. A DID is distinct from a DID URL?

swcurran: I suspect it is because a DID resolved to a DIDDoc, and a DID URL, resolves to whatever the resource is pointed to by the URL (AnonCreds object, OCA bundle, status list VC, etc). The DID still needs to be resolved (perhaps from a cache) before getting the resource, but the resolution result object will be different.

dbluhm: For clarity (forgive the pedantism), we resolve DIDs but dereference DID URLs. I think it's valuable to keep the two operations distinct since they're distinct in the DID Core spec. Resolving a DID results in a DID Document. Dereferencing a DID URL results in a "resource." So rather than altering the existing resolve endpoint to accept a DID URL, I'd recommend a separate endpoint for that.

jamshale: So the resolver endpoint should fail for something like `did:tdw:QmZiKXwQVfyZVuvCsuHpQh4arSUpEmeVVRvSfv3uiEycSr:example.com%3A5000#whois`, and pass for `did:tdw:QmZiKXwQVfyZVuvCsuHpQh4arSUpEmeVVRvSfv3uiEycSr:example.com%3A5000`? I guess I was a bit confused because the did_tdw library resolves `did:tdw:QmZiKXwQVfyZVuvCsuHpQh4arSUpEmeVVRvSfv3uiEycSr:example.com%3A5000#whois` 

I think updating this to just resolve dids shouldn't be a problem. I can't remember if I was getting validation errors with a valid did:tdw without the `#` or not. Will refactor it back and check.

andrewwhitehead: The `resolve_did` method which is used here does not perform dereferencing (it returns a `ResolutionResult` not a `DereferencingResult`). It does not currently reject a DID URL with a path or extra parameters, although I might make that change to avoid misuse.

swcurran: @jamshale is using `#whois` in the examples above, but I think (I hope) they should be `/whois`.  To me, `<did>#` should be for getting a fragment of the DIDDoc.  The `<did>/` is for getting a resource separate from the DIDDoc. And `whois` is an external resource.

That said, I have wondered why the `did:tdw` spec says the implicit service is `#whois`. I wrote what I (think I) was told.  Is that right to get the `/` behaviour?

andrewwhitehead: `#whois` does indeed dereference to the service definition, not the whois document itself. I believe `/whois` should be translated to `?service=whois` by the resolver and dereferenced that way.

jamshale:
Removed the did url usage in the resolver.

I think a dereference ticket should be created next as a separate task.

dbluhm (approved):
LGTM. I agree with Jamie that we can worry about dereferencing next as a separate change.

dbluhm:
@jamshale heads up on the conflict in the poetry.lock

dbluhm (approved):

PR #3244, opened 2024-09-19
feat: verify creds signed with Ed25519VerificationKey2020

dbluhm:
This PR adds support for verifying JSON-LD Credentials using a verification method of type `Ed25519VerificationKey2020`.

I'm not thrilled about this approach but I think it's the least disruptive change possible right now. Less disruptive seems like a good idea given @PatStLouis work on supporting VC-DI is already underway.

Sidebar, Patrick, hopefully the VC-DI and VCDM 2.0 implementation handles working with verification methods better than the original LDP-VC code does :sweat_smile: Working on a verification method that has been framed does not feel right.

### DIFF
--- a/aries_cloudagent/vc/ld_proofs/crypto/tests/test_wallet_key_pair.py
+++ b/aries_cloudagent/vc/ld_proofs/crypto/tests/test_wallet_key_pair.py
@@ -76,4 +76,4 @@ class TestWalletKeyPair(IsolatedAsyncioTestCase):
 
         with self.assertRaises(LinkedDataProofException) as context:
             key_pair.from_verification_method({})
-        assert "no publicKeyBase58" in str(context.exception)
+        assert "Unrecognized" in str(context.exception)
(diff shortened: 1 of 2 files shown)


dbluhm:
I would rather be able to do something like this code in the didcomm-messaging-python library:

https://github.com/Indicio-tech/didcomm-messaging-python/blob/a0f6b88d39b90552d1a639334b71cb44ce07634e/didcomm_messaging/crypto/backend/askar.py#L76-L108

PatStLouis:
sorry for my confusion but can't aca-py already do this?

jamshale (commented):
Changes look good and make sense, but I don't have much knowledge of the crypto verification method stuff so I will defer to other reviewers.

dbluhm:
> sorry for my confusion but can't aca-py already do this?

The Ed25519Signature2020 is supported for issuance and verification but only if used with an Ed25519VerificationKey2018, as it turns out. I thought the same before I ran into the problem while doing some interop testing.

jamshale (approved):

PR #3252, opened 2024-09-24
refactor: remove ICv1, PPv1

dbluhm:
This PR removes Issue Credential v1 and Present Proof v1.

This is a rough cut of these changes that will need to be further refined.

Notes on sharp edges to be added.

### DIFF
--- a/aries_cloudagent/protocols/issue_credential/v2_0/routes.py
+++ b/aries_cloudagent/protocols/issue_credential/v2_0/routes.py
@@ -1,7 +1,8 @@
 """Credential exchange admin routes."""
 
-import logging
 from json.decoder import JSONDecodeError
+import logging
+import re
 from typing import Mapping, Optional
 
 from aiohttp import web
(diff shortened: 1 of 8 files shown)


dbluhm:
I stewed on how best to handle the OOB invitation attachments for these v1 protocols when they're no longer in the core repo. I think the way I would approach this is:
- Subclass `InvitationCreator` from OOB package and override the `create_attachment` or add a method that handles creating an OOB attachment for a v1 cred issuance or proof presentation.
- Create a new endpoint in each of the respective plugins that does the OOB invite generation. I think the `InvitationCreator` should encapsulate enough of the hard stuff that having multiple endpoints for OOB invite generation shouldn't be too bad.

dbluhm:
I've rebased this PR on current main and resolved merge conflicts. My last comment about restoring the ability to attach v1 credential offers or presentation requests to OOB are still relevant. This would need to be done from the plugin side. I'll make sure my comment is present on the corresponding PR in acapy-plugins.

I believe this is otherwise ready for review. cc @swcurran @jamshale

dbluhm:
Stale; might be useful as a reference but no need to keep this open

PR #3253, opened 2024-09-24
Add did:indy transaction version 2 support

jamshale:
This adds the ability to create a `did:indy` with transaction version 2 algorithm. https://hyperledger.github.io/indy-did-method/#nym-transaction-version. 

- It's created in a new /did/indy/create endpoint. If the controller tries to create it in the /wallet/did/create endpoint it will get a `indy` method not supported (same as before) and tell them to use the new endpoint.
- Adds seed support with allow-insecure-seed configuration
- When starting up an agent wither a did:indy or did:sov seed can be used.
- Does a refactor on wallet startup function to reduce complexity.


jamshale:
These sonarcloud issues are due to the scenarios directory. I'm going to try and make it ignore them.

jamshale:
The security alerts are nothing. http used in the scenario tests. Not sure how to ignore it yet.

dbluhm:
I like what you're doing in this PR -- really appreciate the wallet startup cleanup as well. Question: how do we get the DID onto the ledger? Are we saying this is handled out of band?

jamshale:
It's the same way as a did:sov. You use the /did/indy/create and then post it to /wallet/did/public. There's no way to start up a fresh agent with a seed and a did:indy currently. Doing that with a seed still creates a did:sov.

aries_cloudagent/wallet/did_parameters_validation.py
dbluhm, line 1: You're doing as the Romans do here but I think this logic ought to be captured directly in DID registration/creation endpoints. This "centralized" logic made sense when there was a single endpoint for creating DIDs but, "decentralizing" the logic, as we've talked about doing and as you've done here for did:indy, I don't think it's necessary for this to all be in one place.

aries_cloudagent/wallet/askar.py
```diff
@@ -582,8 +582,10 @@ async def set_did_endpoint(
 
         """
         did_info = await self.get_local_did(did)
-        if did_info.method != SOV:
-            raise WalletError("Setting DID endpoint is only allowed for did:sov DIDs")
+        if did_info.method != SOV and did_info.method != INDY:
+            raise WalletError(
+                "Setting DID endpoint is only allowed for did:sov or did:indy DIDs"
+            )
```
dbluhm, line 588: Setting DID Endpoint in this way will get picked up by the legacy resolution support in Indy VDR but doesn't get picked up by the universal resolver, just as an aside. Ideally, we set this by updating the nym with diddocContent rather than with an attrib. We can address that in the future, though, of course.

dbluhm (dismissed):
Couple of quick comments but otherwise looks good!

aries_cloudagent/did/indy/routes.py
```diff
@@ -0,0 +1,76 @@
...
+        return web.json_response(
+            (await DidIndyManager(context.profile).register(body.get("seed")))
+        )
+    except WalletError as e:
+        raise web.HTTPBadRequest(reason=str(e))
+
+
+async def register(app: web.Application):
+    """Register routes."""
+    app.add_routes([web.post("/did/indy/create", create_indy_did)])
```
esune, line 59: Would it make sense (and would it be a manageable/reasonable overhead) to have a path-parameter based route to handle creation of different did types? Maybe this is a non-issue, but I think it would be nicer (and easier for users to consume) to be able to get to DID creation by just invoking `/did/{type}/create` rather than having to find the right endpoint to call.

Currently there aren't too many options (but we already have a couple different endpoints to create DIDs), however I wonder whether we're going to end-up in a situation where there is a proliferation of endpoints due to growing number of options in did method types.

jamshale: In some ways that would be nicer. One factor is how consistent the payload options are going to be. Having them separate make it easier to give detailed examples related to that did method. The idea here is there would be multiple endpoints but the did method is somewhat like a path variable. Like if I want to make a did:indy it would be /did/indy/create and if I want a did:web it would be did/web/create and this would be consistent for all did methods supported by core acapy and official plugins.

esune: Right. Your point about the consistent payload is valid. Is there a way for us to make sure the payload is consistent for new DID method registrations? I am not too familiar with options other than indy/sov right now.

At the very least we should maybe keep this pattern in mind as we move forward so we can implement it IF we see it makes sense and it is possible (I really am not a big fan of many endpoints that basically do the same thing like we already have for several functions).

jamshale: This idea is from here https://hackmd.io/06YnFvnVQLyP_BzGQjEdIg?view#Registrars but I haven't implemented this exactly this way. The endpoints would have a consistent `features` dictionary with options to support. I'm not sure if the `seed` option would fall under features. I think it's going to be hard to create a simple options payload for all these did methods. That's a main reason why we want to move away from the current `wallet/did/create` endpoint.

Possibly it would be better to get this *interface* narrowed down before adding this endpoint.

swcurran:
Given the 5 errors found (examples with an “http” protocol instead of “https” — would it be easiest to just add an “s” in the indicated places, even though it is irrelevant?

jamshale:
> Given the 5 errors found (examples with an “http” protocol instead of “https” — would it be easiest to just add an “s” in the indicated places, even though it is irrelevant?

I think we should be able to disable this rule in the sonarcloud account or `sonar-project.properties` file, or the workflows. I'll try and look into this. I don't think using https should be required here.

jamshale:
I think the sonarcloud should just be safe to ignore. It's annoying, because we should be able to mark things a safe in sonarcloud but it requires an admin configuring things. Might try and figure it out on a personal account and then ping Ry.

acapy_agent/wallet/crypto.py
```diff
@@ -63,20 +65,24 @@ def create_ed25519_keypair(seed: Optional[bytes] = None) -> Tuple[bytes, bytes]:
...
 
     """
     seed = validate_seed(seed)
     verkey, _ = create_ed25519_keypair(seed)
-    did = bytes_to_b58(verkey[:16])
-    return did
+    if method == SOV:
+        return bytes_to_b58(verkey[:16])
+    if method == INDY:
+        return f"did:indy:{bytes_to_b58(hashlib.sha256(verkey).hexdigest()[:16])}"
```
zoblazo, line 84: why hexdigest rather than digest ?

zoblazo:
* [seed parameter](https://github.com/openwallet-foundation/acapy/pull/3253/files#diff-06a0dca2d668d84c40103c92175f252f998e807e8e4d049ef9ae351a57d2a21aR1558) should not make any reference to did:indy as it only supports did:sov. It should be documented explicitely as such.
 
 * /wallet/did/create 'method' parameter is somewhat useless as did method indy has its own endpoint (and, from what I understand, so will any other new did method). so /wallet/did/create is purely for the did:sov method (and probably should become /did/sov/create to stay coherent with other did/*/create)
 * /did/indy/create cannot specify either the did or the encryption key type (as we can with /wallet/did/create). We should be able to fill out the wanted did as parameter.
 * /did/indy/create needs --wallet-allow-insecure-seed or else it won't work. That should be documented OR avoided.
 
 [seed_to_did: hexdigest vs digest](https://github.com/openwallet-foundation/acapy/pull/3253/files#diff-8ed68c8f6b9d303ed109aa1073c4bddf830f64c98c5986fb9d9ea44a7ed4cd02R84) Currently, the generated NYM does not seem to follow the specification (and does not generate same NYM as askar lib generates for a given seed). Again, we should have a way to provide the wanted did (and thus wanted NYM) to aca-py rather than have aca-py generate it for us. Please see my comments on [issue 3240](https://github.com/openwallet-foundation/acapy/issues/3240)

acapy_agent/wallet/crypto.py
jamshale: I wasn't aware askar used digest, and honestly not super confident about some stuff I'm doing here. I'll change it to digest and make sure it's working.

jamshale:
I agree with pretty much all the points here, except a couple small points:

 - the seed parameter will work for a did:indy type did that was created with the `/did/indy/create` endpoint. It checks if if the seed works for a did:sov and then tries again with did:indy. If either passes it will use that did to startup.
 - Your right about `/did/indy/create` not having a did or encryption key type param. My thinking was to get a minimal create endpoint complete and the approach verified first. I can try adding the additional parameters in this PR.
 - I agree about the `wallet/did/create` endpoint being non intuitive and more like a `/did/sov/create` endpoint. I think this endpoint should be deprecated after we define the new did management endpoints fully.

zoblazo:
> the seed parameter will work for a did:indy type did that was created with the `/did/indy/create` endpoint. It checks if if the seed works for a did:sov and then tries again with did:indy. If either passes it will use that did to startup.

What's the use-case for this ? As we already feed the seed when calling /did/indy/create, and, thus, the key pair already generated, it seems redundant to feed it again at startup after the did has been provisionned post-deployment via the API. Or am I missing something here ?

jamshale:
If you have multiple dids, you can tell it which one to use as the public did with the seed when starting up.

Do you want to start a brand new agent with the seed option and have it create the did locally at the same time? This is so you can avoid doing it with the endpoints when you initialize an agent?

zoblazo:
> If you have multiple dids, you can tell it which one to use as the public did with the seed when starting up.
> 
> Do you want to start a brand new agent with the seed option and have it create the did locally at the same time? This is so you can avoid doing it with the endpoints when you initialize an agent?

That's how we use --seed and how I interpreted its intent. I thought the public did status was stored and persisted and once a did had been promoted to public, it would stay public even after agent was restarted.

zoblazo:
> If you have multiple local dids, you can tell it which one to use as the public did with the seed when starting up. So you could create a did:sov and a did:indy and tell it which one should be the public (active) did on startup.

Even there, if I have multiple local did, and I want to specify the one I want to use as public, then I should specify the did to promote public rather than the seed.

You're right that this is probably better suited in 3240. In the scope of this PR, I'd just exclude anything related with --seed (and indy) and consider --seed to be a pure did:sov parameter for now. And then see what happens in 3240. Otherwise it easily gets confusing for those who do use --seed to provision a new (public) did and expect it to work for a did:indy.

jamshale:
I don't really think the --seed parameter should be used, but I'm not really sure of the history of it. I think that's what that ticket https://github.com/openwallet-foundation/acapy/issues/3240 is trying to figure out. I think the agent controller should create or ensure the did's are correct on startup. That's what the demo and integration tests do.

The public did is persisted, but it can be changed from one did to another. If you wanted to ensure a particular did:indy was the public did on startup you could use the --seed parameter. But yes, if you want to startup and create the did using the --seed parameter on a brand new agent then it would still create a did:sov. 

This PR was mostly focused on adding the ability to create and use did:indy dids and allowing the seed parameter was a bit of an aside. So maybe that could get removed.

zoblazo:
When I create a NYM transaction on an indy ledger with NYM transaction version == 2
When I call the `/did/indy/create` endpoint with a 'did' argument including the NYM previously created but no 'seed' argument
Then aca-py creates the requested did with a randomly generated seed and corresponding keypair
Then the generated verkey does not match the validation algorythm as documented in the specification ( did = Base58(Truncate_msb(16(SHA256(publicKey))))) )

To keep it simple for now, a suggestion would be for the 'seed' parameter to be mandatory when the 'did' parameter is included in a given `/did/indy/create` request.

zoblazo:
Most probably this is out of scope for this PR but here's my thoughts. Please let me know if I should document this elsewhere.

Given I start acapy not providing the `--wallet-allow-insecure-seed` parameter
Given I then call the api endpoint `/did/indy/create` providing a seed
Then I get an error response `400: Insecure seed is not allowed`
Expected : DID is created, no error response.

A did:indy is necessarely a public DID. Unless I'm mistaken, there is no use-case (and technically no logic) in a private did:indy. Since now `/did/indy/create` has its own endpoint, it is to be assumed that the intent is to create a public did. So this `--wallet-allow-insecure-seed` doesn't really make sens since specifically intended for custom seed used to create a local (private) DID.

One could even argue that having to POST on `/wallet/did/public` afterwards is redundant but some use case might require to 'switch' between multiple public dids at some point in time (and it really does seem out of scope of this PR) so ... I'll keep this part out of the discussion

jamshale:
The public did concept does need to be discussed. My goal here was to maintain the same process as did:sov, but I do agree with your points. The did registration in general is something we are going to be looking at improving in the near future. Likely with ideas from this spec https://identity.foundation/did-registration/.

I don't want to do too much in this PR as there will be changes based on the upcoming discussions. 

I will address the comments about the seed option and maybe we can get it merged and then have follow up work around the did registration and public did stuff. I think the `/wallet/did/create` and public did concepts will likely become depreciated.

jamshale:
I'm going to update this and try and get it merged. I think this one did:indy method could be added to the core for now. Eventually removed with the other did:indy stuff as a plugin.

jamshale:
I'm reopening this as there was another request to get it included in acapy. Allows the creation of did:indy did's which are generated with a slightly different algorithm then did:sov's.

Because I don't know when the did and key management stuff will be tackled I think we should just get this in for now and it can be addressed along with the other types in the future.

acapy_agent/messaging/valid.py
```diff
@@ -352,8 +352,8 @@ def __init__(self):
 class IndyDID(Regexp):
     """Validate value against indy DID."""
 
-    EXAMPLE = "WgWxqztrNooG92RXvxSTWv"
-    PATTERN = re.compile(rf"^(did:sov:)?[{B58}]{{21,22}}$")
+    EXAMPLE = "did:sov:WgWxqztrNooG92RXvxSTWv"
```
esune, line 355: Should the example be `did:indy`?

esune (commented):
Code looks good to me. I can't speak about the specifics though as I am not close enough to the concepts to assess the implementation itself so will leave final approval to another maintainer.

acapy_agent/wallet/crypto.py
```diff
@@ -63,20 +65,24 @@ def create_ed25519_keypair(seed: Optional[bytes] = None) -> Tuple[bytes, bytes]:
...
 
     """
     seed = validate_seed(seed)
     verkey, _ = create_ed25519_keypair(seed)
-    did = bytes_to_b58(verkey[:16])
-    return did
+    if method == SOV:
+        return bytes_to_b58(verkey[:16])
+    if method == INDY:
+        return f"did:indy:{bytes_to_b58(hashlib.sha256(verkey).digest()[:16])}"
```
ff137, line 84: single lines with this many nested step is usually hard to read and difficult to maintain. Let's split to multi line
```py
    if method == INDY:
        # Hash the verkey, take the first 16 bytes, and convert to a base58 string
        hashed_verkey = hashlib.sha256(verkey).digest()
        did = bytes_to_b58(hashed_verkey[:16])
        return f"did:indy:{did}"
```

ff137 (commented):
LGTM :+1: just a few recommendations

acapy_agent/config/wallet.py
```diff
@@ -136,6 +109,74 @@ async def wallet_config(
...
+        public_did = await _initialize_with_public_did(
+            public_did_info, wallet, settings, wallet_seed
+        )
+    elif wallet_seed:
+        await _initialize_with_seed(
+            settings, wallet, provision, create_local_did, wallet_seed
+        )
+
+    if provision and not create_local_did and not public_did:
+        print("No public DID")
```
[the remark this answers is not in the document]
jamshale: Oh yes. I think this was how it was in this file and I didn't think to change it. I'll fix this.

jamshale:
Changed quite a few other error responses in wallet routes to BadRequest. I agree Forbidden didn't make sense for almost all of them.

ff137:
I think it may be better to handle the HTTPForbidden -> BadRequest refactoring in a separate PR. There are quite a few instances being updated, and it's technically a breaking change ... since it's a change to the 'server contract'. Idk how significant it is, given I can't tell when those errors would be raised. But that's why it feels worth isolating to make it clearer and easier to review. It was mainly just the new error being raised that I thought was best as a 400

jamshale:
Sounds good. I reverted the other error responses other than the one I added in this PR. Will create separate issue to review the Forbidden responses.

acapy_agent/config/wallet.py
```diff
@@ -136,6 +111,74 @@ async def wallet_config(
...
+            LOGGER.info(f"Created new local DID: {local_did}")
+            LOGGER.info(f"Verkey: {local_did_info.verkey}")
+    else:
+        public_did_info = await wallet.create_public_did(
+            method=SOV, key_type=ED25519, seed=seed
+        )
+        public_did = public_did_info.did
+        if provision:
+            LOGGER.info(f"Created new public DID: {public_did}")
+            LOGGER.info(f"Verkey: {public_did_info.verkey}")
```
ff137, line 139: DID and Verkey is better printed together in one log line.

I think the duplication can be avoided.

But actually, I'm not sure what `if provision` is doing here? That's more behaving like `if verbose`, because the only difference is whether it gets logged or not! So not sure if something is missing here ...

acapy_agent/config/wallet.py
```diff
@@ -30,104 +30,79 @@
...
-    public_did_info = await wallet.get_public_did()
-    public_did = None
+        LOGGER.info("Replacing public DID due to --replace-public-did flag")
+        replace_did_info = await wallet.create_local_did(
+            method=SOV, key_type=ED25519, seed=wallet_seed
+        )
+        public_did = replace_did_info.did
+        await wallet.set_public_did(public_did)
+        LOGGER.info(
+            f"Created new public DID: {public_did}, with verkey: {replace_did_info.verkey}"  # noqa: E501
```
ff137, line 98: noqa can be avoided by just splitting string into multiline

ff137 (commented):
Added some pedantic notes ...
The one about the "StartupError" -- Not sure if there's anything wrong there, was kinda just thinking out loud.

jamshale:
I'll go over this wallet startup stuff again. I was intending to refactor it without changing any behavior. But from the comments there is still some improvements that should be made.

acapy_agent/config/wallet.py
```diff
@@ -30,104 +30,79 @@
...
+        if settings.get("auto_provision", False):
+            profile = await profile_manager.provision(context, profile_config)
+            provision = True
         else:
-            print("Opened existing profile")
-        print("Profile backend:", profile.backend)
-        print("Profile name:", profile.name)
+            raise StartupError(
+                "Profile not found. Use `aca-py start --auto-provision` to create."
+            )
```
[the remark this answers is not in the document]
ff137: @jamshale I think what would make sense is just for there to be a `logger.error` line just before the StartupError gets raised. Just to ensure that if it occurs, there will always be a meaningful error log above any stack trace. If we want to avoid stack traces on startup, I think the Conductor set up method, or wherever it's called, just need to be wrapped in a try-except block with some generic exception handling and logging.

dbluhm (approved):

ff137:
This PR introduces the following exception in our existing e2e tests, when creating a connection between endorser and mt-agent
```sh
2025-04-11 14:14:32,470 aiohttp.server ERROR Error handling request from 127.0.0.6
  File "/home/aries/.local/lib/python3.12/site-packages/aiohttp/web_protocol.py", line 480, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/admin/server.py", line 146, in ready_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/admin/server.py", line 238, in debug_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/admin/server.py", line 360, in setup_context
    return await task
           ^^^^^^^^^^
  File "/usr/local/lib/python3.12/asyncio/futures.py", line 289, in __await__
    ^^^^^^^^^^
  File "/usr/local/lib/python3.12/asyncio/tasks.py", line 385, in __wakeup
    future.result()
  File "/usr/local/lib/python3.12/asyncio/futures.py", line 202, in result
    raise self._exception.with_traceback(self._exception_tb)
  File "/usr/local/lib/python3.12/asyncio/tasks.py", line 314, in __step_run_and_handle_result
    result = coro.send(None)
             ^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/admin/server.py", line 222, in upgrade_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/aiohttp_apispec/middlewares.py", line 51, in validation_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/admin/decorators/auth.py", line 84, in tenant_auth
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/protocols/out_of_band/v1_0/routes.py", line 312, in invitation_create
    invi_rec = await oob_mgr.create_invitation(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/protocols/out_of_band/v1_0/manager.py", line 651, in create_invitation
    return await creator.create()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/protocols/out_of_band/v1_0/manager.py", line 251, in create
    result = await self.handle_public(attachments, mediation_record)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/protocols/out_of_band/v1_0/manager.py", line 414, in handle_public
    return await self.handle_did(public_did, attachments, mediation_record)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/protocols/out_of_band/v1_0/manager.py", line 357, in handle_did
    endpoint, recipient_keys, routing_keys = await self.oob.resolve_invitation(
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/connections/base_manager.py", line 501, in resolve_invitation
    doc, didcomm_services = await self.resolve_didcomm_services(did, service_accept)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/.local/lib/python3.12/site-packages/acapy_agent/connections/base_manager.py", line 450, in resolve_didcomm_services
    raise BaseConnectionManagerError(
acapy_agent.connections.base_manager.BaseConnectionManagerError: Cannot connect via DID that has no associated services
2025-04-11 14:14:32,474 aiohttp.access INFO 127.0.0.6 [11/Apr/2025:14:14:32 +0000] "POST /out-of-band/create-invitation?auto_accept=true HTTP/1.1" 500 247 "-" "OpenAPI-Generator/1.2.1-20250327/python"
```

Failing tests: https://github.com/didx-xyz/acapy-cloud/actions/runs/14404985029/job/40399536477?pr=1462
(Above log from "Governance Agent Logs")

All tests pass using our acapy fork image, with this PR commit dropped:
https://github.com/didx-xyz/acapy-cloud/actions/runs/14444169894/job/40501056475?pr=1462

At a glance, I can't see exactly what change in this PR would cause the issue

ff137:
@jamshale @swcurran

jamshale:
I'm still on vacation until Wednesday. 

I thought this error was when the did doesn't have an endpoint associated with it. I don't know why the did:indy stuff would have an effect on it. 

I thought this was expected with the removal of connection v1 protocols. I changed some tests in the interop tests to create a use of a did with an associated endpoint. 

I can try to look at it when I get back.

PR #3254, opened 2024-09-25
1.0.1rc0

swcurran:
A release candidate for Release 1.0.1. Easy to create a new release if needed.

Attention should be paid to the notice in the README about the move to OWF, and the deprecation notices in both the ChangeLog and Supported RFCs file. Is the coverage sufficient?

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,104 @@
 # Aries Cloud Agent Python Changelog
 
+## 1.0.1rc0
+
+### September 25, 2024
+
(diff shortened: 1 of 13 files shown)


jamshale:
I need to adjust the integration test workflows. They aren't running on doc changes anymore and I set them to run when the `poetry.lock` file changed but not `pyproject.toml`. 

Doesn't need to block this though. The nightly runs are pretty much the same as this would have been.

jamshale (dismissed):
LGTM

dbluhm (approved):

PR #3256, opened 2024-09-26
[BUG] Handle get key operation when no tag has been set

PatStLouis:
Handle cases where a key doesn't have a tags property.

@dbluhm @jamshale small but critical bug introduced by my last PR, I was under the impression every key had tags by default but it doesn't seem like so, this will handle such cases.

### DIFF
--- a/aries_cloudagent/wallet/askar.py
+++ b/aries_cloudagent/wallet/askar.py
@@ -97,7 +97,7 @@ class AskarWallet(BaseWallet):
         if kid:
             tags = {"kid": kid}
         else:
-            tags = None
+            tags = {}
 
         try:
             keypair = _create_keypair(key_type, seed)
--- a/aries_cloudagent/wallet/askar.py
+++ b/aries_cloudagent/wallet/askar.py
@@ -192,8 +192,14 @@ class AskarWallet(BaseWallet):
         if not key:
             raise WalletNotFoundError("Unknown key: {}".format(verkey))
         metadata = json.loads(key.metadata or "{}")
-        # FIXME implement key types
-        kid = key.tags["kid"] if "kid" in key.tags else None
+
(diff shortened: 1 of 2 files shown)


aries_cloudagent/wallet/askar.py
```diff
@@ -192,8 +192,14 @@ async def get_signing_key(self, verkey: str) -> KeyInfo:
         if not key:
             raise WalletNotFoundError("Unknown key: {}".format(verkey))
         metadata = json.loads(key.metadata or "{}")
+
+        # Provisioned keys don't get tags by default
+        try:
+            kid = key.tags["kid"] if "kid" in key.tags else None
```
jamshale, line 198: You can do `key.tags.get('kid')` instead of this. The default if it doesn't exist is `None`.

PatStLouis:
another fix could be to change this line:
https://github.com/OpSecId/aries-cloudagent-python/blob/036f108a5ea4fbc4805b49be6bdc8fd02b1511f6/aries_cloudagent/wallet/askar.py#L100

from:
```python
if kid:
    tags = {"kid": kid}
else:
    tags = None
```

to 

```python
if kid:
    tags = {"kid": kid}
else:
    tags = {}
```

I believe this would be more consistent as a fix since all keys will have a tags object. Any opinion?

aries_cloudagent/wallet/askar.py
PatStLouis: thanks, I prefer this solution which I've just commited

PatStLouis: actually this might not fix it as the problem is that tags doesn't exist. See my second comment.

PatStLouis:
I made a combination of @jamshale suggestion and the latter idea.

jamshale (approved):

PR #3259, opened 2024-09-26
Adjust sonarcloud and integration test workflows

jamshale:
This adjusts the soanrcloud to ignore coverage and duplication on the entire folders for the existing folders and scenarios.

Also runs integration tests when there is a pyproject.toml file change (during releases).

The security reports still generate for the entire repo. I don't think I can filter these. We could allow users to review or mark them as safe, or only scan the aries_cloudagent source code. I think it's ok to deal with this for now.

Also change to workflow to only run on hyperledger repo.

### DIFF
--- a/.github/actions/run-unit-tests/action.yml
+++ b/.github/actions/run-unit-tests/action.yml
@@ -30,7 +30,7 @@ runs:
     - name: Tests
       shell: bash
       run: |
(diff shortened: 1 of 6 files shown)


jamshale:
The sonarcloud stuff needs to be merged into main before having effect.

swcurran (approved):

PR #3260, opened 2024-09-26
fix: enable refreshing did endpoint using mediator info

dbluhm:
This PR makes it possible to update endpoint info with mediator information after the initial endpoint publish.

The `POST /wallet/public/did` endpoint was capable of setting the endpoint with mediator info (routing keys, mediator endpoint). However, if your mediator changes after the initial publish, there wasn't a (convenient) way for the controller to refresh the endpoint with the new mediator info. This adds this functionality to the `POST /wallet/set-did-endpoint` to do just that.

### DIFF
--- a/aries_cloudagent/wallet/routes.py
+++ b/aries_cloudagent/wallet/routes.py
@@ -43,6 +43,8 @@ from ..messaging.valid import (
     NON_SD_LIST_VALIDATE,
     SD_JWT_EXAMPLE,
     SD_JWT_VALIDATE,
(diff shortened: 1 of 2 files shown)


jamie-quartech (commented):
Looks good. Think you just have to change that inject_or in the unit tests.

jamshale (approved):
Accidently commented with my other account. Looks good other than updating the unit tests.

jamshale (approved):

PR #3261, opened 2024-09-27
Data integrity routes

PatStLouis:
This supersedes #3181 with a better architecture based on the MultikeyManager class.

@jamshale @dbluhm this is mostly the same code as the other pr so review should be light.

I would like to get this in the up comming release if possible.

Adds support for 

- eddsa-jcs-2022 cryptosuite
- Data integrity proof issuance/verification
- Data integrity proof set issuance/verification

Here's the tracelog of reviews from the previous PR:

@dbluhm said:

> Nitpick but I think I'd rather have these be closer to the request handlers that use them rather than being in a separate file.

The reason I had separated them is because the wallet routes was a very bloated file. In this PR I went with the plugin model for data integrity which is a lot more manageable. The schemas are now closer to the request handlers.

@dbluhm said:

> When would we look up this context by the string data-integrity-v2? I get the feeling it might be simpler to just have DIV2_CONTEXT = "...".

This has been addressed by removing linked data support for the time being, therefore this URI is no longer needed for validation. We will review / re haul LD support when implementing eddsa-edfc-2022, it's due for improvement.

Other notable improvements:
The crypto suite has now been split into functions, each reflecting a specific algorithm form the spec (can almost be followed line by line). Links are in the doc string.

It now lives at the `/vc/di/*` endpoints, since this is the vc-data-integrity specification.

Future improvements:

- Add proof chains
- Add other cryptosuite support


aries_cloudagent/vc/data_integrity/cryptosuites/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .eddsa_jcs_2022 import EddsaJcs2022
+
+CRYPTOSUITES = {
+    "eddsa-jcs-2022": EddsaJcs2022,
+}
```
dbluhm, line 5: I don't have a problem with this as is but just to note: having a static dictionary of crypto suites means this won't be extensible by plugin. Having the ability to plugin this sort of thing has been pretty handy in the past but I think it's fair to wait until there's a need before making a pluggable cryptosuite registry.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+        https://www.w3.org/TR/vc-di-eddsa/#proof-configuration-eddsa-jcs-2022.
+        """
+        proof_config = options.copy()
+
+        assert (
+            proof_config["type"] == "DataIntegrityProof"
+        ), 'Expected proof.type to be "DataIntegrityProof'
+        assert (
+            proof_config["cryptosuite"] == "eddsa-jcs-2022"
+        ), 'Expected proof.cryptosuite to be "eddsa-jcs-2022'
```
dbluhm, line 70: Do we anticipate user input potentially failing these assertions? Or do these constitute a violation of the "interface contract?"

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+        """Transformation algorithm.
+
+        https://www.w3.org/TR/vc-di-eddsa/#transformation-eddsa-jcs-2022.
+        """
+        assert (
+            options["type"] == "DataIntegrityProof"
+        ), "Expected proof.type to be `DataIntegrityProof`"
+        assert (
+            options["cryptosuite"] == "eddsa-jcs-2022"
+        ), "Expected proof.cryptosuite to be `eddsa-jcs-2022`"
```
dbluhm, line 96: Do these assertions need to be in both proof config and transformation?

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+        Args:
+            session: ProfileSession to use during crypto operations.
+
+        """
+        super().__init__()
+        self.session = session
+        self.wallet = session.inject(BaseWallet)
+        self.key_manager = MultikeyManager(session)
+
+    async def create_proof(self, unsecured_data_document: dict, options: dict):
```
dbluhm, line 36: I think I'd like to see a model representing options rather than a plain dictionary.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        # If the verification method is a did:key: URI,
+        # we derive the signing key from a multikey value
+        if options["verificationMethod"].startswith("did:key:"):
+            multikey = options["verificationMethod"].split("#")[-1]
+            key_info = await self.key_manager.from_multikey(multikey)
+
+        # Otherwise we derive the signing key from a kid
+        else:
+            key_info = await self.key_manager.from_kid(options["verificationMethod"])
```
dbluhm, line 123: I think the special case of did:key should be encapsulated by a method on key manager, probably `from_kid`.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+
+        return await self.wallet.sign_message(
+            message=hash_data,
+            from_verkey=self.key_manager._multikey_to_verkey(key_info["multikey"]),
+        )
```
dbluhm, line 128: I think we should consider updating the wallet interface to accept a multikey. Also a bit odd to call an underscore method outside of the key manager class.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+        # If verification method is a did:key URI,
+        # we derive the multikey directly from the value.
+        if kid.startswith("did:key:"):
+            return kid.split("#")[-1]
+
+        # Otherwise we resolve the verification method and extract the multikey.
+        else:
+            verification_method = await DIDResolver().dereference(
+                profile=self.session.profile, did_url=kid
+            )
```
dbluhm, line 146: DIDResolver must be retrieved from the profile context. I don't believe this `_get_multikey` would work for anything other than did:key right now since the DIDResolver must have registered resolvers, which a fresh instance would not have.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+    async def _get_multikey(self, kid: str):
+        """Derive a multikey from the verification method."""
+
+        # If verification method is a did:key URI,
+        # we derive the multikey directly from the value.
+        if kid.startswith("did:key:"):
+            return kid.split("#")[-1]
+
+        # Otherwise we resolve the verification method and extract the multikey.
+        else:
```
dbluhm, line 143: Why bother special casing did:key? "Resolution," as it were, for did:key is cheap so might as well handle it consistently with other DID Methods, IMO.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+            verification_method = await DIDResolver().dereference(
+                profile=self.session.profile, did_url=kid
+            )
+            assert (
+                verification_method["type"] == "Multikey"
+            ), "Expected Multikey verificationMethod type"
```
dbluhm, line 149: This seems like an inappropriate use of an assertion. Do we have control over whether the resolved verification method will always be a `Multikey`? If not, this should be a more specific exception.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+
+            return verification_method["publicKeyMultibase"]
+
+    async def verify_proof(self, secured_document: dict):
```
dbluhm, line 153: I would prefer to see the return value of this method be a typed object rather than just a dictionary.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,201 @@
...
+    ):
+        """Proof verification algorithm.
+
+        https://www.w3.org/TR/vc-di-eddsa/#proof-verification-eddsa-jcs-2022.
+        """
+        multikey = await self._get_multikey(options["verificationMethod"])
+        return await self.wallet.verify_message(
+            message=hash_data,
+            signature=proof_bytes,
+            from_verkey=self.key_manager._multikey_to_verkey(multikey),
```
dbluhm, line 199: Similar comment to the sign; I think we should consider updating the wallet to directly accept a multikey value.

dbluhm (commented):
Quick pass, I'll have to come back for more review.

aries_cloudagent/vc/data_integrity/cryptosuites/__init__.py
PatStLouis: I will consider this when adding the next cryptosuite. The intent here is that the DataIntegrity plugin invokes the cryptosuite instance given the `proof.cryptosuite` string or the `options.cryptosuite` string provided

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: It's here because the spec says to check for these, I tried to transform each spec line into a code line. It should never actually fail at this point since the DataIntegrity manager will parse the data before invoking this cryptosuite.

PatStLouis: > If proofConfig.type is not set to DataIntegrityProof or proofConfig.cryptosuite is not set to eddsa-jcs-2022, an error MUST be raised that SHOULD convey an error type of [PROOF_GENERATION_ERROR](https://www.w3.org/TR/vc-data-integrity/#PROOF_GENERATION_ERROR).

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: I share the same feeling however I opted to put my feelings aside and just follow the specification. They do check for this in both algorithms:

> If options.type is not set to the string DataIntegrityProof and options.cryptosuite is not set to the string eddsa-jcs-2022, an error MUST be raised that SHOULD convey an error type of [PROOF_VERIFICATION_ERROR](https://www.w3.org/TR/vc-data-integrity/#PROOF_VERIFICATION_ERROR).

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: I can do this yes, good suggestion.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: Resolvers are inconsistent with how they return a `did:key:`, we specifically want a multikey value here. For example, the uniresolver will return a JWK from a `did:key:` resolving step. This cryptosuite defines a multikey as the verification method and this is what the manager will be expecting. did key is an easy shortcut for this since it's literally the multikey value.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: The spec defines the verification method as Multikey
https://www.w3.org/TR/vc-di-eddsa/#verification-methods

It seems like a fair assertion.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: I can make a verification result object, Is this what you are hinting at?

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: Agreed, this has been abstracted by the multikeyManager. I think the wallet should just use multikey instead of verkey, since you can derive alot more information from a multikey.

Is this something that should happen in this pr? The manager is carrying this burden for now, we can look at porting some functionalities from the multikey manager to the Key class directly in a subsequent pr

PatStLouis: verkey seems like an old artifact from indy / Ed25519VerificationKey2018

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: Ok I will have to look more into this. I'm mostly concerned for did web for now as this will be the main use case

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
dbluhm: Very interesting... That will have some fun DID Method interop implications. Even given the spec requirement, I get the feeling a more specific exception would be beneficial because there is still the opportunity for the VM type to not be a Multikey, isn't there?

An exception denoting the expectations of the spec have been violated is different than an assertion error denoting the expectations of the programmer/program have been violated. I would apply these same comments to the other instances of assertions being used in this PR.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
dbluhm: Indeed, definitely an Indy artifact. I'm cool with leaving that update for a future PR.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
dbluhm: We have control over the resolvers in ACA-Py; we can make it always return a multikey, even if we have registered the uniresolver as a resolver in ACA-Py. "Native" resolvers are given preference over "remote" resolvers.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
dbluhm: Oh, the one change I would make in this PR is accessing underscored methods. `_multikey_to_verkey` and `_verkey_to_multikey` don't even need to be methods on the MultikeyManager since they aren't accessing self. I think I'd recommend popping those out into just a function in the key manager module that can be reused inside and outside of the manager until we give the wallet the ability to handle multikeys directly.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: DIOptions model added

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: I tested the implementation and its working fine, please have a look at the changes I've made

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: got rid of this check, we are now resolving all verification method

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: I decided to accept a couple more verification method and no longer restrict on multikey

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: implemented a verification response object

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: extracted both method as well as `key_type_from_multikey`

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: This is the only thing left to address, I think this is also part of a subsequent PR to make the wallet natively recognize multikey as a key reference.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: @dbluhm it works when deploying the agent but now it's failing when calling it from the unit tests, could you provide me a hint how to resolve this issue? I'm getting the error:
`aries_cloudagent.config.base.InjectionError: No instance provided for class: DIDResolver`

When calling the verify functions from the unit tests but it works as intended when running the agent

dbluhm: You need to bind an instance of DIDResolver to the context used in the unit tests. There are several examples of this in other tests if you do a search for DIDResolver and then narrow the results to test files.

PatStLouis:
@dbluhm @jamshale I could use a hand here, I re-based on main and now 2 tests are failing which seem unrelated to the work in this PR. I had to re-lock the poetry file since I'm adding a new dependency `canonicaljson`.

PatStLouis:
I just ran the test locally and I don't get a single failure.

PatStLouis:
The error I'm getting is 
```
aries_cloudagent/protocols/issue_credential/v2_0/tests/test_routes.py::TestV20CredRoutes::test_credential_exchange_send_bound_offer_linked_data_error - TypeError: argument of type 'LinkedDataProofException' is not iterable
```

Can someone shed some light as in how my PR is affecting the `LinkedDataProofException` type?

jamshale:
I'm not working today and can have a closer look tomorrow. For the unit tests it looks like there's an error. I'm not sure why it wouldn't happen locally. But there's also a warning and the the github action fails on most warnings. There's something going on here.

For the interop integration tests the AATH just upgraded promp_toolkit and it's conflicting with the version in acapy :/ https://github.com/hyperledger/aries-agent-test-harness/pull/873. Apparently both projects use it. We'll either have to upgradeit in aca-py or revert it in AATH for this workflow to work.

jamshale:
I noticed you did a general update to the poetry.lock file. Probably using `poetry install`. Sometimes this can have unintended consequences as it will upgrade all libraries. Not sure if it is a problem here but when adding a library I'd suggest using `poetry lock --no-update` instead to only add the one library.

PatStLouis:
@jamshale thanks for the pointer, I bumped the prompt-toolkit version to match the version required by AATH (`~=3.0.48`), and it's causing even more problems since there's a lot of breaking changes. I think reverting would be preferable.

PatStLouis:
I copied the lock file from main and applied a `poetry lock --no-update` as suggested.

PatStLouis:
@jamshale ussing `--no-update` fixed the issue, thank you

PatStLouis:
Commented on the aath PR about the prompt toolkit version:
https://github.com/hyperledger/aries-agent-test-harness/pull/873#issuecomment-2384272893

jamshale:
Ya. I think I looked at prompt-toolkit a while ago and decided to stay on major version 2. I think reverting it in AATH would be the best. I'll talk to Stephen about it tomorrow.

PatStLouis:
@dbluhm all tests are now passing except the BDD which are currently broken due to a version upgrade of a package in aath. If I could have a final review before merging this would be appreciated.

jamshale:
There's a few minor issues reported by sonarcloud. I think most of them can be ignored but the redundant exception class could easily be fixed. https://sonarcloud.io/project/issues?id=hyperledger_aries-cloudagent-python&pullRequest=3261&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

PatStLouis:
@dbluhm @jamshale can I get a thumbs up so we can merge this? thank you

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
```diff
@@ -0,0 +1,210 @@
...
+            transformed_data = self.transformation(unsecured_document, proof_options)
+            proof_config = self.proof_configuration(proof_options)
+            hash_data = self.hashing(transformed_data, proof_config)
+            verified = await self.proof_verification(
+                hash_data, proof_bytes, proof_options
+            )
+            if not verified:
+                raise CryptosuiteError("Invalid signature.")
+
+        except (AssertionError, CryptosuiteError) as err:
```
dbluhm, line 175: As I mentioned over Discord the other day, catching an AssertionError is a red flag. Can you tell me about why we expect to catch AssertionErrors here?

See this discussion from ChatGPT:

> Catching an `AssertionError` in Python is generally considered an anti-pattern because assertions are meant to serve as a development-time tool, not as a mechanism for handling run-time errors. Here’s why:
> 
> 1. **Assertions are for debugging, not control flow:**
>    - Assertions are intended to catch bugs during development by verifying that conditions that "should never happen" are true. If an assertion fails, it indicates a flaw in the code that needs to be fixed. Catching the `AssertionError` implies that the code expects such situations to occur, which defeats the purpose of the assertion.
>    
> 2. **Assertions can be disabled:**
>    - Python allows assertions to be disabled in optimized mode (using the `-O` flag when running Python). When this happens, the `assert` statements are not even executed, meaning that any `AssertionError` handling code will be skipped. This can lead to inconsistent behavior when running the code in different environments.
>    
> 3. **Confusing intent:**
>    - Catching `AssertionError` may lead to confusion about the intention of the code. An assertion failure should be a sign that something has gone wrong, and attempting to handle it programmatically can obscure the root cause of the problem.
>    
> 4. **Better alternatives exist:**
>    - If you need to handle predictable or recoverable conditions, it’s better to use explicit checks (e.g., `if` statements) and raise appropriate exceptions, such as `ValueError` or `TypeError`. This approach makes the code more readable and predictable, as it clearly distinguishes between code that handles errors and code that verifies assumptions during development.
> 
> For example, consider this anti-pattern:
> 
> ```python
> try:
>     assert some_condition
> except AssertionError:
>     handle_error()
> ```
> 
> A better approach would be to explicitly check the condition:
> 
> ```python
> if not some_condition:
>     handle_error()
> ```
> 
> This makes it clear that the condition is part of the program's expected flow, rather than an assertion that something has gone fundamentally wrong.
> 
> In summary, assertions are designed to catch programmer errors during development and testing, not for error handling in production code. Catching `AssertionError` is therefore discouraged because it misuses assertions as part of control flow.

aries_cloudagent/vc/data_integrity/manager.py
```diff
@@ -0,0 +1,150 @@
...
+            proof_options.pop("proofValue")
+            proof_options = DataIntegrityProofOptions.deserialize(proof_options)
+            try:
+                self.validate_proof_options(proof_options)
+                suite = self.select_suite(proof_options)
+                input_document = unsecured_document.copy()
+                input_document["proof"] = proof
+                verification_result = await suite.verify_proof(input_document)
+
+            except (AssertionError, DataIntegrityManagerError, DIDNotFound) as err:
```
dbluhm, line 83: Same comment, catching AssertionError is questionable.

aries_cloudagent/vc/data_integrity/cryptosuites/eddsa_jcs_2022.py
PatStLouis: removed the assertion catch

PatStLouis:
@jamshale I applied a couple of the fixes suggested by sonarcloud

dbluhm (approved):
Thanks for working through my comments. I think these changes look good :+1:

PatStLouis:
Agreed and I think the responses from the api calls are a lot more intuitive.

jamshale:
@PatStLouis Are you able to merge this yourself now?

PR #3267, opened 2024-10-01
General repo updates post OWF move.

jamshale:
This is a very large pull request that changes the source directory from `aries_cloudagent` to `acapy_agent` and changes the github actions to prepare for the repo rename from `aries-cloudagent-python` to `acapy`. Also readys all the github actions to have the owner of the repo be `openwallet-foundation`.

Steps:
1. Move the repo to `openwallet-foundation`.
2. Rename repo to `acapy`.
3. Update secret variables.
4. Merge this PR.

***Repo Secrets:*** Repo secrets do get copied when moving a repo but many of them need or should to be updated.
- GITHUB_TOKEN - Should be fine.
- PYPI_USERNAME - The account where the pypi package will be hosted
- PYPI_PASSWORD - The password for the pypi account
  - There's actually a new way to do this with github actions that is preferred and easy to implement. See this PR. https://github.com/decentralized-identity/trustdidweb-py/pull/9. Should consider changing it.
 - SNYK_TOKEN - I think a token for the new account will be needed for this. 
 - SONAR_TOKEN - Sonarcloud token pointing to the correct project name `openwallet-foundation_acapy` will be needed. Free tier works fine for us. Allows opensource repos PR and main branch scanning.

PR #3270, opened 2024-10-03
Update MDs and release publishing files to reflect the repo move to OWF

swcurran:
A complementary PR to the one that @jamshale made - #3267. His moves the technical components -- code and GHAs -- while this one focuses on documentation and the release/publishing process. I don't think there is a lot of overlap between them -- hopefully none.


WadeBarnes (commented):
Would it be possible to use relative links to the various files rather than an explicit URL?  Since you're having to update all the references anyway, it may be a good time to do this.

swcurran:
It is not always straight forward. The links have to work in both the GitHub repo context and on https://aca-py.org, so there are places where I just use the hardcoded link (as you have seen), some where there is an update made to the link in generating the gh-pages site, and others that “just work”.   Things that ONLY exist in the GitHub repo MUST be absolute links for them to work on the gh-pages site.   I did clean up a bunch of them in this PR, but not all can be done.

So, bottom line, no.

sonar-project.properties
```diff
@@ -1,5 +1,5 @@
-sonar.projectKey=hyperledger_aries-cloudagent-python
-sonar.organization=hyperledger
-sonar.projectName=aries-cloudagent-python
+sonar.projectKey=penwallet-foundation_acapy
```
jamshale, line 1: I'd revert this change. It's covered in my PR. If you do include it please fix the spelling mistake.

swcurran: Doh!! Should have known better.  Reverted.

jamshale (approved):

PR #3279, opened 2024-10-09
More robust verification method selection by did

dbluhm:
This PR is intended to supersede #3138 by adding a default verification method selection strategy for all DID methods.

This strategy relies on proof type and proof purpose to select the most appropriate verification method. For example, if proof type is `Ed25519Signature2020` and purpose is `assertionMethod`, the changes in this PR will select the first matching method in the `assertionMethod` relationship that can produce a `Ed25519Signature2020`.

This places more expectations on the DID -- specifically that it is well formatted (which many aren't). By well formatted I mean that verification methods used for issuance are properly referenced in `assertionMethod`, methods used to authenticate the DID owner are in `authentication`, etc. In practice, this may cause challenges but it is possible for users who depend on a DID method/document that isn't well formatted to create their own strategy and plug it in. So I'm comfortable making the assertion that we can expect the DIDs we're working with to be well formatted.

This PR does adjust the interface for the VerificationKeyStrategy to better account for this. This might be painful to some plugin users who've added DID Methods by plugin and added a strategy to match. However, my hope is that those same users won't need to plug in their own strategy anymore with these changes since they should be robust enough to cover most use cases.

Shout out to @aritroCoder for his original work on #3138!

### DIFF
--- a/acapy_agent/wallet/default_verification_key_strategy.py
+++ b/acapy_agent/wallet/default_verification_key_strategy.py
@@ -7,6 +7,8 @@ from acapy_agent.core.profile import Profile
 from acapy_agent.did.did_key import DIDKey
 from acapy_agent.wallet.key_type import KeyType
 
+from acapy_agent.resolver.did_resolver import DIDResolver
+
 
 class BaseVerificationKeyStrategy(ABC):
     """Base class for defining which verification method is in use."""
@@ -60,5 +62,11 @@ class DefaultVerificationKeyStrategy(BaseVerificationKeyStrategy):
         elif did.startswith("did:sov:"):
             # key-1 is what uniresolver uses for key id
             return did + "#key-1"
-
+        elif did.startswith("did:web:"):
+            did_resolver = profile.inject(DIDResolver)
+            did_document = await did_resolver.resolve(profile=profile, did=did)
+            verification_method_list = did_document.get("verificationMethod", {})
+            # taking the first verification method from did document
+            verification_method_id = verification_method_list[0].get("id")
+            return verification_method_id
         return None
--- a/acapy_agent/wallet/default_verification_key_strategy.py
+++ b/acapy_agent/wallet/default_verification_key_strategy.py
@@ -77,7 +77,7 @@ class DefaultVerificationKeyStrategy(BaseVerificationKeyStrategy):
                 verification_method_types = self.key_types_mapping[proof_type]
                 verification_method_list = did_document.get("verificationMethod", None)
                 for method in verification_method_list:
-                    if method.get("type") == verification_method_type:
+                    if method.get("type") in verification_method_types:
                         return method.get("id")
             else:
                 # taking the first verification method from did document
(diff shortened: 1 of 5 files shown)


dbluhm:
@PatStLouis please reivew (I couldn't formally request your review for some reason)

dbluhm:
Because of the changes to the selection strategy from 3138, it's possible I haven't fully accounted for the original issue yet. I will take a closer look. I suspect we just need the prover side of the exchange to use the correct proof purpose when determining vm id.

dbluhm:
@jamshale your input on BDD interop test failure would be appreciated. Is this an OWF migration artifact or something else?

jamshale:
> @jamshale your input on BDD interop test failure would be appreciated. Is this an OWF migration artifact or something else?

Ya. I changed a reference I shouldn't have. Should have a fix ready soon.

acapy_agent/wallet/default_verification_key_strategy.py
```diff
@@ -55,10 +88,50 @@ async def get_verification_method_id_for_did(
         :params proof_purpose: the verkey relationship (assertionMethod, keyAgreement, ..)
         :returns Optional[str]: the current verkey ID
         """
+        proof_type = proof_type or "Ed25519Signature2018"
```
PatStLouis, line 91: I would like to suggest a slight nudge towards `Ed25519Signature2020` as default option simply to hint at driving the community forward. `Ed25519Signature2018` is a thing of the past.
```suggestion
        proof_type = proof_type or "Ed25519Signature2020"
```

PatStLouis (changes requested):
Looks good to me. A few assumptions are being made about proof purposes but I think the decisions here makes sense.

PatStLouis:
@dbluhm thanks for doing this work. @aritroCoder does this cover your use case?

acapy_agent/wallet/default_verification_key_strategy.py
dbluhm: I don't disagree that we should be heading towards bigger and better; this default was motivated by backwards compatibility. I don't know how much of an impact it would have to change the default so I'm reluctant to do so until that's well known.

PatStLouis: That's fair. This is not likely a feature I will be using so I'm not too opinionated about it.

jamshale (commented):
Looks good to me but I'm not very familiar with the use case or consequences. The approach seems to make sense.

dbluhm:
I need to fix unit tests. Changes should be captured and tested at least in the json-ld scenario integration test but might be interesting/a good idea to have a more directed test of the functionality.

swcurran:
Want to defer a merge until we get 1.1.0 release out.

dbluhm:
Unit tests fixed; this PR is ready for final review and merge

acapy_agent/wallet/default_verification_key_strategy.py
```diff
@@ -55,10 +88,55 @@ async def get_verification_method_id_for_did(
...
+        # Filter methods by type expected for proof_type
+        methods = [vm for vm in methods if vm.type in method_types]
+        if not methods:
+            raise VerificationKeyStrategyError(
+                f"No matching verification method found for did {did} with proof "
+                f"type {proof_type} and purpose {proof_purpose}"
+            )
+
+        if len(methods) > 1:
+            LOGGER.info(
```
PatStLouis, line 131: This is great

PatStLouis (approved):
LGTM

PR #3282, opened 2024-10-10
Update interop fork location after AATH update

jamshale:
AATH got updated so the text replacement needed to be changed again when running as a pull request. Other runs will simply use the location defined in AATH.

### DIFF
--- a/.github/workflows/bdd-interop-tests.yml
+++ b/.github/workflows/bdd-interop-tests.yml
@@ -67,7 +67,7 @@ jobs:
           if [ ${{ github.event_name }} == 'pull_request' ]; then
             echo ${{ fromJson(steps.get_pr_data.outputs.data).head.repo.html_url }}
             echo ${{ fromJson(steps.get_pr_data.outputs.data).head.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bdd-interop-tests.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           if [ ${{ github.event_name }} == 'pull_request' ]; then
             echo ${{ fromJson(steps.get_pr_data.outputs.data).head.repo.html_url }}
             echo ${{ fromJson(steps.get_pr_data.outputs.data).head.ref }}
-            sed -i 's|aries-cloudagent\[indy, bbs, askar\]@git+https://github.com/hyperledger/aries-cloudagent-python@main|acapy-agent[indy, bbs, askar]@git+${{ fromJson(steps.get_pr_data.outputs.data).head.repo.html_url }}@${{ fromJson(steps.get_pr_data.outputs.data).head.ref }}|g' ./aries-agent-test-harness/aries-backchannels/acapy/requirements-main.txt
+            sed -i 's|@git+https://github.com/openwallet-foundation/acapy@main|@git+${{ fromJson(steps.get_pr_data.outputs.data).head.repo.html_url }}@${{ fromJson(steps.get_pr_data.outputs.data).head.ref }}|g' ./aries-agent-test-harness/aries-backchannels/acapy/requirements-main.txt
```
swcurran, line 70: Since AATH has move to become OATH — should the “aries-agent-test-harness” stuff be changed as well?

jamshale: Ya. I could change that now. It seemed to work fine because of the redirects but would rather change it now.

jamshale:
Updated to clone from https://github.com/openwallet-foundation/owl-agent-test-harness.git

swcurran (approved):

PR #3291, opened 2024-10-14
Change pypi upload workflow to use pypa/gh-action-pypi-publish

jamshale:
This changes the pypi upload workflow to use the action `pypa/gh-action-pypi-publish` that doesn't require a username and password secret.

Note that it does require `id-token: write` privileges. It should have this already.

@ryjones will need to fill this in when creating a publisher in pypi.

image


ryjones:
@jamshale this is set up at pypi


Pending project name | Publisher | Repository | Workflow
-- | -- | -- | --
acapy-agent | GitHub | [openwallet-foundation/acapy](https://github.com/openwallet-foundation/acapy)| pythonpublish.yml

swcurran (approved):

jamshale:
@swcurran Re-running it runs the same cached workflow file which has changed now. Unfortunately we'll need to do another release candidate because this workflow is triggered by `on: release: types: [created]`.

I was considering making this workflow able to be run manually, but I didn't want to mess with the original design too much.

swcurran:
I wondered about that :-).  OK - no problem, I’ll get a new release done today.

PR #3296, opened 2024-10-15
:arrow_up: Update lock file

ff137:
Includes many new patch releases.
Some noteworthy ones appear to be:
- `aiohttp` (3.10.5 -> 3.10.10)
- `pydantic` (2.8.2 -> 2.9.2)
- `yarl` (1.9.4 -> 1.15.2)


ff137:
Looks like the dependency upgrades has unearthed a buggy unit test:
```
FAILED acapy_agent/protocols/issue_credential/v2_0/tests/test_routes.py::TestV20CredRoutes::test_credential_exchange_send_bound_offer_linked_data_error - TypeError: argument of type 'LinkedDataProofException' is not iterable
```

Taking a look

ff137:
aiohttp probably had a minor change to their exception handling.
We were passing a `LinkedDataProofException` object as the "reason" for the HTTPBadRequest. Which now results in a TypeError. This is fixed by converting the exception to a string

jamshale (approved):

PR #3309, opened 2024-10-23
Fix endorsement setup with existing connection

jamshale:
Moves the invitation url check until after checking for a previous endorser connection.


esune (approved):
LGTM 👍🏻

PR #3311, opened 2024-10-24
Remove in memory wallet

jamshale:
- Removes the in-memory wallet and updates tests to use an in-memory sqlite askar wallet.
- Doesn't flatten the wallet abstraction in this PR. Will do this in an additional PR.
- Runs the tests in parallel to speed up execution time.
- Updates all tests that created an in-memory (non-async) profile (wallet) to use the async askar wallet.
- test_dispatcher had a few tests that are skipped. It seems like there might be bug in this code preventing them from working.
- there was one problematic test in test_conductor that I skipped. It has a bad mock somewhere and was causing other tests to fail when run sequentially.


jamshale:
Switching to a raw key dramatically sped up the tests. Whole PR Tests take less than 3 minutes now.

jamshale:
This should be ready. It is a breaking change for external plugins that use the old in_memory_profile for testing.

swcurran:
Wow that is a big change.  I hope you had some editor automation to help with that! LGTM, but will defer to @dbluhm @ff137 (or other maintainers) for an approval.

dbluhm:
@jamshale if a wallet type is not specified on startup, do we spin up with an in-memory askar wallet? (if so, this should also probably be using raw key method and `Store.generate_key`)

dbluhm:
This is great! I'm interested in taking this one step further after this PR is merged and flattening some of the wallet related abstractions now that there is just Askar.

dbluhm (approved):
I reviewed every non-test file in depth and skimmed the test updates. Nice work!

acapy_agent/admin/tests/test_admin_server.py
```diff
@@ -112,161 +103,7 @@ async def test_ready_middleware(self):
             with self.assertRaises(KeyError):
                 await test_module.ready_middleware(request, handler)
 
-    async def test_ready_middleware_http_unauthorized(self):
-        """Test handling of web.HTTPUnauthorized and related exceptions."""
-        with mock.patch.object(test_module, "LOGGER", mock.MagicMock()) as mock_logger:
-            mock_logger.info = mock.MagicMock()
-
```
ff137, line 119: Seems that the rebase dropped some of the unit tests I added :-) from #3327 
I'll re-add in follow-up PR 👍

PR #3313, opened 2024-10-25
:art: Refactor string concatenation in model descriptions

ff137:
Just some minor artsy cleanup --

I noticed that using the backslash for line continuation includes all the whitespace of the code indentation in the final string. So, the openapi docs would generate a description as:

```
The value is used once for a particular domain and window of time.                     This value is ...
```

So, quick fix: replace \ for line continuation with string concatenation

edit: also fixes mediation typo, while I'm at it


ff137:
@jamshale Thanks! Forgot to check

jamshale (approved):

PR #3318, opened 2024-10-29
:wrench: set default fixture scope for pytest-asyncio

ff137:
Resolves `PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,6 +137,7 @@ markers = [
 ]
 junit_family = "xunit1"
 asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "module"
 
 [tool.coverage.run]
 


jamshale:
I was going to change this with an upcoming PR. But we can change it here. I was going to change it to `session` scope as default. Not sure if there's a reason you chose `module` or not? I don't think it will really matter.

ff137:
> I was going to change this with an upcoming PR. But we can change it here. I was going to change it to `session` scope as default. Not sure if there's a reason you chose `module` or not? I don't think it will really matter.

Ah, the only reason for setting it to module level is because it's the lowest available scope.

I think the default in the next pytest-asyncio release will be "function" scope, so I first tried that. But then there are module-scope fixtures that depend on function-scope fixtures, which isn't allowed. So, module scope is a default that works. I guess that was the only reason.

The main difference will be in the lifecycle of the test fixtures. I think session-scope fixtures should be explicitly marked as such, since they have the longest lifecycle (created once and destroyed at end of test run). And those should probably be few and far between. Probably much of a muchness. But it makes sense that since the tests are highly modular (each module has its own tests folder), most test fixtures that don't have an explicitly defined scope should probably be function or module level. 

I think function-level would make most sense, but that entails a review of all the fixtures that don't have a defined scope, and then setting only the ones that need it to module scope.

jamshale (approved):

Issue #2944, opened 2024-05-10
Migrate from `PyLD` to `RDFLib`


ff137:
I'd like to propose that our usage of the [PyLD](https://github.com/digitalbazaar/pyld) dependency be swapped out for the more versatile and better maintained [RDFLib](https://github.com/RDFLib/rdflib).

Let me bullet point some reasons for this suggestion:
- the current state of the PyLD project is not clear to me. The most recent comment on their most recent PR basically says "let's merge this now", and that was 3 months ago.
- v2.0.3 was released in 2020. And then v2.0.4 was released in Feb this year. 
- the repo has no automated testing. When I tested it on my local, I get ~20 test failures, so as far as I can tell it's in a broken state.
- their [jsonld.py](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/jsonld.py) file is 6700 lines, which was "ported from the JavaScript implementation of JSON-LD"
- their requirements.txt has unpinned dependencies

Long story short, the impression I get is that PyLD is not maintained or maintainable. 

RDFLib, on the other hand, seems to be the gold standard for Resource Description Framework (RDF) processing in python. They offer:
- parsers and serializers for RDF/XML, N3, NTriples, N-Quads, Turtle, TriX, Trig and JSON-LD

This repo has much more adoption and activity, and so I think it would be better to migrate to this instead. This is just an idea and open for discussion.

PyLD is currently used in quite a few places in ACA-Py, and probably represent quite a refactoring job to replace it. But I think this may be a worthwhile operation, to reduce technical debt and improve the functionality of a core feature.

ff137:
Relates to #2941

PatStLouis:
PyLd is a library from the authors of the json-ld specification, Digital Bazaar. It aims to be compliant with the specification by conformity with the [json-ld processor test-suite](https://json-ld.github.io/json-ld.org/test-suite/reports/). We updated v2.0.4 since we uncovered a bug in the library and submitted a PR. Going forward there are other bugs in the library which creates interoperability issues with aca-py at the moment for some specific cases.

I'm not opposed to changing the library but I think more tests are needed to see if its a suitable replacement.

Pyld can use an [asynchronous document loader](https://github.com/digitalbazaar/pyld?tab=readme-ov-file#document-loader) as well

ff137:
[...]

PR #3330, opened 2024-11-06
:white_check_mark: Re-add ready_middleware unit tests

ff137:
Re: https://github.com/openwallet-foundation/acapy/pull/3311/files#r1830649910

### DIFF
--- a/acapy_agent/admin/tests/test_admin_server.py
+++ b/acapy_agent/admin/tests/test_admin_server.py
@@ -3,9 +3,11 @@ import json
 from typing import Optional
 from unittest import IsolatedAsyncioTestCase
 
+import jwt
 import pytest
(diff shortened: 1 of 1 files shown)


jamshale:
Sorry about this. I must have hit the wrong button by mistake and not realized.

An unrelated test had a fail. I'm not really sure why this happened. If this happens too often I might need to disable the parallelism. Or make some tests run sequentially.

jamshale (approved):

ff137:
> An unrelated test had a fail. I'm not really sure why this happened. If this happens too often I might need to disable the parallelism. Or make some tests run sequentially.

Ah, I see it's a test that uses the PyLD library. I've seen that error before, and it's partially what motivated posting this issue: #2944

PR #3344, opened 2024-11-21
Restore `--base-wallet-routes` flag functionality

### LINKED ISSUE #3283: `--base-wallet-routes` flag no longer works
After the swap to using decorators to delineate routes accessible by tenants and routes accessible by admins, the ability to grant access to the base wallet to additional routes was lost.

This option made it possible for a base wallet to form a didcomm connection with a mediator and then use that as a base mediator for all tenants, among other things.

cc @esune @jamshale

esune:
Resolves #3283

The `tenant_authentication` has been updated to also allow access to the base wallet when the route matches a path defined using `--base-wallet-routes`.

Please note that, when compared to the previous implementation, the matcher has been made more greedy to tighten security: if an extra route of `/test` is specified, the matcher will only match that and not `/testA` or `/test-something-else` as it appears it would have done before.

One drawback of having to use this matcher inside the decorator is that I could not think of an elegant way of caching the compiled pattern for reuse - suggestions on how to achieve that, if desirable/required, will be welcome.

### DIFF
--- a/acapy_agent/admin/decorators/auth.py
+++ b/acapy_agent/admin/decorators/auth.py
@@ -97,8 +97,10 @@ def _base_wallet_route_access(additional_routes: str, request_path: str) -> bool
 def _build_additional_routes_pattern(pattern_string: str) -> Optional[Pattern]:
     """Build pattern from space delimited list of paths."""
     # create array and add word boundary to avoid false positives
-    paths = pattern_string.split(" ")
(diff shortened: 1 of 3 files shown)


jamshale (dismissed):
LGTM

jamshale (approved):

dbluhm:
Thanks @esune! I will test this in the scenario we're using and report back whether it works as expected.

PR #3346, opened 2024-11-21
Fix tails upload for anoncreds multitenancy

jamshale:
During some detailed testing I found that the tails file wasn't uploading for anoncreds tenants in multitenancy mode. It ended up injectecting the IndyTailsServer class.

This will be covered in a integration test in another PR.


sownak (approved):
LGTM

swcurran (approved):

PR #3347, opened 2024-11-23
fix: handle range proofs in presentation

hallazzang:
## Problem

If I specify multiple predicates on the same attribute when çreating a presentation request, like this:
```
requested_predicates: {
  age_gte: {
    name: "age",
    p_type: ">=",
    p_value: 20
  },
  age_lt: {
    name: "age",
    p_type: "<",
    p_value: 30
  }
}
```

The agent crashes due to this:
https://github.com/openwallet-foundation/acapy/blob/3d1682cd2cf6a485122d8e36ec47d1f5f37c1709/acapy_agent/protocols/present_proof/v2_0/formats/indy/handler.py#L295-L302

It happens when `reft` is `age_lt`(the second one) and `ge_proof["predicate"]` is `{ "attr_name": "age", "p_type": "GE", "value": 20 }`(the first one).
It should just skip the first `ge_proof` but the current implementation raises an exception instead.

## Solution

By checking both `"attr_name"` and `"p_type"` of `proof_pred_spec`, it is now possible to specify more than one predicate on a single attribute, enabling doing a range proof on an attribute.

### DIFF
--- a/acapy_agent/protocols/present_proof/v2_0/formats/indy/handler.py
+++ b/acapy_agent/protocols/present_proof/v2_0/formats/indy/handler.py
@@ -287,12 +287,12 @@ class IndyPresExchangeHandler(V20PresFormatHandler):
                     "primary_proof"
                 ]["ge_proofs"]:
                     proof_pred_spec = ge_proof["predicate"]
-                    if proof_pred_spec["attr_name"] != canon(req_name):
-                        continue
                     if not (
-                        Predicate.get(proof_pred_spec["p_type"]) is req_pred
-                        and proof_pred_spec["value"] == req_value
+                        proof_pred_spec["attr_name"] == canon(req_name)
+                        and Predicate.get(proof_pred_spec["p_type"]) is req_pred
                     ):
+                        continue
+                    if proof_pred_spec["value"] != req_value:
                         raise V20PresFormatHandlerError(
                             f"Presentation predicate on {req_name} "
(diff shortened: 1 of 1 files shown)


swcurran:
Thanks for the contribution.  As per note on Discord, I'm not sure this is needed (can use restrictions instead), but I don't think it is harmful either.

Can you please fix the DCO checks that are failing (above).  See DCO - Developer Certificate of Origin - https://github.com/apps/dco. To fix, see: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md and the details link beside the failed checks.

Perhaps also add a test for this?

swcurran:
Hi @hallazzang -- thanks for the contribution.  Given that the DCO is not signed off (and must be) and that there has not been a lot of help in getting this pushed through, I think it is best that we not plan on merging this.  I like the idea, and the constraint preventing support for this seems to be in ACA-Py and not in AnonCreds.  However, I'd like to see more testing of this with other components in the community (e.g. Credo-TS, Bifold) to make sure that support for multiple predicates against a single attribute is across the board.

What I propose to do is close this PR, and open an Issue that points to this PR, in case anyone wants to pick it up and complete it.

Hope that makes sense -- thanks again.

hallazzang:
Oh I completely missed your previous comment, sorry for that. I agree with you that creating an issue that points this PR would be better for the time being. Thank you for taking your time into it!

PR #3354, opened 2024-11-25
Fix/universal resolver

jamshale:
The `https://dev.uniresolver.io/1.0/properties` endpoint which is the default url for the universal resolver changed recently and acapy failed to handle the response.

The patterns are now the keys of the json payload.


swcurran (dismissed):

jamshale:
I need to figure out how to turn off the https check in sonarcloud. Doesn't really apply to this project very often.

Looks like the interop tests are working now...

jamshale:
I just realized that if a deployment is using a different universal resolver that hasn't changed then this would probably break it.

I'm going to change it to have compatibility support.

swcurran (approved):

PR #3355, opened 2024-11-26
Add test wallet config option

jamshale:
With the removal of the in-memory wallet we lost the ability to quickly spin up an in-memory agent without a wallet name or key. This adds the ability back if you use the new `--wallet-test` config. It will create an in-memory askar wallet.

If you don't use `wallet-test` and fail to supply name and key you now get a wallet-name and wallet-key required for persistent wallet error on startup.


acapy_agent/config/argparse.py
```diff
@@ -1627,11 +1627,24 @@ def add_arguments(self, parser: ArgumentParser):
...
-                "Supported internal storage types are 'basic' (memory), "
-                "'default' (sqlite), and 'postgres_storage'.  The default, "
+                "Supported internal storage types are 'default' (sqlite), "
+                "and 'postgres_storage'.  The default, "
                 "if not specified, is 'default'."
             ),
         )
+        parser.add_argument(
+            "--wallet-test",
+            type=str,
```
dbluhm, line 1637: Should probably just be a `store_true` rather than a str

PatStLouis (approved):
Thanks for looking into this

acapy_agent/askar/tests/test_profile.py
```diff
@@ -107,15 +108,17 @@ async def test_profile_manager_transaction():
...
+        settings=config,
+    )
+    await AskarProfileManager().provision(
+        context=context,
+        config=config,
+    )
+    await AskarAnonProfileManager().provision(
+        context=context,
+        config=config,
+    )
```
ff137, line 124: Test no longer has assertions

PR #3356, opened 2024-11-26
Make vc libraries mandatory

jamshale:
Currently both vc libraries are required to be installed and they should not be extras.

This is mainly due to compatibility reason. When indy is fully retired indy-credx should be able to be an extra or completely removed and contained within a plugin.


swcurran (approved):

dbluhm (dismissed):

swcurran (approved):

PatStLouis (approved):

swcurran:
@dbluhm — can you please fix the conflicts?

dbluhm (approved):

swcurran:
Sorry about that @dbluhm — I thought it was your PR :-).  I wish I knew the git magic to deal with stuff like that…

dbluhm:
> I thought it was your PR :-)

I wondered lol but figured, sure, I'll take care of it :smile:

dbluhm:
Sonarcloud issue pretains to the Dockerfile.demo

> This image might run with root as the default user. Make sure it is safe here.

I think we're okay with the the demo container not having a separate, non-root user. Unless you'd like to take some time to address that, @jamshale -- I have no objections either way

jamshale:
Ya. It's flagging it because I made a change in the file but I think it's fine in the demo conatiner.

My sonarcloud linter flagged a bunch of stuff in the dockerfiles that I think should be cleaned up, but I didn't want to do it in this PR.

PR #3358, opened 2024-11-27
:art: Sync Ruff version in configs and apply formatting

ff137:
Following up on #3348

### DIFF
--- a/acapy_agent/commands/tests/test_help.py
+++ b/acapy_agent/commands/tests/test_help.py
@@ -5,11 +5,10 @@ from .. import help as command
 
 class TestHelp(IsolatedAsyncioTestCase):
     def test_exec_help(self):
-        with mock.patch.object(
-            command.ArgumentParser, "print_help"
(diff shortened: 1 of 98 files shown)


acapy_agent/commands/tests/test_help.py
```diff
@@ -5,22 +5,20 @@
...
             mock_print.assert_called_once_with(command.__version__)
 
     def test_main(self):
-        with mock.patch.object(
-            command, "__name__", "__main__"
-        ) as mock_name, mock.patch.object(
-            command, "execute", mock.MagicMock()
-        ) as mock_execute:
+        with (
+            mock.patch.object(command, "__name__", "__main__") as mock_name,
```
jamshale, line 20: I've started removing these variables (`mock_name`) when they aren't used. It's not important but makes the code slightly cleaner. Doesn't have to happen in this PR.

jamshale (approved):

PR #3362, opened 2024-12-02
:bug: Fix IndyVdrLedgerPool RuntimeWarning

### LINKED ISSUE #3328: 🐛 `RuntimeWarning: coroutine 'IndyVdrLedgerPool.context_close.<locals>.closer' was never awaited`
When operating under load, the following RuntimeWarning gets generated in our logs:
```sh
RuntimeWarning: coroutine 'IndyVdrLedgerPool.context_close.<locals>.closer' was never awaited
```

As far as I can tell, this likely occurs when `IndyVdrLedgerPool`'s async close task gets garbage collected. Or, perhaps the background task gets created too many times when acapy is under load, and doesn't always get scheduled properly. It's not yet clear why it occurs. And it doesn't _appear_ to cause any other issues, but it's worth investigating, as the code might be behaving in unexpected ways.

The issue can also manifest as:
```sh
Exception ignored in: <coroutine object IndyVdrLedgerPool.context_close.<locals>.closer at 0x7ff607befe20>
File "<string>", line 1, in <lambda>
KeyError: '__import__'
```

I can't really tell what's going on here, but wanted to record it so long as we're debugging a fix.

Edit:
The exception associated with the runtime warning appears to be a red herring - it just depends on what task was running concurrently. This RuntimeWarning seems to only appear on DataDog. So, as far as I can tell, the `coroutine .. was never awaited` issue would usually occur silently, but due to DataDog / ddtrace injection, it picks up on it.

I have a fix for this, but I'll try include some before/after testing to validate that the fix is really necessary.

ff137:
Resolves #3328

Still in progress -- cleaning up test logs, and still want to improve on the new async get_or_create approach

### DIFF
--- a/acapy_agent/askar/profile.py
+++ b/acapy_agent/askar/profile.py
@@ -41,15 +41,27 @@ class AskarProfile(Profile):
         *,
         profile_id: Optional[str] = None,
     ):
-        """Create a new AskarProfile instance."""
+        """Private constructor. Use 'create' to instantiate."""
         super().__init__(
             context=context, name=profile_id or opened.name, created=opened.created
         )
         self.opened = opened
         self.ledger_pool: Optional[IndyVdrLedgerPool] = None
         self.profile_id = profile_id
-        self.init_ledger_pool()
-        self.bind_providers()
+
+    @classmethod
+    async def create(
+        cls,
+        opened: AskarOpenStore,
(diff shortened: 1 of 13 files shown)


ff137:
I'll pick this one up again at some point. The problem/solution here is quite esoteric

PR #3369, opened 2024-12-02
Anoncreds create credential

jamshale:
When `cheqd` was attempting anoncreds issuance they were getting a `entropy required` error when using the anoncreds library to create the credential. After some research it looked like we should be using the entropy parameter, but were using the `prover_did` parameter.  https://hyperledger.github.io/anoncreds-spec/#constructing-the-credential-request

I don't know why this works for an indy credential request and not the cheqd. Maybe something internally in the anocnreds library. But it looks like using the holder_did value in the entropy works and I tested with indy and cheqd and issuance succeeds.

Then noticed some indy parsing when storing the credential that shouldn't be in the anoncreds holder. I couldn't figure out why we would need these extra tags so I removed them. If we did need them then the parsing should go into the actual registry.

### DIFF
--- a/acapy_agent/anoncreds/default/legacy_indy/registry.py
+++ b/acapy_agent/anoncreds/default/legacy_indy/registry.py
@@ -80,6 +80,7 @@ from ...models.schema import (
     SchemaResult,
     SchemaState,
 )
+from ...models.schema_info import AnoncredsSchemaInfo
 from ...revocation import (
     CATEGORY_REV_LIST,
     CATEGORY_REV_REG_DEF,
@@ -1230,11 +1231,11 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
         except LedgerError as err:
             raise AnonCredsRegistrationError(err.roll_up) from err
 
-    async def get_schema_info_by_id(self, schema_id: str) -> dict:
+    async def get_schema_info_by_id(self, schema_id: str) -> AnoncredsSchemaInfo:
         """Get schema info by schema id."""
         schema_id_parts = re.match(r"^(\w+):2:([^:]+):([^:]+)$", schema_id)
-        return {
-            "issuer_id": schema_id_parts.group(1),
-            "name": schema_id_parts.group(2),
-            "version": schema_id_parts.group(3),
-        }
+        return AnoncredsSchemaInfo(
+            issuer_id=schema_id_parts.group(1),
+            name=schema_id_parts.group(2),
+            version=schema_id_parts.group(3),
+        )
--- a/acapy_agent/anoncreds/default/legacy_indy/registry.py
+++ b/acapy_agent/anoncreds/default/legacy_indy/registry.py
@@ -56,6 +56,7 @@ from ...base import (
 )
 from ...events import RevListFinishedEvent
 from ...issuer import CATEGORY_CRED_DEF, AnonCredsIssuer, AnonCredsIssuerError
+from ...models.cred_def_info import AnoncredsCredDefInfo
 from ...models.credential_definition import (
     CredDef,
     CredDefResult,
@@ -1238,4 +1239,10 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
             issuer_id=schema_id_parts.group(1),
             name=schema_id_parts.group(2),
             version=schema_id_parts.group(3),
+        )
+
+    async def get_cred_def_info_by_id(self, cred_def_id: str) -> AnoncredsCredDefInfo:
+        """Get cred def info by cred def id."""
+        return AnoncredsCredDefInfo(
+            issuer_id=cred_def_id.split(":")[-5],
         )
--- a/acapy_agent/anoncreds/default/legacy_indy/registry.py
+++ b/acapy_agent/anoncreds/default/legacy_indy/registry.py
@@ -1229,3 +1229,12 @@ class LegacyIndyRegistry(BaseAnonCredsResolver, BaseAnonCredsRegistrar):
                 )
         except LedgerError as err:
             raise AnonCredsRegistrationError(err.roll_up) from err
+
+    async def get_schema_info_by_id(self, schema_id: str) -> dict:
+        """Get schema info by schema id."""
+        schema_id_parts = re.match(r"^(\w+):2:([^:]+):([^:]+)$", schema_id)
+        return {
+            "issuer_id": schema_id_parts.group(1),
+            "name": schema_id_parts.group(2),
+            "version": schema_id_parts.group(3),
+        }
--- a/acapy_agent/anoncreds/default/legacy_indy/registry.py
+++ b/acapy_agent/anoncreds/default/legacy_indy/registry.py
@@ -56,7 +56,6 @@ from ...base import (
 )
 from ...events import RevListFinishedEvent
 from ...issuer import CATEGORY_CRED_DEF, AnonCredsIssuer, AnonCredsIssuerError
-from ...models.cred_def_info import AnoncredsCredDefInfo
 from ...models.credential_definition import (
     CredDef,
     CredDefResult,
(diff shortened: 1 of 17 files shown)


swcurran:
Frustrating — sorry that got messy.  The orginal AnonCreds implementation (in Indy) used “holder_did”, but semantically, that was wrong — the “did” had nothing to do with the operation. Clients were using the “DIDComm DID” of the relationship between the holder and issuer — but it was not checked and could be anything.  We looked into the code, and it was determined that the purpose of the data was simply to add entropy into the operation.  A very minor change, but not everyone got the memo, and all it does is create friction. Likely the best way to handle it is for the issuer to accept either, and pass it on to AnonCreds with what it expects.  Painful - sorry about that.  The “clarification” of the purpose does not make up for the interop problems this has caused.

jamshale:
Need to look into failing tests... Thought I had tested this.

jamshale:
I think we can ignore the sonarcloud report. Most of the duplication is because of the extension of the registry base class and there's not really much to cover with unit tests, and it's covered by the scenario test.

gmulhearn:
FWIW, i also ran into compatibility issues when testing VCX against cheqd's acapy fork/branch. The issue was that VCX (using anoncreds-rs) was creating cred request attachments that had `entropy` instead of `prover_did`, but then the ACApy schema/model validation for the cred request was upset at this. My quick fix was to change the model from `prover_did` to `entropy`.

Looks like this is the more permanent fix - cheers!

question; is it still necessary to include both `prover_did` & `entropy` in the anoncreds cred request attachment model? Would it be too aggressive to make the assumption that if you're using the [0771 attachment format](https://github.com/hyperledger/aries-rfcs/blob/main/features/0771-anoncreds-attachments/README.md#credential-request-format), you're abiding by the anoncreds spec, which i _believe_ means you must be using `entropy` instead of `prover_did`?

Maybe that's too aggressive, as i don't think the anoncreds spec explicitly denies usage of `prover_did`, infact the cred request example uses `holder_did`: https://hyperledger.github.io/anoncreds-spec/#constructing-the-credential-request 👀. Should this be updated? @swcurran

gmulhearn:
> I don't know why this works for an indy credential request and not the cheqd. Maybe something internally in the anocnreds library. But it looks like using the holder_did value in the entropy works and I tested with indy and cheqd and issuance succeeds

yeah i ran into this too. here's the source where all the validation rules are applied: https://github.com/hyperledger/anoncreds-rs/blob/main/src/data_types/cred_request.rs#L27

different behaviours depending on whether or not the cred-def-id is qualified ("legacy") or not

acapy_agent/anoncreds/holder.py
```diff
@@ -231,25 +230,17 @@ async def _finish_store_credential(
...
-            "schema_name": schema_id_parts[2],
-            "schema_version": schema_id_parts[3],
-            "issuer_did": cdef_id_parts[1],
-            "cred_def_id": cred_def_id,
+            "schema_id": credential_data["schema_id"],
+            "schema_issuer_id": schema_info.issuer_id,
+            "schema_name": schema_info.name,
+            "schema_version": schema_info.version,
+            "issuer_id": credential_definition["issuerId"],
+            "cred_def_id": cred_recvd.cred_def_id,
```
dbluhm, line 243: Is the `schema_issuer_did` to `schema_issuer_id` and `issuer_did` to `issuer_id` going to cause issues in wallets with existing credential records? Do we need an upgrade script for this?

jamshale: I reverted this change. I don't want to mess up the holder tags or require a migration for this.

dbluhm:
These changes will get hit by the anoncreds scenario, right?

jamshale:
Yes. The anocnreds scenario test is testing askar (issuer) --> askar-anoncreds (holder) and askar-anoncreds (issuer) --> askar (holder). And the verifier scenario.

The anoncreds holder now uses entropy instead of prover_did, and the indy credx holder still uses prover_did.

The only issue would be an old version of an indy issuer, issuing to a new anoncreds holder. In this case it wouldn't know what to do with the `entropy` attribute.

swcurran:
If it helps — the “holder_did” property that would come from an Indy request can be used as the value of “entropy” — they are the same value.

swcurran:
Updated that last comment to be “prover_did” — ~~not “holder_did”~~

DaevMithran:
@jamshale we need to `get_schema_info_by_id` even in the presentation anoncreds handler to replace the indy parsing. Maybe we also need a `get_credential_definition_info_by_id`
 
https://github.com/jamshale/acapy/blob/anoncreds-issuance/acapy_agent/protocols/present_proof/v2_0/formats/anoncreds/handler.py#L239

jamshale:
> @jamshale we need to `get_schema_info_by_id` even in the presentation anoncreds handler to replace the indy parsing. Maybe we also need a `get_credential_definition_info_by_id`

👍 Ok. Thanks for pointing this out. I had tried to run through the presentation protocol as well with your branch but was having other issues so I never got through it. This looks like an obvious place to apply the method. Let me know if you see any others please.

acapy_agent/anoncreds/default/legacy_indy/registry.py
```diff
@@ -1229,3 +1231,18 @@ async def txn_submit(
                 )
         except LedgerError as err:
             raise AnonCredsRegistrationError(err.roll_up) from err
+
+    async def get_schema_info_by_id(self, schema_id: str) -> AnoncredsSchemaInfo:
```
DaevMithran, line 1235: @jamshale I think we should use the existing methods

- get_schema
- get_credential_definition

instead of these new methods as it's not guaranteed that the name, version or other attributes required are within the identifiers. For cheqd we have to make a ledger query inorder to get them. The above functions already handle this. I created a reference PR [here](https://github.com/jamshale/acapy/pull/178)

DaevMithran (changes requested):

acapy_agent/anoncreds/default/legacy_indy/registry.py
jamshale: That had been my original thought, but this is used only by the `holder`. The holder shouldn't necessarily need to go to the ledger or wallet to get the issuer id and schema version, etc., if the information is available in the schema id and cred def id itself.

For `cheqd` this `get_schema_info_by_id` method could call the the get_schema method and extract the correct info for the holder if it's required that the holder needs to contact the ledger to get the required info.

I'm not sure what did methods will be able to put the required info into the schema and cred_def id's themselves.

It'd be nice to get some feedback from others on this...

jamshale: I could change the handler method implementations like you did as these are done by the `verifier` and they should be able to contact the ledger. My comment more applies to the usage in https://github.com/openwallet-foundation/acapy/blob/main/acapy_agent/anoncreds/holder.py#L224 when the holder save the credential.

I could remove the `get_cred_def_info_by_id` method by changing the handler like you did.

jamshale: I updated this to only use the `get_schema_info_by_id` method with the holder when saving the credential.

acapy_agent/protocols/present_proof/v2_0/formats/anoncreds/handler.py
```diff
@@ -175,7 +176,7 @@ async def create_pres(
     async def receive_pres(self, message: V20Pres, pres_ex_record: V20PresExRecord):
         """Receive a presentation and check for presented values vs. proposal request."""
 
-        def _check_proof_vs_proposal():
```
jamshale, line 178: This class should really get some refactoring but don't want to address it in this PR.

acapy_agent/anoncreds/default/legacy_indy/registry.py
DaevMithran: In that case the profile should be one of the parameters within get_schema_info_by_id.

jamshale: I thought it would be available through the Registry class, but looks like all the other functions pass it as a parameter so i'll follow the pattern.

acapy_agent/protocols/present_proof/v2_0/formats/anoncreds/handler.py
```diff
@@ -292,15 +303,21 @@ def _check_proof_vs_proposal():
                         f"Proposed request predicate on {req_name} not in presentation"
                     )
 
+                sub_proof_index = attr_spec["sub_proof_index"]
```
sownak, line 306: This line introduces a bug (its already defined correctly in line 285

jamshale: Fixing. Ya this function should really be refactored later.

DaevMithran (approved):

esune (approved):
👍🏻

PR #3371, opened 2024-12-02
fix: check routing keys on endpoint refresh

dbluhm:
Previously we were only checking if the endpoint value had changed but it is possible for the routing keys to update and not the endpoint. This resulted in new attrib txns not getting published when there were new routing keys.


dbluhm:
Would like to see this in the patch release if possible, @swcurran

swcurran (approved):

PR #3375, opened 2024-12-03
feat: add did management design doc

### LINKED ISSUE #3343: DID Management Proposed Update
In light of our current push to add support for more DID Methods to ACA-Py, some primitives within ACA-Py need some updates. This issue outlines the updates I propose. I plan to update this further as the topic is discussed or as implementations better inform decisions.

# Proposed Updates

### Current State
At present, the DIDInfo object looks like this:

https://github.com/openwallet-foundation/acapy/blob/f5c49b0710dd180ea31c45f73bc82ef06f9523b4/acapy_agent/wallet/did_info.py#L20-L29

This is stored in the wallet with a category of `did`, the primary identifier being the DID value, and the following tags:
- `method`: the method name
- `verkey`: the `verkey` element of the tuple where it is the base58 encoding of the public key
- `verkey_type`: the key type of the verkey (e.g. `ed25519`)

As currently used, metadata will include:
- `posted`: a boolean value representing whether this did has been published to an indy network
- `endpoint`: a string value representing associated with the endpoint attrib of this did on an indy network.

### Evaluation

As is plain to see, the structure, tags, and metadata of the `DIDInfo` object are very Indy-oriented. This structure has been in use for years.

Currently, ACA-Py will retrieve a `DIDInfo` object in order to use the key associated with the "DID." It will do this by taking a "DID" as input (usually, actually more like a "nym" value, i.e. 16 base58 encoded bytes without a `did:` prefix), then using the `verkey` value to retrieve a `Key` object that it can then use to perform a signature or pack a DIDComm message.

### Solution

DIDs should have multiple keys associated with them rather than a single key. To achieve this while also having an efficient lookup mechanism, we should reorient our storage as outlined below.

#### Quick background on Askar

[Askar](https://github.com/openwallet-foundation/askar) is a secure storage solution used by ACA-Py. Askar encrypts all data and provides a tagging mechanism to enable lookup of encrypted records. An entry in Askar is composed of the following elements:

- **Category:** The major group or "bucket" that the entry belongs to.
- **Name:** The primary identifier for the record; this is roughly equivalent to primary keys on a traditional DB table. The most efficient lookup possible is by name.
- **Value:** The value stored in the entry. This is usually a serialized JSON object.
- **Tags:** A mapping of strings to strings or lists of strings. These values can be used with the ["Wallet Query Language (WQL)"](https://github.com/hyperledger-archives/indy-sdk/tree/main/docs/design/011-wallet-query-language) to look up encrypted Askar entries efficiently.

Askar has a dedicated API for storage and retrieval of keys. However, this API is conceptually just a shorthand for record storage and retrieval from a "private" `key` category with the key itself as the value of the entry. Key entries behave almost exactly the same as non-key entries, including names and tags.

#### Key Storage

Building off of Patrick's contributions of managing keys by multikey instead of "verkey," the multikey representation of a key should be the default identifier for keys in the wallet.

- **Name:** multikey representation of the key
- **Tags:**
    - Implicit tag for the `KeyAlg` (automatically included on every key)
    - `did`: the DID (or a list of DIDs) the key is associated with
    - `vm_id`: an absolute DID URL (or a list of DID URLs) representing the verification method ID(s) of the key
    - `rel`: A list of verification relationships as defined by the DID Core spec; e.g. `["authentication", "assertionMethod"]`. This represents the intended use of this key.
    - `alias`: A human-friendly alias (or list of aliases) that can help identify a key to a user

These sets of tags enable us to look up keys with a combination of `did` and `rel`; when these tags are lists, Askar will return all keys that contain the tag filter value in their respective list. This permits the controller to continue to specify just a DID as the issuer/signer/sender of a value without having to know exactly which key ACA-Py should use to perform the operation. This also permits the controller to continue to use the verification method ID directly to specify a key that might not normally be selected first. Additionally, when a specific proof type is desired, Askar can also filter by `KeyAlg` so a simple mapping from proof type to appropriate `KeyAlg`s can efficiently accomplish this filtering.

#### DID Storage

DIDs should be altered to be stored in a way that simply acknowledges that we own the DID and not as the primary key retrieval mechanism.

- **Category:** `did`
- **Name:** the DID itself
- **Value:**
    - ...
- **Tags:**
    - `method`: a string representing the DID Method
    - (Maybe?) `features`: the list of features the DID is capable of
    - Other things it would be valuable to use to look up the DID?

## Migration

Existing ACA-Py wallets should have the following migration performed to accommodate this reorientation:

- Migrate all existing records in the `did` category to the new structure and also duplicate the record to a legacy `nym` category
  - If the `did` value is unqualified, "move" to the new `did` category and map old values onto new, adding `did:sov:` to the front.
  - If the `did` value is unqualified, also create a record in the new `nym` category with a structure matching the original `DIDInfo` object, where a verkey is closely associated with the nym. This should enable us to continue using the builtin Indy support in a way that distinguishes the old from the new.
  - For `did` values that are qualified, move to the new did category and make sure the associated key entries are properly tagged. This will depend on the DID Method. Plugged in DID Methods may need to account for their own DID methods in a separate migration process.

dbluhm:
This PR closes #3343.

This adds a design document synthesized from the discussion and comments of #3343.

Feedback is welcome!


docs/design/DIDManagement.md
```diff
@@ -0,0 +1,187 @@
...
+Bound keys are associated with a specific DID.
+
+When an unbound key is used to generate a DID for a method like did:peer, bound key representations MUST be stored in the wallet. The unbound representation MAY be removed after the bound representation is added.
+
+- **Category:** `key`
+- **Name:** Verification Method ID (absolute DID URL, e.g., `did:example:123#key-1`)
+- **Value:** The key material
+- **Tags:**
+  - `KeyAlg`: Implicit tag indicating the key algorithm (e.g. `ed25519`)
+  - `did`: The DID associated with the key
```
PatStLouis, line 100: is this limited to 1 did or can it be a list of dids?

PatStLouis (commented):
Great document!

docs/design/DIDManagement.md
dbluhm: The intent is for this one to be a single DID. If this key is bound to multiple DIDs, there will be multiple key records, each one uniquely identified by the verification method ID of the key in that DID's document.

jamshale:
Looks good to me. I'm going to read it over a couple more times this week, but I think it addresses concerns I've had with my limited experience of the problem.

docs/design/DIDManagement.md
```diff
@@ -0,0 +1,186 @@
...
+
+As part of our initiative to support a wider range of DID Methods in ACA-Py, we need to update the primitives related to DID and key storage within ACA-Py. This document outlines the proposed changes to how DIDs and keys are stored and managed, the rationale behind these changes, and the migration strategy from the current implementation.
+
+## Background
+
+### Askar Records and Lookup Mechanisms
+
+[Askar](https://github.com/openwallet-foundation/askar) is the secure storage solution used by ACA-Py. Askar encrypts all data and provides a tagging mechanism to enable lookup of encrypted records. An entry in Askar is composed of the following elements:
+
+- **Category:** The major group or "bucket" that the entry belongs to.
```
swcurran, line 13: Nice to have examples of what these are. This comes later, but…. :-)

docs/design/DIDManagement.md
```diff
@@ -0,0 +1,186 @@
...
+
+- **Indy-Centric Design:** The current structure is closely tied to Indy, making it less suitable for other DID Methods.
+- **Single Key per DID:** Assumes a one-to-one relationship between DIDs and keys, which is not the case for many DID Methods.
+- **Inefficient for Multiple Keys:** Lacks support for DIDs with multiple verification methods and keys.
+
+## Proposed Updates
+
+### Goals
+
+- **Support Multiple Keys per DID:** Allow DIDs to have multiple associated keys and verification methods.
```
swcurran, line 68: Should we add that we would like to track whether a key is active or historical/archived/no longer in use.  When we add `did:webvh` we will have keys referenced only in older versions of the DID/DIDDoc that should no longer be used for signing.

docs/design/DIDManagement.md
```diff
@@ -0,0 +1,186 @@
...
+- **Value:** The key material (private or public key)
+- **Tags:**
+  - `KeyAlg`: Implicit tag indicating the key algorithm (e.g., `ed25519`)
+  - `alias`: A list of human-friendly aliases for the key
+
+#### Bound Keys
+
+Bound keys are associated with a specific DID.
+
+When an unbound key is used to generate a DID for a method like did:peer, bound key representations MUST be stored in the wallet. The unbound representation MAY be removed after the bound representation is added.
```
swcurran, line 93: “While” ?

swcurran:
Looks good. How do we get this implemented?  Is this a one person job, with support for review and testing, or is it worth trying to split up the work?

Would also like to understand how an AnonCreds issuer interacts with this.  Presumably, it is just about the DID, and the rest of the objects are “standalone” and will have different implementation methods.

docs/design/DIDManagement.md
dbluhm: Interesting question... If they should no longer be used for signing, do we have a need to keep the private key material around? If yes, then marking a key as inactive makes sense. If not, we could just remove the key from storage.

dbluhm:
> Would also like to understand how an AnonCreds issuer interacts with this. Presumably, it is just about the DID, and the rest of the objects are “standalone” and will have different implementation methods.

DIDs feed into the AnonCreds issuer interface by supplying `issuer_id` to the creation of various AnonCreds objects. The "global" AnonCreds interface uses that issuer ID to determine which method-specific AnonCreds registry should handle the request. Then, the method specific implementation will (most likely) use the issuer ID to generate signatures or do whatever operations are required for the method.

So the method specific registries would need to be aware of how to look up the right key material to perform those operations. A good default operation would be to look up the first key in the authentication relationship of the issuer ID (DID). But aside from that, I believe DID management and AnonCreds object management have very little overlap.

jamshale (approved):

jamshale:
I'm going to merge this. I think there's enough to get going on, and can always update it later.

PR #3389, opened 2024-12-09
Switch to COPY commands in dockerfiles

jamshale:
> Using the ADD instruction instead of COPY for local resources in Dockerfiles can lead to several issues, including unexpected behavior, increased complexity, and security risks. The ADD instruction has additional features that can introduce unintended side effects, such as automatically extracting compressed files and fetching remote resources. This can make the behavior of the instruction less predictable and harder to understand or even lead to security issues, if, for example, due to a typo in the source path, the ADD instruction could fetch a remote resource instead of copying a local file.
> 
> If you only need to copy local files or directories into your Docker image, it is recommended to use the COPY instruction instead. Only use the ADD instruction when you need its additional features, such as fetching remote resources or extracting compressed files. See also the rule [S7026](https://sonarcloud.io/organizations/openwallet-foundation/rules?open=docker%3AS7026&rule_key=docker%3AS7026) for more information on using the ADD instruction to fetch remote resources.

Easy fix for the low rated security issues in Sonarcloud.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ FROM python:${python_version}-slim-bullseye AS build
 
 WORKDIR /src
 
-ADD . .
+COPY . .
 
 RUN pip install --no-cache-dir poetry
 RUN poetry build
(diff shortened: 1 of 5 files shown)


swcurran (approved):
Fine to do. Looks like it didn’t get rid of all of the Warnings :-(

jamshale:
No there's other warnings. I'll try and look at these in another ticket. These are the easiest ones to fix and won't have any side effects.

PR #3393, opened 2024-12-09
Only copy agent code in dockerfiles

jamshale:
The dockerfiles don't need to copy the entire repo into the base, run, and test images. They should only copy the project files and the actual agent code.

Found some old aries_cloudagent naming that should be updated.


WadeBarnes (commented):
The other way to handle this, which is possibly more manageable long term, is with `.dockerignore` files, which seem to already be in use; https://github.com/openwallet-foundation/acapy/blob/main/.dockerignore

image

jamshale:
> It might be an idea to add a comment in the dockerfiles referencing the `.dockerignore` filtering their content. That way it's clear where the content of the dockerfile is controlled.

Sounds good. I wasn't aware of that. I will do an overview of this.

WadeBarnes:
> I'm somewhat confused as to why the demo code is included. Seems like there's a lot of code there that shouldn't be.

That screenshot is just a list showing all of the places a `.dockerignore` file are being used in the project.

swcurran (approved):

jamshale:
I'll rebase and test this a bit more. I think it would be good to have. 

The .dockerignore file doesn't really help us because it applies to every Dockerfile in the docker folder, and different files are included for the different images. A comment mentioning those files are also ignored would be good to have, though.

jamshale:
I tried to test all usages of these dockerfiles I could think of. There's a chance that a process I don't know about expects the files I removed to be there and will break and would need to be added back.

The most important ones are used in automatic testing.

PR #3395, opened 2024-12-10
Patch PR 3391 - 0.11.lts

jamshale:
For this to work someone with access to https://pypi.org/project/aries-cloudagent/ needs to set a new publisher.

Then we also need someone from hyperledger to set the new HYPERLEDGER_GHCR_PAT token in the openwallet-foundation repo. See https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic.

> secrets.HYPERLEDGER_GHCR_PAT

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,12 +51,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ inputs.ref || '' }}
-
-      - name: Gather image info
-        id: info
-        run: |
-          echo "repo-owner=${GITHUB_REPOSITORY_OWNER,,}" >> $GITHUB_OUTPUT
-
       - name: Cache Docker layers
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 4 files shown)


swcurran:
@ryjones can you help with this?  See notes from @jamshale above.

@WadeBarnes anything you can do?  I’m guessing not, but figured I’d ask and make sure you knew about this.

🤞

WadeBarnes:
I can help with the aries-cloudagent pypi project, but I'm unclear on what, exactly, needs to be done.

jamshale:
@WadeBarnes I might need to change this workflow more. It's using an old username/password authentication that is no longer supported. It the https://pypi.org/project/aries-cloudagent/ account somewhere I thought it might be available for maintainers.

I think I can change this workflow to the new way and give better instructions.

swcurran:
We’re trying to publish LTS releases of ACA-Py — back when it was in Hyperledger.  So we need to publish to the old PyPi project “aries-cloudagent” and the old Hyperledger GHCR.  This is the first time we’ve had to do this.

WadeBarnes:
> and the old Hyperledger GHCR.

@ryjones will need to assist with the PAT for that part.

jamshale:
So this pypi part should be easier now. Someone just needs to add a new publisher with the openwallet information.

image

WadeBarnes:
Done

ryjones:
is there more for me to do here?

WadeBarnes:
> is there more for me to do here?

If you've already setup an access token that will allow workflows in this project to publish images to the `ghcr.io/hyperledger/aries-cloudagent-python` repo, then no.

ryjones:
@WadeBarnes done

jamshale:
I'm going to run all the tests manually on this branch and then I will try and make the release.

jamshale:
So.. I should have tested this better before. The problem doesn't actually affect 0.11.lts . Soft binding was introduced after. This version should be safe.

All this work will be used for 0.12.lts patch though. I'm going to close this and move to that release.

swcurran:
Thanks, Jamie.  Good to know we don’t have to do 0.11.  That’s helpful.

PR #3396, opened 2024-12-10
Patch PR 3391 - 0.12.lts

jamshale:
We need someone from hyperledger to set the new HYPERLEDGER_GHCR_PAT token in the openwallet-foundation repo. See https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic

> secrets.HYPERLEDGER_GHCR_PAT

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,12 +51,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || '' }}
-
-      - name: Gather image info
-        id: info
-        run: |
-          echo "repo-owner=${GITHUB_REPOSITORY_OWNER,,}" >> $GITHUB_OUTPUT
-
       - name: Cache Docker layers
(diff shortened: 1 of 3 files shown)


jamshale:
All the tests passed for this patch, so I think we can go ahead.

swcurran (approved):
LGTM

jamshale:
The image succeeded 🚀 https://github.com/orgs/hyperledger/packages/container/package/aries-cloudagent-python

I'm not sure what to do about the pypi package though. I think we could try creating a 0.12.3 release if required. Because the image is the preferred way to deploy acapy, maybe we can get away with just having the patched image? 

Not really sure what happened with the package for previous patches...

swcurran:
I can create a new release on the 0.12 branch — is that the next step?  That should publish the PyPi package, if the scripts work as expected, right?  Let me know and I’ll create the PR.

PR #3398, opened 2024-12-11
Remove sonar cov report move step

jamshale:
Something changed with github actions and this step to move the coverage report is no longer needed. It was done already for merge to main, but it affected the PR workflow as well. 

It needs to be merged to main before taking effect.


ianco (approved):

jamshale:
Ugh.. This didn't work. It's getting a different error about the test report. I'll need to look at this closer.

PR #3399, opened 2024-12-11
Adjust coverage location for sonarcloud

jamshale:
Ok. This should actually fix the sonar coverage reporting. 

Github actions did change so that last PR was needed. However, the coverage report gets saved at the current location now. It no longer needs to be moved into a test-reports folder. 

It actually probably didn't need to do this before. Just needed to move from the runner location to the workspace location. That's what changed.


swcurran (approved):
I’ll believe it when I see it :-) .

jamshale:
It worked 😌

Issue #3269, opened 2024-10-03
Scenario testing - Restarts and upgrades


jamshale:
In the acapy tools repo there is some really useful testing using docker containers that allows testing of upgrade scripts and restarts. https://github.com/hyperledger/aries-acapy-tools/tree/main/askar_tools/tests/e2e.

It should be possible to support this in the scenarios directory from acapy itself. It would be awesome to be able to test a supported upgrade path from the last LTS version and test a few different restarts with changed configurations such as making public dids and using seeds etc.

ianco:
@jamshale  are there instructions for running the e2e tests?

jamshale:
In `[acapy-tools](https://github.com/openwallet-foundation/acapy-tools)` I don't think there is any instructions :/ . It's pretty simple though. I just `poetry install` and then navigate to `askar_tools` and run `poetry run pytest .`.

You can use other pytest commands to run an individual test.

ianco:
ok thanks, I'll add this to the README

ianco:
Probably more useful to be able to test this with a back-end postgres database rather than having to worry about import/export for sqlite wallets (and then the tests could run in a separate docker container, and just point to the external database).

jamshale:
I agree. For this scenario testing against a postgres db would be easier and better. The sqlite testing was much more difficult with restarts in acapy-tools. Also would be good to have postgres getting tested more here in general.

ianco:
@jamshale I'm getting errors running the tests.

I ran `poetry install --with=dev` (to include the dev dependencies, basically `acapy-controller`, which failed with the following error:

```
HTTPSConnectionPool(host='github.com', port=443): Max retries exceeded with url: /indicio-tech/acapy-minimal-example.git/info/refs?service=git-upload-pack (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)')))
```

I updated the project to use `acapy-controller = "~=0.2.0"` (the published version of the library) and then the install worked.

After running `poetry run pytest .` however everything failed:

```
[...]

PR #3400, opened 2024-12-13
Aca-Py test scenario including a container restart (with aca-py version upgrade)

ianco:
Test for Issue #3269.

Script starts agents using docker-compose - alice agent is based on the Aca-Py 1.0.1 release image.  Alice connects with and then issues a credential to bob and then the test shuts down the alice agent, restarts based on the `acapy-test` image (based on the local aca-py codebase) and then asks for a proof.

This is working!  I'm going to release this PR as a simple test-with-restart and do a separate test with an askar-to-anoncreds wallet upgrade.


jamshale:
I believe you can add assertions into the example.py script. Basically if it exits with exit code 1 it fails and if it exists with exit code 0 it passes. There's some caveats with this though. I had a test exiting with code 0 locally but it failed with github actions because the acapy-controller library never received an event it was expecting.

If the example raises an exception then it will fail.

jamshale (approved):
Looks good to me! 

Some of those docker attrs are pretty complicated and might be fragile to upgrading the docker package, but upgrading it won't be necessary for a testing project so that ok.

ianco:
> Some of those docker attrs are pretty complicated and might be fragile to upgrading the docker package, but upgrading it won't be necessary for a testing project so that ok.

Yah I agree.  For some of the docker stuff it was a bit of trial and error so I'm not sure how fragile these tests are going to be :-S

dbluhm (approved):
I'm interested in potentially pulling some of that docker management complexity into an extra of the acapy_controller library in the future. That could be a convenient way to develop more of this style of test. As it stands, I think this is a great addition and I'm eager to get this kind of testing into the pipeline.

ianco:
> I'm interested in potentially pulling some of that docker management complexity into an extra of the acapy_controller library in the future. That could be a convenient way to develop more of this style of test. 

Agree.  I've started a second test (to include an askar-to-anoncreds) upgrade.  For now I'll refactor the common stuff into a `util.py` (in acapy repo) but agree this should eventually be a feature of the `acapy_controller`.

PR #3407, opened 2024-12-17
Add explicit token for publishing

ryjones:
[Based on this](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#advanced-release-management)

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -27,3 +27,5 @@ jobs:
           poetry build
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_HYPERLEDGER }}


WadeBarnes (commented):
Does the repository url need to be passed to `pypa/gh-action-pypi-publish`?  I see the repository url defined in the `environment` section, but it's not used in the publish step.

jamshale:
This would only need to be done for the 0.12.lts branch. 

I'm not sure if we need to do this. I'm going to look at manually publishing for the backporting instead. I think it should be easy and we won't need to do it unless there's a critical problem for these LTS branches.

swcurran:
For the 0.11.lts branch, we’ll not need to do so again — LTS EoL is January 2025.

For the 0.12.lts branch, we have at least 9 more months. We’ll probably declare 1.1.x as the LTS branch, which will start the clock on ending 0.12.lts support.

jamshale:
I think this should work if we target 0.12.lts branch instead of main. I can open a new PR and try it. If it does work then we don't need to bug Ry to do it manually everytime.

jamshale:
Yay. This worked!

image

Thanks @ryjones . Going to close this PR.

PR #3410, opened 2024-12-18
Scenario test with anoncreds wallet upgrade and restart

ianco:
Also a bit of refactoring ...

Alice (issuer - askar) issues 2 credentials and revokes 1 to each of 3 holders (2 askar and 1 anoncreds).

The revocation registry size is set to `5` to ensure there will be a `full` registry.

Alice and one of the askar holders are then upgraded to anoncreds wallets.

### DIFF
--- a/scenarios/examples/util.py
+++ b/scenarios/examples/util.py
@@ -45,6 +45,21 @@ def wait_until_healthy(client, container_id: str, attempts: int = 350, is_health
     raise TimeoutError("Timed out waiting for container")
 
 
+def update_wallet_type(agent_command: List, wallet_type:str) -> str:
+    for i in range(len(agent_command)-1):
+        if agent_command[i] == "--wallet-type":
+            agent_command[i+1] = wallet_type
+            return wallet_type
+    raise Exception("Error unable to upgrade wallet type to askar-anoncreds")
+
+
+def get_wallet_name(agent_command: List) -> str:
+    for i in range(len(agent_command)-1):
+        if agent_command[i] == "--wallet-name":
+            return agent_command[i+1]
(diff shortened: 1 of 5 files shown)


jamshale:
Looks good 👍 Happy to have this scenario tested better.

scenarios/examples/util.py
```diff
@@ -0,0 +1,427 @@
...
+    inspect_results = container.attrs
+    return not inspect_results["State"]["Running"]
+
+
+def wait_until_healthy(client, container_id: str, attempts: int = 350, is_healthy=True):
+    """Wait until container is healthy."""
+    container = client.containers.get(container_id)
+    print((container.name, container.status))
+    for _ in range(attempts):
+        if (is_healthy and healthy(container)) or unhealthy(container):
```
swcurran, line 40: I gotta ask... "healthy" or "unhealthy" is fine??

ianco: "unhealthy" means we're shutting down the agent and we're checking that it's actually down (we let it shutdown cleanly).

swcurran: Sounds good.

jamshale (approved):
Looks really good 👍This is what I was hoping for. Such a useful testing tool to have.

Only thing I'm not sure about is if the credentials actual revocation status is being tested after the upgrade. I remember being a bit concerned that the indexing would line up and the revocation size was identical. I can't remember exactly. There was an issue with this at one point that got fixed.

ianco:
> Only thing I'm not sure about is if the credentials actual revocation status is being tested after the upgrade.

Good point, I'll add some more post-upgrade testing to confirm everything is upgraded properly (for both issuer and holder).

PR #3413, opened 2024-12-20
Fix indy fallback format in presentation from holder

jamshale:
I'm not exactly sure how this was missed by the previous tests. The was a spot in the presentation protocol handling where the anoncreds holder wasn't sending the correct format to a indy verifier.

Good news is the testing is better.

My ide did some automatic linting on the test files but there were no actual changes to the tests.


jamshale:
Don't think this really requires unit test coverage because it's well covered by the scenario test.

swcurran (approved):

PR #3173, opened 2024-08-16
:sparkles: Add ordering options to askar scan and fetch_all methods

ff137:
:sparkles: Adds `order_by` and `descending` options to paginated scan or fetch_all methods

❗ This PR is waiting for next official Askar release

:construction: At present, uses TestPyPI package to test ordering functionality introduced in Askar: https://github.com/hyperledger/aries-askar/pull/291

To-do: test coverage, and point to next aries-askar release that includes this change.

Notes: currently it only appears to ever make sense to order by the "id" column, since record field values are encrypted.
So, it does feel a bit weird allowing users to select an "order_by" column when there's only one valid choice. But, I still believe it's better than nothing. We at least add the functionality for control over ordering, and it can be disabled (for whatever reason) by requesting None. Ordering by id is effectively the same as ordering by time created (AFAICT), so the newest records can be received at the top by selecting descending=True. Perhaps this can be made default behavior.

At present (before this Askar PR is merged), there is no guaranteed ordering when fetching records, for the scan (paginated queries) or fetch all methods. This means there's no solid guarantee that all records will be retrieved while scanning (with increasing offsets), and ordering of records may be inconsistent across many fetch_all calls. So, resolving that is the motive for this contribution.

Will close #3001

PR #3429, opened 2025-01-07
Change did:tdw resolver naming to did:webvh

jamshale:
Changes the `tdw` resolver to the new `webvh` naming.


dbluhm (approved):

ff137:
`did-webvh` is a blocker for running with the latest askar version, as noted here: https://github.com/decentralized-identity/didwebvh-py/pull/18

A new release for `did-webvh` is not immediately possible, so a `did-tdw` release was prepared instead: https://github.com/decentralized-identity/didwebvh-py/pull/20

did-tdw 0.2.2 is on pypi but did-webvh is still 0.2.1

With all that said -
There is a critical scalability fix in the latest askar release: https://github.com/openwallet-foundation/askar/pull/350
And we can't upgrade askar along with did-webvh, yet.

So @jamshale @swcurran @dbluhm I would like to propose reverting this PR, meaning the old did-tdw name will be used in the meantime, so that we can upgrade askar to the latest release instead.

A side-effect is this will unblock https://github.com/openwallet-foundation/acapy/pull/3173, which has been waiting 6 months for a new askar release (!)

If it causes unnecessary work to revert this, and a new did-webvh release can be prepared in due time, then I'd say that's ok, and up to you whether we wait or not. I can pull all the upgrades into our acapy fork manually. I'm just looking forward to rebasing our fork being less of a headache 😄 since we're using #3173 and a bunch of other customized things.

jamshale:
We should be able to get a new release of did_webvh out soon. I'd rather not revert this. I'm not sure what's taking so long to get a simple askar upgrade for did_webvh.

This has been a bit more annoying than I thought, though. It might be better to completely remove the resolver from acapy and only have it in the plugin instead.

ff137:
Alright -- I see that https://github.com/decentralized-identity/didwebvh-py/pull/19 has gotten a few approvals since I last looked yesterday, so hopefully a new did-webvh release is just around the corner

jamshale:
I don't know why but the did_webvh 0.2.2 library publish failed. Just trying to get it published and then we can get askar upgraded. We don't need the big PR finished to move ahead on this.

PR #3431, opened 2025-01-08
Add some more functionality and checks to the restart/upgrade test

ianco:
Fleshes out the upgrade & restart scenario test (checks record counts, issues and revokes credentials after restart, etc).

There is one event type I'm not sure of (issue credential from anoncreds issuer agent to askar holder agent and one of the holder events is `_anoncreds` rather than `_indy`.  I patched the test to allow this?

### DIFF
--- a/scenarios/examples/util.py
+++ b/scenarios/examples/util.py
@@ -67,9 +67,9 @@ def anoncreds_presentation_summary(presentation: V20PresExRecord) -> str:
         {
             "state": presentation.state,
             "verified": presentation.verified,
-            "presentation_request": request.model_dump(by_alias=True)
-            if request
-            else None,
+            "presentation_request": (
+                request.model_dump(by_alias=True) if request else None
+            ),
         },
         indent=2,
         sort_keys=True,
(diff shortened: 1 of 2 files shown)


scenarios/examples/util.py
```diff
@@ -301,7 +301,7 @@ async def anoncreds_issue_credential_v2(
     )
     holder_indy_record = await holder.event_with_values(
         topic="issue_credential_v2_0_anoncreds"
-        if is_holder_anoncreds
+        if (is_holder_anoncreds or is_issuer_anoncreds)
```
ianco, line 304: This change in `util.py` allows the event `issue_credential_v2_0_anoncreds` from an askar holder (this is what I observed in the test, and also in the alice/faber demo).

jamshale: Tests look great. I do remember this and wasn't exactly sure what to do. I'll need to think about it for a bit.

jamshale (approved):
Looks good! I'm glad this is tested more thoroughly. I meant to do this for a long time.

For the changing the event name I'll think about it a bit. This already requires both parties being on the latest version of acapy so I think adding the anoncreds event handler to an indy agent for forward compatibility would solve the issue.

That's not an issue that needs to be addressed in this PR. I'll create a seperate issue for it.

PR #3432, opened 2025-01-09
Follow up from Release 1.2.0 -- including LTS change

swcurran:
Update references to the previous release follow the new 1.2.0 Release.

Update the snyk configuration to use the 1.2.lts branch.

@jamshale -- take a look at the snyk config. For the 0.12.lts branch, there is no `aries-agent` path. Should we add the `aries-cloudagent`  path that doesn't existin in main? I assume that is OK, since it has been working fine even though there is no `aries-agent` folder in the 0.12.lts branch.

### DIFF
--- a/.github/workflows/snyk-lts.yml
+++ b/.github/workflows/snyk-lts.yml
@@ -5,6 +5,7 @@ on:
     # These branches represent the LTS releases
     branches:
       - 0.12.lts
+      - 1.2.lts
     paths:
       - acapy_agent/**
       - docker/**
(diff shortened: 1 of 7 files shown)


jamshale (dismissed):
The 0.12.lts branch would need to scan it's `aries_cloudagent/**` path for the source code. I haven't looked at this before. Maybe we can just add it here. Not sure what happens when it scans a path that doesn't exist.

jamshale (approved):

PR #3442, opened 2025-01-14
Support P256 keys & did:keys

gmulhearn:
* adds p256 did:key resolution (using Multikey VM type)
* update askarwallet to support sign & verify with p256
* update JWT signing to get the DID key's JWS Alg rather than hardcode to EdDSA
* update parts of `acapy_agent/wallet/keys/manager.py`, removing some hardcoded ed25519 key types


jamshale (commented):
Looks well done to me 👍  Will let others get a chance to review.

acapy_agent/wallet/askar.py
```diff
@@ -878,7 +890,7 @@ def _create_keypair(key_type: KeyType, seed: Union[str, bytes, None] = None) ->
         raise WalletError(f"Unsupported key algorithm: {key_type}")
     if seed:
         try:
-            if key_type == ED25519:
+            if key_type == ED25519 or key_type == P256:
```
ff137, line 893: `if key_type in (ED25519, P256)` for easier expansion

acapy_agent/wallet/askar.py
```diff
@@ -878,7 +890,7 @@ def _create_keypair(key_type: KeyType, seed: Union[str, bytes, None] = None) ->
...
                 # not a seed - it is the secret key
                 seed = validate_seed(seed)
                 return Key.from_secret_bytes(alg, seed)
```
ff137, line 896: Not part of your changes, but I notice in lines 899--901:
```py
        except AskarError as err:
            if err.code == AskarErrorCode.INPUT:
                raise WalletError("Invalid seed for key generation") from None
```
This will swallow AskarError's that don't match that error code, and silently fail. i.e. any other error code will go unlogged / unreported.
And I'm not sure why it's `from None` and not `from err`.

Could we perhaps, while we're here, change that to:
```py
        except AskarError as err:
            if err.code == AskarErrorCode.INPUT:
                raise WalletError("Invalid seed for key generation") from err
            else:
                logger.error(f"Unhandled Askar error code: {err.code}")
                raise
```

acapy_agent/wallet/jwt.py
```diff
@@ -67,11 +69,19 @@ async def jwt_sign(
         if not did:
             raise ValueError("DID URL must be absolute")
 
+    async with profile.session() as session:
+        wallet = session.inject(BaseWallet)
+        did_info = await wallet.get_local_did(did_lookup_name(did))
```
ff137, line 74: The session is now opened twice, and `did_info = await wallet.get_local_did(did_lookup_name(did))` is called twice (later again in line 93)

Perhaps the code can be refactored to just open the session once, and remove the 2nd `get_local_did` call

acapy_agent/wallet/jwt.py
```diff
@@ -143,33 +155,48 @@ async def resolve_public_key_by_kid_for_verify(profile: Profile, kid: str) -> st
             "Dereferenced resource is not a verification method"
         )
 
-    if not isinstance(vmethod, SUPPORTED_VERIFICATION_METHOD_TYPES):
```
ff137, line 146: The `SUPPORTED_VERIFICATION_METHOD_TYPES` var is now unused in the codebase. 

Perhaps just for maintainability / "record keeping" of supported methods, we can keep using this var with the following check:
```py
    if not isinstance(vmethod, SUPPORTED_VERIFICATION_METHOD_TYPES):
        raise InvalidVerificationMethod(
            f"Dereferenced method {type(vmethod).__name__} is not supported"
        )
```

So, return the above check before handling the vmethod, and replace the final `# unsupported` comment and line with an unimplemented exception (for cases where it's in the `SUPPORTED_VERIFICATION_METHOD_TYPES` list, but not implemented in the handling here).
i.e. just change "not supported" to "not implemented" in line 171

I know this sounds pedantic, but just my way of thinking about it, as supported methods will grow over time. Open to alternatives.

acapy_agent/wallet/jwt.py
```diff
@@ -143,33 +155,48 @@ async def resolve_public_key_by_kid_for_verify(profile: Profile, kid: str) -> st
...
 
 async def jwt_verify(profile: Profile, jwt: str) -> JWTVerifyResult:
     """Verify a JWT and return the headers and payload."""
     encoded_headers, encoded_payload, encoded_signature = jwt.split(".", 3)
     headers = b64_to_dict(encoded_headers)
-    if "alg" not in headers or headers["alg"] != "EdDSA" or "kid" not in headers:
-        raise BadJWSHeaderError("Invalid JWS header parameters for Ed25519Signature2018.")
+    if (
+        "alg" not in headers
+        or (headers["alg"] != "EdDSA" and headers["alg"] != "ES256")
```
ff137, line 181: `or (headers["alg"] not in ("EdDSA", "ES256"))`

^ perhaps that should be its own var somewhere. `SUPPORTED_JWT_ALGS`? Idk.

acapy_agent/wallet/jwt.py
```diff
@@ -143,33 +155,48 @@ async def resolve_public_key_by_kid_for_verify(profile: Profile, kid: str) -> st
...
+        or "kid" not in headers
+    ):
+        raise BadJWSHeaderError("Invalid JWS header parameters")
```
ff137, line 184: There are a lot of different ways this error can be raised. Perhaps to assist users with debugging, we can be more descriptive.
Such as:
```py
    if "alg" not in headers:
        raise BadJWSHeaderError("Missing 'alg' parameter in JWS header")
    if "kid" not in headers:
        raise BadJWSHeaderError("Missing 'kid' parameter in JWS header")
    if headers["alg"] not in SUPPORTED_JWT_ALGS:
        raise BadJWSHeaderError(
            f"Unsupported 'alg' value in JWS header: '{headers['alg']}'. "
            f"Supported algorithms: {', '.join(SUPPORTED_JWT_ALGS)}"
        )
```

acapy_agent/wallet/askar.py
gmulhearn: good find - added that in

acapy_agent/wallet/jwt.py
gmulhearn: agreed

acapy_agent/wallet/jwt.py
gmulhearn: yea fair enough. let me know if changes is what you had in mind

acapy_agent/wallet/jwt.py
gmulhearn: nice. fixed

acapy_agent/wallet/jwt.py
ff137: I was thinking to do the "if not supported" check ahead of the specific vmethod checks, but in effect it's the same thing 👍

acapy_agent/wallet/jwt.py
```diff
@@ -67,20 +71,25 @@ async def jwt_sign(
...
-    encoded_payload = dict_to_b64(payload)
-
     async with profile.session() as session:
         wallet = session.inject(BaseWallet)
-        LOGGER.info(f"jwt sign: {did}")
         did_info = await wallet.get_local_did(did_lookup_name(did))
+
+        header_alg = did_info.key_type.jws_algorithm
+        if not header_alg:
+            raise ValueError("DID key type cannot be used for JWS")
```
ff137, line 80: Perhaps include the key type in the error message, for clarity:
```            raise ValueError(f"DID key type '{did_info.key_type}' cannot be used for JWS")```

ff137 (dismissed):
Nice, looks good to me 👍

ff137 (approved):

acapy_agent/wallet/jwt.py
ff137: I see, that makes sense. It was mainly just an idea for maintainability / consistency, but in this case, wallet/jwt functionality does now diverge from jsonld/routes, so I think it's good as it is.

ff137 (approved):

dbluhm (approved):

PR #3450, opened 2025-01-17
Pass the correct key for multitenant single wallets

jamshale:
For multitenant single wallet deployments the `multitenant_sub_wallet` needs the admin key. I don't exactly know how I messed this up. I think, I thought that the subwallet key was blank because I was trying to fix a problem where it had been created with a blank key.

This passes the admin wallet key to the subwallet and allows it to get rekey'd along with the admin wallet key.


jamshale:
The upgrade scenario test that Ian added a little while age would have caught this if it tested multitenancy. I'm going to create add this as a separate PR.

PR #3455, opened 2025-01-20
:arrow_up: Upgrade dependencies

ff137:
Includes major upgrades, where I've eyeballed changelogs to confirm they don't include relevant breaking changes. We'll let the tests be the judge.

:heavy_minus_sign: Removes `async-timeout`, as it's been pulled into `asyncio`

___

The only packages that remain outdated:
- apispec (6.6.1  vs 6.8.1 )
- webargs (8.4.0  vs 8.6.0)

^ these are held back by `aiohttp-apispec-acapy`
@jamshale I've upgrade the fork to be more permissive for these dependencies, so they can be upgraded: https://github.com/ff137/aiohttp-apispec

- aries-askar (0.3.2 vs 0.4.1)
- cached-property (1.5.2 vs 2.0.1)

^ held back by did-webvh, which needs more relaxed restriction on askar: https://github.com/decentralized-identity/didwebvh-py/pull/18

- prompt-toolkit  (2.0.10 vs 3.0.50)
^ This is the only one that seems to include some breaking changes. They don't seem significant, but it requires some focus


jamshale (dismissed):
I'm happy to move forward on this but we need to get a patch release done before these changes go in. Will block until that is complete.

ff137:
Will also have to debug why agent in interop test doesn't start up

ff137:
@swcurran I included unpinning of the remaining dependencies that were set to 1 specific version, and some last improvements.

Tried upgrading to next marshmallow release, but 1 pytest broke, and didn't have time to dig into it now. So leaving marshmallow upgrade for later

jamshale:
https://pypi.org/project/aiohttp-apispec-acapy/3.0.3/ Is now available with the unpinned dependencies if you want to include it in this PR.

ff137:
> https://pypi.org/project/aiohttp-apispec-acapy/3.0.3/ Is now available with the unpinned dependencies if you want to include it in this PR.

Awesome, thank you! Included

One step closer to running with 3.13 :eyes:

acapy_agent/messaging/jsonld/tests/test_routes.py
```diff
@@ -303,6 +303,10 @@ async def asyncSetUp(self):
             headers={"x-api-key": "secret-key"},
         )
 
+    async def asyncTearDown(self):
+        # Ensure the event loop is closed
+        await self.profile.close()
```
jamshale, line 308: 👍

jamshale (approved):
Looks good to me and tests are all passing so should be safe 👍

PR #3457, opened 2025-01-20
Add Multi-tenancy single wallet upgrade test

jamshale:
This would have caught the recent regression with a single wallet multi-tenancy deployment.

It adds a tenant to the upgrade scenario and re-uses the existing token and connection to do a basic issuance..

Another test should be added for multi-wallet multi-tenancy deployment as well but that can be done later.


jamshale:
I confirmed that this test failed with the error with version 1.2.0

ianco (approved):

PR #3461, opened 2025-01-24
Restore connection route tests

dbluhm:
This is a follow up to #3184 to address an issue @ff137 spotted; I failed to move the connection route tests from the connections protocol package to the core connections package. The routes themselves had been moved so the tests should have accompanied them. Several of those tests were no longer needed since they dealt with the connections protocol specifically so those have been removed.


acapy_agent/connections/tests/test_routes.py
```diff
@@ -0,0 +1,506 @@
...
+
+            await test_module.connections_retrieve(self.request)
+            mock_response.assert_called_once_with({"hello": "world"})
+
+    async def test_connections_endpoints(self):
+        self.request.match_info = {"conn_id": "dummy"}
+
+        with (
+            mock.patch.object(
+                test_module, "ConnectionManager", autospec=True
```
ff137, line 164: I think this should now be `"BaseConnectionManager"`

dbluhm:
Security "hotspots" are a couple of fake IP addresses used in the test as a mock return value. Safe to ignore.

ff137 (approved):

PR #3470, opened 2025-01-27
fix: ensure profile names are unique

dbluhm:
The `profile.name` property is expected to be unique. When using single-wallet-askar, it is not unique and is always `multitenant_sub_wallet` or whatever the wallet for all the profiles is called. This causes problems in a few key areas. This fix uses the profile_id (when present, i.e. when using single-wallet-askar) to make sure profile.name is unique. I will open an issue to further detail the extent of this problem so we can discuss how far back we need to go.

### DIFF
--- a/acapy_agent/askar/profile.py
+++ b/acapy_agent/askar/profile.py
@@ -42,7 +42,9 @@ class AskarProfile(Profile):
         profile_id: Optional[str] = None,
     ):
         """Create a new AskarProfile instance."""
-        super().__init__(context=context, name=opened.name, created=opened.created)
+        super().__init__(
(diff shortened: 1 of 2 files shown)


jamshale:
I'll look into why that test is failing...

jamshale:
So... Maybe this could be done better but the `AskarAnoncredsProfile` class needs to be updated with this change as well. https://github.com/openwallet-foundation/acapy/blob/main/acapy_agent/askar/profile_anon.py#L57

This is why the failing test is happening. It's reloading the sub-wallet profile as Anoncreds and then getting the sub-wallet name again instead of the id.

jamshale (approved):

PR #3472, opened 2025-01-27
Add multiple kid per key feature, kid removal and get by kid routes

PatStLouis:
The goal of this PR is to enable assigning multiple kid to a keypair, and unbind them.

This serves 2 purposes:
- Publish the same key in 2 different locations, such as a `did:webvh` and parallel `did:web`
- Publish the same key in 2 different format, such as `Multikey` and `JsonWebKey`


acapy_agent/wallet/askar.py
```diff
@@ -131,14 +135,58 @@ async def assign_kid_to_key(self, verkey: str, kid: str) -> KeyInfo:
         if not key_entry:
             raise WalletNotFoundError(f"No key entry found for verkey {verkey}")
 
+        try:
+            existing_kid = key_entry.tags.get("kid")
```
jamshale, line 139: key_entry.tags.get("kid", [])

acapy_agent/wallet/askar.py
```diff
@@ -131,14 +135,58 @@ async def assign_kid_to_key(self, verkey: str, kid: str) -> KeyInfo:
...
+        key = cast(Key, entry.key)
+        verkey = bytes_to_b58(key.get_public_bytes())
+        metadata = cast(dict, entry.metadata)
+        key_types = self.session.inject(KeyTypes)
+        key_type = key_types.from_key_type(key.algorithm.value)
+        if not key_type:
+            raise WalletError(f"Unknown key type {key.algorithm.value}")
+        try:
+            existing_kid.remove(kid)
+        except Exception:
```
jamshale, line 184: Should catch the correct type of exception for removing entry that doesn't exist. Can't remember of the top of my head.

In general try to only catch the base Exception class when nessecary.

swcurran:
Replaced by #3762

PR #3474, opened 2025-01-28
Upgrade askar and did_webvh

jamshale:
Upgrades aries-askar to 0.4.2 and did-webvh library which also has an askar dependency.


ianco (approved):
This is working for me.

ff137:
Thank you 🙏

PR #3475, opened 2025-01-28
Askar upgrade and fix profile unique names

jamshale:
Upgrades aries askar to 0.4.2 and fixes the unique names for multitenant single-wallet.


swcurran (approved):

swcurran:
Looks like there is another Askar fix going in that might be useful to have. Let’s check on that and perhaps update this PR first. https://github.com/openwallet-foundation/askar/pull/354

jamshale:
Updated aries-askar to 0.4.3

dbluhm:
Whats the difference between this one and #3477?

swcurran:
The LTS version — 0.12 vs. 1.2.  It’s all in the details.  😃

PR #3485, opened 2025-01-31
fix typo in error message of indy credential offer

zoblazo:
anoncreds rather than anonreds


swcurran (approved):

swcurran (changes requested):
The Commit needs to be DCO signed - Developer Certificate of Origin - https://github.com/apps/dco. To fix, see: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md.

See the Details beside the failed DCO testt for host to fix the failed check.

PR #3495, opened 2025-02-06
Fix Class import for AnonCreds Registry routes

PatStLouis:
The wrong validator was being imported for 2 of the AnonCreds revocation management routes, validating the ID against the indy method.

### DIFF
--- a/acapy_agent/anoncreds/routes.py
+++ b/acapy_agent/anoncreds/routes.py
@@ -25,7 +25,10 @@ from ..messaging.valid import (
     UUIDFour,
 )
 from ..revocation.error import RevocationNotSupportedError
-from ..revocation_anoncreds.routes import RevocationModuleResponseSchema, RevRegIdMatchInfoSchema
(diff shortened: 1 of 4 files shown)


jamshale:
Actually. Forget that comment. I see that those are the same. Should be renamed but not important.

jamshale:
Interop tests have been a problem today. This is small enough I can merge it without them passing.

PatStLouis:
@jamshale I'm not sure what a cred_rev_id is and there doesn't seem to be the equivalent for anoncreds. I'll investigate

jamshale:
@PatStLouis You can ignore that. It could be renamed to remove the INDY'ism but the cred_rev_id object is the same in anoncreds and indy.

PatStLouis:
@jamshale there seems to be a circular import issue, I'll look into it

PatStLouis:
@jamshale I've renamed them for consistency's sake

jamshale (approved):

PR #3498, opened 2025-02-09
Upgrade to bookworm

jamshale:
Same as https://github.com/openwallet-foundation/acapy/pull/3494 but with some libraries removed that we don't need installed into the base image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -46,23 +46,15 @@ RUN apt-get update -y && \
     apt-get install -y --no-install-recommends \
     apt-transport-https \
     ca-certificates \
-    build-essential \
-    bzip2 \
-    curl \
-    git \
-    less \
     libffi-dev \
     libgmp10 \
-    liblzma5 \
     libncurses5 \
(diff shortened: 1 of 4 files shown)


jamshale:
The OATH Interop tests have been having issues. They aren't failing because of this PR.

swcurran (approved):

docker/Dockerfile
```diff
@@ -46,24 +46,17 @@ RUN apt-get update -y && \
     apt-get install -y --no-install-recommends \
     apt-transport-https \
     ca-certificates \
-    build-essential \
-    bzip2 \
     curl \
-    git \
```
ff137, line 52: Just a heads-up -- removing `git` here means that installation instructions for acapy-plugins now won't work:
```dockerfile
# install plugins as binaries
RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@main#subdirectory=basicmessage_storage
```

Ran into this now when testing with main. Might want to re-include git for convenience

PR #3508, opened 2025-02-10
Grouped upgrades - Week 7, 2025

jamshale:
Grouped upgrades reported by dependabot.


swcurran (approved):

jamshale:
The lock file has quite a few changes because I updated it with poetry version 2 for the first time.

swcurran:
Funny — just to test things out, I checkout out this PR and then ran `poetry update` and got a number of differences from your poetry.lock file.  I’m running `Poetry (version 2.0.1)`. 

Should there have been differences?

Differences:

- `aiohappyeyeballs`
- `attrs`
- `certifi`
- `coverage`
- `did-webvh`
- `identify`
- `lxml`
- `pydantic`

jamshale:
I ran poetry lock after manually updating the libraries reported by dependabot.

Poetry update will try and update all libraries based on the version tagging.

We can possibly try and update, but it's not really necessary and can sometimes have problems you weren't expecting.

PR #3515, opened 2025-02-13
:heavy_plus_sign: Re-add `git` to Dockerfile

ff137:
See https://github.com/openwallet-foundation/acapy/pull/3498/files#r1954246291


jamshale (approved):
I was trying to think of why we would need git in the base image but forgot about installing plugins. Fine to add it back.

PR #3520, opened 2025-02-13
:sparkles: Add regex validators for Anoncreds models

### LINKED ISSUE #3519: :art: Update regex validators for AnonCreds models
There are 3 AnonCreds model validators:
```py
class AnoncredsCredDefId(Regexp):
    """Validate value against anoncreds credential definition identifier specification."""

    EXAMPLE = "did:(method):3:CL:20:tag"
    PATTERN = r"^(.+$)"

    def __init__(self):
        """Initialize the instance."""

        super().__init__(
            AnoncredsCredDefId.PATTERN,
            error="Value {input} is not an anoncreds credential definition identifier",
        )


class AnoncredsRevRegId(Regexp):
    """Validate value against anoncreds revocation registry identifier specification."""

    EXAMPLE = "did:(method):4:did:<method>:3:CL:20:tag:CL_ACCUM:0"
    PATTERN = r"^(.+$)"

    def __init__(self):
        """Initialize the instance."""

        super().__init__(
            AnoncredsRevRegId.PATTERN,
            error="Value {input} is not an anoncreds revocation registry identifier",
        )


class AnoncredsSchemaId(Regexp):
    """Validate value against indy schema identifier specification."""

    EXAMPLE = "did:(method):2:schema_name:1.0"
    PATTERN = r"^(.+$)"

    def __init__(self):
        """Initialize the instance."""

        super().__init__(
            AnoncredsSchemaId.PATTERN,
            error="Value {input} is not an anoncreds schema identifier",
        )
```

These patterns (`^(.+$)`) are just matching on non-empty strings.

There should be a to-do assigned to defining these validation patterns, with some tests asserting that they pass on expected values, and fail on bad ones.

Potentially what the patterns should be, as generated for me by GPT:
```py
class AnoncredsCredDefId(Regexp):
    """Validate value against anoncreds credential definition identifier specification."""

    EXAMPLE = "did:(method):3:CL:20:tag"
    PATTERN = r"^did:(?P<method>[a-zA-Z0-9]+):3:CL:(?P<schema_id>\d+):(?P<tag>[a-zA-Z0-9_-]+)$"


class AnoncredsRevRegId(Regexp):
    """Validate value against anoncreds revocation registry identifier specification."""

    EXAMPLE = "did:(method):4:did:<method>:3:CL:20:tag:CL_ACCUM:0"
    PATTERN = r"^did:(?P<method>[a-zA-Z0-9]+):4:did:(?P<method>[a-zA-Z0-9]+):3:CL:(?P<schema_id>\d+):(?P<tag>[a-zA-Z0-9_-]+):CL_ACCUM:(.+)$"


class AnoncredsSchemaId(Regexp):
    """Validate value against indy schema identifier specification."""

    EXAMPLE = "did:(method):2:schema_name:1.0"
    PATTERN = r"^did:(?P<method>[a-zA-Z0-9]+):2:(?P<schema_name>[a-zA-Z0-9_-]+):(?P<version>[0-9.]+)$"
```

I would just need assistance / input wrt these AI guesses. If someone can provide a range of possible, valid values for these, then we can write some tests too

ff137:
(Resolves #3519)

### DIFF
--- a/acapy_agent/messaging/valid.py
+++ b/acapy_agent/messaging/valid.py
@@ -467,10 +467,10 @@ class IndyCredDefId(Regexp):
     EXAMPLE = "WgWxqztrNooG92RXvxSTWv:3:CL:20:tag"
     PATTERN = (
         rf"^([{B58}]{{21,22}})"  # issuer DID
-        f":3"  # cred def id marker
-        f":CL"  # sig alg
+        ":3"  # cred def id marker
+        ":CL"  # sig alg
         rf":(([1-9][0-9]*)|([{B58}]{{21,22}}:2:.+:[0-9.]+))"  # schema txn / id
-        f":(.+)?$"  # tag
+        ":(.+)?$"  # tag
     )
 
     def __init__(self):
@@ -555,7 +555,7 @@ class IndyRevRegId(Regexp):
         rf"^([{B58}]{{21,22}}):4:"
         rf"([{B58}]{{21,22}}):3:"
         rf"CL:(([1-9][0-9]*)|([{B58}]{{21,22}}:2:.+:[0-9.]+))(:.+)?:"
-        rf"CL_ACCUM:(.+$)"
+        r"CL_ACCUM:(.+$)"
     )
 
     def __init__(self):
--- a/acapy_agent/messaging/valid.py
+++ b/acapy_agent/messaging/valid.py
@@ -486,7 +486,9 @@ class AnoncredsCredDefId(Regexp):
     """Validate value against anoncreds credential definition identifier specification."""
 
     EXAMPLE = "did:(method):3:CL:20:tag"
-    PATTERN = r"^(.+$)"
+    PATTERN = (
+        r"^did:(?P<method>[a-zA-Z0-9]+):3:CL:(?P<schema_id>\d+):(?P<tag>[a-zA-Z0-9_-]+)$"
+    )
 
     def __init__(self):
         """Initialize the instance."""
@@ -531,7 +533,10 @@ class AnoncredsSchemaId(Regexp):
     """Validate value against indy schema identifier specification."""
 
     EXAMPLE = "did:(method):2:schema_name:1.0"
-    PATTERN = r"^(.+$)"
+    PATTERN = (
+        r"^did:(?P<method>[a-zA-Z0-9]+):2:(?P<schema_name>[a-zA-Z0-9_-]+):"
+        r"(?P<version>[0-9.]+)$"
+    )
 
     def __init__(self):
         """Initialize the instance."""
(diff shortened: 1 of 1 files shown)


acapy_agent/messaging/valid.py
```diff
@@ -485,8 +485,11 @@ def __init__(self):
 class AnoncredsCredDefId(Regexp):
     """Validate value against anoncreds credential definition identifier specification."""
 
-    EXAMPLE = "did:(method):3:CL:20:tag"
-    PATTERN = r"^(.+$)"
+    EXAMPLE = "did:method:example:3:CL:12:tag1"
+    PATTERN = (
```
jamshale, line 489: I left a comment on the issue but I don't think we can use a pattern like this. The `3:CL:12` stuff is indy specific. For the cheqd and the upcoming webvh method these identifiers are much different and can be really almost anything. That's why I changed it to this. I meant to just match anything.

I think every pattern with atleast have the did method at the beginning. So we could have a `did:<any sting only letters>:<any string>` pattern here.

Same with the other object id's.

ff137: Ah yes, but a validator that matches anything is not really a validator :-)

If it must start with a qualified did, that's at least something. The pattern after that can be more random, but it assists in understanding the code. So I think it's worth sussing out the true restrictions on what the patterns are -- and then expanding those patterns later if we run into issues

jamshale: Yes. But I remember the legacy indy(sov) method that still persists (try to change this), doesn't even have the `did:sov` part at the beginning.

I think I was lost on what to do for the validator and that's why I had it this way. Maybe not having a validator at all would be better for now?

ff137: I figure that's an option too. Skipping validation, just to save on regex compute and make the earth greener 😄

Of course it's much of a muchness, doing `.+` regex check vs skipping it. 
I figured it's at least worth marking this as a to-do, so that it's tracked as an unanswered question: what is the correct validation for anoncreds cred def / schema id / rev reg id. If we as a community can't answer that, then something's wrong 😄 if we're still dragging technical debt of needing to handle unqualified dids, maybe that should be prioritised

PatStLouis: I'm starting to look at unqualified dids and the such. I would like for acapy to let go did:sov entirely (qualified & unqualified), and use the latest set of did methods to enforce regex. It should be something like `did:{method}:{identifier}/{resourcePath}`. Sov used semicolons instead of path for the resource identifier, where other did methods (indy, cheqd, webvh) all use did uris. This would make a more compelling regex that ensures that the resourceId is a did uri instead of an abstract per method string.

https://hyperledger.github.io/indy-did-method/#did-urls-for-indy-object-identifiers
https://identity.foundation/didwebvh/v1.0/#did-url-resolution
https://docs.cheqd.io/product/studio/did-linked-resources

PatStLouis: DID sov simply doesn't have a specification, it's just legacy code from the initial sovrin codebase

ff137:
OBE

PR #3521, opened 2025-02-13
:construction_worker: Dependabot: don't ignore major releases

ff137:
Because they are good to know about
___
As mentioned in ACA-Pug meeting notes: https://lf-openwallet-foundation.atlassian.net/wiki/spaces/ACAPy/pages/82706433/2025-02-04+-+ACA-Py+Users+Group+Community+Meeting

>Enable major version update recommendations by Dependabot: review open PRs and pick-and-choose which ones could be merged as-is and which ones should be actioned by someone (more as a recommendation to update than as a solution). Project is in a testing state that allows us to move forward with this.
___
edit: now ignores patch releases instead

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,9 @@ updates:
       day: "monday"
       time: "04:00"
       timezone: "Canada/Pacific"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
 
   # Maintain dependencies for Python Packages
   - package-ecosystem: "pip"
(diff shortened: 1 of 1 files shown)


ff137:
By the way, I wanted to say this when the dependabot files were first uploaded to begin with:
if the objective is to minimise on dependency noise, then _patch_ releases should be ignored. Not major ones. Patches are obviously more frequent, and they _can_ be ignored, so long as poetry lock file is updated regularly. Patch releases will automatically be pulled in every time the poetry lock file is regenerated, because pyproject.toml typically uses minor version restrictions.

So, instead of weekly dependabot announcements of patch updates, just run poetry lock every week. A github action can be created to do that too ... might help reduce the tedious dependabot PR reviews. @jamshale @dbluhm @swcurran

swcurran:
@WadeBarnes — adding you to this.

If the approach that @ff137 proposes could replace the current Manic Monday PRs, would be awesome.  What do you think about a manual poetry lock update manually for now, and automated thereafter.  I would love to see that on ACA-Py and ACA-Py Plugins — proactive, but with much less email noise.

WadeBarnes (approved):

PR #3522, opened 2025-02-13
:zap: Remove `--cov` from pytest.ini_options

ff137:
Speeds up local test runs, especially when only wanting to run a few tests at a time
___
Previously, running a single unit test took me 7.5 seconds, and now it takes 0.5 seconds.

This is because it takes ~7 seconds to generate the coverage report (in terminal and xml file) for the whole acapy_agent folder, and the pyproject.toml (`[tool.pytest.ini_options]`) configures that all pytest commands include a coverage request.

This is hugely unnecessary to be forced on the developer every time they type pytest

If a developer wants to see a coverage report, they can add --cov at the end.
___
Impact: running tests locally are faster, and a developer will have to now type --cov if they ever want to see a coverage report.

Also, test output is more readable now, since it won't always include 100s of lines of coverage info.

No modification necessary for the github workflows, since they manually specify to run with --cov


swcurran:
This is for developers to weigh in on, but makes total sense to me!  Nice!

pyproject.toml
```diff
@@ -152,6 +152,7 @@ omit = ["*/tests/*", "demo/*", "docker/*", "docs/*", "scripts/*"]
 data_file = "test-reports/.coverage"
 
 [tool.coverage.report]
+# include = ["acapy_agent"]
```
jamshale, line 155: Why is this commented out? Won't it only consider the acapy_agent code when a developer does decide to produce a report?

ff137: Ah, that was accidentally added. It didn't work as expected

PR #3530, opened 2025-02-13
:arrow_up: Upgrade sphinx versions in docs

ff137:
To help with the major bumps for `sphinx`, I went through /docs/UpdateRTD.md, and generated the docs locally using the up-to-date versions. I scanned for errors, and everything appears to be good.

It looked like the sphinx dependencies can be removed from the main pyproject.toml file, since they're are separately installed using docs/requirements.txt

@swcurran please review - in the readme it mentions docs are not auto-generated, so a manual re-generated is required. Maybe that just means to run `sphinx-apidoc` / update the `docs/generated` folder, which is included here


swcurran (approved):
Looks good to me.  I was able to repeat my “normal” steps with a bit of a glitch on the update to the sphinx version. I’ll figure that out — might add a dockerfile to make it easier. 

I’m not sure anyone is using the ReadTheDocs, but it takes all of 3 minutes per release, so I keep doing it :-).  I do get a whole bunch of autodoc warnings when I generated that are annoying, but I’ve not dug into them.

My only concern with this is if the upgrade to Sphinx and the RTD Themes works at readthedocs.org — but worst case we just adjust.

Awesome to get the sphinx requirements out of the core — should have been done long ago.

Thanks!

swcurran:
Success with the readthedocs.org update -- RTD generation still working.

Thanks!

PR #3531, opened 2025-02-13
:white_check_mark: Fix demo playground example tests

ff137:
There's a dependabot PR to upgrade pytest in this folder. So I checked out if it all worked.

Turns out these tests were failing out of the box, so here are some fixes


swcurran (approved):

ff137:
Note: I didn't run the tests with the docker images running... but I fixed the ruff errors, and pytest breaking changes (renamed arg)

PR #3545, opened 2025-02-28
fix: tenant access to endpoints leading to access the base wallet

thiagoromanos:
When using acapy with `multitenant` enabled **and** `admin-insecure-mode`, tenant endpoints (like `[GET] /wallet/did` for example) could be accessed **without a bearer token**.

With this fix, single tenant agents will access agent's endpoints with no problem, and multi-tenant agents will require a bearer token to access those endpoints.

### DIFF
--- a/acapy_agent/admin/decorators/auth.py
+++ b/acapy_agent/admin/decorators/auth.py
@@ -78,7 +78,7 @@ def tenant_authentication(handler):
             (multitenant_enabled and authorization_header)
             or (not multitenant_enabled and valid_key)
             or (multitenant_enabled and valid_key and base_wallet_allowed_route)
-            or insecure_mode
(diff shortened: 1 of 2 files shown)


dbluhm (dismissed):

dbluhm:
Looks like one of the tests was perhaps encoded to expect the original behavior

thiagoromanos:
Found the test that was failing. Now, the question is, `insecure_mode` + `multitenant` wasn't supposed to work without a bearer token for tenant endpoints, right? (or am I missing something)

jamshale (approved):

jamshale (commented):
I think this change makes sense as the insecure mode is only supposed to apply to the admin agent.

PR #3553, opened 2025-03-05
Fix prompt for alice/faber demo

ianco:


jamshale (approved):

jamshale:
I tested this change and pasting works for me with the demo.

swcurran:
Any idea when this problem was introduced?  And more importantly — do we need to backport this to the 1.2.lts branch?

esune (approved):

jamshale:
It was the prompt-toolkit major upgrade 2 weeks ago. The demo uses it even though it has it's own requirements.txt file. No need to backport, it was done after release 1.2.2.

esune:
> Any idea when this problem was introduced? And more importantly — do we need to backport this to the 1.2.lts branch?

The change prior to this one on the file was related to updating the `prompt-toolkit` dependency, with the same fix @ianco applied here in other parts of the code. Probably a library update that went unnoticed in the tests?

PR #3554, opened 2025-03-06
Fix revocation accum sync when endorsement txn fails (#3547)

jamshale:
Cherry picks the fix for revocation accum error recovery during endorsement transaction. `630563d`

Not sure if there's any additional commits we want to backport now.


jamshale:
Have to fix the test profile for this. Might need to remove the unit test for this older backport.

jamshale:
I removed the unit test for this patch. Too much has changed with the test profile. Would need to completely re-write it.

esune (approved):

Issue #3556, opened 2025-03-06
Proof request fails with anoncreds wallet once a credential has been revoked


ianco:
This scenario works wit askar wallet but not with anoncreds:

- issue a credential
- request proof (proof validates)
- revoke credential
- request proof (proof fails)
- issue another credential
- request proof

For the last step, with askar the proof validates but with anoncreds it fails.

jamshale:
Should be fixed via https://github.com/openwallet-foundation/acapy/pull/3570.

PR #3557, opened 2025-03-06
Anoncreds proof validation issue (once credential has been revoked)

ianco:
WIP - integration test to reproduce error

See issue #3556 

To run the test:

```
AGENT_PORT_OVERRIDE=9030 ./run_bdd -t @T003-RFC0454.4b
```

The `askar` scenario passes and the `anoncreds` scenario fails.

### DIFF
--- a/.github/workflows/sonar-pr.yml
+++ b/.github/workflows/sonar-pr.yml
@@ -15,7 +15,7 @@ jobs:
         with:
             fetch-depth: 0
       - name: Download PR number artifact
-        uses: dawidd6/action-download-artifact@v8
+        uses: dawidd6/action-download-artifact@v9
         with:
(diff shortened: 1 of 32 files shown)


ianco:
> Is this replace by #3570 ? I think it is the same issue, right?

Same issue.  This PR contains an integration test that demonstrates the issue.

ianco:
> > Is this replace by #3570 ? I think it is the same issue, right?
> 
> Same issue. This PR contains an integration test that demonstrates the issue.

... and the fix in PR #3570 fixes this integration test

jamshale (commented):
Looks good 👍  We could add the `@Release` tag so these get tested when a release is created.

jamshale (approved):

PR #3558, opened 2025-03-06
0.12.lts Patch the fix_ledger_entry improvements

jamshale:
This patches the fix_ledger_entry refactor and updating the local wallet accum value that is currently available in 1.2.lts and improves the revocation downtime by not needing to constantly re-sync the local wallet to post revocations.


jamshale:
Need to fix a small problem with the unit test mocking.

esune (approved):
👍🏻 LGTM

PR #3562, opened 2025-03-08
Fix missing log_timer import in acme.py

parth5805:
The script runners/acme.py was missing an import for log_timer,  which caused an error during execution. This function was not  mentioned in AcmeDemoWorkshop.md, but was required for proper functionality.

Changes:
- Added log_timer to the import list in runners/acme.py.

This ensures that the script runs without import-related errors.

### DIFF
--- a/demo/runners/acme.py
+++ b/demo/runners/acme.py
@@ -14,6 +14,7 @@ from runners.support.utils import (  # noqa:E402
     check_requires,
     log_msg,
     log_status,
+    log_timer,
     prompt,
     prompt_loop,
 )


swcurran:
Thanks for the PR.

I’m sure this is good, but @ianco can you take a quick look please?  Thanks!

ianco (commented):
Is this needed?  I don't see `log_timer` used at all in this code.

parth5805:
@ianco , `log_time` is used in [AcmeDemoWorkshop.md](https://github.com/openwallet-foundation/acapy/blob/main/docs/demo/AcmeDemoWorkshop.md) in this demo, but it is not present in acme.py. 
If we follow the instructions in AcmeDemoWorkshop.md, it will result in an error. To fix this, we should either add log_time to AcmeDemoWorkshop.md or include it in acme.py.
image

ianco:
> @ianco , `log_time` is used in [AcmeDemoWorkshop.md](https://github.com/openwallet-foundation/acapy/blob/main/docs/demo/AcmeDemoWorkshop.md) in this demo

Got it, thanks!

ianco (approved):

PR #3568, opened 2025-03-10
fix post-vcdm 2.0 fetch all calls

PatStLouis:
vcdm 2.0 removed the default issuance date field from all credential and as a result it is no longer possible to sort stored vc. There is still a call to the `fetch_all` askar wallet with a `descending` ordering, which is removed here. This was causing the agent to crash.


PatStLouis:
it was a local config issue, apologies

PR #3569, opened 2025-03-10
Create OOB invitation using an existing key as recipient key

PatStLouis:


jamshale (commented):
I think this makes sense. This is related to https://github.com/openwallet-foundation/acapy/issues/3536, correct? Maybe someone with interest in that issue could provide some feedback.

I would like to see this path unit tested for the oob manager. I think adding it to the `simple` scenario test would also be worth the time.

PatStLouis:
@jamshale no this is not the same, this just adds a way to use an existing key as the `connection_key` when creating a new connection invitation. This doesn't add a new path, it adds a new parameter to an existing path.

jamshale:
Ok. Sounds good. I that case I still think you could add a unit test to make sure the error handling is protected if you use `use_key` when the provided key isn't valid. I'm not sure if it could throw an error or not for the `multikey_to_verkey(self.use_key)` and `await wallet.get_signing_key(verkey)` lines.

A integration test isn't really necessary if it's covered by a unit test for this.

PatStLouis:
ok I will add a couple negative tests and a positive test for this feature.

I just opened #3572 for the issue you referred to in your first comment. It will also need some tests

ff137 (commented):
Not sure, but I think `get_signing_key` needs to be mocked in `test_create_invitation_use_key`

ff137:
> Not sure, but I think `get_signing_key` needs to be mocked in `test_create_invitation_use_key`

@PatStLouis

jamshale:
Going to close this PR as won't do.

PR #3570, opened 2025-03-10
Repair anoncreds holder revocation list request

jamshale:
I'm still not exactly sure when this happened but the anoncreds api changed slightly for fetching the revocation list. The request from the holder wasn't using named parameters and this caused the `to` value to become the `from` value. This would return a list without any indexes revoked in the list. Then because the accumulator value didn't match the verification would fail.


ianco (approved):
This works for me (and it fixes the integration test in PR #3557)

Issue #3536, opened 2025-02-19
Unable to Retrieve Invitation Object After Creation


MonolithicMonk:
### **Feature Request: Reliable Invitation Object Retrieval**

**Description**

Currently, ACA-Py lacks a consistent and reliable method to retrieve invitation objects after creation, especially when using `did:peer:2` or `did:peer:4` DID methods. While invitation URLs are returned upon creation, there is no standard endpoint to fetch the full invitation object later if not initially stored.

**Current Behavior**

* **Inconsistent `/connections/{conn_id}/invitation` endpoint:**  Traction ACA-Py's endpoint fails with `AttributeError: 'InvitationMessage' object has no attribute 'endpoint'` when using `did:peer:2` or `did:peer:4`.
* **Missing Generic Retrieval:**  ACA-Py lacks a standard endpoint or method to retrieve invitation objects based on connection ID or invitation ID.

**Expected Behavior**

Provide a reliable mechanism to retrieve the full invitation object (including the invitation URL and properties like `accepted`, `imageUrl`, `label`) for any connection, regardless of the DID method used.

**Steps to Reproduce**

1. Create an OOB or legacy connection invitation.
2. Do not store the returned invitation object.
3. Attempt to retrieve the invitation object.

**Environment**

* ACA-Py Version: All
* Traction Deployment: 1.1
* DID Method: All (especially `did:peer:2`, `did:peer:4` in Traction)

**Proposed Solutions**

1. **Standardize `/connections/{conn_id}/invitation`:**  Make this endpoint a standard ACA-Py feature and ensure compatibility across all DID methods.
2. **Dedicated `/invitations/{invitation_id}` Endpoint:** Introduce a new endpoint to retrieve invitations by their unique ID.
3. **Embed Invitation in Connection Object:** Include the invitation object as a property within the existing connection object for easy access.

**Workaround**

Manual reconstruction of the invitation object from the connection object is possible but incomplete, as key properties like `accepted`, `imageUrl`, and `label` may be missing.

MonolithicMonk:
I have created a fix within the Traction plugin (`traction_innkeeper`) for the `/connections/{conn_id}/invitation` endpoint.

The issue was that the endpoint was failing to generate the correct invitation URL for peer DIDs, leading to an `AttributeError` and incomplete URLs.

**Fix Details:**

A pull request has been created in the Traction repository to address this: [PR #1510](https://github.com/bcgov/traction/pull/1510).

[...]

PR #3572, opened 2025-03-10
Fetch existing invitation route

PatStLouis:
This is to address #3536


acapy_agent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -1172,6 +1172,18 @@ async def delete_stale_connection_by_invitation(self, invi_msg_id: str):
             for conn_rec in conn_records:
                 await conn_rec.delete_record(session)
 
+    async def fetch_conn_and_oob_record_invitation(self, invi_msg_id: str):
+        """Delete conn_record and oob_record associated with an invi_msg_id."""
```
ff137, line 1176: docstring

acapy_agent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -1172,6 +1172,18 @@ async def delete_stale_connection_by_invitation(self, invi_msg_id: str):
...
+        async with self.profile.session() as session:
+            oob_records = await OobRecord.query(
+                session,
+                tag_filter={
+                    "invi_msg_id": invi_msg_id,
+                },
+                post_filter_positive={},
+            )
```
ff137, line 1184: Can be collapsed to:
```py
            oob_records = await OobRecord.query(
                session, tag_filter={"invi_msg_id": invi_msg_id}
            )
```

acapy_agent/protocols/out_of_band/v1_0/manager.py
```diff
@@ -1172,6 +1172,18 @@ async def delete_stale_connection_by_invitation(self, invi_msg_id: str):
...
+            return oob_records[0].serialize()
```
ff137, line 1185: Is it safe to take the 0th element? Maybe worth adding something like:
```py
if not oob_records:
    raise StorageNotFoundError(f"No record found with invi_msg_id {invi_msg_id}")
```
which will get handled in the route method to return 404

And is it possible to fetch 2 records with the same invitation id?
Even if it's thought to be impossible, I'm an advocate for checking and logging those possibilities (lest it ever happens silently).
```py
if len(oob_records) > 1:
    logger.warning("More than one OobRecord fetched with the same invi_msg_id: %s", invi_msg_id)
```
But that's just pedantic - it's not necessary if it's truly not possible (idk how the invitation id values are generated)

acapy_agent/protocols/out_of_band/v1_0/manager.py
PatStLouis: I changed to the `retrieve_by_id` which is a more direct access to the record.

acapy_agent/protocols/out_of_band/v1_0/manager.py
PatStLouis: I implemented the first suggestion, now fetching with the `oob_id` directly so only 1 record is expected.

acapy_agent/protocols/out_of_band/v1_0/manager.py
PatStLouis: fixed

ff137:
Note:
```
            mock_oob_mgr.return_value.fetch_oob_invitation_record_by_id = (
                mock.CoroutineMock(return_value={"oob_id": "dummy"})
            )
...
>       return web.json_response(record.serialize())
E       AttributeError: 'dict' object has no attribute 'serialize'
```

The mock should just return an OobRecord type instead

acapy_agent/protocols/out_of_band/v1_0/routes.py
```diff
@@ -365,6 +404,11 @@ async def register(app: web.Application):
         [
             web.post("/out-of-band/create-invitation", invitation_create),
             web.post("/out-of-band/receive-invitation", invitation_receive),
+            web.get(
+                "/out-of-band/invitations/{oob_id}",
+                invitation_fetch,
+                allow_head=False,
+            ),
             web.delete("/out-of-band/invitations/{invi_msg_id}", invitation_remove),
```
loneil, line 412: Does this cause some REST confusions? (not that I'm too much a stickler for that but wonder what others think)
If I call
GET `/out-of-band/invitations/abc-123`
that returns a different thing than would be deleted by the same route on
DELETE `/out-of-band/invitations/abc-123`
oob_id and invi_msg_id being different identifiers (right?)

Probably pretty minor, maybe the existing DELETE one being on `/out-of-band/invitations/{connection record invi_msg_id to look up` is what's more of a mismatch

ff137: @loneil I agree that it's a bit confusing as it stands if the resource identifiers are different. And consistency is worth it

Initially this new endpoint was going to fetch by `invi_msg_id`, and now it's using `oob_id` instead. So it might make sense to change this new endpoint to `GET /out-of-band/{oob_id}`. (Maybe also some renaming of methods to indicate it's not using invitation id anymore - e.g. `fetch_oob_invitation_record_by_id` -> `fetch_oob_record_by_id`)

The existing DELETE endpoint does seem like the mismatch (since there's no existing delete by oob_id functionality), but it makes sense to be on /invitations/ because it's deleting by invitation msg id, and that can refer to multiple records.

And a possible to-do for another time, is to add endpoints to delete by `oob_id`, and fetch records by `invi_msg_id`. So that things are more symmetrical. But that's just my 2c worth - there's maybe reason to delete by invi_msg_id instead

PatStLouis: I agree with everything above. To shed some lights, the `oob_id` is the id by which the invitation record is stored. The `invi_msg_id` is only a tag. The DELETE route does more that just delete that invitation, it also deletes all connections associated by querying the `invi_msg_id`. I think `oob_id` is the proper id to use from a RESTful approach on this endpoint. It's also aligning with didcomm v2 url shortening:
https://identity.foundation/didcomm-messaging/spec/v2.0/#short-url-message-retrieval

@ff137 its still fetching the invitation record. The `invi_msg_id` was querying this record using a tag instead of directly accessing it by its `id`. IMHO `invi_msg_id` should be used as a query parameter, not a path parameter from a RESTful standpoint, since its not the invitation record id, but is instead a tag associated with it which you can query.

When fetching the resource (the invitation record), `oob_id` is the more deterministic way to do it.

ff137 (approved):

swcurran:
@PatStLouis -- an you please update this PR so we can merge it?  Thanks!

PR #3573, opened 2025-03-11
:art: Rename Anoncreds to AnonCreds

ff137:
The names "Anoncreds" and "AnonCreds" are naturally used interchangably, but with code, consistency is a beautiful thing. It's best to choose a name and stick to it.

When searching the *.py files for "AnonCreds", there were about ~860 results. Searching for "Anoncreds", there were about ~580 results.

So the capital C is more common, as it is the official name, but it's still pretty mixed. About a 60-40 split.

If one didn't have auto-complete to help, then it's a bit of headache to know when a capital C should be used in a model name, and when it shouldn't. (It's mainly a bit of headache when working with the OpenAPI models.)

So, it's a simple change - just rename all instances (except for AnoncredsError, because that's how it is in the anoncreds package), and enjoy the consistent naming pattern.

>NB: Worthwhile to note that it will break plugin imports that expect "Anoncreds" in the name, so they will need a similar find-replace-all change.

Let me know if it's too much of a headache to rename it. I think it's worth it, but can maybe be held back if it breaks integrations


swcurran:
I’m a fan of this one...

jamshale:
I think this is good to do but I'd like to get a release out before having a massive change that effects a lot of files. Even if it's trivial. We're targeting a release by the end of the week and then we could rebase this and get it merged.

ff137:
> I think this is good to do but I'd like to get a release out before having a massive change that effects a lot of files. Even if it's trivial. We're targeting a release by the end of the week and then we could rebase this and get it merged.

Understood - that's all good. It will impact plugins, plus it's low priority. So will move to draft for now, til it's ready

ff137:
I've prepped a PR in the plugins repo, for when this is merged and in the next release:
https://github.com/openwallet-foundation/acapy-plugins/pull/1671

ff137:
Hmm, if you ask me, now is as good a time as any. It's a breaking change to rename objects, so it's suitable to include in this minor version bump. Or the next one. It is a low priority change, so can be held back, especially if there's enough noteworthy changes in the release already. At the same time, may as well do it now. Your call! :-)

The rename fix is quick. Should only impact python projects that import AnonCreds objects from acapy agent, like the plugins. And the PR for that is ready - not many changes needed there

esune:
I think my vote goes to doing it as part of `1.3.0`: we don't know how many patch releases we will go through before we go to `1.4.0` which might mean a LONG time before we can merge and a lot of changes that can potentially conflict with this.

swcurran:
Agreed.  I’ll take a scan through later today (after a winery tour near Florence…) and approve.  Have a good day!

swcurran (approved):
LGTM! Is there a way to add a check via lint or custom command to prevent the text from leaking back into the code base?  Worth the effort?

ff137:
Best is probably code review, looking out for any new AnonCreds additions, and enforcing the standard

PR #3575, opened 2025-03-11
A couple of tweaks to the demo docs

ianco:


swcurran:
@ianco — should we just switch entirely to localtunnel and not even talk about ngrok?

ianco:
> @ianco — should we just switch entirely to localtunnel and not even talk about ngrok?

Probably.  Ngrok is quite pervasive though, and will probably still work with a licensed version ...

samcrichard:
Hey, Sam from ngrok here. Would love to understand what you're running into and see if I can help out.

ianco:
> Hey, Sam from ngrok here. Would love to understand what you're running into and see if I can help out. @ianco

Hey @samcrichard thanks for reaching out.  We've used ngrok quite a lot in the past, to test agents running on servers that have to communicate with agents running on mobile devices.  Awhile back (maybe even a couple year ago?) there were some changes to ngrok that broke some of our tests.  I don't recall all the details, and I'm not sure how much investigation anyone did as we found other options (now we use mainly localtunnel).

Today when I was testing some of the old ngrok code I get the following error:

`acapy_agent.core.dispatcher ERROR Message parsing failed: Message does not contain '@type' parameter, sending problem report`

It's a messaging issue between 2 agents, but it doesn't occur when I run without ngrok (or if I use localtunnel) so it looks like ngrok is removing some of the message attributes?

Issue #3567, opened 2025-03-10
1.2.lts - cherry-pick fixes from main


thiagoromanos:
There was bugs fixed on `main` branch that should be fixed on LTS branches as well. 

in `1.2.lts`
- #3543 
- #3545

in `0.12.lts`
- #3543

esune:
It looks like this could be closed - final word to @jamshale who actually did the picking 🙂

PR #3577, opened 2025-03-12
fix: cherry-pick fixes from main to 1.2.lts

thiagoromanos:
Adding fixes from `main` branch to `1.2.lts`. See #3567


jamshale (approved):
👍 I also want to add this https://github.com/openwallet-foundation/acapy/pull/3570. But I'll do that as a separate PR.

thiagoromanos:
@jamshale , I'm a little late here, but one thing to notice is that it would be better to not use "squash", but "rebase", so the commits remain like the original commits in `main` branch.

jamshale:
Good call. I hadn't thought of that.

PR #3578, opened 2025-03-12
fix: cherry-pick fixes from main to 0.12.lts

thiagoromanos:
Adding fixes from main branch to 0.12.lts #3567


jamshale (approved):
Also want to add https://github.com/openwallet-foundation/acapy/pull/3570. Will do as a separate PR.

PR #3585, opened 2025-03-14
:bug: Fix: Register both askar and anoncreds plugins for multitenancy

### LINKED ISSUE #3584: :bug: Multitenant agents should register both askar and askar-anoncreds plugins
Currently, both sets of plugins are only registered when the multitenant admin API is enabled:

```py
        if context.settings.get("multitenant.admin_enabled"):
            plugin_registry.register_plugin("acapy_agent.multitenant.admin")
            register_askar_plugins()
            register_anoncreds_plugins()
```

This should be changed to check if multitenancy is enabled: `multitenant.admin_enabled` -> `multitenant.enabled`

This is to allow tenants to work with Indy or Anoncreds endpoints in a multitenancy scenario

ff137:
Resolves #3584

Includes some deduplication and added logging

### DIFF
--- a/acapy_agent/config/default_context.py
+++ b/acapy_agent/config/default_context.py
@@ -137,22 +137,29 @@ class DefaultContextBuilder(ContextBuilder):
         if not self.settings.get("transport.disabled"):
             plugin_registry.register_package("acapy_agent.protocols")
 
-        # Currently providing admin routes only
-        plugin_registry.register_plugin("acapy_agent.holder")
+        # Define core plugins
+        core_plugins = [
+            "acapy_agent.holder",
+            "acapy_agent.ledger",
+            "acapy_agent.connections",
+            "acapy_agent.messaging.jsonld",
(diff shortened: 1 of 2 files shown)


jamshale:
I'm a bit confused by this. I'm sure the change is fine but how is this different than if they are registered for `multitenant.admin_enabled` ?

ff137:
> I'm a bit confused by this. I'm sure the change is fine but how is this different than if they are registered for `multitenant.admin_enabled` ?

`multitenant.admin_enabled` (--multitenant-admin) specifies to enable the multitenant admin API. 
`multitenant.enabled` (--multitenant) is whether multitenancy mode is enabled or not.

Picked it up because we run multitenancy without the admin API enabled

jamshale:
Sounds good. I hadn't even realized you could do that. I've always had the admin api enabled.

jamshale (approved):

PR #3587, opened 2025-03-14
:sparkles: Refactor TailsServer injection pattern

### LINKED ISSUE #3586: :bug: Creating Indy revocation registries fails when multitenant agent is askar-anoncreds
We are currently migrating from Indy to AnonCreds. Our existing test suite in [acapy-cloud](https://github.com/didx-xyz/acapy-cloud) runs everything using Indy, and we want to run the exact same test suite with AnonCreds.

In the following PR we try to change our multitenant agent from `askar` to `askar-anoncreds` wallet type:
https://github.com/didx-xyz/acapy-cloud/pull/1385

We would expect everything to function as before.

First we encountered #3584, where endpoints were "Not Found", since they weren't being registered. 
Then, testing the fix (#3585), we notice that credential definitions that support revocation do not have their revocation registries being created successfully. Everything else seems to work.

So, there you have it. I'll be investigating a bit more to see why the registries are not being created.

Edit: I've got a fix: #3587 -- tested on the above-linked acapy-cloud PR, and everything passed 🎉

ff137:
Resolves #3586

Removes the `context.injector.bind_provider` pattern for BaseTailsServer in the `DefaultContextBuilder.bind_providers` method. This is to support both IndyTailsServer and AnonCredsTailServer in a multitenancy scenario

It's a simple fix because in each context where the tails server is used, you know if it should be Indy or AnonCreds, so the dependency injection pattern is unnecessary.

### DIFF
--- a/acapy_agent/anoncreds/revocation.py
+++ b/acapy_agent/anoncreds/revocation.py
@@ -30,9 +30,7 @@ from ..askar.profile_anon import AskarAnoncredsProfile, AskarAnoncredsProfileSes
 from ..core.error import BaseError
 from ..core.event_bus import Event, EventBus
 from ..core.profile import Profile, ProfileSession
-from ..multitenant.base import BaseMultitenantManager
 from ..tails.anoncreds_tails_server import AnonCredsTailsServer
-from ..tails.base import BaseTailsServer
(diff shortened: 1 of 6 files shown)


jamshale (approved):
Makes sense to me 👍

PR #3588, opened 2025-03-14
:memo: Update README

ff137:
I'd like to propose the following updates to the readme page:

- drops the 🚨 emojis from "ACA-Py is part of the OWF" (which is more appropriate for "busy moving")
- moves badges to the top (and no longer wrapped in html)
  - Drops (Scanned by SonarQube badge) - because it's oversized, and it's already implied by the SonarQube coverage and security badges
  - adds Python version badge
  - moves OpenSSF scorecard to end
- updates summary from:
  - An easy to use enterprise wallet for building decentralized trust services using any language that supports sending/receiving HTTP requests.
  - to: ACA-Py is an easy to use enterprise SSI agent for building decentralized trust services using any language that supports sending/receiving HTTP requests.
  - main change: "enterprise wallet" to "enterprise SSI agent".
  - ❔ I can change that to "enterprise SSI agent and wallet"

Feedback and recommended amendments are welcome

___
Before:
image

After:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,14 @@
 # ACA-Py -- A Cloud Agent - Python  <!-- omit in toc -->
 
-🚨 **ACA-Py is part of the [OpenWallet Foundation] (OWF)!** 🚨
+![Python](https://img.shields.io/badge/python-3.12-blue.svg)
+[![PyPI version](https://img.shields.io/pypi/v/acapy-agent)](https://pypi.org/project/acapy-agent/)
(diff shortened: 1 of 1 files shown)


jamshale (commented):
Looks good to me. Will let Stephen give an opinion. 

I did these badges in html. For some reason I remember not being able to format them properly. But this looks better.

swcurran (approved):
LGTM

PR #3592, opened 2025-03-16
Indicate when interop tests fail

jamshale:
This will cause the interop test run to indicate fail status. It outputs the test logs to a temporary file and then greps for `Failing scenarios:` which only happens when tests fails.

It also runs the anoncreds tests on the nightly runs. Currently there is one failing test, so this will fail currently. Running both the indy and anoncreds tests on PR's takes too long. (45mins)

I would like to parse the failed tests and output them to PR's as a comment but this is an improvement for now.


jamshale:
WIP... Testing.

ianco (approved):

esune (approved):

PR #3604, opened 2025-03-19
1.3.0rc0

swcurran:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,14 +26,13 @@ Release 1.3.0 is a significant release that adds many updates, fixes and an impo
 
 In this release, the DiDComm [RFC 0160 Connections] is removed, in favour of the newer, more complete [RFC 0434 Out of Band] and [RFC 0023 DID Exchange]. Those still requiring [RFC 0160 Connections] protocol support must update their startup parameters to include the [Connections Protocol Plugin]. See the documentation for details, but once the ACA-Py instance startup options are extended to include the Connections protocol plugin, Controllers using the Connections protocol should continue to work as they had been. That said, we highly recommend implementers move to the [RFC 0434 Out of Band] and [RFC 0023 DID Exchange] Protocols as soon as possible.
 
-In the ACA-Py Controller API and models:
+### 1.3.0 ACA-Py Controller API Changes:
 
-- **BREAKING**: The `properties` field has been renamed to `additionalProperties` in many request/response models.
 - Added: `did:indy` support, including a new `POST /did/indy/create` endpoint
 - Routes that support pagination (such as endpoints for fetching connections or credential/presentation exchange records), now include `descending` as an optional query parameter.
-- `validFrom` and `validUntil` have been added to some models
+- `validFrom` and `validUntil` added to the `Credential` and `VerifiableCredential` objects
 
-Specifics of the majority of the can be found by looking at the diffs for the `swagger.json` and `openapi.json` files that are part of the [1.3.0.rc Release Pull Request](https://github.com/openwallet-foundation/acapy/pull/3604). Later pull requests might have additional changes.
+Specifics of the majority of the can be found by looking at the diffs for the `swagger.json` and `openapi.json` files that are part of the [1.3.0.rc Release Pull Request](https://github.com/openwallet-foundation/acapy/pull/3604). Later pull requests might introduce some additional changes.
 
 ### 1.3.0 Categorized List of Pull Requests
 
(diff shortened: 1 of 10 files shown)


jamshale (dismissed):
LGTM

dbluhm (approved):

ff137 (approved):

swcurran:
I updated the Changelog to add those to the Breaking Changes section (although some are additions).

Two thoughts come to mind:

- Should we be updating the swagger.json and openapi.json files as changes are made to main vs. only as part of the release ?
- I’ve added a link to this PR so people can look at the openapi.json and swagger.json diffs to see the details of what has changed.  However, I think I should be making that as a compare between the last the release and the current release., right?  I’ll try to find the right GitHub URL for that...

ff137:
> Should we be updating the swagger.json and openapi.json files as changes are made to main vs. only as part of the release ?

Either way is probably fine. One idea I've had for acapy-cloud, which will work here too, is to have a GHA for detecting diffs in the openapi spec. So, it's automatically generated by a workflow, and if a diff is detected, it'll fail, or comment a note on the PR. 

What we haven't decided on is if the openapi spec should be forced/encouraged to be updated manually, or if a GHA should automatically update it.

I think manual is good, so that a PR can get notified that this impacts the openapi spec, and it can require or recommend manually committing the updated spec, so the developer shows it's an intentional change to the spec.

Just mentioning the idea - something we can maybe contribute after adding it on our side.

> I’ve added a link to this PR so people can look at the openapi.json and swagger.json diffs to see the details of what has changed. 

The following would be a way to show diff between 1.2.4 and 1.3.0rc0: 
`https://github.com/openwallet-foundation/acapy/compare/1.2.4...1.3.0rc0`

So it'll only be viewable after 1.3.0rc0 is tagged. Unfortunately it shows all changes, and idk if there's a way to show single-file changes. Should be fine - they'd just have to manually scroll to the openapi spec.

CHANGELOG.md
```diff
@@ -26,6 +26,15 @@ Release 1.3.0 is a significant release that adds many updates, fixes and an impo
 
 In this release, the DiDComm [RFC 0160 Connections] is removed, in favour of the newer, more complete [RFC 0434 Out of Band] and [RFC 0023 DID Exchange]. Those still requiring [RFC 0160 Connections] protocol support must update their startup parameters to include the [Connections Protocol Plugin]. See the documentation for details, but once the ACA-Py instance startup options are extended to include the Connections protocol plugin, Controllers using the Connections protocol should continue to work as they had been. That said, we highly recommend implementers move to the [RFC 0434 Out of Band] and [RFC 0023 DID Exchange] Protocols as soon as possible.
 
+In the ACA-Py Controller API and models:
+
+- **BREAKING**: The `properties` field has been renamed to `additionalProperties` in many request/response models.
```
ff137, line 31: Apologies - I was mistaken with this. It seems to be an openapi generator change. It's not part of the actual models - just the `properties` field name in the spec has changed

CHANGELOG.md
```diff
@@ -26,6 +26,15 @@ Release 1.3.0 is a significant release that adds many updates, fixes and an impo
...
+- Added: `did:indy` support, including a new `POST /did/indy/create` endpoint
+- Routes that support pagination (such as endpoints for fetching connections or credential/presentation exchange records), now include `descending` as an optional query parameter.
+- `validFrom` and `validUntil` have been added to some models
```
ff137, line 34: `validFrom` and `validUntil` added to Credential and VerifiableCredential objects

ff137 (changes requested):
I just double checked the spec changes (because we've been tracking it in PRs for our controller). Looks like I was mistaken about the `properties` rename - sorry about the hassle

swcurran:
I like the idea of a warning that the PR changes the API. Not sure we should require it be updated manually by all contributors. Mind you, “drive-by” contributors are less likely to alter the API.

A question that comes to mind if we do that.  What do we set the API version between releases. Right now, we update the API version to match the tag that will be created.  But I don’t think we want to use that version after the tag has been created.

I did some searching for a way to do a single file diff across two tags and there is kind of a way, but it’s really not reliable (using the filehash).  The best way is to just include a handy reference to the `git diff` command that a dev can run locally, I think.

jamshale (commented):
Looks good. Just a couple more small comments from @ff137 for the changelog that don't look updated?

jamshale (approved):

jamshale:
We don't have to wait for the integration tests to pass for only the changelog change.

PR #3608, opened 2025-03-24
:bug: Fix unchanged endpoint being rewritten to ledger

### LINKED ISSUE #3607: :bug: Agent re-publishes endpoint to ledger upon restart, despite no change
As the title describes -- when an agent is restarted, and `ledger_config` gets called in the startup `provision` logic, there is a faulty check for updating the endpoint for a did (indy_vdr module):
```py
    async def update_endpoint_for_did(
        self,
        did: str,
        endpoint: str,
        endpoint_type: Optional[EndpointType] = None,
        write_ledger: bool = True,
        endorser_did: Optional[str] = None,
        routing_keys: Optional[List[str]] = None,
    ) -> bool:
...
        all_exist_endpoints = await self.get_all_endpoints_for_did(did)
        exist_endpoint_of_type = (
            all_exist_endpoints.get(endpoint_type.indy, None)
            if all_exist_endpoints
            else None
        )
        existing_routing_keys = (
            all_exist_endpoints.get("routingKeys") if all_exist_endpoints else None
        )

        if exist_endpoint_of_type != endpoint or existing_routing_keys != routing_keys:
...
                # Build and send attrib request
```

existing_routing_keys can be `[]`, and routing_keys can be `None`, leading to `[] != None` -> `True`, and a new attrib request being published to the ledger, despite no actual change in endpoint or routing keys.

ff137:
Resolves #3607

___

I think this might've popped up when the method args were refactored to no longer contain `routing_keys = []`. Because that raises a pylint/ruff warning to initialise an empty list in a method's args.

So it was possible replaced with default None, and that broke the list comparison. Which is anyway faulty, cuz lists can be the same but with different ordering... So I added set comparison to be safe

### DIFF
--- a/acapy_agent/ledger/indy_vdr.py
+++ b/acapy_agent/ledger/indy_vdr.py
@@ -731,6 +731,7 @@ class IndyVdrLedger(BaseLedger):
             endorser_did: DID of the endorser to use for the transaction
             routing_keys: List of routing keys
         """
+        routing_keys = routing_keys or []  # Ensure list type if None was passed
         public_info = await self.get_wallet_public_did()
         if not public_info:
             raise BadLedgerRequestError(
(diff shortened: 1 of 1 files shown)


jamshale:
The interop tests have been pretty flaky lately. It's more a problem with the test project with a github actions machine as they pass locally.

jamshale (approved):

ianco (approved):

PR #3611, opened 2025-03-24
feat(demo): remove broken aip 10 and fix aip 20

davidchaiken:
The mobile phone demo (AliceGetsAPhone) documentation says to run with AIP 10, which is deprecated and no longer works. Starting the faber agent results in [this error](https://gist.github.com/davidchaiken/cf047fa6be02ba00f339bca267b7bc7f). Instead of trying to repair the AIP 10 option, it seemed best to try to roll forward to AIP 20.

I found that using AIP 20 works with the BC Wallet mobile agent, except that the connectionless proof request interaction (CLI option 2a) with the faber agent was broken.

This pull request:
* Removes the deprecated and broken AIP 10.
* Fixes the connectionless proof request interaction for AIP 20.

I have tested these fixes manually in the Play With Docker (PWD) environment with the BC Wallet.

I also got it working running the faber agent in a docker container on my development machine, but that involved some ngrok configuration hacks and a small change to the demo code. I'd be happy to submit the code and ngrok configuration documentation in a separate PR.

### DIFF
--- a/demo/runners/support/agent.py
+++ b/demo/runners/support/agent.py
@@ -1004,17 +1004,21 @@ class DemoAgent:
         decorator = {
             "recipientKeys": [agent_public_did["result"]["verkey"]],
             # "routingKeys": [agent_public_did["result"]["verkey"]],
-            "serviceEndpoint": agent_endpoint["endpoint"],
+            "serviceEndpoint": agent_endpoint["endpoint"] or (
+                "http://"
+                + os.getenv("DOCKERHOST").replace(
+                    "{PORT}", str(self.admin_port + 1)
+                )),
         }
         return decorator
 
     async def _send_connectionless_proof_req(self, request: ClientRequest):
         pres_req_id = request.match_info["pres_req_id"]
-        url = "/present-proof/records/" + pres_req_id
+        url = "/present-proof-2.0/records/" + pres_req_id
         proof_exch = await self.admin_GET(url)
         if not proof_exch:
             return web.Response(status=404)
-        proof_reg_txn = proof_exch["presentation_request_dict"]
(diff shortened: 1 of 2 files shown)


swcurran (approved):

swcurran:
Oooppss…I shouldn’t have approved that.  The GHA’s needed approval and that’s what I meant to do.  The changes look good, but I didn’t run through them…

swcurran (dismissed):
Please update the DCO so the changesd can be accepted.  See DCO - Developer Certificate of Origin - https://github.com/apps/dco. To fix, see: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md, and the “details” beside the failed checks.

davidchaiken:
Thanks for having a look, @swcurran. I fixed the DCO.

As a follow-up to this pull request, I could:
1. Add the code and documentation required to run the mobile demo locally in docker.
2. Do some clean-up by removing some or all of the AIP 10 code from the demo.

Please let me know if either are of interest -- or if there is a better place to be discussing changes. For example, I could open a couple of issues to discuss.

ianco (approved):
@davidchaiken  thanks for the contribution it looks good!

I think the documentation that you suggest would be valuable.  IMHO we should leave the AIP 1.0 stuff in the demo for now, but @swcurran has a better pulse on the community and how much other agents still rely on the older protocols ...

jamshale:
I think it would be good to remove the 1.0 stuff. It would greatly simplify the demo code which has become quite bloated. If we should keep it though, we should be able to install and configure the plugin for it to keep it working.

davidchaiken:
Conversation above continued in Issue #3617.

PR #3615, opened 2025-03-26
fix: align ledger config schema with API response

### LINKED ISSUE #3614: 🐛  OpenAPI Spec Mismatch for `/ledger/config` Endpoint
#### **Problem Description**  
The `/ledger/config` endpoint currently returns a response with the structure:  
```json
{
  "production_ledgers": [...],
  "non_production_ledgers": [...]
}
```  
However, the **OpenAPI schema** (auto-generated documentation and interfaces) incorrectly defines the response as:  
```json
{
  "ledger_config_list": [...]
}
```  
and omits critical fields like `is_write`, `keepalive`, and `pool_name`, etc from `LedgerConfigInstance`.  

This mismatch causes:  
1. **Confusion for Developers**: Documentation/Swagger UI displays an inaccurate schema, leading to incorrect assumptions about the API response.  
2. **Broken Client Code Generation**: Tools like `openapi-generator` produce invalid interfaces, forcing downstream workarounds (e.g., manual data transformation).  
3. **Runtime Errors**: Clients expecting the documented schema will fail to parse actual responses.  

---

#### **Why This Matters**  
- **Trust in Documentation**: Accurate OpenAPI specs are critical for developers integrating with acapy.  
- **Automation Compatibility**: Mismatched schemas break automated codegen workflows, increasing maintenance overhead.  

---

#### **Proposed Solution**  
Update the OpenAPI schemas to match the actual API response:  

1. **Fix `LedgerConfigListSchema`**:  
   - Replace `ledger_config_list` with `production_ledgers` and `non_production_ledgers`.  
   - Ensure both fields are required and properly nested.  

2. **Extend `LedgerConfigInstanceSchema`**:  
   - Add missing fields (`is_write`, `keepalive`, `pool_name`, etc.) with correct metadata.  
   - Remove deprecated/unused fields (e.g., `genesis_transactions`).  

3. **Type Corrections**:  
   - Ensure `is_production` is a `boolean` (previously incorrectly typed as `str`).  

**Code Changes**:  
```python  
# Updated LedgerConfigListSchema  
class LedgerConfigListSchema(OpenAPISchema):  
    production_ledgers = fields.List(...)  # Replaces ledger_config_list  
    non_production_ledgers = fields.List(...)  

# Updated LedgerConfigInstanceSchema with missing fields  
class LedgerConfigInstanceSchema(BaseModelSchema):  
    is_write = fields.Bool(required=True)  
    keepalive = fields.Int(required=True)  
    # ... other fields  
```  

---

#### **Testing**  
**No New Tests Required** (existing tests cover `/ledger/config` response validation):  
- The fix is a schema alignment and does not alter API behavior.  
- Existing tests should validate the endpoint’s structure implicitly.  

**Verification Steps**:  
1. Deploy acapy agent with admin enabled and confirm the `/ledger/config` schema matches the actual response.  
2. Run existing integration tests to ensure no regressions.  

---

#### **Impact**  
- **Developers**: Documentation/Swagger UI will reflect the true API behavior.  
- **Codegen Tools**: Auto-generated clients (TypeScript, Python, etc.) will work out-of-the-box.  
- **Maintainability**: Reduces technical debt caused by client-side workarounds.

MonolithicMonk:
- Replace ledger_config_list with production/non_production_ledgers
- Add missing fields to LedgerConfigInstanceSchema
- Update OpenAPI documentation

Resolves #3614


jamshale:
You're the first guinea pig for Copilot 🙂

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -20,18 +20,26 @@ class Meta:
...
-        genesis_transactions: Optional[str] = None,
-        genesis_file: Optional[str] = None,
-        genesis_url: Optional[str] = None,
+        id: str = None,
+        is_production: bool = True,  # Fixed type (was str)
+        is_write: bool = None,
+        keepalive: int = 5,
+        read_only: bool = False,
+        socks_proxy: Optional[str] = None,
+        pool_name: str = None,
```
ff137, line 29: `id`, `is_write`, and `pool_name` are all default null here, so should be typed Optional

However, in the `LedgerConfigInstanceSchema`, they are all marked required, so should probably not be optional / set to default None -- so, e.g. `is_write` default should be one of False or True (whichever is most appropriate)

Or, `required` should be changed in the Schema

And, nitpick, the `  # Fixed type (was str)` comment can be dropped

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -62,8 +74,15 @@ def validate_id(self, data, **kwargs):
...
-        fields.Nested(LedgerConfigInstanceSchema(), required=True), required=True
+    production_ledgers = fields.List(  # Changed from ledger_config_list
+        fields.Nested(LedgerConfigInstanceSchema(), required=True),
+        required=True,
+        metadata={"description": "Production ledgers"},
+    )
+    non_production_ledgers = fields.List(  # Added new field
+        fields.Nested(LedgerConfigInstanceSchema(), required=True),
+        required=True,
+        metadata={"description": "Non-production ledgers"},
```
ff137, line 85: Marking all of these required - the nested fields too - may lead to some overly strict validation. And for a response model that could lead to confusing 422 errors.

i.e. you have to pass a list instead of None, and you can't pass an empty list either - both of these must contain a LedgerConfigInstanceSchema -- as far as I understand it. 

(Comments can be removed too)

acapy_agent/ledger/tests/test_routes.py
```diff
@@ -1,4 +1,6 @@
+import uuid
```
ff137, line 1: Nitpick: We have uuid_utils, which is slightly faster.
This is just a test module, so doesn't really matter - but for sake of consistency we may as well us `uuid_utils` instead. (Should work fine if you say `import uuid_utils as uuid`)

MonolithicMonk:
Funny thing is that I just built the changed acapy code and included it in my project.  And of course I go to use the interface and noticed it HAHAHAHAH

demo/runners/acme.py
```diff
@@ -14,7 +14,6 @@
     check_requires,
     log_msg,
     log_status,
-    log_timer,
```
ff137, line 17: I don't think this should be removed (I presume it was auto-removed by ruff?). I recall there was a PR to re-add it: https://github.com/openwallet-foundation/acapy/pull/3562

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
MonolithicMonk: I marked those as required because they always have default values.  Meaning, if no values are supplied, they will will be populated with values.  `id` is always auto generated.  `is_write` defaults to false if logic for true isn't met.  `pool_name` is also defaulted value.  I may have a different interpretation of those values.

MonolithicMonk:
I will address all of these tomorrow morning

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
ff137: It's just to help with intentionality / readability. 

I can see that `id` will be pre-filled with a uuid if it's not set.
But how is default for `is_write` and `pool_name` being set? I don't see that in the changes.
___
Ah, I see now the initial setting logic:
```py
        is_write_ledger = (
            False if config.get("is_write") is None else config.get("is_write")
        )
        ledger_id = config.get("id") or str(uuid4())
        if is_write_ledger:
            write_ledger_set = True
        config_item = {
            "id": ledger_id,
            "is_production": (
                True
                if config.get("is_production") is None
                else config.get("is_production")
            ),
            "is_write": is_write_ledger,
            "genesis_transactions": txns,
            "keepalive": int(config.get("keepalive", 5)),
            "read_only": bool(config.get("read_only", False)),
            "socks_proxy": config.get("socks_proxy"),
            "pool_name": config.get("pool_name", ledger_id),
        }
        if "endorser_alias" in config:
            config_item["endorser_alias"] = config.get("endorser_alias")
        if "endorser_did" in config:
            config_item["endorser_did"] = config.get("endorser_did")
```

What I wanted to avoid was making the `get_ledger_config` route only respond with 422 errors if someone didn't specify `is_write` or `pool_name` in their initial config.

I can see now that `is_write` will default to False, and `pool_name` will default to the ledger id (which defaults to a uuid if it wasn't set).

What that's telling me is that this `validate_id` step is not needed:
```py
    @pre_load
    def validate_id(self, data, **kwargs):
        """Check if id is present, if not then set to UUID4."""
        if "id" not in data:
            data["id"] = str(uuid4())
        return data
```
since the id will always be set during the initial config setup.

Ideally we'd move this model closer to the initial setup logic ... i.e. that the config and response model share the same Marshmallow definition, so it's more maintainable / less chance of diff being introduced in the future.

For this PR that's probably optional... Up to you if you want to refactor the initial `load_multiple_genesis_transactions_from_config` method as well

But at the least I would set id, is_write, and pool_name to not be default None, since they cannot be None.

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -43,13 +51,17 @@ class Meta:
...
+    keepalive = fields.Int(required=True, metadata={"description": "keepalive"})
+    read_only = fields.Bool(required=True, metadata={"description": "read_only"})
+    socks_proxy = fields.Str(required=False, metadata={"description": "socks_proxy"})
+    pool_name = fields.Str(required=True, metadata={"description": "pool_name"})
+    endorser_alias = fields.Str(
+        required=False, metadata={"description": "endorser_alias"}
     )
-    genesis_file = fields.Str(required=False, metadata={"description": "genesis_file"})
-    genesis_url = fields.Str(required=False, metadata={"description": "genesis_url"})
+    endorser_did = fields.Str(required=False, metadata={"description": "endorser_did"})
```
ff137, line 64: These descriptions are just equal to the field names. Which is not the purpose of a description :-)

Maybe we can just auto-generate generic definitions for what each field represents. And then include the detail, such as `ledger_id` indicating that it will default to a UUID if not set, and pool name will equal the `ledger_id` if not set. That's the kind of helpful info we want in a model description.

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -43,13 +51,17 @@ class Meta:
...
 
     @pre_load
     def validate_id(self, data, **kwargs):
```
ff137, line 67: This method can either be removed, since it will never be run in practice. Or, maybe better is to just use this Marshmallow model in the initial setup of the ledger config (instead of a dictionary). So that this validate_id step is called during the initial creation of the config. And then one can add the same logic for pool_name, etc

ff137 (changes requested):
The nitpicks regarding the `LedgerConfigInstance` init default args are not that important.

But I think the `required=True` condition for all the fields in `LedgerConfigListSchema` might be too strict, because it means the `get_ledger_config` will respond with a 422 if `non_production_ledgers` are not set.

Also, I think it's worth using the `LedgerConfigInstance` model in the initial config. So that there's more coupling between the initial ledger setup and the response model for fetching the config. And also just to streamline the defaults and model validation being defined in 1 place. Makes it more maintainable

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -43,13 +51,34 @@ class Meta:
         model_class = LedgerConfigInstance
         unknown = EXCLUDE
 
-    id = fields.Str(required=False, metadata={"description": "ledger_id"})
-    is_production = fields.Bool(required=False, metadata={"description": "is_production"})
-    genesis_transactions = fields.Str(
-        required=False, metadata={"description": "genesis_transactions"}
+    id = fields.Str(
+        required=True, metadata={"description": "Auto-generated UUID of none is provided"}
```
ff137, line 55: `"Auto-generated UUID of none is provided"`
->
`"String identifier for ledger. A UUID is generated if not provided."`

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -43,13 +51,34 @@ class Meta:
...
+    )
+    is_production = fields.Bool(
+        required=True, metadata={"description": "Production-grade ledger (true/false)"}
+    )
+    is_write = fields.Bool(
+        required=False,
+        metadata={"description": "Write capability enabled (default: False)"},
+    )
+    keepalive = fields.Int(
+        required=False, metadata={"description": "Connection ping interval in seconds"}
```
ff137, line 65: `"Connection ping interval in seconds"`

This is the duration in seconds that the connection to the ledger is kept open, since the last operation, before the connection is closed. 

Internally a default of 5 is used, so could specify that here.

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
```diff
@@ -43,13 +51,34 @@ class Meta:
...
+    )
+    read_only = fields.Bool(
+        required=False, metadata={"description": "Read-only access (default: False)"}
+    )
+    socks_proxy = fields.Str(
+        required=False, metadata={"description": "SOCKS proxy URL (optional)"}
+    )
+    pool_name = fields.Str(
+        required=False, metadata={"description": "Defaults to ledger ID if not provided"}
```
ff137, line 74: "The name for the ledger pool. Defaults to ledger ID if not provided"

ff137 (commented):
just some notes for the descriptions

MonolithicMonk:
I'm think I'm good with my version of the description.   This is getting a little bit ridiculous

MonolithicMonk:
When you have a change please merge the changes.  I'm sure a lot of downstream can use it.

acapy_agent/ledger/tests/test_routes.py
```diff
@@ -869,6 +877,110 @@ async def test_get_ledger_config_x(self):
...
+                "read_only": True,
+                "pool_name": "non_prod_pool",
+                "socks_proxy": None,
+            },
+        ]
+
+        with mock.patch.object(
+            test_module.web, "json_response", mock.Mock()
+        ) as json_response:
+            await test_module.get_ledger_config(self.request)
```
ff137, line 915: Can you write another test please for the scenario where only one element is set in the config list - either just a production ledger, or not.

I suspect that the response will fail with a 422 due to the `required=True` restrictions being set on both list types. But I may be mistaken. I think it's worth covering for that case in a test, just to make sure different config will succeed

ff137 (commented):
Looking neat. Using the schema in the initial config loading can be separate work. For this, I just want to request an extra test to ensure the schema response won't raise 422s

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
ff137: I still think this can now be removed

acapy_agent/ledger/multiple_ledger/ledger_config_schema.py
ff137: @MonolithicMonk

MonolithicMonk:
@ianco, @jamshale, @esune , can you please review this so we can move along.  Thanks

MonolithicMonk:
To avoid over-engineering toward an unattainable standard of perfection, let’s adhere to agile principles by leveraging the tools and standards already in place:

SonarQube for objective code quality metrics.

The attached issue to finalize measurable criteria aligned with the project’s acceptance criteria.

This ensures contributors work toward clear, agreed-upon goals while iterating efficiently.

ff137:
Yeah, no, the broader refactoring idea is completely optional. That's going above and beyond.

The request for an extra test case was because I think the implementation, as it stands, is faulty. I may be wrong, but my comment here hasn't been addressed yet (https://github.com/openwallet-foundation/acapy/pull/3615#discussion_r2015075624), and the tests don't currently dispel my suspicion that it's faulty.

So ... I'll pull a patch for the PR to write that test and see if it's really faulty or not

ff137 (approved):

Issue #3617, opened 2025-03-26
Demo: remove AIP 10 code


davidchaiken:
The discussion in PR #3611 indicates agreement that it will be good to remove the AIP 10 code from the demo, but we didn't converge on whether now is a good time to do this work.

@jamshale wrote:
> I think it would be good to remove the 1.0 stuff. It would greatly simplify the demo code which has become quite bloated. If we should keep it though, we should be able to install and configure the plugin for it to keep it working.

@ianco wrote:
> IMHO we should leave the AIP 1.0 stuff in the demo for now, but @swcurran has a better pulse on the community and how much other agents still rely on the older protocols ...

This issue tracks the work to be done -- either now or at some point in the figure.

@swcurran - We would be interested in your thoughts. If you want the work done now (or soon), I'd be happy to put in some time removing the AIP 10 code.

swcurran:
That would be awesome!  The more we can remove technical debt the better — and removing deprecated items from the demo definitely falls into that category.

davidchaiken:
Completed with PR #3619 with this caveat from @jamshale:
> I feel like there's probably other places that still reference v1 connections in other modules and documentation but we can continue to remove stuff as we find it.

PR #3619, opened 2025-03-27
chore(demo): remove aip 10 code

davidchaiken:
This PR completes the work documented in Issue #3617. It removes the AIP 10 code from conditional statements and a few function signatures, but leaves the scaffolding in place so that new versions can be added in the future. (If there's no chance that there will be a new AIP version in the future, please let me know so that I can do an even deeper clean.)

The default view of the diffs looks more complicated than it actually is, because git diff sometimes scrambles the code between the half of the conditionals that are removed (aip == 10) and the code that remains (aip == 20). To see a cleaner version of the diffs, you can view them [ignoring whitespace](https://github.com/openwallet-foundation/acapy/compare/main...davidchaiken:acapy:dc_issue_3617?w=1).

I tested these changes manually using local docker on my MacBook and in the PWD environment. I ran both the alice/faber and the "Alice Gets a Phone" version of the demo. I also verified that `run_bdd` works, but noted that it doesn't actually use the faber.py runner.

### DIFF
--- a/demo/runners/agent_container.py
+++ b/demo/runners/agent_container.py
@@ -1376,7 +1376,7 @@ def arg_parser(ident: str = None, port: int = 8020):
         type=str,
         default=20,
         metavar=("<api>"),
-        help="API level (10 or 20 (default))",
+        help="API level (20)",
     )
     parser.add_argument("--timing", action="store_true", help="Enable timing information")
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


davidchaiken:
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=openwallet-foundation_acapy&pullRequest=3619)

SonarQube is complaining about a couple lines of `raise Exception` statements, which are consistent with the preexisting code. I'm okay with fixing this complaint, but the change requires deciding what convention for Exceptions to use and then applying it uniformly. IMO trying to fix the problem just for the one line of code that I added would add more tech debt than it would remove.

jamshale:
We can ignore sonarcloud in situations like this. I think this is fine for the demo. I actually thought it was configured to ignore the demo code.

jamshale (approved):
LGTM 👍

I feel like there's probably other places that still reference v1 connections in other modules and documentation but we can continue to remove stuff as we find it.

PR #3622, opened 2025-03-31
(fix) VM resolution strategy correction

gmulhearn:
The current VM resolution strategy will get potential VMs by getting the set VMs of that `proof_purpose` (e.g. "assertionMethod", "authentication") from the `pydid` formatted `DIDDocument`. It does this with a `getattr` (e.g. `getattr(doc, "assertionMethod", [])`.

The problem is that the class attributes of the `DIDDocument` class are snake case (`assertion_method`), so attempts to use the proof purpose `assertionMethod` result in an empty result when you try to get the `assertionMethod` attribute.

I've just changed this to `.get` the attribute from the original raw doc dict instead, which has the attributes/keys in the camelCase format.

The new test is a regression test - it fails before the fix


gmulhearn:
Discovered this while testing W3C VCDM issuance (over DIDComm) with a cheqd DID.

swcurran:
Thanks!  Can you please take a look at the Ruff/lint failure? 

I’ve asked @ff137 and @esune to review this.

ff137 (approved):
Makes sense 👍 

___
I think that the camelCase vs snake_case headaches should be avoided by aliasing the Pydantic fields, but that requires changes in pydid 
@dbluhm 

e.g.
```py
class DIDDocumentRoot(Resource):
    assertion_method: Optional[List[Union[DIDUrl, VerificationMethod]]] = None
```
could be:
```py
class DIDDocumentRoot(Resource):
    assertion_method: Optional[List[Union[DIDUrl, VerificationMethod]]] = Field(None, alias="assertionMethod")
```
___
Looks like Pydantic has a way to generate aliases automatically: https://docs.pydantic.dev/2.1/usage/model_config/#alias-generator

I'll post as an issue on pydid

esune (approved):
👍🏻

PR #3626, opened 2025-04-03
:bug: Fix publishing all pending AnonCreds revocations

### LINKED ISSUE #3625: :bug: AnonCreds: publishing all pending revocations does not revoke anything
In acapy-cloud, we have a test for asserting that the revocation record updates to `state: revoked`, after a pending revocation is published. This test did not work when it was migrated to AnonCreds.

Test scenario:
- Faber (anoncreds issuer) issues 3 revocable credentials to Alice
- Alice accepts - she has 3 credentials
- Faber revokes all 3 credentials (without auto-publishing, so they are pending)
- Faber then calls `/anoncreds/revocation/publish-revocations` with `rrid2crid = []` (i.e. publish all pending)
- After this, for each credential exchange record, we assert the revocation record now has `state: revoked` (instead of `issued`)

This was not working. The record state remained `issued`.

So, I added some logging to the `publish_pending_revocations` method:
```py
    async def revoke_pending_credentials(
        self,
        revoc_reg_id: str,
        *,
        additional_crids: Optional[Sequence[int]] = None,
        limit_crids: Optional[Sequence[int]] = None,
    ) -> RevokeResult:
        """Revoke a set of credentials in a revocation registry.
...
        """
        LOGGER.info(
            "Starting revocation process for registry %s with "
            "additional_crids=%s, limit_crids=%s",
            revoc_reg_id,
            additional_crids,
            limit_crids,
        )
...
            if limit_crids is None:
                skipped_crids = set()
            else:
                skipped_crids = rev_crids - set(limit_crids)
            rev_crids = rev_crids - skipped_crids

            LOGGER.info(
                "Revoking %d credentials, skipping %d credentials for registry %s",
                len(rev_crids),
                len(skipped_crids),
                revoc_reg_id,
            )
```

and then I could see:
```
2025-04-03 14:36:20,180 acapy_agent.anoncreds.revocation INFO Starting revocation process for registry ... with additional_crids=None, limit_crids=[]
2025-04-03 14:36:20,247 acapy_agent.anoncreds.revocation INFO Processing 3 credential revocation IDs for registry ...
2025-04-03 14:36:20,247 acapy_agent.anoncreds.revocation INFO Revoking 0 credentials, skipping 3 credentials for registry ...
2025-04-03 14:36:20,278 acapy_agent.anoncreds.revocation INFO Successfully updated revocation list for registry ...
2025-04-03 14:36:20,338 acapy_agent.anoncreds.revocation INFO Completed revocation process for registry ...: 0 revoked, 0 failed
```

Some things to note: 
- the method is called with `limit_crids=[]`, instead of None (which would typically mean publish all pending)
- "Revoking 0 credentials, skipping 3 credentials"
- the `if limit_crids is None:` check is false, so skipped_crids = all the rev_crids that should be revoked ...

This appears to stem from the parent method which has:
```py
            if rrid2crid:
                if rrid not in rrid2crid:
                    continue
                limit_crids = [int(crid) for crid in rrid2crid[rrid]]
            else:
                limit_crids = None
```

It's possible for limit_crids to be an empty list, instead of None, breaking the later `if limit_crids is None` check.

So, simple solution: the if check should add `or limit_crids == []`

ff137:
Resolves #3625

In short, there was another `[] is None` comparison, which leads to unintended behaviour (previously it would only publish all if None was passed, but it should do so when an empty list is passed too)

Includes improved info/debug logging in the publish pending revocation method


swcurran (approved):
Nice. Logging looks so much better.

PR #3640, opened 2025-04-10
Add BLS12381G2 keys to multikey manager

gmulhearn:

PR #3641, opened 2025-04-10
(fix) W3C LDP Fixes for alternative VMs

### LINKED ISSUE #3639: 🐛  Cannot issue W3C LDPs with other keys/VMs owned by the issuer DID
Encountering 2 very related problems to more complicated DIDs & W3C LDP issuance

## Problem 1
If you have an issuer DID in acapy, then attempting to issue a W3C VCDM (via the vc-api or issue-credential-2.0 APIs) with a proofType other than the _DID's default key_type_ will fail - even if the specified verificationMethod has the correct proof type.

For instance:
`POST issue-credential-2.0/send-offer` with options:
```json
"options": {
    "proofType": "BbsBlsSignature2020",
    "verificationMethod": "did:cheqd:testnet:c052cf29-bd2f-4598-8c62-248bf707d274#bbs-1"
}
```
The acapy VcLdpManager fails to create the offer, as Unable to issue credential with issuer id did:cheqd:testnet:a84138fd-cca0-4182-86dc-c24fb114fe52 and proof type BbsBlsSignature2020. DID only supports proof types {'Ed25519Signature2020', 'Ed25519Signature2018'}.

This issue comes from `VcLdpManager`'s `assert_can_issue_with_id_and_proof_type` check before any cred issue or offer.
```py
did = await self._did_info_for_did(issuer_id)
# Raise error if we cannot issue a credential with this proof type
# using this DID from
did_proof_types = KEY_TYPE_SIGNATURE_TYPE_MAPPING[did.key_type]
if proof_type not in did_proof_types:
    raise VcLdpManagerError(
        f"Unable to issue credential with issuer id {issuer_id} and proof "
        f"type {proof_type}. DID only supports proof types {did_proof_types}"
    )
```

## Problem 2
If you can dodge problem 1 (in my case, i just commented out the assert checks), then you will still run into issues further down the code

If you have an issuer DID in acapy, then attempting to issue a W3C VCDM from this DID (via the vc-api or via issue-credential-2.0 APIs) will always result in a VC signature produced by the _DID's default verkey*_, regardless of whatever verificationMethod you request it be signed with.

This is problematic for DIDs which have multiple VMs, and you want to use the other VMs. For instance, i ran into this when issuing W3C VCDMs from my acapy cheqd DID which has multiple VMs (including some for BBS): sample DID - [did:cheqd:testnet:c052cf29-bd2f-4598-8c62-248bf707d274](https://resolver.cheqd.net/1.0/identifiers/did:cheqd:testnet:c052cf29-bd2f-4598-8c62-248bf707d274).

The source of this issue comes from `VcLdpManager`, which constructs a signature suite with the with the DIDInfo's `verkey`:
```py
did_info = await self._did_info_for_did(issuer_id)
suite = await self._get_suite(
  proof_type=proof_type,
  verification_method=verification_method,
  proof=proof.serialize(),
  did_info=did_info,
)
# ...
# where self._get_suite(..) does:
# Get signature class based on proof type
SignatureClass = PROOF_TYPE_SIGNATURE_SUITE_MAPPING[proof_type]
# Generically create signature class
return SignatureClass(
    verification_method=verification_method,
    proof=proof,
    key_pair=WalletKeyPair(
        profile=self.profile,
        key_type=SIGNATURE_SUITE_KEY_TYPE_MAPPING[SignatureClass],
        public_key_base58=did_info.verkey if did_info else None,
    ),
)
```

### Solution?

IMO a better solution would be if the verification method is resolved/dereferenced and the WalletKeyPair is extracted. e.g.
```py
resolver = self.profile.inject(DIDResolver)
full_vm = await resolver.dereference_verification_method(self.profile, verification_method)
# Get signature class based on proof type
SignatureClass = PROOF_TYPE_SIGNATURE_SUITE_MAPPING[proof_type]
# Generically create signature class
return SignatureClass(
    verification_method=verification_method,
    proof=proof,
    key_pair=WalletKeyPair(
        profile=self.profile,
        key_type=SIGNATURE_SUITE_KEY_TYPE_MAPPING[SignatureClass],
        public_key_base58=full_vm.public_key_base58,
    ),
)
```
(this fixes the issue in my case, but note ugliness with assuming the VerificationMethod has public_key_base58).

## Bigger Picture Problem
This problem did reveal a bigger picture problem (which i'm sure has been mentioned before, perhaps someone could cross-reference if there's any existing docs), which is that `DidInfo` is structured a bit awkwardly for what a DID really is;
```
[
  ("did", str),
  ("verkey", str),
  ("metadata", dict),
  ("method", DIDMethod),
  ("key_type", KeyType),
]
```
the main problem here is that it's assuming a DID has 1 associated key (verkey + key_type), and that's it. But we know it can have multiple. 

On another note, even if that problem was fixed, i still like the idea of being able to sign documents with a DID that isn't stored in the acapy wallet (e.g. as a DIDInfo), or even just isn't in-sync with the acapy wallet. As long as acapy has the associated keys, then it should be able to sign.

gmulhearn:
Fixes #3639 

## Problem 1 fix:
~soften the assertions made by `assert_can_issue_with_id_and_proof_type`, which is used in 2 places: 1. before creating an aries/ldproof offer, 2. before issuing a credential. I've softened it by removing the check that the DID has a key that is capable of producing the given proof type.~

~TODO - is there a better way? ideally what we might want this method to check: "does the DID have a VM with a key type that can satisfy the proof type?". But even then, that is skipping the case where a VM was explicitly specified.~

I've removed `assert_can_issue_with_id_and_proof_type`, and replaced it with a rehaul to the verification method strategy. Instead of `assert_can_issue_with_id_and_proof_type`, now the offer & issue flows with try to resolve a suitable VM, which in the process will check all the conditions that `assert_can_issue_with_id_and_proof_type` intended to check:
* appropriate relationship (e.g. assertionMethod),
* appropriate keytype for proof type,
* & VM key exists in the acapy wallet.

However in the case that an `options.verificationMethod` was specified, we _skip_ the relationship/proofpurpose check - we assume the consumer _knows_ this verificationMethod is the appropriate proofpurpose... this is imperfect, but the motivation is to avoid having to resolve the DID Doc every time, just to check the VM relationship.

## Problem 2 fix:
~modify `_get_suite`, such that it resolves the verification method to get the verkey, rather than using the one within the `DIDInfo`. My only issue with this fix, is that it may be seen as inefficient to resolve the doc here, and it's wasted effort if the VM was already resolved recently - e.g. by the verkey strategy if VM ID not specified~

create a new method, `idempotent_resolve_and_store_from_kid` for `MultikeyManager` which resolves & caches (if not already cached) the VM<->multikey mapping. Modify `_get_suite` such that it uses this method to get the _correct_ key to use in the signature (rather than the `verkey` in the `DIDInfo`).

BLOCKED - branched from #3640 , merge that first


PatStLouis:
@gmulhearn have you looked how the [DataIntegrityManager](https://github.com/anonyome/aries-cloudagent-python/blob/gm/bbs-multikey/acapy_agent/vc/data_integrity/manager.py) handles this? I've reflected on all these issues and would like to have a chat to determine how to tackle this moving forward.

I partially addressed it to meet my requirements.

For problem 1, I think the first check is if a `verificationMethod` id is provided, acapy looks it up locally using the multikey manager. From then, it will be able to derive the key type from the multikey and assert if its compatible with the proofType/cryptosuite. If it's being checked from the did value itself,  it needs to resolve the did document and find the first verification method that matches. I'm the opinion that the second option is less desirable and would discourage it, unless using a declarative did method like did:key: and I would no recommend it when using any did method that can have more than one verification method listed in the did document.

For problem 2, the keypair should have been [appropriately binded](https://github.com/anonyome/aries-cloudagent-python/blob/531fcf1107dc2367d0f4955411c6a5b59f0167f8/acapy_agent/wallet/keys/manager.py#L166) to the `verificationMethod` prior to issuance. This should happen 1) automatically if its a managed did method (cheqd plugin would ensure the keypair exists/bind it when updating the did document) 2) manually using the admin api if the controller is managing the did document itself.

I would rather to start working with multikeys than verkey as they carry more information.

gmulhearn:
hey @PatStLouis , thanks for having a look at this initial PR.

Having a look at the data_integrity/manager -> eddsa_cs_2022 code now, i see how you're using the MultiKeyManager to pull the locally stored keyinfo (i.e. the multikey) mapped to the input VM ID.

So am i understanding correctly; this works on the assumption that the key <-> VM binding/mapping has been pre-established - either automatically by the did method, or manually (controller ensures VMs are synced/accurate via admin APIs), as you said. OR, in the eddsa_jcs_2022 case, it works if the VM is a did:key, as you provide a shortcut:
```py
  if options.verification_method.startswith("did:key:"):
      multikey = options.verification_method.split("#")[-1]
      key_info = await self.key_manager.from_multikey(multikey)
```

passing comment, but a few minor things i'm picking up from this;
* there doesn't seem to be validation of the verificationMethod's suitability for the suite, particularly: ensuring a suitable key type (ed25519) & ensuring a suitable purpose (e.g. assertionMethod).
* there's no *guarantee* that the key that the multikey manager fetches for the VM ID is the actual correct key (according to the source of truth (the resolved DID Document)), so there's potential for issuing unverifiable creds. but maybe this is a non-problem, i suppose it's own the controller to ensure the multikeys they store are correct representations.

anyway, this is an interesting approach which i don't have experience with, my thinking was more aligned with credo-ts's approach of "give me a VM ID, i'll resolve it and try find the matching keypair in my wallet for signing". Pros: uses the real source of truth (the DID Doc), Cons: expensive to resolve the DID Doc frequently.

Usage of the multikey manager avoids that resolution expense, but introduces that VM synchronisation problem, which would occur if A) the acapy DID manager/plugin doesn't update your keys, B) your DID is also being controlled outside of acapy. But atleast you offer a good amount of admin APIs where a good controller can ensure synchronisation themselves.

but i'll definitely have a go at playing with this strategy. although i'm a little worried about backwards compatibility, as these areas of acapy code are typically used with `did:sov` & `did:key` issuers - which i can't assume there are registered multikey manager entries for.

FWIW, the cheqd plugin doesn't utilize the multikey manager. The implementation of cheqd DID Doc in acapy is fairly trivial; you give it a complete DID Document structure and it'll sign+submit it to the ledger (no checks for the VMs inside that DID Doc or anything like that). So in my case, i've been creating BBS keys with the multikey manager, then updating my cheqd DID Document with a new assertionMethod VM that has that BBS key, then calling acapy to sign credentials with that VM (`did:cheqd:abc#bbs-1`). It's probably out of my scope to try update the did:cheqd plugin with logic for checking the VMs and synchronising with multikey manager.

gmulhearn:
@PatStLouis that most recent comment i made includes a strategy for problem 2:
"if the multikey manager has a local key entry for the VM ID, use that multikey for signing, otherwise resolve the signer multikey from ledger".

Both branches for the strategy are working in my acapy cheqd testing environment (one where i'm taking the effort to ensure that the new VMs i update my DID Doc with are also registered to the multikeymanager, and one where i don't)

PatStLouis:
In the case where its not created locally and you are resolving from the ledger, couldn't you then tag it? sort of a `bind_if_unbound` function, so the next time it needs to use that VM, it will be bound? This would avoid constantly resolving did documents

gmulhearn:
@PatStLouis happy to do this - i still have minor concerns about the local <-> remote VM/DIDDoc state syncing, but i understand it may be too pedantic to worry about or sacrifice performance for.

acapy_agent/wallet/keys/manager.py
```diff
@@ -66,6 +75,20 @@ def __init__(self, session: ProfileSession):
         self.session: ProfileSession = session
         self.wallet: BaseWallet = session.inject(BaseWallet)
 
+    async def idempotent_resolve_and_store_from_kid(self, kid: str):
+        """Fetch key if exists, otherwise resolve and store it"""
```
PatStLouis, line 79:
```suggestion
    async def resolve_and_bind_kid(self, kid: str):
        """Fetch key if exists, otherwise resolve and store it. This function is idempotent."""
```

PatStLouis: Since its not "storing" a key, I think the term "bind" is preferable. It's binding a key id to a key which already exists.

acapy_agent/wallet/keys/manager.py
```diff
@@ -66,6 +75,20 @@ def __init__(self, session: ProfileSession):
...
+        if await self.kid_exists(kid):
+            LOGGER.debug(
+                f"Key for kid {kid} already exists in storage, will not resolve from ledger."
```
PatStLouis, line 82:
```suggestion
                f"kid {kid} already bound in storage, will not resolve from ledger."
```

acapy_agent/wallet/keys/manager.py
```diff
@@ -66,6 +75,20 @@ def __init__(self, session: ProfileSession):
...
+            )
+            return await self.from_kid(kid)
+        else:
+            multikey = await self.resolve_multikey_from_verification_method(kid)
+            LOGGER.debug(
+                f"Key for kid {kid} did not exists in storage, storing resolved multikey {multikey}."
```
PatStLouis, line 88:
```suggestion
                f"kid {kid} binding not found in storage, binding to resolved multikey {multikey}."
```

PatStLouis (changes requested):

PatStLouis:
I like the idea here. I'd like to hear your concern about the state going out of sync and which scenario you could see this happening?

PatStLouis:
In the resolve/bind function you added, I would first do a check if the key itself exists. If it doesn't, it can't be binded and it should raise an error.

The multikey manager is interfacing with the askar wallet (or whichever wallet type is deployed). If the key doesn't exist in the wallet, it won't be able to bind it/sign a document. This would happen if you add a third party verification method in a did document.

acapy_agent/wallet/keys/manager.py
PatStLouis: could maybe still resolve to ensure its the right multikey value?

acapy_agent/vc/vc_ld/manager.py
```diff
@@ -170,16 +171,7 @@ async def assert_can_issue_with_id_and_proof_type(
...
-
-            # Raise error if we cannot issue a credential with this proof type
-            # using this DID from
-            did_proof_types = KEY_TYPE_SIGNATURE_TYPE_MAPPING[did.key_type]
-            if proof_type not in did_proof_types:
-                raise VcLdpManagerError(
-                    f"Unable to issue credential with issuer id {issuer_id} and proof "
-                    f"type {proof_type}. DID only supports proof types {did_proof_types}"
-                )
+            _ = await self._did_info_for_did(issuer_id)
```
gmulhearn, line 174: TODO - better solution for option 1?

gmulhearn:
@PatStLouis 
> In the resolve/bind function you added, I would first do a check if the key itself exists. If it doesn't, it can't be binded and it should raise an error.
> 
> The multikey manager is interfacing with the askar wallet (or whichever wallet type is deployed). If the key doesn't exist in the wallet, it won't be able to bind it/sign a document. This would happen if you add a third party verification method in a did document.

The MultikeyManager is calling `self.wallet.assign_kid_to_key`, which fails if the key doesn't exist in the wallet, but still, doesn't cost much to do a more explicit check in the multikeymanager.

acapy_agent/wallet/keys/manager.py
gmulhearn: still resolving would defeat the purpose of using the multikeymanager right? as then we are _always_ resolving the VM - not taking advantage of the multikeymanager's storage of kid<->multikey

gmulhearn:
Hmmm, i'm probably leaning towards the compromise for solving problem 1:
* if you specify a VM (options.VM), then fetch the key locally with `resolve_and_bind_kid`,
  * perform basic "suitable VM" checks: is it the correct key type for the required proofType?
  * we are _skipping_ the check that the VM is the correct relationship/purpose (e.g. assertionMethod)
* if you don't specify a VM, then we have no choice but to resolve the DID Doc and find a "suitable" VM
  * perform the full checks for this suitable VM - checking correct relationship/purpose

This could be replaced once DID Management is improved in acapy

gmulhearn:
Update: i've gone with the compromise approach:
> if a VM ID is specified, then assume the consumer knows the relationship is correct and don't bother checking it - letting us skip DID resolution in that case.

acapy_agent/wallet/keys/manager.py
```diff
@@ -59,6 +63,30 @@ def key_type_from_multikey(multikey: str) -> KeyType:
...
+        )
+
+    elif verification_method.type == "Ed25519VerificationKey2020":
+        multikey = verification_method.public_key_multibase
+
+    elif verification_method.type == "Bls12381G2Key2020":
+        multikey = verkey_to_multikey(
+            verification_method.public_key_base58, alg="bls12381g2"
+        )
+
```
PatStLouis, line 83:
```suggestion
# TODO address JsonWebKey based verification methods
```

PatStLouis (dismissed):

PatStLouis:
> this matches the DataIntegrityManager/eddsa_jcs_2022 - which, as far as i can tell, is not doing any checks that the specified options.VM is actually the correct relationship/proofPurpose.

Correct, since this is declarative by the controller, it's the controller's responsibility to ensure the correct relationship is set in the did document, otherwise verifiers might reject the proof. 

Consider cases where 1000+ credentials need to be issued in a single batch, resolving 1000 times might not be the best approach.

This check should ideally be done internally. If the did document is cached it might be more realistic. I would say to leave the responsibility to the controller in this case.

PatStLouis (approved):

PatStLouis:
Looks great, thanks @gmulhearn !

swcurran:
@gmulhearn — can you please update the branch to enable merging?

PR #3649, opened 2025-04-15
Fix: Use correct profile context for ledger signing during revocation setup (#3624)

### LINKED ISSUE #3624: Error during creation of credential definition with revocation enabled
#### **Problem Description**  

I get the error an error during the creation of a credential definition with revocation enabled using anoncreds.  The credential definition is actually created however, the error occurs during the revocation operation.  I still haven't been able to determine what the cause of the error is.  Here is a partial log:

```python
File "/home/aries/.venv/lib/python3.12/site-packages/acapy_agent/wallet/askar.py", line 398, in get_local_did
    raise WalletNotFoundError("Unknown DID: {}".format(did))
acapy_agent.wallet.error.WalletNotFoundError: Unknown DID: REDACTED
```
I can create the credential definition in a multitenant - endorsement setup. However, after the successful creation and endorsement of the Revocation Registry Definition (RevRegDef), the automatic process to create the corresponding initial Revocation Registry Entry (the list state) fails with the WalletNotFoundError shown above.

To clarify the sequence based on my logs:

A tenant agent (using askar-anoncreds wallet type, configured with an endorser) successfully creates a CredDef via endorsement.
The same tenant agent successfully creates the associated RevRegDef via endorsement. The anoncreds::revocation-registry-definition::finished event is emitted.
The event handler revocation_setup.py::on_rev_reg_def triggers automatically upon receiving this event.
This handler successfully uploads the tails file and then calls revocation.py::create_and_register_revocation_list.
This leads down a call stack eventually calling indy_vdr.py::send_revoc_reg_entry, which needs to look up the tenant's issuer DID using askar.py::get_local_did to prepare the transaction for endorsement.
Failure: get_local_did raises WalletNotFoundError for the tenant's own DID at this point in the execution flow.
It seems that while the initial request context (handled via API and endorsement flow) correctly resolves the tenant's wallet and DID, the execution context available within the on_rev_reg_def event handler (running asynchronously after the RevRegDef write is acknowledged) is somehow unable to access the same DID within the tenant's wallet.

Possibly related to #3586

MonolithicMonk:
Solution:

- Added profile parameter and conditional context usage to relevant methods in acapy_agent/ledger/indy_vdr.py.
- Updated calls in acapy_agent/anoncreds/default/legacy_indy/registry.py to pass the profile parameter.
- Added new tests in acapy_agent/ledger/tests/test_indy_vdr.py and acapy_agent/anoncreds/default/legacy_indy/tests/test_registry.py specifically verifying the correct propagation and usage of the passed profile parameter.

Testing:

- Existing ledger and registry tests remain unchanged and pass, verifying backward compatibility / default behavior.
- New tests added to specifically validate that the explicit profile parameter is correctly passed through the registry and ledger layers and used for signing/TAA context within IndyVdrLedger._submit and related methods.

Considerations:

- This solution directly fixes the identified failure point by ensuring the correct context is available for critical operations.
- The underlying behavior of MultiIndyVDRLedgerManager returning potentially admin-bound ledger instances remains unchanged. A future refactor of ledger instance management/scoping might provide a more fundamental solution, but is outside the scope of this immediate fix.

Resolves #3624


.gitignore
```diff
@@ -193,4 +193,17 @@ _build/
...
+diode-start
+SHOWCASE_1
+SHOWCASE_1.bck
+SHOWCASE_2
+SHOWCASE_2.bck
+SHOWCASE_3
+SHOWCASE_3.bck
+testprocess.sh
+ledgers.yml
+acapy_logs*
```
ff137, line 209: Are all of these files generated by the project?

.vscode-sample/alice.yml
```diff
@@ -1,32 +0,0 @@
-# Alice agent configuration
```
ff137, line 1: Why is the .vscode-sample directory being deleted?

poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
ff137, line 2: Adding this to the repo is not strictly necessary. I think this can be configured locally, globally using `~/.config/pypoetry/config.toml` or something like that

According to a poetry developer: 
> The `poetry.toml` contains the local configuration of the poetry application. One should not check-in this file into a repository, because these configuration are just personal flavors.

acapy_agent/ledger/indy_vdr.py
```diff
@@ -1226,25 +1254,59 @@ async def send_revoc_reg_entry(
         issuer_did: Optional[str] = None,
         write_ledger: bool = True,
         endorser_did: Optional[str] = None,
+        profile: Optional[Profile] = None,
     ) -> dict:
         """Publish a revocation registry entry to the ledger."""
-        async with self.profile.session() as session:
+        LOGGER.debug(">>> Entered send_revoc_reg_entry")
+        LOGGER.debug(
+            f"Initial parameters - issuer_did: {issuer_did}, write_ledger: {write_ledger}, endorser_did: {endorser_did}"  # noqa: E501
```
ff137, line 1262: f-string formatting should generally not be used in logging, especially at debug level.

Many of these seem like logs for local testing and should be polished

acapy_agent/ledger/indy_vdr.py
```diff
@@ -1253,53 +1315,92 @@ async def send_revoc_reg_entry(
...
+            LOGGER.debug(">>> Entering default Indy-VDR path")
             try:
+                LOGGER.debug("Building standard revoc_reg_entry request")
                 request = ledger.build_revoc_reg_entry_request(
                     did_info.did,
                     revoc_reg_id,
                     revoc_def_type,
                     json.dumps(revoc_reg_entry),
                 )
+                LOGGER.debug(f"Request type created: {type(request).__name__}")
```
ff137, line 1368: Between prev comment and this one, please review and polish logging to only what's meaningful

docker/Dockerfile.test
```diff
@@ -11,7 +11,8 @@ WORKDIR /usr/src/app
 
 RUN pip install --no-cache-dir poetry==2.1.1
 
-COPY ./README.md pyproject.toml ./poetry.lock ./
+COPY ./README.md pyproject.toml ./
+# COPY ./README.md pyproject.toml ./poetry.lock ./
```
ff137, line 15: Probably don't want to commit this?

ff137 (changes requested):
PS: sign commits with DCO, can be done with `git rebase -i HEAD~7` and reword commits

acapy_agent/ledger/indy_vdr.py
```diff
@@ -1253,17 +1278,22 @@ async def send_revoc_reg_entry(
...
+                resp = await legacy_indy_registry.txn_submit(
+                    self,
+                    revoc_reg_entry_req,
+                    sign=True,
+                    sign_did=did_info,
+                    write_ledger=write_ledger,
+                    profile=current_profile,
+                )
+            except Exception:
+                raise
```
ff137, line 1296: try-except can be removed if exception is rethrown, or just an error log can be added in the except case

acapy_agent/ledger/indy_vdr.py
```diff
@@ -1226,24 +1241,34 @@ async def send_revoc_reg_entry(
...
+        async with current_profile.session() as session:
             wallet = session.inject(BaseWallet)
-            if issuer_did:
-                did_info = await wallet.get_local_did(issuer_did)
-            else:
+            try:
                 did_info = await wallet.get_public_did()
-            del wallet
+            except Exception:
+                raise
```
ff137, line 1254: As far as I can tell, get_public_did does exception handling, and will return None if something goes wrong. It's possible that the try-except block isn't necessary. Would make sense if the finally block did important cleanup, but `del wallet` isn't necessary in case of an exception. So imo may as well just `did_info = await wallet.get_public_did()`

acapy_agent/ledger/indy_vdr.py
```diff
@@ -1226,24 +1241,34 @@ async def send_revoc_reg_entry(
...
-            )
+        wallet_type = current_profile.context.settings.get("wallet.type")
+        if wallet_type == "askar-anoncreds":
+            try:
+                from acapy_agent.anoncreds.default.legacy_indy.registry import (
+                    LegacyIndyRegistry,
+                )
+
+            except ImportError:
+                raise
```
ff137, line 1271: This try-except is also redundant

acapy_agent/ledger/indy_vdr.py
```diff
@@ -312,6 +312,7 @@ async def _submit(
         taa_accept: Optional[bool] = None,
         sign_did: DIDInfo = sentinel,
         write_ledger: bool = True,
+        profile: Optional[Profile] = None,  # <-- ADD profile parameter
```
ff137, line 315: Remove comment

ff137 (commented):
Good work and test coverage 👍

ff137:
Looks like the rebase was out of sync. Try `git fetch upstream` and `git rebase upstream/main` - hopefully fixes the extra changes

jamshale:
I think this is ok for now. Thanks for figuring it out. Maybe we can just add some notes where the profile is added as a parameter that this was done because of an injection issue.

We should create an issue for looking into fixing the root of the injection problem.

dbluhm:
I'm not able to reproduce; @jamshale I recommend holding off on merging until we know what's going on.

acapy_agent/ledger/indy_vdr.py
MonolithicMonk: This pr will likely be rejected due to further developments in the root cause of the error.

MonolithicMonk:
Can this be merged now pending a preferred resolution?

jamshale:
So this is a temporary fix for the `multi-ledger` with a `genesis-transactions-list` configuration scenario? I'd like to see some more comments on the reason that the profile option is added to these functions so it doesn't get forgotten about forever.

I'm not opposed to allowing this for now when the root of the problem gets looked into. That's not an area I have looked at, so I'm not sure of the difficulty in finding a solution.

MonolithicMonk:
Yes it is specific to the multi-ledger configuration scenario.  Also are you referring to inline comments?  Or does this pr + issue suffice?

acapy_agent/ledger/indy_vdr.py
```diff
@@ -321,9 +322,10 @@ async def _submit(
             taa_accept: whether to apply TAA acceptance to the (signed, write) request
             sign_did: override the signing DID
             write_ledger: whether to write the request to the ledger
-
+            profile: The profile context used for the request
```
jamshale, line 325: I was thinking a comment here would be good. Some type of note saying why this argument was added and that it shouldn't be needed. The abstract class would be good to have as well but it doesn't look like it has argument descriptions.

jamshale:
I was thinking some in code comments. We definitely want to make it clear to avoid this pattern and the PR and issues are easier to miss than having it recorded in the code.

jamshale:
I see there's an effort going on to help find the root cause of that issue. Let's hold off on this for now.

jamshale:
Closing as stale...

PR #3650, opened 2025-04-15
:art: Add type hints to `messaging/jsonld`

ff137:
Minor improvement. Was looking at the jsonld code, and none of the methods had typed args. Some of the pyld functions return types like `(Any | list | dict | None)`, when we always expect dict. So just to help with readability/maintainability, added type hints to the methods in messaging/jsonld

### DIFF
--- a/acapy_agent/messaging/jsonld/create_verify_data.py
+++ b/acapy_agent/messaging/jsonld/create_verify_data.py
@@ -9,6 +9,7 @@ import hashlib
 
 from pyld import jsonld
 
+from ...vc.ld_proofs import DocumentLoader
 from .error import (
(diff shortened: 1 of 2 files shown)


swcurran (approved):
LGTM — again, not the expert, but looks useful and safe.

Question: Can the data type and result data type be required in lint?  Worthwhile?

Lined up behind the rest of the approved PRs merge...

ff137:
It can be added to the ruff rules 👍 named: ANN001, and ANN201
But, currently more than 1000 individual errors 👀 
So, will have to chip away at it piece by piece

PatStLouis (approved):

PR #3652, opened 2025-04-15
:art: Add type hints to anoncreds module

ff137:
Ruff can check for when type hints or return types are missing, e.g.:
```sh
ruff check acapy_agent/anoncreds/ --select ANN001,ANN201 --exclude '**/test*/**','*/routes.py'
```

ANN201 is for return types. ANN001 checks for method arg types. Ignoring test files, and routes.py (because they all just need -> web.Response | None)

In Python, it's really worth it to add type hints. For maintainability/readability - it improves clarity (e.g. it's better to know a method returns `list[dict]`, instead of just `list`), and intentionality - you can much easier see, oh, this method should actually be returning something.

But there's more than 1000 errors when scanning the codebase.
In the long run I think it would be good to add ANN001/ANN201 to the set of ruff rules that are enforced. But it'll take some work to get there. 
Chunking is better than doing it all in one go. So, I've started the effort by going through the acapy_agent/anoncreds module, and adding type hints + return types

### DIFF
--- a/acapy_agent/anoncreds/__init__.py
+++ b/acapy_agent/anoncreds/__init__.py
@@ -7,7 +7,7 @@ from .registry import AnonCredsRegistry
 LOGGER = logging.getLogger(__name__)
 
 
-async def setup(context: InjectionContext):
+async def setup(context: InjectionContext) -> None:
     """Set up default resolvers."""
     registry = context.inject_or(AnonCredsRegistry)
     if not registry:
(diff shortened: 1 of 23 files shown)


ff137:
"2 New Issues" = 2 new todos

swcurran (approved):
LGTM — useful!  Lined up behind the others to be merged...

PR #3654, opened 2025-04-15
👷 Split Docker Builds

rblaine95:
* Build Standard image on native CPU architectures
  * `linux/arm64` builds on `ubuntu-24.04-arm` (`arm64`)
  * `linux/amd64` builds on `ubuntu-24.04` (`x86_64`)
* Split build and publish
  * Build docker images on native runners, don't push, cache to GHA
  * Combine multi-arch, read from cache, push to GHCR.io
* Due to a version mismatch in buildx between `ubuntu-24.04` and 
`ubuntu-24.04-arm`, have to ensure the latest version of `buildx` is installed
* Due to how the platforms are being handled, I've removed the `platforms`
input to the workflow

---

With zero cache, purely building the images goes from ±5 minutes to ±1 minute.
Full pipeline run is ±3 minutes.
https://github.com/didx-xyz/acapy/actions/runs/14472669065/attempts/1

With cache, purely building is ±30 seconds.
Full pipeline run is ±1.5 minutes
https://github.com/didx-xyz/acapy/actions/runs/14472669065/attempts/2

If I enable caching of the Buildx binary it'll reduce the build time by ±30 seconds,
but zizmor doesn't like it due to risk of cache poisoning.
Might still be worth doing though?

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -56,4 +56,3 @@ jobs:
         tag: ["nightly-${{needs.setup_and_check_pub.outputs.date}}", nightly]
     with:
       tag: ${{ matrix.tag }}
-      platforms: "linux/amd64,linux/arm64"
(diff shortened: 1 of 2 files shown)


WadeBarnes (approved):
Nice!

PR #3656, opened 2025-04-15
:test_tube: Fix test warnings

ff137:
- ignores the expected `create_did_document` deprecation warning in relevant tests
- correctly tear down ClientSession in TestAdminServer
- use `pytest_asyncio.fixture` for async test fixtures (allows tests to successfully run with `--asyncio-mode=strict`)
___
There remains 1 test warning:
```
acapy_agent/messaging/jsonld/tests/test_routes.py::TestJSONLDRoutes::test_verify_credential
  /root/anaconda3/envs/acapy/lib/python3.12/site-packages/pytest_asyncio/plugin.py:788: DeprecationWarning: pytest-asyncio detected an unclosed event loop when tearing down the event_loop
  fixture: <_UnixSelectorEventLoop running=False closed=False debug=False>
  pytest-asyncio will close the event loop for you, but future versions of the
  library will no longer do so. In order to ensure compatibility with future
  versions, please make sure that:
      1. Any custom "event_loop" fixture properly closes the loop after yielding it
      2. The scopes of your custom "event_loop" fixtures do not overlap
      3. Your code does not modify the event loop in async fixtures or tests
  ```

I've tried - I can't figure out what's causing an unclosed event loop here. It doesn't occur when the test, or the module is run in isolation, but for a full test run it appears.

I believe it's related to `nest_asyncio` being used in the json-ld logic, but I'm not sure.

Nevertheless, 1 warning is better than 8. Progress

### DIFF
--- a/acapy_agent/admin/tests/test_admin_server.py
+++ b/acapy_agent/admin/tests/test_admin_server.py
@@ -5,6 +5,7 @@ from unittest import IsolatedAsyncioTestCase
 
 import jwt
 import pytest
+import pytest_asyncio
 from aiohttp import ClientSession, DummyCookieJar, TCPConnector, web
 from aiohttp.test_utils import unused_port
 from marshmallow import ValidationError
@@ -577,7 +578,7 @@ class TestAdminServer(IsolatedAsyncioTestCase):
             await test_module.upgrade_middleware(request, handler)
 
 
-@pytest.fixture
+@pytest_asyncio.fixture
 async def server():
     test_class = TestAdminServer()
     await test_class.asyncSetUp()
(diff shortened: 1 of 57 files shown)


swcurran (approved):
Not knowledgeable enough on this, but I can see this doesn’t do damage, so approving. Feel free to merge (after the next one in line…).

ff137:
It's a shame I can't solve the `TestJSONLDRoutes::test_verify_credential` event loop warning, but I'm pretty sure it's to do with `nest_asyncio`, which I've been trying to see if that can be replaced (since the project is archived, and prevents usage of uvloop, the "faster" asyncio event loop library. (Relevant issue https://github.com/openwallet-foundation/acapy/issues/2941)

PR #3661, opened 2025-04-16
:art: Fix swagger tag names for AnonCreds endpoints

### LINKED ISSUE #3658: Missing AnonCreds endpoints in swagger ui
Recent changes to acapy code has resulted in missing `AnonCreds` endpoints namely:

- AnonCreds - Credential Definitions
- AnonCreds - Schemas
- AnonCreds - Wallet Upgrade
- AnonCreds - Revocation (some are missing)

✅:

Image

❌:

Image

ff137:
Resolves #3658

Some duplicated tag names were not caught in the renaming PR: #3573

___
Note: there's a bit of a knot created by the fact that there's `revocation_anoncreds/routes.py`, and then some revocation endpoints being defined in `anoncreds/routes.py`. Not gonna try fix that here. Ideally should be cleaned up by the original contributors, or logged as a to-do

### DIFF
--- a/acapy_agent/anoncreds/routes.py
+++ b/acapy_agent/anoncreds/routes.py
@@ -49,6 +49,9 @@ from .util import handle_value_error
 
 LOGGER = logging.getLogger(__name__)
 
+CRED_DEF_TAG_TITLE = "AnonCreds - Credential Definitions"
+SCHEMAS_TAG_TITLE = "AnonCreds - Schemas"
+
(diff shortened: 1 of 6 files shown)


swcurran (approved):
Minor points about the text wording (e.g. should be upgrading from “askar” to “askar-anoncreds”), but I think the point is sufficient.  Not sure that wallet-type was the right flag to use for the change of a library, but ¯\_(ツ)_/¯.  Move on...

PR #3664, opened 2025-04-17
:art: Make ledger config more readable

ff137:
Was reviewing the argparse and ledger config code, and some of it is really unnecessarily unreadable!

e.g.
```py
            "is_production": (
                True
                if config.get("is_production") is None
                else config.get("is_production")
            ),
```
... that's just `"is_production": config.get("is_production", True)` !

And some of the code is too tightly spaced for how much is going on, so just spent a minute trying to make it more readable. No changes to functionality

Readability = maintainability

### DIFF
--- a/acapy_agent/revocation_anoncreds/routes.py
+++ b/acapy_agent/revocation_anoncreds/routes.py
@@ -879,10 +879,10 @@ async def update_rev_reg_revoked_state(request: web.BaseRequest):
             ledger_manager = context.injector.inject(BaseMultipleLedgerManager)
             write_ledger = context.injector.inject(BaseLedger)
             available_write_ledgers = await ledger_manager.get_write_ledgers()
-            LOGGER.debug(f"available write_ledgers = {available_write_ledgers}")
-            LOGGER.debug(f"write_ledger = {write_ledger}")
+            LOGGER.debug("available write_ledgers = %s", available_write_ledgers)
+            LOGGER.debug("write_ledger = %s", write_ledger)
             pool = write_ledger.pool
-            LOGGER.debug(f"write_ledger pool = {pool}")
+            LOGGER.debug("write_ledger pool = %s", pool)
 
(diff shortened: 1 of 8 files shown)


dbluhm (approved):
Definitely an improvement!

acapy_agent/revocation_anoncreds/routes.py
```diff
@@ -879,10 +881,10 @@ async def update_rev_reg_revoked_state(request: web.BaseRequest):
             ledger_manager = context.injector.inject(BaseMultipleLedgerManager)
             write_ledger = context.injector.inject(BaseLedger)
             available_write_ledgers = await ledger_manager.get_write_ledgers()
-            LOGGER.debug(f"available write_ledgers = {available_write_ledgers}")
-            LOGGER.debug(f"write_ledger = {write_ledger}")
+            LOGGER.debug("available write_ledgers = %s", available_write_ledgers)
```
jamshale, line 884: I'm just curious why using f-strings is less preferrable?

jamshale (approved):

acapy_agent/revocation_anoncreds/routes.py
ff137: In this case, it really doesn't matter. Mainly, I see it as adopting a pattern and sticking to it

f-formatting means the string is evaluated before being sent to the log function. So if agent is running at info level, it unnecessarily compiles the string first, before seeing debug won't be logged

Really only makes a difference if the function in the string is expensive, or if the log is called very often. On the other hand, f-string is sometimes just more readable, and that outweighs needing to change it.

Mainly, I can't help myself. I tell myself it's because code-patterns propagate, and want to encourage using the better, lazy log pattern for low log levels. But really it's kinda just OCD. Trying to squeeze out every optimisation possible, even the microsecond improvements 😂

For error or warning logs it doesn't really matter, because those are almost always logged. But if it's at debug level I kind just auto change it

PR #3669, opened 2025-04-23
:construction_worker: Update dependabot file

ff137:
Flattens duplication by specifying "directories", instead of multiple individual `directory` configs.

Adds `/scenarios` directory to pip and docker management, which was missing.


jamshale (approved):
Nice! Way better. The plugins repo is like this too.

PR #3678, opened 2025-04-23
:lock: Update Token Permissions in GitHub Actions

ff137:
Attempts to resolve the following Token Permission warnings from the OpenSSF Scorecard:
```
Warn: no topLevel permission defined: .github/workflows/bdd-integration-tests.yml:1
Warn: no topLevel permission defined: .github/workflows/bdd-interop-tests.yml:1
Warn: no topLevel permission defined: .github/workflows/format.yml:1
Warn: no topLevel permission defined: .github/workflows/nightly.yml:1
Warn: no topLevel permission defined: .github/workflows/pr-tests.yml:1
Warn: topLevel 'contents' permission set to 'write': .github/workflows/publish-docs.yml:11
Warn: no topLevel permission defined: .github/workflows/scenario-integration-tests.yml:1
Warn: no topLevel permission defined: .github/workflows/sonar-merge-main.yml:1
Warn: no topLevel permission defined: .github/workflows/sonar-pr.yml:1
```
___
Most of them get the following added:
```yml
permissions:
  contents: read
  pull-requests: read
  checks: write
```

and the publish-docs one just moves the write permission from top-level to job-level
___
This aligns with security best practices, to explicitly define the permissions that the workflow needs. It'll hopefully bump the OpenSSF score to ~8.5

### DIFF
--- a/.github/workflows/bdd-integration-tests.yml
+++ b/.github/workflows/bdd-integration-tests.yml
@@ -9,6 +9,11 @@ on:
       - main
     types: [opened, synchronize, reopened, ready_for_review]
 
+permissions:
+  contents: read
+  pull-requests: read
+  checks: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/bdd-interop-tests.yml
+++ b/.github/workflows/bdd-interop-tests.yml
@@ -9,6 +9,11 @@ on:
       - main
     types: [opened, synchronize, reopened, ready_for_review]
 
+permissions:
+  contents: read
+  pull-requests: read
+  checks: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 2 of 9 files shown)


swcurran (approved):

ff137:
@swcurran Yay! That helped move it from 7.9 to 8.4, with a 6/10 for token permissions now instead of 0

There's still some points being deducted for some reason.

It's possibly due to write permissions:
```
Warn: jobLevel 'packages' permission set to 'write': .github/workflows/nightly.yml:63
Warn: jobLevel 'contents' permission set to 'write': .github/workflows/publish-docs.yml:1
```
But those permissions are of course intended for publishing. [Their docs](https://github.com/ossf/scorecard/blob/cd152cb6742c5b8f2f3d2b5193b41d9c50905198/docs/checks.md#token-permissions) mention that "points are not reduced if the job utilizes a recognized packaging action or command." So that could be why; it doesn't recognise the publishing tools.

Also, I added `checks: write` permissions when it might not have been needed. It logs a warning for those, but as far as I can tell, it says it doesn't deduct points for that.

That's just to debrief. Any further refinements are probably optional, and I'll leave that for someone else :-)

Issue #3659, opened 2025-04-15
Webhook triggered before an item is saved in DB


gmulhearn:
I have http webhook handlers, which upon receiving the webhook for a issue-credential v1.0 exchange, would immediately re-lookup the item (i.e. via acapy admin API) to get the full item (+ look up additional items) to perform handling on.

However, i was finding that (somewhat) unique to issue-credential v1.0, there is a race condition, where after receiving the webhook, it's not _guaranteed_ that the acapy DB is updated yet, so a look up of the item means that it's non-existent or not yet updated.

Looking at the issue-credential v1 code, it seems that the usage of a transaction is likely the candidate:
https://github.com/openwallet-foundation/acapy/blob/main/acapy_agent/protocols/issue_credential/v1_0/manager.py#L384
```py
            await cred_ex_record.save(txn, reason="receive credential offer")
            await txn.commit()
```

`.save` invokes the webhook, then the webhook processing logic is racing against the next python lline `txn.commit`.

The other acapy protocols i have webhook handlers for are not using DB transactions as liberally as issue credential v1, so the sync problem hasn't occurred yet. However i can see that generally the use of DB transactions in protocol handlers across acapy is common.

So looking for some advice/thoughts;
* is this behaviour expected? - IMO it'd be nice if the webhooks (i.e. `post_save`) were only triggered _after_ the DB transaction commit.
* is there any documentation on the expected behaviour of webhooks? - e.g. i was interested to discover that the webhooks are done in a "broadcast and forget" manner - acapy doesn't seem to await the HTTP response from the webhook handler.
* is what i'm currently doing an _anti pattern_? (receive webhook and re-fetch data). I believe we originally went with this pattern because the webhooks may not have been returning adequate data at some point in time (e.g. i vaguely recall parts of the proof exchange not being returned in the webhook). I acknowledge as well that re-fetching data like this may be risky for other reasons, e.g. if by the time the webhook re-fetches it's already in the next state (due to auto-issue/auto-accept).

swcurran:
> Side-point (out of scope for this issue): I think it's wrong for an API to respond with a 200, when there is still background processing happening. The correct response code is a 202, to make it clear that the server accepted the work, but stuff is still happening. That would be a good change to some endpoints.

💯 on this.

ff137:
I'm noticing this may have been attempted to be resolved in:
- https://github.com/openwallet-foundation/acapy/pull/2760

[...]

PR #3690, opened 2025-04-29
:recycle: Refactor EventBus notify method

ff137:
:construction:
___
Relates to #3659


ff137:
Note: the PR tests in `acapy_agent/protocols/present_proof/dif/tests/*` are taking exceptionally long to complete.
(Not related to these changes.)

Full pytest run usually takes ~3 minutes, and it currently takes around 25 minutes, since about 2 hours ago. 
Not sure why, but presumably there's some network load. And it's probably good at some point to mock the network calls for these unit tests, so the network is not a blocker for PRs

ff137:
The problem with scenario tests is that now the revocation registries are not guaranteed to be ready by the time the tests tries to use them.

I'll see what can be done about this. I presume that creating a revocable cred def would previously wait for events associated with rev-reg-def's to be complete before returning. So that behaviour can maybe be preserved somehow.

ff137:
`EventBus.notify` can now be made synchronous, but it would require updates to plugins that call it (would be breaking change if async is removed, since it shouldn't be awaited). So, I just added a todo

ff137:
Finally, took a long time to fix the tests, but now it's actually ready for review!

acapy_agent/core/tests/test_event_bus.py
```diff
@@ -185,23 +204,229 @@ async def test_wait_for_event_multiple_do_not_collide(event_bus: EventBus, profi
...
+
+    async def uncancellable_processor(profile, event):
+        """A processor that ignores cancellation."""
+        try:
+            await asyncio.sleep(10)  # Long sleep
+        except asyncio.CancelledError:
+            # Ignore cancellation and keep running
+            await asyncio.sleep(10)
+
+    event_bus.subscribe(re.compile(".*"), uncancellable_processor)
```
jamshale: Is this unit test going to take 10 seconds?

jamshale (dismissed):
This is very nice 👍

I also think we should change it to not be asynchronous in a separate task later. Maybe as a targeted release. Just to make the plugin updates more isolated and if there is any potential problems with the work around this it will be easier to find.

We also want to get the Kanon profile stuff completed. I'm not sure how many merge conflicts this will cause with it, but if it's ready now, it might be better to merge that first.

acapy_agent/core/tests/test_event_bus.py
ff137: Hmm, I see this test is actually instant, because `asyncio.wait` is patched to raise a TimeoutError. Meaning the "uncancellable task" is never actually awaited.

This one is thanks to vibe coding... Just adds coverage to assert a warning is logged if a timeout were to occur. But it's a bit dubious if it doesn't actually ensure an uncancellable task raises a timeout error! And if I remove the patch, it doesn't actually do that ... which is confusing.
___

So, today I learned, about `asyncio.wait`:
```py
async def wait(fs, *, timeout=None, return_when=ALL_COMPLETED):
    """Wait for the Futures or Tasks given by fs to complete.
...

    Note: This does not raise TimeoutError! Futures that aren't done
    when the timeout occurs are returned in the second set.
    """
```

It doesn't raise a timeout error! I always thought that's what it does.

Ok, so, I'm removing the test, and the TimeoutError handling. Because it's not really meaningful to get a warning log that some tasks didn't complete. It's not actionable info in any way, so I think it's better just to remove that.

ff137:
> I also think we should change it to not be asynchronous in a separate task later. Maybe as a targeted release. Just to make the plugin updates more isolated and if there is any potential problems with the work around this it will be easier to find.

> We also want to get the Kanon profile stuff completed. I'm not sure how many merge conflicts this will cause with it, but if it's ready now, it might be better to merge that first.

Agreed on both 👍 
Good to narrow the scope and do `async def` -> `def` change separately
And I think good to merge the Kanon stuff first. It'll be a smaller headache for me to fix merge conflicts here, than yet again making them fix merge conflicts there. So I'm happy to wait for that merge first.

esune (commented):
Did a pass, changes look great - much cleaner code in many sections. Will look out for a rebase after the other PR is merged for final review

ff137:
Rebased post-Kanon and ready to review again

jamshale (approved):

PR #3691, opened 2025-04-29
:construction_worker: Skip sonar-merge-main workflow if github actor is dependabot

ff137:
This is because GitHub Actions does not provide repository secrets (such as SONAR_TOKEN) to workflows triggered by dependabot[bot] for security reasons. As a result, the SonarCloud scan step fails when dependabot merges to main (i.e., using `@dependabot squash and merge`, instead of manually merging oneself). By skipping the scan when the actor is dependabot, we avoid unnecessary workflow failures.

Signed-off-by: ff137 <ff137@proton.me>

### DIFF
--- a/.github/workflows/sonar-merge-main.yml
+++ b/.github/workflows/sonar-merge-main.yml
@@ -12,7 +12,7 @@ jobs:
   sonarcloud:
     name: SonarCloud
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ff137:
Sample of failing workflow when dependabot is actor: https://github.com/openwallet-foundation/acapy/actions/runs/14729296725

jamshale (approved):

PR #3692, opened 2025-04-30
fix: multiuse invite derived conns should have msg id

dbluhm:
The connection completed webhook for connections created from multi-use public invitations (and probably other multi-use configurations) were lacking `invitation_msg_id` for the inviter role. This made it difficult to determine if a connection originated from a particular invitation. For too long, we've relied on the `invitation_key` value, which did happen to be properly populated. But, especially with the move to did:key representations, this became increasingly burdensome in some cases where only the base58 representation of the key was known. Additionally, it meant that different multi-use invitations for the same public DID would be indistinguishable from each other since the key of the DID was used as the invitation key.

I added a scenario to demonstrate the issue.

### DIFF
--- a/acapy_agent/protocols/didexchange/v1_0/manager.py
+++ b/acapy_agent/protocols/didexchange/v1_0/manager.py
@@ -756,6 +756,7 @@ class DIDXManager(BaseConnectionManager):
         """
         new_conn_rec = ConnRecord(
             invitation_key=conn_rec.invitation_key,
+            invitation_msg_id=conn_rec.invitation_msg_id,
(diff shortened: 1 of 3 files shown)


dbluhm:
Would really like to see this fix make it into a release soon, if at all possible. Thanks!

dbluhm:
Sonar cloud analysis can be disregarded; points to use of http urls in the scenario test.

jamshale:
They still print out in the `Check if tests failed` step.

We do occasionally get random fails with the interop tests. Looks like it could be the case here.

ff137 (approved):
👍

PR #3719, opened 2025-05-13
:loud_sound: Improve logging related to public DIDs and routing keys

ff137:
General addition of debug / info / warning / error logs throughout the steps related to promoting public DIDs and routing verkeys, with some improvements to surrounding methods.

Improved logging = happier developers

These logs would've saved me 10 hours of debugging 😅

___
Developer notes (to share some of the thinking behind these changes):
- Any method that does important work should debug/info log when it is entered.
- Log-lines improve readability / reduce complexity (e.g. in if-else blocks), and reduce cognitive load when debugging.
- Don't use `pass`; rather debug-log the intention, so that if/else/exceptions don't pass silently.
- It's good practice to always log in the line before an exception is raised, because sometimes a parent method will silently swallow that exception, and tracing code becomes much harder.
  - Therefore not all exceptions will need error/warning logs, some are just informative.

___
Includes some test coverage, just to resolve the sonarcloud quality gate (thank you GPT 4o)

### DIFF
--- a/acapy_agent/anoncreds/verifier.py
+++ b/acapy_agent/anoncreds/verifier.py
@@ -199,9 +199,11 @@ class AnonCredsVerifier:
                                 f"{uuid}"
                             )
                             LOGGER.info(
-                                f"Timestamp {timestamp} from ledger for item"
-                                f"{uuid} falls outside non-revocation interval "
-                                f"{non_revoc_intervals[uuid]}"
+                                "Timestamp %s from ledger for item %s falls outside "
+                                "non-revocation interval %s",
+                                timestamp,
+                                uuid,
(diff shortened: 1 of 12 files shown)


swcurran (approved):
Scanned it -- real Devs would know better, but this is awesome.  Great work!

swcurran:
Guess I should have noticed the test failing and the WIP flag...

ff137:
I slipped in some modification to warning logs on startup -- no longer logs warnings if ledger is disabled, or if `--from-version` startup argument is not set / wallet type record not found. Feel like warnings shouldn't be logged for standard operation. Those are just info logs now

jamshale (approved):
Nice improvement 👍

Issue #2818, opened 2024-02-28
Multiuse invitations with a mediator cause keylist update webhook to fail


dbluhm:
When using a mediator, there is a connection lookup occurring in the keylist update handler in order to emit a webhook/event with some additional info. This can fail when using multiuse invitations:

```python
2024-02-28 00:29:39,580 aries_cloudagent.core.conductor ERROR Exception in message handler:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/asyncio/tasks.py", line 256, in __step
    result = coro.send(None)
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/core/dispatcher.py", line 253, in handle_message
    await handler(context, responder)
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/coordinate_mediation/v1_0/handlers/keylist_update_response_handler.py", line 31, in handle
    await self.notify_keylist_updated(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/coordinate_mediation/v1_0/handlers/keylist_update_response_handler.py", line 45, in notify_keylist_updated
    key_to_connection = {
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/coordinate_mediation/v1_0/handlers/keylist_update_response_handler.py", line 46, in <dictcomp>
    updated.recipient_key: await route_manager.connection_from_recipient_key(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/protocols/coordinate_mediation/v1_0/route_manager.py", line 292, in connection_from_recipient_key
    conn = await ConnRecord.retrieve_by_tag_filter(
  File "/home/aries/.local/lib/python3.9/site-packages/aries_cloudagent/messaging/models/base_record.py", line 265, in retrieve_by_tag_filter
    raise StorageDuplicateError(
aries_cloudagent.storage.error.StorageDuplicateError: Multiple ConnRecord records located for {'invitation_key': 'HGkqsn48rpcWawXWAHiikVo26AcTEbhAr8Mh1iupTivA'}
```

The keylist update is still correctly handled; only the emission of the webhook is failing. This webhook is not widely used, which is why it went undetected. I would classify this as an annoyance more than anything but should be fixed regardless.

jamshale:
This issue reported by Sonarcloud looks like it could be related?

[...]

PR #3721, opened 2025-05-14
Regress test to check #2818 issue

andrepestana-aot:
Bug for #2818 is already fixed.
This PR adds a guard test so we never regress.


dbluhm (approved):
Nice! Thanks!

Issue #3627, opened 2025-04-03
🎨 ACA-Py should not log credential attributes


ff137:
Credential attributes can contain private information, and it should be obfuscated in logs. Examples of where

PR #3722, opened 2025-05-15
:loud_sound: Improve logging

ff137:
Just a few commits:

Currently, the handlers for credentials and proofs log every message at info level. 

- This is more appropriate for a debug level -- for one, because messages can contain private data (relates to #3627), and two, because these logs are only meaningful when somebody is actively observing logs and trying to trace stuff down. For standard, production-level operation of the agent, these logs are not informative; and if someone wants to see all messages, then debug will be the way to run it.

- (Bearing in mind that debug level will then log credential attributes -- someone still needs to implement a solve for the above-linked issue, e.g. dropping credential attributes from log messages altogether, perhaps manually removing each call, or adding some Pydantic config.)

I searched for each Handler, but may have missed some info logs, but it's a good change to start and anyone can pick up other log improvements they'd like to see.

Some handler debug logs used f-string to add context (so, strings are concatenated even when not being logged, when running at >=info level), so there's also a commit to use lazy logging instead, just a minor optimisation

___
Side note: 
Moving more helpful logs to debug level does further incentivise implementing a trace level, because we want debug level to be useful for developers, but not too noisy. There are many mundane yet important operations that happen frequently e.g. CRUD operations, which should be logged, but moving to a trace level makes the most sense, as it's reserved for more serious troubleshooting.


swcurran (approved):
LGTM -- but don't know if you want to do anything about the "code duplication" issue -- I don't see what you would do, but...

ff137:
> don't know if you want to do anything about the "code duplication" issue

If I could, I would bump the quality gate's duplication threshold up to 10 or 20% 😄 -- 3% is very stiff! Because just touching existing code, and having e.g. 1 duplicate log line out of 30 lines changed, will flag the quality gate

jamshale:
We can just ignore the duplication. It's not relevant every time. I'm not sure of a way to change the level atm. I should be able to change it, if it lets you.

jamshale (approved):

PR #3736, opened 2025-05-23
Add websocket outbound debug log

jamshale:
There was no logging for when the agent is sending outbound websocket messages. This makes it easier to understand when the agent is sending over ws as opposed to http transport.


swcurran:
@ff137 -- best for you to take a look at this.  Aligned with the evolving log handling?

ff137 (approved):

Issue #3637, opened 2025-04-08
migrate demo from ngrok to localtunnel


davidchaiken:
In PR #3623, we started discussing moving the demo from ngrok to localtunnel. This issue tracks the discussion and work required to do this migration.

davidchaiken:
Here are the files in the demo directory that mention ngrok...
```bash
demo $ grep -irl ngrok .
./run_demo
./multi-demo/Dockerfile.acapy
./multi-demo/README.md
./multi-demo/docker-compose.yml
./multi-demo/ngrok-wait.sh
./playground/.env.sample
./playground/README.md
./playground/docker-compose.yml
./playground/start.sh
./docker-agent/Dockerfile.acapy
./docker-agent/README.md
./docker-agent/docker-compose.yml
./docker-agent/ngrok-wait.sh
./run_bdd
./ngrok-wait.sh
./elk-stack/README.md
```

I'm interested in whether localtunnel has comparable functionality to ngrok. I've found the localhost API on port 4040 useful for automating configuration and the Traffic Inspector functionality on the ngrok website useful for debugging.

davidchaiken:
@swcurran I had a look at using localtunnel instead of ngrok...

The main thing that I want to confirm is that it's okay that the tunnel between the acapy development host (e.g. my laptop) and the public localtunnel servers (on the domain loca.lt) run on cleartext TCP connections, not over secure connections. I realize that the [localtunnel web site](https://theboroer.github.io/localtunnel-www/) claims "Secure https for all tunnels," but that doesn't quite tell the whole story. Yes, the client connection (e.g. from the BC Wallet) uses https, but the public localtunnel servers terminate the https and then forward the connection over an insecure tunnel to the server running localtunnel. In the reverse direction (server to client), the connection is cleartext to the loca.lt server and then encrypted between loca.lt over the https session. I've confirmed this behavior using the [official localtunnel js client](https://github.com/localtunnel/localtunnel), watching the connection to the tunnel (loca.lt) server using `tcpdump -X -i en0 ip host localtunnel.me` on my laptop.

I had a look at both the localtunnel client code and the [localtunnel server code](https://github.com/localtunnel/server). I may be missing something, but I don't see any support for secure connections in either repo -- except that the server needs to know whether to tell the client to use the http or the https schema. It looks like the folks running the public server put a proxy that handles secure https connections in front of their localtunnel server.

[...]

PR #3737, opened 2025-05-24
alice/faber demo supports Microsoft dev tunnels

davidchaiken:
See Issue #3637 for discussion. Please do not merge this PR until the discussion in the issue confirms that supporting dev tunnels is actually a good idea.

* Add Microsoft dev tunnels support to the Alice/Faber demo
* Ports forwarded by devtunnel override ngrok, but do not override the environment when run_demo is started
* Any combination for devtunnel and ngrok forwarding should work
* The ngrok port detection code now does just one call to the ngrok API and then reuses the results


davidchaiken:
From ACA-Pug meeting... Need to add some instructions to documentation on when to use devtunnel and when to use ngrok.

PR #3744, opened 2025-05-26
Remove unnecessary hash pinning

jamshale:
Pinning to a hashes version of these docker images isn't necessary and is causing additional dependabot maintenance.


swcurran:
It was added here by step-security-bot -- https://github.com/openwallet-foundation/acapy/pull/3668, which in turn was enabled by @ff137 .

ff137 (approved):

ff137:
There are fortunately lots of other dependencies that can be pinned too, to make up for the scorecard. I think it's mainly docker images like python/postgres that you don't necessarily want on latest, which can be skipped

PR #3751, opened 2025-05-29
Demo: Change mediation connection to out-of-band

jamshale:
Changes the mediation connection in the demo to use out-of-band and removes some old connections code.


demo/runners/support/agent.py
```diff
@@ -1504,14 +1504,20 @@ async def get_invite(
                 invi_params = {
                     "auto_accept": json.dumps(auto_accept),
                 }
-                payload = {"mediation_id": self.mediator_request_id}
+                payload = {
+                    "mediation_id": self.mediator_request_id,
+                    "handshake_protocols": ["https://didcomm.org/connections/1.1"],
```
swcurran, line 1509: Sorry for the dumb question, but should this still be connections?

jamshale: Nice catch. I meant to change this to `did-exchange/1.1`. Not exactly sure why this worked.

shiseki:
I git cloned the updated acapy and executed ./run_demo run alice --mediation ,which ended up with the following error.
It seems that there is a small error with the usage of PromptSession, so I changed the line 162 of utils.py with
                tmp = await session.prompt(*args, **kwargs)
then it worked as expected.
**************************************************************************************
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/aries/demo/runners/alice.py", line 265, in <module>
    asyncio.get_event_loop().run_until_complete(main(args))
  File "/usr/local/lib/python3.12/asyncio/base_events.py", line 691, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/aries/demo/runners/alice.py", line 159, in main
    await input_invitation(alice_agent)
  File "/home/aries/demo/runners/alice.py", line 75, in input_invitation
    async for details in prompt_loop("Invite details: "):
  File "/home/aries/demo/runners/support/utils.py", line 172, in prompt_loop
    option = await prompt(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aries/demo/runners/support/utils.py", line 162, in prompt
    tmp = await PromptSession().prompt_async(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'PromptSession' object has no attribute 'prompt_async'
**************************************************************************************

swcurran:
Thanks for the feedback.  @jamshale -- do we need a fix for this PR?

jamshale:
I don't think so. Maybe the error occurred when using the wrong version of `prompt-toolkit`?

This was changed recently here. https://github.com/openwallet-foundation/acapy/pull/3713.

shiseki:
@jamshale Thank you for your comment. You are absolutely right. I was using promt-toolkit version 2. I upgraded it to version 3 (3,0.51), and then I was able to build and run alice --mediation successfully.

PR #3753, opened 2025-05-30
Remove  header from http/ws responses

jamshale:
This removes the `Server` header from http/ws responses. `Server` is where the python and aiohttp are automatically added by aiohttp library.

To remain in accordance with HTTP specification we could instead override it with a custom name (https://datatracker.ietf.org/doc/html/rfc9110#name-server).

NOTE: 
 - This does not effect requests. When the agent initiates a request it has a `User-Agent` header which also contains the python and aiohttp version. I'm not sure if we'll need to remove this info as well. `User-Agent` is important from some interactions but probably not for a acapy agent. We'd need to make sure to only remove the correct info.


ff137 (approved):
Approving because I missed the linked issue re: OWASP hardening

PR #3759, opened 2025-06-02
Repair lts workflow

jamshale:
It's hard to test this without setting up a bunch of publishing stuff but I believe this will fix the broken workflow.

### DIFF
--- a/.github/workflows/tag-recreate-lts.yml
+++ b/.github/workflows/tag-recreate-lts.yml
@@ -6,6 +6,7 @@ name: Tag and Recreate LTS Release
 on:
   release:
     types: [published]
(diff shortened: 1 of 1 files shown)


swcurran (approved):
Let's give it a go...

jamshale:
It's still triggered via publish action. I'm not sure if it will succeed without a new version or not. I'll try and run it manually.

jamshale:
I think this would need to be an actual release and publish. I tried manually with the same tag and it never ran.

swcurran:
We'll do an rc1 to see how it works.  And we can do likewise with the next LTS releases.

PR #3761, opened 2025-06-03
:zap: Skip upgrade check for status checks

### LINKED ISSUE #3694: :zap: Improve performance by reworking "upgrade middleware"
Relates to changes introduced here: https://github.com/openwallet-foundation/acapy/pull/2922
___
_Originally posted by @ff137 in https://github.com/openwallet-foundation/acapy/issues/2922#issuecomment-2841594532_

> The upgrading middleware will now use a cache during upgrades and after the upgrade has been completed. Agents that have the upgrade middleware but never upgrade will still have a very minor performance hit, due to needing to query the is_upgrading record.

With regards to that "minor performance hit" -- I'm noticing that because `upgrade_middleware` is always added to the web application's set of middleware, it means that askar is queried for `acapy_upgrading` records every single time there is any incoming request.

I'm picking this up after testing with more debug logging (#3689)

And I can see with every `/status/live` or `/status/ready` check that we do, we get:
```
2025-04-30 10:52:50,958 acapy_agent.admin.server DEBUG Incoming request: GET /status/live
2025-04-30 10:52:50,963 acapy_agent.storage.askar DEBUG Fetching all records for acapy_upgrading with tag query None
2025-04-30 10:52:50,971 aiohttp.access INFO 127.0.0.6 [30/Apr/2025:10:52:50 +0000] "GET /status/live HTTP/1.1" 200 194 "-" "kube-probe/1.32"
2025-04-30 10:52:50,977 acapy_agent.admin.server DEBUG Incoming request: GET /status/ready
2025-04-30 10:52:50,982 acapy_agent.storage.askar DEBUG Fetching all records for acapy_upgrading with tag query None
2025-04-30 10:52:50,988 aiohttp.access INFO 127.0.0.6 [30/Apr/2025:10:52:50 +0000] "GET /status/ready HTTP/1.1" 200 194 "-" "kube-probe/1.32"
```

That means an extra, unnecessary askar session and connection is opened up _every single API call_.

I think this isn't just a minor performance hit, but really not the way that this should have been implemented.

Can we please revisit this, and see if there's a different way to handle the "upgrading" logic?

ff137:
The blocking middleware for upgrades is only required for certain requests, and can be skipped for status checks.

Edit: I noticed that the real problem for #3694 is that the base wallet does not get added to the AnonCreds singleton if it's already anoncreds - it only gets added if it went through the upgrade process - and because this singleton is used to check if a wallet can be upgraded or not, it would query the DB for every request, even though the wallet can't be upgraded.

Resolves #3694


ff137:
@jamshale Should the upgrade check only be run for wallets that are of `askar` type?

In which case, I think that new tenants being created with `askar-anoncreds` type should also be added to the singleton. So that their requests don't have to be checked if an upgrade is in progress.

Would it be possible to do away with the singleton, and just check the incoming request context's wallet type setting? Or is that not reliable

jamshale (approved):
This is a good improvement 👍 

Yes, currently the upgrade middleware is only for anoncreds wallets so adding it like this and when a new wallet is created is a good plan. I guess I had missed that. 

We can merge this and I can add the wallet to the singleton when askar-anoncreds is created in another PR.

I had tried to try and make this middleware re-usable in the future if we ever needed to lock the api for any critical migrations. However, it would still have this limitation for multi-instance deployments needing to check the database. 

I think we should consider adding a pub/sub system to acapy  with postgres/sqlite support via hooks. It could be very useful for situations like this and transactions which need to be done atomically.

jamshale:
We may be able to do away with the singleton. I can't remember if there was a practical reason why I did that instead of just checking the wallet-type. I had done this before I had very much experience with multi-tenancy.

ff137:
I think it's worth mentioning that NATS / JetStream will solve a lot of ACA-Py's state and event bus problems. But that will entail a _lot_ of refactoring and more microservices / domain driven design principles. It might as well be a new project. But I see no reason why it can't be a design for ACA-Py 2.0

PR #3764, opened 2025-06-04
chore: Remove `did:indy` Stub

TheTechmage:
We're working on implementing the `did:indy` resolver/registry with anoncreds via the https://github.com/Indicio-tech/did-indy-py library that @dbluhm setup. We're planning to implement it as a plugin, rather than as part of core. The code that I've removed has caused a bit of a development hurdle since, once the plugin is loaded, there is now "More than one registrar found for identifier did:indy" which blocks the plugin from running.

So that's all I did. I just removed the `did:indy` registry class to allow the plugin version of the class to load. It's just a bunch of stubbed out methods that return `NotImplemented` exceptions anyways so there should be 0 negative impact.


jamshale:
Looks good. I think there's a couple tests that must be asserting the default plugins that are loaded. Hopefully it's not much trouble to update them.

jamshale (approved):

PR #3772, opened 2025-06-10
Add recreate lts workflow to 1.2.lts branch

jamshale:
***Needs more testing...***

After studying this further and doing some testing. The workflow needs to be added to the lts branches because it needs to be available for the released tag. Then for main which isn't LTS yet it will need to skip this workflow.

When a new LTS is created it will need to add the if condition with the major versions to the initial commit of the LTS.


jamshale:
Can we merge these and get them out of the way for the next patch release? It doesn't require a approval but maybe a quick review.

PR #3773, opened 2025-06-10
Disable lts-recreate workflow for main branch

jamshale:
This disables the workflow for the main branch and fixes a couple things so the next LTS branch can use it properly. 

LTS branches will need to enable it by adding a condition to match the lts version on the published tag.

### DIFF
--- a/.github/workflows/tag-recreate-lts.yml
+++ b/.github/workflows/tag-recreate-lts.yml
@@ -6,9 +6,6 @@ name: Tag and Recreate LTS Release
 on:
   release:
     types: [published]
-  push:
-    branches: 
-      - '*.lts'
       
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-recreate-lts.yml
```diff
@@ -6,16 +6,15 @@ name: Tag and Recreate LTS Release
...
-      - '*.lts'
       
 permissions:
   contents: write
   packages: write
 
 jobs:
   recreate-lts-release:
+    # This job is disabled by default for main, should be enabled for LTS branches and tags
+    if: false
```
swcurran, line 22: So I understand this -- the "push" directive was intended to have the effect of not running this on main, but for some reason it didn't work.  So this `if` is hardcoded to `false` on main, and will be hardcoded to `true` on the LTS branches.  Right?

jamshale: Correct. The push directive didn't work and run on main.

The PRs for the LTS branches have a hardcoded if statement to only run when the release tag matches the lts.

swcurran: How about adding the lines used on the LTS branch to this, commented out, so the main and LTS versions (almost) match, except for what's commented out and the LTS details. Makes it easier to know what to do when creating a new LTS.

PR #3784, opened 2025-06-17
Fix the rev_list read, increment index and update transaction

jamshale:
So, I'm quite pissed off at myself but I was looking at adding the distributed lock to the old credx issuance and noticed the problem with the index wasn't reproducing. 

After studying it further, I noticed that the entire read, index increment and update was happening inside a single transaction while the anoncreds did the read and update in separate transactions. I didn't think doing this with askar prevented another process from reading the rev_list at the same time but apparently that is the case.

The hardest and most consuming part of this was creating the testing for multiple instances which I think will still be valuable going forward with this project.

I'm not sure if the distributed locking mechanism will ever be needed or not. It wasn't too hard to implement and could potentially be useful in the future. It did solve this issue in another way, but I don't think it's required for this problem anymore...

### DIFF
--- a/scenarios/examples/clustered/docker-compose.yml
+++ b/scenarios/examples/clustered/docker-compose.yml
@@ -11,6 +11,12 @@
     depends_on:
       - faber
       - alice
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+      start_period: 15s
 
   faber:
     image: acapy-test
@@ -50,6 +56,8 @@
     depends_on:
       tails:
         condition: service_started
+      wallet-db:
+        condition: service_healthy
 
   alice:
     image: acapy-test
@@ -80,6 +88,11 @@
       interval: 7s
       timeout: 5s
       retries: 5
+    depends_on:
+      tails:
+        condition: service_started
+      wallet-db:
+        condition: service_healthy
 
   example:
     container_name: controller
@@ -98,6 +111,8 @@
         condition: service_healthy
       wallet-db:
         condition: service_healthy
+      nginx:
+        condition: service_started
 
   tails:
     image: ghcr.io/bcgov/tails-server:latest
--- a/scenarios/examples/clustered/docker-compose.yml
+++ b/scenarios/examples/clustered/docker-compose.yml
@@ -35,12 +35,14 @@
         --wallet-storage-type "postgres_storage"
         --wallet-storage-config "{\"url\":\"wallet-db:5432\",\"max_connections\":5}"
         --wallet-storage-creds "{\"account\":\"DB_USER\",\"password\":\"DB_PASSWORD\",\"admin_account\":\"DB_USER\",\"admin_password\":\"DB_PASSWORD\"}"
+        --auto-accept-invites
+        --auto-accept-requests
         --auto-provision
         --log-level info
         --debug-webhooks
         --notify-revocation
     healthcheck:
-      test: curl -s -o /dev/null -w '%{http_code}' "http://nginx/status/live" | grep "200" > /dev/null
+      test: curl -s -o /dev/null -w '%{http_code}' "http://localhost:4000/status/live" | grep "200" > /dev/null
       start_period: 30s
       interval: 7s
       timeout: 5s
@@ -67,6 +69,8 @@
         --wallet-name alice
         --wallet-key insecure
         --auto-provision
+        --auto-accept-invites
+        --auto-accept-requests
         --log-level info
         --debug-webhooks
         --monitor-revocation-notification
--- a/scenarios/examples/clustered/docker-compose.yml
+++ b/scenarios/examples/clustered/docker-compose.yml
@@ -1,0 +1,202 @@
+ 
+ services:
+  nginx:
+    image: nginx:latest
+    container_name: test_nginx
+    ports:
+      - "8080:80"  # Host:Container for Admin API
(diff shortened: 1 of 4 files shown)


swcurran:
@ff137 -- what do you think?  This would explain why this hadn't been reported before.

The question I have is whether or not we should include the Distributed Lock work that @jamshale worked on in the PR that he closed.  It seems to me that it might be a valuable feature. On the other hand, it seems we haven't had a use case for it yet.  Should we add it anyway?

Comments on this PR and the second question welcome from others -- @esune @thiagoromanos @andrewwhitehead @dbluhm @cjhowland @mepeltier

jamshale:
I do think the distributed lock is good in some ways. I still don't know if this got a ton of concurrent issuance requests if it's truly atomic or not. The scenario test can only test concurrency to limited load.

The distributed lock mechanism does add more complexity and the chance of stuck locks, etc, so it's better to avoid it if we don't need it.

I don't think it was a complete waste of time if only to generate the test and think about concurrency in a clustered deployment more.

esune:
I agree with @jamshale, given the discovery and the possibility that wrapping the required steps in a `transaction` will heavily mitigate or solve the issue I would at least wait and reassess before we implement the additional locking mechanism - the added complexity might not be worth the benefit, if very/too small.

scenarios/examples/clustered/docker-compose.yml
```diff
@@ -0,0 +1,202 @@
...
+      - "5000:5000"  # Host:Container for didcomm
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf:ro
+    depends_on:
+      - faber_0
+      - faber_1
+      - faber_2
+      - alice
+
+  faber_0:
```
esune: Wondering if it might be worth using the [scale](https://docs.docker.com/reference/compose-file/services/#scale) and/or [replicas](https://docs.docker.com/reference/compose-file/deploy/#replicas) in docker-compose rather than defining separate services? This might also eliminate the need for a reverse proxy.

jamshale: I'll have a look at that. Makes sense that would be a better option.

jamshale:
I think we should go ahead with this solution for now and get a release out. 

From what I understand is askar transactions aren't atomic for multiple instances and connection pools. So, while this greatly reduces the chance 2 processes grab the same index at the same time it doesn't prevent multiple instances grabbing a transaction at the same time. At large scale this could still occur. 

This is such a critical process because it prevents multiple credentials having the same revocation index and also if it fails than the issuance is cancelled. Perhaps for a large scaled deployment having a fully atomic lock is the correct solution. 

Still I think this solves the problem for 99.9+% of the time.

jamshale:
ChatGPT sums this up really well. https://chatgpt.com/s/t_6852e134862081918dd705fd92b46205

scenarios/examples/clustered/docker-compose.yml
jamshale: This is updated to use replicas but the nginx proxy remains as it's recommended for routing traffic to the replicas in docker.

swcurran:
Not really understanding the ChatGPT response in combination with the "this will be good enough for now" assessment.  Let's talk about this at Tuesday's ACA-Pug meeting.  Not to hold up this PR -- it's necessary and should go ahead, but rather to discuss the larger issue about why this is "good enough" vs. the heavyweight (but not really helpful?) distributed lock solution.

jamshale:
Sounds good. I think it's more like this is a big improvement over basically what is a bug with it currently. However, the distributed lock is what is needed for a fully scalable heavy load deployment.

The distributed lock is a more complicated feature so getting the bug fixed for a release would probably be a good thing to do while we discuss and fully test out the distributed lock if we decide it is needed.

jamshale:
Need to look at the failing scenario test. Thought it was working.

scenarios/examples/clustered/docker-compose.yml
```diff
@@ -0,0 +1,129 @@
...
+    volumes:
+      - ./example.py:/usr/src/app/example.py:ro,z
+    command: python -m example
+    depends_on:
+      faber:
+        condition: service_healthy
+      alice:
+        condition: service_healthy
+      wallet-db:
+        condition: service_healthy
```
ff137, line 113: Looks like scenario fails with:
`Cannot connect to host nginx:80 ssl:default [Name or service not known]`

Not sure, but should this also depend on nginx?

ff137: And perhaps depends_on: wallet-db should be moved to alice and faber

ff137:
By the way - just mentioning because I'm busy debugging some revocation stuff - it looks like cred_rev_id starts from 1, and this means that the `revocation_list` resources have an unused slot (the 0th index of the list).

i.e. after revoking some credentials, I see in the logs: `revocation_list=[0, 1, 1, 1, ...]`. Which makes it look like rev reg size of n is actually n-1, since the first slot is unused.

Not a big deal, but because we're looking at this code now, would it make sense to start counting from 0 instead of 1? Or is it intentional like it is

jamshale:
This is because of the anoncreds revocation algorithm. The 0th index is reserved. It's caused a lot of confusion. 

It's different from the legacy indy implementation because it it uses an actual list data type. I think we could change the block size to size + 1 to make it so it actually had the same amount of credentials as you think you are creating in a block.

jamshale:
I'm having a lot of trouble with github actions and this scenario test. I'm going to get a separate PR open for only the fix to move this forward when I figure out how to fix it.

PR #3794, opened 2025-06-24
0.12.7rc0

swcurran:
First release candidate for 0.12.7 LTS release. Adds only the LTS publishing job.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.12.7rc0
 
-### June 24, 2025
+### June 25, 2025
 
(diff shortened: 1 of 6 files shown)


swcurran:
@pradeepp88 -- here is the first RC release.  In doing this, I realize that the publication should only happen for final releases -- not RCs.  This PR will let us test the process, but do we need to update the GHAs to bypass RCs?  Or will that already happen?

Thanks

jamshale:
Ugh. I think it will do it for rc releases. We should be able to put a simple if and exit statement for the tag name.

jamshale:
I tried to address this with https://github.com/openwallet-foundation/acapy/pull/3796. I haven't tested it out yet. Need to adjust all the publish workflows for my fork.

jamshale:
Ok this PR to fix it is tested now. Wasn't actually hard to do.

swcurran:
@jamshale -- added your PR to the release. Please review again and once approved, I'll merge and do a release.

jamshale (approved):

PR #3798, opened 2025-06-25
:test_tube: Skip tests with jsonld url resolution failures

ff137:
The jsonld tests will always fail if w3.org is not resolvable.

This PR contributes a decorator that can be added to jsonld tests. It will automatically check if test failures matches a JsonLd error for URL resultion failure, in which case it will skip the test.
Other test failures will re-raise as normal.

Pytests are required for CICD, so this helps avoid development being blocked when w3.org is down

___
Shoutout to Cursor + claude-4-sonnet for doing this in less than 5 minutes

### DIFF
--- a/acapy_agent/messaging/jsonld/tests/test_credential.py
+++ b/acapy_agent/messaging/jsonld/tests/test_credential.py
@@ -3,7 +3,7 @@
 import json
 from unittest import IsolatedAsyncioTestCase
 
-from ....utils.testing import create_test_profile
+from ....utils.testing import create_test_profile, skip_on_jsonld_url_error
 from ....wallet.base import BaseWallet
 from ....wallet.key_type import ED25519
 from .. import credential as test_module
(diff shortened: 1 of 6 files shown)


jamshale (approved):

jamshale:
I think the interop tests will fail but we can merge it anyway.

dbluhm:
I'm late to the party but shouldn't the fix for this be to not resolve the JSON-LD context dynamically but rather have all contexts downloaded and statically cached in the document loader? This is the "real" way to use JSON-LD Contexts (with mild sarcasm on those quotes)

swcurran:
@dbluhm (hey!!! hope you are well!!) is right.  The correct thing to do is to download and hold locally the JSON-LD context--perhaps with a check to see if our cached copy matches the remote and fail if not (but not if the remote is inaccessible -- ignore that).  I thought we had already done that for at least some of our JSON-LD contexts.  Will look for that...

PR #3800, opened 2025-06-26
1.3.1rc2

swcurran:
Should be the last RC before the final 1.3.1. Contains all completed PRs, and we don't have any other candidates.

Rapid review and verification would be appreciated so we can get 1.3.1 out ASAP.

Thanks


swcurran:
@jamshale --- can you please check why the PR Tests failed.  Nothing I'm doing in this PR...

jamshale:
Not sure. It's passing when I run in locally. This test change with a PR merged yesterday but it passed for the PR :/

jamshale:
I just re-ran it and it passed. They do run in github actions as a swarm of processes so maybe something happened there. 

I think we can just keep an eye on this. I don't have much of a concern about it.

jamshale (approved):

PR #3808, opened 2025-07-02
:art: Code cleanup and :arrow_up: lock file update

ff137:
Includes some minor improvements that I've been stashing away while working on other stuff.

See individual commit messages for more detail. 

Including a poetry lock file update, because it's been a while since the last one. (e.g. aiohttp is upgrading from 3.12.6 to 3.12.13)


swcurran:
Was going to start the morning off with a 1.3.1 release -- should I wait on this or go ahead without it?  Seems both good and safe (famous last words...), so leaning to include it.  The missing 3.1% coverage is annoying :-).

jamshale:
I think we can go ahead with 1.3.1 and merge this after. It's good and seems safe with all the test passing, but it's not critical to get into the release.

jamshale (approved):

ff137:
Really minor changes. Doesn't have to be included in 1.3.1, but imo it may as well.

There's 1 bug fix included here - `get_schema_info_by_id` was missing the profile argument in the anoncreds `DIDWebRegistry`.

And, in general, I think it's worth doing a lock file update before a release, so all dependencies are up to date. Currently:
- there's a security advisory to upgrade pillow; 
- when installing the project there's a warning about configargparse 1.7 being yanked
  - `Warning: The file chosen for install of configargparse 1.7 (ConfigArgParse-1.7-py3-none-any.whl) is yanked. Reason for being yanked: incorrect metadata related to supported python versions`

But it should be inconsequential either way. The lock file update can just be done separately in the 1.3.1 PR. Much of a muchness

jamshale:
Ok. Let's merge it. I'm pretty confident the tests would have found any problems with these changes.

PR #3811, opened 2025-07-03
Add retries when opening the askar store / Refactor store.py

jamshale:
This adds a few re-tries to the open store process. When starting a new clustered agent, multiple isntances would attempt to create the store at the same time and result in race conditions, causing agents to fail. In an environment where agents are forced to restart this isn't really a problem but still this is a cleaner process.

Refactored askar.store.py file.


jamshale:
I am going to self review a bit more and increase the test coverage.

acapy_agent/askar/store.py
```diff
@@ -15,200 +17,180 @@
...
+                        store = await Store.open(
+                            uri, self.key_derivation_method, self.DEFAULT_KEY
+                        )
+                    except AskarError as err:
+                        await self._handle_open_error(err)
+                    await Store.rekey(store, self.rekey_derivation_method, self.rekey)
+                    return AskarOpenStore(self, provision, store)
 
-        return AskarOpenStore(self, provision, store)
+        raise ProfileError("Failed to open or provision store after retries")
```
ff137: This method is too complex. 

It needs comments, debug logging, and/or more modularized methods to reduce the cyclomatic complexity

acapy_agent/askar/store.py
```diff
@@ -15,200 +17,180 @@
...
-            )
-        if err.code == AskarErrorCode.NOT_FOUND:
-            raise ProfileNotFoundError(
-                f"Store '{self.name}' not found",
+    async def _handle_open_error(self, err: AskarError, retry=False):
+        if err.code == AskarErrorCode.BACKEND:
+            LOGGER.warning(
+                """Askar backend error: %s. This may indicate multiple instances
+                attempting to create the same store at the same time or a misconfigured 
+                backend.""",
```
ff137: Using """ means it'll print with the whitespace. So better is concatenating with single quote strings instead:
```py
            LOGGER.warning(
                "Askar backend error: %s. This may indicate multiple instances attempting "
                "to create the same store at the same time or a misconfigured backend."
                err,
            )
```

acapy_agent/askar/store.py
```diff
@@ -15,200 +17,180 @@
...
     DEFAULT_KEY = ""
     DEFAULT_KEY_DERIVATION = "kdf:argon2i:mod"
-    DEFAULT_STORAGE_TYPE = None
-
-    KEY_DERIVATION_RAW = "RAW"
-    KEY_DERIVATION_ARGON2I_INT = "kdf:argon2i:int"
-    KEY_DERIVATION_ARGON2I_MOD = "kdf:argon2i:mod"
+    SUPPORTED_STORAGE_TYPES = ("sqlite", "postgres")
 
     def __init__(self, config: Optional[dict] = None):
```
ff137, line 26: Just a suggestion, but I think this config dictionary could be a dataclass instead. That way it's easier for devs to know what fields can be populated or not. It helps replace all the .get() with defaults, since the defaults can be defined in the dataclass. And then types are preserved, since the dictionary returns Any.

acapy_agent/askar/store.py
```diff
@@ -15,200 +17,180 @@
...
+            raise ProfileDuplicateError(f"Duplicate store '{self.name}'")
+        elif err.code == AskarErrorCode.NOT_FOUND:
+            raise ProfileNotFoundError(f"Store '{self.name}' not found")
+        elif retry and self.rekey:
             return
-
         raise ProfileError("Error opening store") from err
 
-    async def open_store(
+    async def open_or_provision_store(
```
ff137: Maybe the method can keep the same name? Since "open or provision store" is implicit in "open store" -- thanks to the docstring, and the provision: bool arg

acapy_agent/askar/store.py
jamshale: I agree with this but it's a bit out of scope for this PR. I started looking at where this dictionary is used and defined and will try and address it separately.

ff137: All good, I thought it might have too many touchpoints to change quickly

jamshale:
> If we don't want to remove it yet, then perhaps the comment can just be maintained, or tracked in an issue somewhere. Just to avoid confusion about what the rekey is for.

I think we can remove this now, actually. 0.11.lts is end-of-life and if any old deployments have this blank key problem they can upgrade to any other previous version to rekey it. I'll remove it.

acapy_agent/askar/store.py
```diff
@@ -15,200 +17,199 @@
...
+            )
+        store = await Store.open(uri, self.key_derivation_method, self.key)
+        if self.rekey:
+            await Store.rekey(store, self.rekey_derivation_method, self.rekey)
+        return store
 
-        Raises:
-            ProfileNotFoundError: If the store is not found
-            ProfileError: If there is another aries_askar error
+    async def _try_open_with_default_key(self, uri: str):
```
esune: Wondering if this function should be called something different (maybe `_rekey_from_default_key`?) or it can maybe collapsed into the previous function since functionality is very similar.

esune: Nevermind, must have reviewed a version of the file that got updated before I pushed my comment.

esune (approved):
LGTM 👍🏻

PR #3816, opened 2025-07-10
1.2.lts.patch

jamshale:
Backporting via cherry pick:
 - https://github.com/openwallet-foundation/acapy/pull/3793
 - https://github.com/openwallet-foundation/acapy/pull/3753
 - https://github.com/openwallet-foundation/acapy/pull/3699

Open for suggestions on additional commits.

I don't exactly know what happened with https://github.com/openwallet-foundation/acapy/commit/7549643f84014710547bec2598062d7caa034716. I think somehow it ended up getting merge back to my fork. It's the first commit here so it should still work correctly.


swcurran (approved):

jamshale:
I need to run all the integration tests we have on this branch locally still.

swcurran:
Sounds good.  I was leaving it to you to merge -- just giving an approval if you needed it. Although, since the branch is not protected, you don't need an approval.  Perhaps we should add protections on the LTS branches?

jamshale:
JSON-LD tests are failing :/ Probably a library upgrade. I'll figure it out. Maybe just going over any security related library upgrades is better then doing a general one.

swcurran:
Don't take too long digging in, but it is obvious (or easily found from the related updates on `main`) go for it.  Otherwise back out what you need to.  Either path might be a rabbit hole...

jamshale:
It's actually the @BBS and json tests which are usually skipped. I thought they would work on my linux machine but apparently  BbsBlsSignature2020 isn't supported for json-ld. They get a `Unable to sign credential with unsupported proof type BbsBlsSignature2020. Supported proof types: dict_keys(['Ed25519Signature2018', 'Ed25519Signature2020', 'EcdsaSecp256r1Signature2019'])` 400.

I'll check them again excluding @BBS tests.

jamshale:
Ughhh. I should have merged this without squash rebasing to keep the same commit hashes. I don't think it's a big enough deal to revert and redo but I can if we want.

PR #3824, opened 2025-07-15
Add unsecure credential storage option

PatStLouis:
This adds an option to store a W3C VC without verifying it's proof on the `/vc/credentials/store` endpoint.

This option will default to true, and can be turned off (unsecured) for experimenting with unsupported cryptosuites and/or did methods.

The DataModel will still be validated, this only affects the proof verification step.

### DIFF
--- a/acapy_agent/vc/routes.py
+++ b/acapy_agent/vc/routes.py
@@ -154,14 +154,16 @@ async def store_credential_route(request: web.BaseRequest):
     manager = VcLdpManager(context.profile)
 
     try:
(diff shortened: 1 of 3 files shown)


ff137 (commented):
Please only mark ready / tag for review when PR tests are passing. Because it gives a GitHub notification for each new commit. And if I see tests aren't passing, then I know it's not ready to merge / not ready for review.

jamshale:
FYI, if you didn't realize. We fail the test run for most warnings. There is a few of them that need to be fixed.

PR #3826, opened 2025-07-15
Upgrade rust version for indy-build to 1.56

jamshale:
I believe this should fix the build issues for the indy image. The previous rust image version 1.46 used `debian-buster` which has been EOL'd. Upgrading to 1.56 doesn't cause any typing issues but uses `debian-bullseye` which is still available.

### DIFF
--- a/docker/Dockerfile.indy
+++ b/docker/Dockerfile.indy
@@ -1,5 +1,5 @@
 ARG python_version=3.9.18
-ARG rust_version=1.46
+ARG rust_version=1.56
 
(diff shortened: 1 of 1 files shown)


swcurran (approved):

swcurran:
Thanks, Jamie -- looks like that worked.  Now onto the final LTS releases.  I'll plan to do that tomorrow -- anything we should wait on to the trigger the releases, or are we safe enough?

jamshale:
I can't think of anything. Stuff coming up is looking like features that can be in the next version.

PR #3832, opened 2025-07-18
Change ghcr token to hyperledger scoped

jamshale:
The 0.12.lts needs to delete and recreate images on the hyperledger registry. Unfortunately it was using the root github token for openwallet-foundation and hence why this failed.

This workflow runs on a release so annoyingly we will need to create yet another one.


swcurran (approved):

swcurran:
So we'll need to do another release to have this apply, right?  Sigh.  Oh well.  We'll go straight to 0.12.8 without any RCs.

PR #3837, opened 2025-07-23
Remove by_format from standard webhook payloads

jamshale:
***Note***: This will be a breaking change. 

Webhooks shouldn't be sending detailed credential and presentation information. https://github.com/openwallet-foundation/acapy/issues/3803. To enable the full payload, applications will still have the `ACAPY_DEBUG_WEBHOOKS` configuration. However, any applications using webhooks in this way should consider coming up with another approach such as saving the exchange records and querying them through the admin api. Sending information like this out via webhooks and without an api-key is a vulnerability. Only applications that have a `webhook-url` configured are affected by this.

As far as I know, only the https://github.com/openwallet-foundation/acapy-vc-authn-oidc project is using webhook information in this way and will need to use DEBUG_WEBHOOKS until another solution is created.

### DIFF
--- a/.github/actions/run-integration-tests/action.yml
+++ b/.github/actions/run-integration-tests/action.yml
@@ -27,6 +27,7 @@ runs:
         PUBLIC_TAILS_URL: ${{ inputs.IN_PUBLIC_TAILS_URL }}
         LOG_LEVEL: warning
         NO_TTY: "1"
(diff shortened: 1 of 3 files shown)


jamshale:
@esune I'm merging this. I just want it noted that vc-authn will break if this gets merged and `ACAPY_DEBUG_WEBHOOKS` is not enabled.

PR #3838, opened 2025-07-24
Add additional error handling for wrong filter

jamshale:
Just adds a couple checks when the wrong filter is used for the active issuance/presentation handler.


jamshale:
I'm going to look at this one again... There's complexity around when the issuer/verifier is indy and the holder is anoncreds and the reverse, which is tested. But the other filter handling in these functions seems to be blocked by this change without causing problems. There was another PR around this stuff I can't remember.

I'll leave in draft until I understand what's going on.

jamshale:
I'll close this and re-open a new ticket after further investigation. This problem was partly mitigated by a previous PR.

PR #3846, opened 2025-07-30
Only strip did:sov dids to unqualified did in oob receive invitation requests (holder)

jamshale:
I hate to add more did:sov unqualified logic but this condition statement was only for did:sov and did:peer and wasn't considering other did methods being used in oob invitations.

***Note:*** The did:sov needs to be supplied as a fully qualified did:sov by the acapy holder event though it was provided as unqualified in the invitation. I'm not sure how much I should try and remedy this seeing as it's just on the holder and I don't want to keep adding more unqualified did logic.


esune (approved):
LGTM 👍🏻

esune:
I agree with @jamshale assessment - this seems like a the way to go until we can (possibly?) consider `did:sov` retired and do a bigger/different clean-up?

ff137 (approved):
I think the PR title should be something like only strip OOB public_did if it's did:sov

Was first a bit confusing, because it says remove stripping, but I see it's just changing it specifically unqualify sov dids, and not others - which makes sense.

ff137:
PS: I agree that the decision to work with unqualified DIDs, at all, was a big mistake. There's all this technical debt of trying to separate sov-specific functionality from other did functions.

For that reason, I think did:sov should only be stripped at the final point that requires it to be stripped - e.g. at ledger integrations - so it can remain qualified everywhere else.

Keep all dids qualified, always, until they need to be stripped. If legacy users provide unqualified dids, then assume it's did:sov. Eventually drop support for unqualified dids altogether.

To that end, I think all dids should be handled as a rich object, instead of just a string, and that'll help solve these headaches. A did is more than just a str, so why not make a rich class and build utility into it.

```py
class DIDStr:
    """Utility class for working with string representations of DIDs"""

    def __init__(self, did: str):
        self.did = did if did.startswith("did:") else f"did:sov:{did}"
        # Add some validation checks

    # Build out utility functionality
```

It'll be much easier to be able to say `if did.is_did_peer_4: did = did.as_did_peer_2`, etc. Python offers a lot of flexibility to create these helper / rich types, so acapy should really leverage that more to make developer's lives easier!

I don't think any new feature is as important as resolving technical debt and having a maintainable codebase. So I would support such a move, sooner rather than later. But that's just my 2c worth. Fortunately ideas are cheap :-) most important is that someone shows up and does the work!

jamshale:
I agree with all that. It's probably the biggest issue currently with the project. Along with did:sov being so woven into the core. There is this document which I tried to implement at one point https://github.com/openwallet-foundation/acapy/blob/main/docs/design/DIDManagement.md. 

I eventually abandoned the effort due to a few factors, one of which being that the mobile credo agents don't handle did:sov's with the anoncreds issuance/presentation filter. The do this thing, where they migrated all legacy unqualified did's into did:peers with meta data representing the legacy did. This is actually outlined in the document, but I didn't understand the process at the time. They also migrated did:sov's to did:indy's where they could.

I think we need to re-think the approach outline in the document, but I do think fixing these things is critical for the maintainability of this project long term.

PR #3848, opened 2025-07-31
[WIP] Prevent subwallet (tenant) endpoints from needing admin key

jamshale:
This is work-in-progress and tests will fail.

This is what I was thinking for the subwallet create token endpoint. There would be a sperate authorization for the subwallet to create the authentication token for using the api that would use the subwallet wallet key. For managed mode (currently the only available), the admin get wallet endpoints would supply the key, this way admin managed systems could still retrieve the key for the tenant. In an unmanaged system, the tenant would need to keep track of it's own wallet key and not lose it because the admin would not be able to retrieve it.

However, the problem I have with this approach is that we're using the wallet keys as an authorization key. I added a re-key feature for the tenant to try and alleviate some of the concern. But still this means a compromised key could be used to rekey and lock out a tenant. And then this same key is used for the wallet encryption.

Sooo, I'm still not exactly sure what to do about this. It seems like it more needs some fundamental oauth principals applied to this.

### DIFF
--- a/acapy_agent/admin/decorators/auth.py
+++ b/acapy_agent/admin/decorators/auth.py
@@ -6,7 +6,10 @@ from typing import List, Optional, Pattern
 
 from aiohttp import web
 
+from ...multitenant.error import WalletKeyMissingError
+from ...storage.error import StorageNotFoundError
 from ...utils import general as general_utils
+from ...wallet.models.wallet_record import WalletRecord
 from ..request_context import AdminRequestContext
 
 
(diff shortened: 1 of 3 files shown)


jamshale:
@esune I'm going to close this for now. I think this task is more like enable unmanaged mode and I'm not really sure that's something we should really do considering the implied security is never achievable. 

I think a system requiring isolated tenants that manage their own keys should probably spawn their own individual agents and wallets instead of using baked in acapy multitenancy.

PR #3851, opened 2025-08-05
Update scenarios to openwallet acapy-minimal-example repo

jamshale:
acapy-minimal-example used in the scenario tests has moved to openwallet-foundation. The download would still succeed because of a re-direct but removing indicio references is preferred.


jamshale:
Hmm... Looks like a couple breaking changes with this upgrade.

dbluhm (approved):

PR #3852, opened 2025-08-08
fix: update tails server upload methods to return public file URIs

TheTechmage:
We ran into an issue where we had a different tails_server_base_url from the tails_server_upload_url. This makes sense where you want to have a separate base URL for accessing the public files, and a private URL that is used for pushing the changes (where access controls dictate who can upload files, and they are restricted on the public URL).

The `upload_tails_file` methods were returning the private URLs and the calling code was expecting public URLs. The solution was to update the methods to return the public URLs instead, while still uploading to the private URLs.


TheTechmage:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=openwallet-foundation_acapy&pullRequest=3852) **Quality Gate failed**
> Failed conditions  [6 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=openwallet-foundation_acapy&pullRequest=3852&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [41.7% Coverage on New Code](https://sonarcloud.io/component_measures?id=openwallet-foundation_acapy&pullRequest=3852&metric=new_coverage&view=list) (required ≥ 80%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=openwallet-foundation_acapy&pullRequest=3852)

Not sure why Sonar thinks that so much is "not tested". I had to modify the tests to work around failures and I even added checks to make sure we were uploading to the right URL. As far as I know, all the changes (aside from the argparse 2-line addition) are tested.

dbluhm:
Interested to know how others are using the two different tails server url parameters. It seems like @TheTechmage has correctly captured the original intent and fixed the gap but it's possible I'm not understanding the intended use.

Anyone at BC Gov using both base and upload urls for tails servers?

esune:
> 
> Anyone at BC Gov using both base and upload urls for tails servers?

We're not really differentiating in our deployments, in the only scenario we specify both we are setting them to the same (public) value so we would not have encountered this behaviour.

jamshale (approved):

jamshale:
There is missing coverage on anoncreds_tails_server.py, but there wasn't any to begin with. The logic is limited enough that I think we can just go ahead and merge it. We could add a small unit test for it, though. I'll leave it in approved state for now and you can decide whether to add a test.

swcurran:
Maintainers agree that the Sonar report is a false negative -- or at least not a roadblock for this PR. Merging anyway.

PR #3861, opened 2025-08-25
:recycle: Refactor and modularize anoncreds revocation package

ff137:
Currently, there are two main places where anoncreds revocation functionality is maintained:
- `acapy_agent.anoncreds.revocation` module
- `acapy_agent.revocation_anoncreds` package 

The `revocation_anoncreds` package uses several functions from the `anoncreds.revocation` module. It also has its own routes module, mixing in additional revocation endpoints, on top of what is defined in `acapy_agent.anoncreds.routes`.

Overall, this makes things very confusing. Revocation functionality is already complex, and with modules holding 1000+ lines of code, it becomes very unwieldy; both in terms of understanding what's going on, and in trying to improve on what exists and make it more maintainable.

So, this PR is a step in that direction of making things more maintainable.

Two main parts to this refactoring:
- The `revocation_anoncreds` package gets merged into `acapy_agent.anoncreds.revocation`. 
- The `acapy_agent.anoncreds.routes` module is unpacked and modularized into subpackages for each of the primary components: schemas, cred defs, and revocation.
  - Revocation is then modularized further, into packages for: credentials (revoking, publishing revocations); lists (managing revocation lists); registry (rev-reg-def and registry management); and tails specific functions.

The `revocation_anoncreds` package is preserved, with all the same methods being imported, in case there are plugins that use the directory. A deprecation warning is added to notify that the code is moved to `acapy_agent.anoncreds.revocation`.

___

This is a precursor PR for the event-driven rev-reg management PR (#3831), which will benefit from first organizing all the code so that the large-scale changes there are easier to review.

___

Note to reviewers: The bulk of the code changes are exact copies of what previously existed. 

Summary of code reorganization:

Before:
- `acapy_agent.revocation_anoncreds` package
- `acapy_agent.anoncreds.{revocation|revocation_setup}` modules

Now merged into:
- `acapy_agent.anoncreds.revocation.*`
___
Routes, before:
- `acapy_agent.anoncreds.{routes|tests}`
- `acapy_agent.revocation_anoncreds.{routes|tests}`

Now:
- `acapy_agent.anoncreds.routes.{cred_defs|**revocation**|schemas}.{tests|models|routes}`
  - with `..**revocation**.{credentials|lists|registry|tails}..`

In other words -- the routes modules are broken up by category, and revocation gets more subcategories. 
Request + response models  now live in models.py within each subpackage, along with the bespoke tests module.

It's indeed a lot of code to review -- but it's mostly exactly what existed before. So, the main thing is to confirm that the names are consistent -- correct classes in correct modules -- and that the re-organization of endpoints makes sense.

The thinking with "revocation.credentials" is that it's supposed to mainly have endpoints related to revoking credentials, so, revoking and publishing revocation. It also includes fetching cred rev records. Feel free to suggest / recommend other organization patterns. While we're refactoring, may as well measure twice, cut once.


ff137:
Tests before: `4937 passed, 49 skipped, 3 xfailed, 1 warning in 128.04s (0:02:08)`
Tests after:    `4965 passed, 49 skipped, 3 xfailed, 1 warning in 129.13s (0:02:09)`

After manual review, I'm pretty sure no tests went missing.

ff137:
Side notes;
- The reorganization also means that code deduplication could be prioritised (just barely managed to hit 2.9%, under SonarCloud's 3% threshold). So, there's now `routes.common` that holds utils, as well mixins for common model/schema patterns, and common test fixtures. Hopefully makes maintaining the wider set of files easier.

- I realised that revocation_setupy.py holds the `register_events`, to register event patterns on the event bus, for the automated rev-reg management. This was previously in routes.py, which gets automatically picked up by the `PluginRegistry.register_protocol_events()` function. So, I've preserved a routes.py file in the anoncreds.revocation package (just as a shim; has no endpoints), to preserve this automated wiring-up for `register_events`. Perhaps there's a better way to do this -- call register events more explicitly, instead of relying on a "routes" file being picked up.

- Still to be refactored: revocation.py (1551 LOC). I want to follow a similar approach of splitting up existing methods by methods related to revocation registries, lists, tails, credential_helper, etc. But there are already a lot of changes in this PR, so maybe best to do that separately.

ff137:
Doing a quick review again, there's some duplicated tests that I just want to clean up

Maybe also some technical debt that's worth cleaning up.
e.g. `@pytest.mark.xfail(reason="This test never worked but was skipped due to a bug")`

I'll review some of these cases and see if it's best to fix here, or in a separate PR

ff137:
Note: there's a route path inconsistency for the anoncreds tails routes:
```py
async def register(app: web.Application) -> None:
    """Register routes."""

    app.add_routes(
        [
            web.put("/anoncreds/registry/{rev_reg_id}/tails-file", upload_tails_file),
            web.get(
                "/anoncreds/revocation/registry/{rev_reg_id}/tails-file",
                get_tails_file,
                allow_head=False,
            ),
        ]
    )

```

i.e.:
`PUT "/anoncreds/registry/{rev_reg_id}/tails-file"`
vs
`GET "/anoncreds/revocation/registry/{rev_reg_id}/tails-file"`

___
Judging from old routes in acapy_agent.revocation.routes, the `upload_tails_file` method here has the wrong route, and it should be:
`PUT "/anoncreds/revocation/registry/{rev_reg_id}/tails-file"`

@jamshale I propose adding the new route above, and marking the `PUT "/anoncreds/registry/{rev_reg_id}/tails-file"` route as deprecated. Let me know what you think

jamshale:
> @jamshale I propose adding the new route above, and marking the `PUT "/anoncreds/registry/{rev_reg_id}/tails-file"` route as deprecated. Let me know what you think

Yes. Sounds good. Let's get this changed.

ff137:
> Yes. Sounds good. Let's get this changed.

This is how I opted to do it: https://github.com/openwallet-foundation/acapy/pull/3861/commits/6096e757b907b26eb7627d4ae0b8e607b83c0e57

It looked like the `web.put` call doesn't take a deprecated argument, so I added an alias for the method, with flag added there. Might be different way to do it, but seemed good enough

swcurran:
@ff137 -- ready to merge?  I'll leave it to you to push the button!

PR #3881, opened 2025-09-16
fix: Repair Interop tests url

jamshale:
The test.bcovrin.vonx.io ledger used for the interop tests got changed to be https with a re-route for http requests. This broke the AATH test project. Easiest way to fix them is just changing the endpoint configuration.


swcurran (approved):
@WadeBarnes -- can we make it that http still works to reduce the impact of the BCovrin Test change? Or is it not worth it?

WadeBarnes:
From a chat with @jamshale the underlying issue was a `curl` statement that was not using the follow redirects option `curl -L`.  Technically `http` is still supported, it redirects automatically to `https`, though some tools like `curl` will not follow a redirect unless explicitly told.  For example, click on this `http` link to the genesis file; http://test.bcovrin.vonx.io/genesis

swcurran:
Good stuff -- thanks.

PR #3892, opened 2025-10-03
(fix) Properly use VM key when signing [SD-]JWT

### LINKED ISSUE #3891: 🐛 Wrong [SD-]JWT signer used when using non-default DID VerificationMethod
Basically the same problem as here: https://github.com/openwallet-foundation/acapy/issues/3639

The `jwt_sign` logic does not actually use the key associated with the selected `verification_method`, instead it uses whatever `verkey` is associated with the DID - which in some cases (e.g. did:cheqd) can be a completely different key to the key of the selected VM.

bug is here: https://github.com/openwallet-foundation/acapy/blob/9144f76b41ea2f26ba73b73fa2e1cd45c3ba7c1f/acapy_agent/wallet/jwt.py#L85,

the signature should be performed with the key of the verification_method

gmulhearn:
Closes #3891 

Utilizes the existing MultikeyManager logic to reduce code (like is done in the w3c ldp module)

### DIFF
--- a/acapy_agent/wallet/jwt.py
+++ b/acapy_agent/wallet/jwt.py
@@ -2,21 +2,17 @@
 
 import json
 import logging
-from typing import Any, Mapping, Optional, Tuple
+from typing import Any, Mapping, Optional
 
 from marshmallow import fields
-from pydid import DIDUrl, Resource, VerificationMethod
-from pydid.verification_method import Ed25519VerificationKey2018, Multikey
 
-from acapy_agent.wallet.keys.manager import key_type_from_multikey, multikey_to_verkey
+from acapy_agent.wallet.keys.manager import MultikeyManager, key_type_from_multikey, multikey_to_verkey
 
 from ..core.profile import Profile
-from ..messaging.jsonld.error import BadJWSHeaderError, InvalidVerificationMethod
+from ..messaging.jsonld.error import BadJWSHeaderError
 from ..messaging.models.base import BaseModel, BaseModelSchema
-from ..resolver.did_resolver import DIDResolver
 from .base import BaseWallet
 from .default_verification_key_strategy import BaseVerificationKeyStrategy
-from .key_type import ED25519, KeyType
 from .util import b64_to_bytes, bytes_to_b64
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


swcurran:
@PatStLouis @jamshale -- please take a look when you can.  @gmulhearn -- @jamshale is on vacation today, so won't see this until next week.

acapy_agent/wallet/jwt.py
```diff
@@ -64,19 +64,18 @@ async def jwt_sign(
...
-        # We look up keys by did for now
-        did = DIDUrl.parse(verification_method).did
-        if not did:
-            raise ValueError("DID URL must be absolute")
 
     async with profile.session() as session:
         wallet = session.inject(BaseWallet)
-        did_info = await wallet.get_local_did(did_lookup_name(did))
+        key_manager = MultikeyManager(session)
+        key_info = await key_manager.resolve_and_bind_kid(verification_method)
```
PatStLouis, line 71: I don't think this signing function should do any binding, binding should happen before this call. Any reason why you opted for that method?
```suggestion
        key_info = await key_manager.from_kid(verification_method)
```

PatStLouis (changes requested):
Thanks for looking into this, its been on my radar for some time.

acapy_agent/wallet/jwt.py
gmulhearn: was trying to copy the logic of the MR I did for fixing this problem when signing LDPs: https://github.com/openwallet-foundation/acapy/pull/3641/files#diff-fa67467fa6a637a46e04a1791349440d73f9aefd9fc476e86539b878d3609c6bR161 .

I somewhat remember us having conversations around this, but the rough idea was that rather than assume the consumer has correctly registered their kid in the multikey manager prior to calling this function, we can just idempotent-ly bind it for them (`resolve_and_bind_kid`).

e.g. if i create a did:key, and then try to sign with it `did:key:z123#z123` verificationMethod, `from_kid` doesn't know that kid exists

gmulhearn: switching to `from_kid` would be a breaking change in behaviour for the consumer in this sense

PatStLouis: Yeah I remember now. I'm ok with leaving it I think the DID key example is valid.

PatStLouis (approved):

jamshale:
I'm going to wait to merge this. We're really trying to get the Kanon profile stuff in and I don't want to cause any potential merge conflicts withit.

gmulhearn:
@jamshale sounds good! let me know when it's ready and i'll rebase this branch

jamshale:
@gmulhearn Can you please update this with main now and we will get it merged? Thx.

jamshale:
@gmulhearn Sorry. I merged another PR and github isn't letting me update the branch. Could you do it one more time please?

Issue #3894, opened 2025-10-03
Indy proof request returns credentials even when predicate is not satisfied


nb-vivek-bodar:
### Description:

- When a Verifier sends a proof request containing a predicate that the Holder's credential does not satisfy, the Holder agent still returns that credential in the **GET /present-proof-2.0/records/{pres_ex_id}/credentials** API call. This leads to confusion, as the expectation is that only matching credentials (those satisfying all restrictions, including predicates) should be returned.

### Steps to Reproduce:

- **Preconditions:**

1. ACA-Py instance running in Holder role.
2. Holder already has a stored Indy credential in their wallet as shown below
3. **GET /credentials** call should return Indy credential
 ```json
 {
  "results": [
    {
      "referent": "253e2370-284c-4254-9bee-be80b96a51b2",
      "schema_id": "9ZUiWwsNkk2iZZu67BRu3K:2:Driving_license:1.0",
      "cred_def_id": "9ZUiWwsNkk2iZZu67BRu3K:3:CL:2926088:Driving_License",
      "rev_reg_id": "9ZUiWwsNkk2iZZu67BRu3K:4:9ZUiWwsNkk2iZZu67BRu3K:3:CL:2926088:Driving_License:CL_ACCUM:5c2950a7-60ec-49c2-af3a-df8761dd8cec",
      "cred_rev_id": "1",
      "attrs": {
        "dob_dateint": "19900101",
        "first_name": "John",
        "last_name": "Doe"
      }
    }
  ]
}
```

**1. Verifier sends proof request (via POST /present-proof-2.0/send-request) with:**

- Requested attributes: _first_name_, _last_name_
- Predicate: _dob_dateint >= 20000101_ (which is **not** satisfied, as the stored value is 19900101)

```json
{
  "auto_remove": false,
  "auto_verify": false,
  "comment": "string",
  "connection_id": "66c0622e-b33b-450c-9e85-0d6fb7055331",
  "presentation_request": {
    "indy": {
      "name": "Proof request",
      "non_revoked": {
        "from": 1640995199,
        "to": 1640995199
      },
      "nonce": "1",
      "requested_attributes": {
        "additionalProp1": {
          "names": [
            "first_name",
            "last_name"
          ],
          "non_revoked": {
            "from": 1640995199,
            "to": 1640995199
          },
          "restrictions": [
            {
              "cred_def_id": "9ZUiWwsNkk2iZZu67BRu3K:3:CL:2926088:Driving_License"
[...]

PR #3895, opened 2025-10-06
:construction: Apply predicate to presentation request's matching credentials

ff137:
Work in progress.

Relates to #3894


ff137:
@swcurran Somebody else should pick this one up. They can set our fork as an available upstream in their local git and use the commits.

Previous state was: "mystery", WQL results are opposite to what I thought they would be (inverted logic).
Not sure what was going on (maybe it was PEBKAC, maybe there's an issue in askar) and I don't have capacity to investigate.

PR #3912, opened 2025-10-14
1.4.0rc0

swcurran:
Release Candidate ready.

Thoughts/questions:

- Why is their a new `deprecated=true` on an AnonCreds Revocation route in the OpenAPI spec?  I'm sure there is a reason -- I'm just curious.
- Are there any PRs that should be cherry-picked into 1.3.2 for an LTS release?
- Is it time we marked the `askar` wallet type as deprecated? If so, we can do that in the next RC/final release.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.4.0rc0
 
-### October 14, 2025
+### October 15, 2025
 
 ACA-Py 1.4.0 delivers a major internal upgrade centered on the introduction of **Kanon Storage**, a new modular storage architecture that separates cryptographic key management from general data persistence. Kanon moves ACA-Py’s non-key data (connections, credentials, protocol records, etc.) out of the encrypted Askar wallet into a dedicated, database-native storage layer. Askar now functions purely as a **Key Management Service (KMS)**, responsible for secure creation and use of keys and secrets. This shift enables ACA-Py deployments to leverage the full capabilities of their database engines—better indexing, analytics, and scalability—while preserving strong security boundaries around key material.
 
--- a/docs/features/SupportedRFCs.md
+++ b/docs/features/SupportedRFCs.md
@@ -8,7 +8,7 @@ ACA-Py or the repository `main` branch. Reminders (and PRs!) to update this page
 welcome! If you have any questions, please contact us on the #aries channel on
 [OpenWallet Foundation Discord](https://discord.gg/openwallet-foundation) or through an issue in this repo.
 
-**Last Update**: 2025-10-14, Release 1.4.0rc0
+**Last Update**: 2025-10-15, Release 1.4.0rc0
 
 > The checklist version of this document was created as a joint effort
 > between [Northern Block](https://northernblock.io/), [Animo Solutions](https://animo.id/) and the Ontario government, on behalf of the Ontario government.


ff137:
> Why is their a new deprecated=true on an AnonCreds Revocation route in the OpenAPI spec? I'm sure there is a reason -- I'm just curious.

The deprecated route is for uploading a tails file: `PUT /anoncreds/registry/{rev_reg_id}/tails-file`

It does not follow the same naming pattern -- all other revocation-related functions are under `/anoncreds/revocation/`. So the old path is marked as deprecated, and same function was added at `PUT /anoncreds/revocation/registry/{rev_reg_id}/tails-file`

ff137:
Ah ... I notice one other route is on `/anoncreds/registry`: 
`PUT /anoncreds/registry/{rev_reg_id}/active` (set_active_registry)

For consistency, that one should probably also be marked as deprecated, and new one added at:
`/anoncreds/revocation/registry/{rev_reg_id}/active`

PS: There are 7 other routes on `/anoncreds/revocation/registry`, and only those two on `/anoncreds/registry`. Which it why it seemed to be an error. Marked the other one deprecated with the recent refactoring, but I missed this other one

ff137:
@swcurran I made a PR to move the one other route, for consistency: https://github.com/openwallet-foundation/acapy/pull/3915

So, the only additions I would make to the release notes is under deprecations, just noting those two endpoints.
And then that the `acapy_agent.revocation_anoncreds` package is marked as deprecated, and moved to `acapy_agent.anoncreds.revocation`. More of a developer note / should only impact plugins, so probably optional to note that.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "acapy_agent"
-version = "1.3.2"
+version = "1.4.0"
```
jamshale: This should be 1.4.0rc0. Otherwise merging this will run the publish workflows and take the 1.4.0 version in pypi.

swcurran: ARRRRGGGGHHHH -- that's a big screw up on my part. Thanks for catching that. I'm usually better than that!

swcurran:
Corrections made to release ID and deprecations section.

jamshale (dismissed):

jamshale:
Snuck in https://github.com/openwallet-foundation/acapy/pull/3915

swcurran:
Let's try again...

jamshale:
> Regarding Kanon: is any documentation focused on _migrating_ from askar to using Kanon required, or is it all automagical? While not essential right now, I wonder if we should consider switching Kanon to being the default for non-key data and therefore provide a good migration path for existing deployments.

There is no migration path currently. It has been discussed before and @dave-promulgare said they'd likely be developing one for their existing deployments.

jamshale:
> The changelog mentions the `debug-webhooks`: wasn't that a change that happened in `1.3.2` already?

This was just to fix them demo. It was broken because it also used --debug-webhooks functionality. I'm not exactly sure why.

jamshale (dismissed):

swcurran:
Another tweak -- updated the date to today instead of yesterday...

swcurran:
No Kanon docs yet -- as noted in the ChangeLog.  Agree that it should have migration guidance as well.  Right @dave-promulgare ?

PR #3913, opened 2025-10-15
:art: Fix and simplify AnonCreds-backend checks

ff137:
AnonCredsRevocation was missing the proper profile check after the Kanon changes. 

Additionally simplifies unnecessary checks introduced to AnonCredsIssuer and AnonCredsHolder.

Adds an `is_anoncreds` property to the Profile, so it's no longer necessary to check profile instance types.

Lastly, the profile is now only validated once (previously validates on every self.profile use), following the same lazy validation pattern (AnonCreds classes can be instantiated using wrong profile, and will only error once used).


ff137:
Lazy validation can be replaced by validating on class init, but I'm just keeping the existing pattern.

ff137:
Includes a bug fix for `KanonIndyCredxIssuer`. The last methods in the class were incorrectly added to `IndyIssuerRetryableError`

jamshale (approved):
Looks good 👍

PR #3914, opened 2025-10-15
:sparkles: Implement ProfileSessionHandle

ff137:
Abstract interface for profile sessions handles, which contains abstract methods for the functionality that is common to both Askar and Kanon storage interfaces . Assists with type safety and auto-complete. 

(Net result is that all the e.g. `session.handle.fetch` methods now actually resolve in the editor and show correct return types.)


jamshale (approved):
Nice work! We should get this in for the official release.

PR #3915, opened 2025-10-15
:art: Move AnonCreds `set_active_registry` route

ff137:
Marks the old route as deprecated (`PUT /anoncreds/registry/{rev_reg_id}/active`) and adds new one to follow the correct naming pattern (`PUT /anoncreds/revocation/registry/{rev_reg_id}/active`


jamshale:
Let's sneak this in before the release 👍

PR #3922, opened 2025-10-23
fix(kanon):fixed password bug and tests for kanon postgres

vinaysingh8866:


jamshale:
We should really add a scenario integration test doing some basic issuance and verification with sqlite and postgres. It can be done later.

PR #3923, opened 2025-10-24
feat: Remove issuance v1 protocols

jamshale:
This removes the v1 issuance protocols in favor of a plugin. See https://github.com/openwallet-foundation/acapy/pull/3252 for original work. 

The plugin PR https://github.com/openwallet-foundation/acapy-plugins/pull/2288 uses this to re-enable v1 issuance support.

When this is finished and successful the same method will be used for the presentation v1 protocols.

### DIFF
--- a/demo/runners/agent_container.py
+++ b/demo/runners/agent_container.py
@@ -188,72 +188,6 @@ class AriesAgent(DemoAgent):
                         params={"transaction_my_job": connection_job_role},
                     )
 
-    async def handle_issue_credential(self, message):
-        state = message.get("state")
-        credential_exchange_id = message["credential_exchange_id"]
-        prev_state = self.cred_state.get(credential_exchange_id)
-        if prev_state == state:
-            return  # ignore
-        self.cred_state[credential_exchange_id] = state
-
-        self.log(
-            "Credential: state = {}, credential_exchange_id = {}".format(
-                state,
-                credential_exchange_id,
-            )
-        )
-
-        if state == "offer_received":
-            log_status("#15 After receiving credential offer, send credential request")
-            await self.admin_POST(
-                f"/issue-credential/records/{credential_exchange_id}/send-request"
-            )
-
-        elif state == "credential_acked":
-            cred_id = message["credential_id"]
-            self.log(f"Stored credential {cred_id} in wallet")
-            log_status(f"#18.1 Stored credential {cred_id} in wallet")
-            resp = await self.admin_GET(f"/credential/{cred_id}")
-            log_json(resp, label="Credential details:")
-            log_json(
-                message["credential_request_metadata"],
-                label="Credential request metadata:",
-            )
-            self.log("credential_id", message["credential_id"])
-            self.log("credential_definition_id", message["credential_definition_id"])
-            self.log("schema_id", message["schema_id"])
-
-        elif state == "request_received":
-            log_status("#17 Issue credential to X")
-            # issue credentials based on the credential_definition_id
-            cred_attrs = self.cred_attrs[message["credential_definition_id"]]
-            cred_preview = {
-                "@type": CRED_PREVIEW_TYPE,
-                "attributes": [{"name": n, "value": v} for (n, v) in cred_attrs.items()],
-            }
-            try:
-                cred_ex_rec = await self.admin_POST(
-                    f"/issue-credential/records/{credential_exchange_id}/issue",
-                    {
-                        "comment": (
-                            f"Issuing credential, exchange {credential_exchange_id}"
-                        ),
-                        "credential_preview": cred_preview,
-                    },
-                )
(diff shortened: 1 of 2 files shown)


jamshale:
The interop tests are expected to fail without the plugin installed. 

Here is a branch using this PR and installing the plugin. https://github.com/openwallet-foundation/owl-agent-test-harness/pull/919

The tests are passing with this branch and the plugin installed.

acapy_agent/core/oob_processor.py
```diff
@@ -301,10 +300,10 @@ async def handle_message(
     ):
         """Message handler for inbound messages."""
         supported_types = [
-            CREDENTIAL_OFFER,
             CRED_20_OFFER,
             PRESENTATION_REQUEST,
             PRES_20_REQUEST,
+            "issue-credential/1.0/offer-credential",
```
jamshale, line 306: This should be the only reference for `issue-credential/1.0`. I wasn't able to think of a way to inject it into this class without maybe an ugly monkey patch of the function in the plugin.

acapy_agent/revocation/manager.py
```diff
@@ -467,53 +463,17 @@ async def set_cred_revoked_state(
...
                     )
-                    cred_ex_id = rev_rec.cred_ex_id
-                    cred_ex_version = rev_rec.cred_ex_version
                     rev_rec.state = IssuerCredRevRecord.STATE_REVOKED
                     await rev_rec.save(txn, reason="revoke credential")
                     await txn.commit()
             except StorageNotFoundError:
                 continue
 
-            async with self._profile.transaction() as txn:
```
jamshale, line 485: This is a main change. The revocation is done via event handler in the routes file now.

acapy_agent/protocols/issue_credential/v2_0/routes.py
```diff
@@ -1907,3 +1910,35 @@ def post_process_routes(app: web.Application):
...
         }
     )
+
+
+def register_events(bus: EventBus):
+    """Register event listeners."""
+    bus.subscribe(re.compile(r"^acapy::cred-revoked$"), cred_revoked)
+
+
+async def cred_revoked(profile: Profile, event: EventWithMetadata):
```
jamshale, line 1920: This is where the revocation is handled now for the cred_ex record. The plugin has the same thing for V1 IssuerCredRevRecords's.

acapy_agent/protocols/present_proof/v1_0/tests/test_manager.py
```diff
@@ -1319,7 +1316,7 @@ async def test_receive_problem_report(self):
             )
             save_ex.assert_called_once()
 
-            assert ret_exchange.state == V10CredentialExchange.STATE_ABANDONED
+            assert ret_exchange.state == "abandoned"
```
esune, line 1319: There is no enum to use for v2 states?

esune (approved):
Looking good, thank you @jamshale for cleaning this up 👍🏻 
It will be nice to start shedding some old code 🙂

acapy_agent/protocols/present_proof/v1_0/tests/test_manager.py
jamshale: I think I could have used the v2 enum which is probably also `abandoned`. But, this is specifically for a v1 issuance so it was better just to use a string with the state.

These tests are being pulled in the subsequent `present-proof` removal anyway.

jamshale:
I'd like to get approval on the plugin PR before merging this. https://github.com/openwallet-foundation/acapy-plugins/pull/2288.

esune:
> I'd like to get approval on the plugin PR before merging this. [openwallet-foundation/acapy-plugins#2288](https://github.com/openwallet-foundation/acapy-plugins/pull/2288).

On it, going through it currently

PatStLouis:
@jamshale don't we want to remove the associated bdd tests and have those live in the plugin repo instead?

jamshale:
> @jamshale don't we want to remove the associated bdd tests and have those live in the plugin repo instead?

You're right. I forgot to remove the BDD tests in the repo. I'll do that. They only run on nightly and release runs.

We have a basic test for coverage in the plugin PR and a few tests in the OATH project which we can keep. No need to remove those.

jamshale:
So I'm pretty sure all the v1 protocol bdd tests were already removed from the acapy repo. I think this is good to go.

jamshale:
> The failing integration tests are the Kanon ones that I believe we're dragging along currently. It however looks like some BDD tests are failing as well - they are related to the issue-credential steps so will need to get checked.
> 
> Code changes were already approved 👍🏻

The BDD Interop tests are the test harness repo tests. They are expected to fail until the plugin is installed. I have tested OATH with this branch and the plugin installed and the tests all pass. I'll update it after merging this and the plugin.

PR #3927, opened 2025-10-27
Enable remote config

PatStLouis:
Adds a new start up arg to load a remote config file.

### DIFF
--- a/acapy_agent/config/argparse.py
+++ b/acapy_agent/config/argparse.py
@@ -89,6 +89,20 @@ def fetch_remote_config(url: str, timeout: int = 30) -> str:
         ) from e
 
 
+def _is_url(file_path: str) -> bool:
+    """Check if a file path is actually a URL.
+
+    Args:
+        file_path: Path to check
+
+    Returns:
+        True if the path is a URL, False otherwise
+
+    """
+    parsed = urlparse(file_path)
+    return bool(parsed.scheme and parsed.netloc)
+
+
 def preprocess_args_for_remote_config(argv: list) -> list:
     """Preprocess argv to handle --arg-file with URL support.
 
@@ -119,10 +133,7 @@ def preprocess_args_for_remote_config(argv: list) -> list:
                 continue
 
             file_path = processed_argv[i + 1]
-            # Check if it's a URL
-            parsed = urlparse(file_path)
-            if parsed.scheme and parsed.netloc:
-                # It's a URL, download it
+            if _is_url(file_path):
                 temp_file = fetch_remote_config(file_path)
                 processed_argv[i + 1] = temp_file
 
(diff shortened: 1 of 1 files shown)


PatStLouis:
@jamshale any ideas why the pr says 0% test coverage? There's 8 tests added that cover this functionality

jamshale:
There should be coverage. Not sure why it's reporting it's missing. Let's see if it updates after rebasing.

jamshale:
I'm trying to think if this option should go into any documentation...

I am reluctant to add more startup configs whenever possible but I think this is useful enough to add.

The small issues raised by sonarcloud are easy enough to fix. https://sonarcloud.io/project/issues?id=openwallet-foundation_acapy&pullRequest=3927&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

esune:
> I'm trying to think if this option should go into any documentation...
> 
> I am reluctant to add more startup configs whenever possible but I think this is useful enough to add.

Could we, instead of adding ANOTHER startup parameter, update the existing argfile parameter to correctly handle URLs rather than file paths?

PatStLouis:
@jamshale I will look at the documentation and find somewhere to add this

PatStLouis:
@esune yes this was my original plan, I can go back to this idea

PatStLouis:
@jamshale there was actually no documentation for the arg file usage so I added some

@esune the --arg-file can now take a local path or a remote url

acapy_agent/config/argparse.py
```diff
@@ -28,6 +32,119 @@
...
+            f"Remote configuration from {url} is not valid YAML: {e}"
+        ) from e
+
+    # Save to temporary file
+    try:
+        fd, temp_path = tempfile.mkstemp(suffix=".yml", prefix="acapy_remote_config_")
+        try:
+            os.write(fd, text.encode("utf-8"))
+        finally:
+            os.close(fd)
```
esune, line 82: Wondering if this should be saved in the shared directory mounted as `home` so multiple pods can access the file, rather than risk having pods running with different file versions at the same time?

PatStLouis: I'm not sure I understand, isn't it the same?

esune: If a pod starts and fetches the configuration, then a second pod starts and fetches the configuration again - but it has changed - we could have two pods for the same service using different configurations. Maybe it's a bit of an edge case, but I believe it would be possible.

We are not specifying a persisted directory to create the file in, so I *think* it will be created by default in the temporary directory specified by the OS (likely `/tmp`) therefore the file won't be accessible by pods starting up after the first one pulled the file.

I am however not sure how we would deal with persisting the file so it can be loaded/shared by more than one pod, while at the same time having the file updated when necessary. Maybe this should just be a note in the docs?

esune: Giving it some more thought, this could potentially also happen when loading the configuration from a configmap/file that is updated between the start of pod 1 and 2 so we are actually not introducing a different risk/behaviour.

PR #3928, opened 2025-10-27
Add skip verification option for credential storage

PatStLouis:
Supersedes #3824

Enables a flag to store an unverified credential (unsecured)

The purpose is to help developers implement new cryptosuites and data integrity proof types which may be unknown to the agent.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -24,9 +24,9 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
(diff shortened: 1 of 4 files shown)


jamshale (approved):
This looks fine to me as long as it's used correctly.

PatStLouis:
@jamshale it should be treated as an unsecured option for development and testing. From a wallet perspective, its recommended to verify all proofs attached to a credential as part of the storing operation, and stripping proof types the wallet is unfamiliar with. However, it might be useful to bypass this verification when working with new crypto suites or making proof of concepts.

PR #3934, opened 2025-11-05
feat: Add kanon profile sqlite issuance/presentation/revocation scena…

jamshale:
This adds the first scenario test covering the most important scenario with an sqlite database.

TODO: Add postgres database wallet agent and run the same scenario.


PatStLouis (approved):

jamshale:
I'm going to merge this and add the postgres as a subsequent PR.

jamshale:
Looking into why these are being intermittent...

jamshale:
I'm not sure what's going on with this. It passed once and now it's having issues creating the public did consistently. Works consistently when I run it locally.

Will leave this for now and try and figure out what's going on with github actions.

jamshale:
Pretty weird. Tests seem to be passing now.

I am a bit concerned there's a co-routine locking issue happening with the kanon storage. I don't think it should block releasing it but it may be something to look into. I have seen http requests randomly getting stuck when I run these tests locally as well.

swcurran (approved):
Intermittent issue sounds scary, but good to see these tests passing. Perhaps open an issue so we don't lose it?

PR #3943, opened 2025-11-10
Disable kanon profile scenario tests

jamshale:
This comments out the actual tests for the kanon profile issuance/presentation/revocation scenario. They have been intermittent due to a seemingly locking problem where the aiohttp requests hang and timeout when creating storage records. There seems to be no difference between using postgres and sqlite storage.

I'll create an issue with more details and this can be re-enabled when the problem is resolved.


jamshale:
Sonarcloud just doesn't like commented code. We can ignore it.

PR #3950, opened 2025-11-13
Timeout issue fix

vinaysingh8866:
# Fix intermittent timeout issue in Kanon PostgreSQL normalized storage

## Problem

The Kanon issuance and presentation scenario test was experiencing intermittent `asyncio.TimeoutError` failures, especially when:
- Running in GitHub Actions
- Running in low-resource Docker containers
- Making rapid API calls that create many storage records quickly

The test would hang and timeout randomly, but the same test passed consistently with Askar storage.

## Root Cause

The PostgreSQL normalized storage had **redundant transaction management** happening at two layers:

1. **Handler layer** (`generic_handler.py`) - Every method had `try/except/finally` blocks that committed/rolled back transactions
2. **Session layer** (`session.py`) - Also managing transactions and commits

This caused:
- **Double commits** for every database operation
- **Longer connection hold times** due to unnecessary transaction overhead
- **Connection pool exhaustion** when making rapid API calls
- **Race conditions** between the two layers trying to manage the same transaction

The connection pool (max 100 connections, 60s timeout) would get exhausted under load, causing the aiohttp client to timeout waiting for connections.

## Solution

Removed all transaction management from the handler layer. Handlers now only execute SQL queries and return results - they don't manage transactions.

Transaction management now happens only at the session layer, which properly:
- Commits after each operation (in non-transaction mode)
- Commits on success / rolls back on error (in transaction mode)
- Returns connections to the pool cleanly

### DIFF
--- a/acapy_agent/database_manager/databases/postgresql_normalized/handlers/generic_handler.py
+++ b/acapy_agent/database_manager/databases/postgresql_normalized/handlers/generic_handler.py
@@ -535,8 +535,11 @@ class GenericHandler(BaseHandler):
                         tag_value = json.loads(tag_value)
                     except json.JSONDecodeError:
                         # If tag_value is not valid JSON, leave it as the original string.
-                        LOGGER.warning("[%s] Failed to decode tag_value as JSON: %r", 
-                                       operation_name, tag_value)
+                        LOGGER.warning(
+                            "[%s] Failed to decode tag_value as JSON: %r",
+                            operation_name,
+                            tag_value,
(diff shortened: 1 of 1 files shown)


jamshale:
@vinaysingh8866 This looks great. Could you address the 2 very small things that co-pilot picked up. Thx.

vinaysingh8866:
@jamshale removed the extra "the" from this file but it exists in all the examples might we worth to removing in all the files.

PR #3951, opened 2025-11-14
fix: removed redundant SET client_encoding calls

vinaysingh8866:
# Fix intermittent timeout issue - Remove redundant SET client_encoding calls

## Problem Still Persisted

After the first fix (removing redundant transaction management), the intermittent timeout issue still occurred. The test would pass sometimes but fail randomly with the same `asyncio.TimeoutError`.

## The Cause

The handlers were executing **`SET client_encoding = 'UTF8'`** on EVERY database operation, even though the connection pool already sets it once when retrieving connections.

### Why This Causes Timeouts:

- **Each `SET client_encoding` creates an implicit transaction** in PostgreSQL
- Under rapid API calls, these implicit transactions pile up
- The connection gets stuck in transaction states
- Connection pool exhaustion occurs
- Eventually → timeout after 60 seconds

## The Solution

Removed ALL `SET client_encoding` calls from the handlers since the connection pool already sets it once per connection.


jamshale:
Nice. Let's give it a try 👍

PR #3955, opened 2025-11-17
Feature: Add plugin installer

PatStLouis:
This PR adds an auto-install-plugin flag which will dynamically install enabled plugin.

Only works with official acapy-plugins, could be extended into a plugin-registry eventually where users can input the plugin source.

The new flag, `--auto-install-plugins`, can be used as a bool or also provide the version of the plugin to be installed.

ex:
```bash
aca-py start --auto-install-plugins # defaults to the current aca-py version
aca-py start --auto-install-plugins main
aca-py start --auto-install-plugins 1.3.2
```

### DIFF
--- a/acapy_agent/utils/plugin_installer.py
+++ b/acapy_agent/utils/plugin_installer.py
@@ -957,6 +957,90 @@ class PluginInstaller:
             )
             return False
 
+    def _check_and_handle_existing_plugin(
+        self,
+        plugin_name: str,
+        installed_package_version: Optional[str],
+        installed_source_version: Optional[str],
+        target_version: str,
+    ) -> bool:
+        """Check if existing plugin version matches and handle accordingly.
+
+        Args:
+            plugin_name: The name of the plugin module
+            installed_package_version: Installed package version
+            installed_source_version: Installed source version
+            target_version: Target version to match
+
+        Returns:
+            True if plugin is already correctly installed, False if needs installation
+
+        """
+        if not self.plugin_version:
+            # No explicit version - check package version match
+            if self._check_version_matches_implicit(
+                plugin_name, installed_package_version, target_version
+            ):
+                self.installed_plugins.add(plugin_name)
+                return True
+            LOGGER.info(
+                "Plugin '%s' exists but version check inconclusive. "
+                "Reinstalling to ensure correct version (%s)...",
+                plugin_name,
+                target_version,
+            )
+        else:
+            # Explicit version - check source version match
+            if self._check_version_matches_explicit(
+                plugin_name,
+                installed_source_version,
+                installed_package_version,
+                target_version,
+            ):
+                self.installed_plugins.add(plugin_name)
+                return True
+            self._log_version_mismatch(
+                plugin_name,
+                installed_source_version,
+                installed_package_version,
+                target_version,
+            )
+        return False
+
+    def _attempt_plugin_installation(
+        self, plugin_name: str, plugin_exists: bool, target_version: str
+    ) -> bool:
+        """Attempt to install or upgrade a plugin.
+
+        Args:
+            plugin_name: The name of the plugin module
+            plugin_exists: Whether plugin already exists
+            target_version: Target version
+
+        Returns:
+            True if installation succeeded, False otherwise
+
+        """
+        if not self.auto_install:
+            LOGGER.warning(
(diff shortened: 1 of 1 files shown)


swcurran:
Two questions on the syntax from the description above:

- you use both "`--auto-install-plugin`" and "`--auto-install-plugins`" -- which is right, or both?
- I would have expected the name of the plugin (or plugins?) to be included in the parameter -- not just the branch.  Or does this get used with other parameters?

Thanks

PatStLouis:
> Two questions on the syntax from the description above:
> 
> * you use both "`--auto-install-plugin`" and "`--auto-install-plugins`" -- which is right, or both?
> * I would have expected the name of the plugin (or plugins?) to be included in the parameter -- not just the branch.  Or does this get used with other parameters?
> 
> Thanks

there is only 1 flag, `--auto-install-plugins`, will fix the singular typing.

For the plugin list, it will look at the "enabled" plugins from the existing config. If no plugins configured, no plugins get installed. The main purpose is to avoid having to build a separate image with the plugin pre-installed.

acapy_agent/utils/plugin_installer.py
```diff
@@ -0,0 +1,420 @@
...
+        vcs_info = direct_url_data.get("vcs_info", {})
+        if vcs_info.get("vcs") == "git":
+            revision = vcs_info.get("requested_revision")
+            if revision and (
+                "." in revision or revision in ["main", "master", "develop"]
+            ):
+                return revision
+
+        url = direct_url_data.get("url", "")
+        if "@" in url and "github.com" in url:
```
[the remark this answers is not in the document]
PatStLouis: fixed

acapy_agent/commands/start.py
```diff
@@ -60,6 +62,48 @@ async def run_app(argv: Sequence[str] = None):
...
+
+        if auto_install:
+            version_info = (
+                f"version {plugin_version}"
+                if plugin_version
+                else f"current ACA-Py version ({acapy_version})"
+            )
+            # Always print to console for visibility
+            plugins_str = ", ".join(external_plugins)
+            print(
```
jamshale: Please remove print statement.

ff137:
Copilot comments are good 👌

acapy_agent/commands/start.py
```diff
@@ -60,6 +62,48 @@ async def run_app(argv: Sequence[str] = None):
...
+            )
+
+            failed_plugins = install_plugins_from_config(
+                plugin_names=external_plugins,
+                auto_install=auto_install,
+                plugin_version=plugin_version,
+            )
+
+            if failed_plugins:
+                LOGGER.error(
```
jamshale, line 90: It would be a bit nicer to just combine both of these into a single log statement with a new line if you were going for nicer formatting.

acapy_agent/commands/start.py
PatStLouis: print statement removed

acapy_agent/commands/start.py
PatStLouis: Both log statements have been combined

jamshale:
This is looking good. There's one new small issue https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3955&id=openwallet-foundation_acapy&open=AZq4C4FtGp2JzAxo9_zl that should be fixed.

Other than that I'll approve.

jamshale (approved):

PatStLouis:
@jamshale just the one kanon test failing again, otherwise I've addressed all SonarQube/copilot comments as well as yours!

jamshale:
I'll probably wait until after the maintainers meeting tomorrow and then disable the problematic test. Will give others a chance to review but I've approved it 👍

acapy_agent/admin/routes.py
```diff
@@ -12,6 +13,16 @@
...
 
+LOGGER = logging.getLogger(__name__)
+
+
+# Lazy import to avoid import-time issues
+def _get_plugin_version(plugin_name: str):
+    """Lazy import wrapper for get_plugin_version."""
+    from ..utils.plugin_installer import get_plugin_version
+
+    return get_plugin_version(plugin_name)
```
ff137: Is this still needed? I can't see why there would be an import issue, but maybe there's a cyclic import somewhere

PR #3958, opened 2025-11-21
fix: minor fix to avoid pool exhaustion and deadlocks

vinaysingh8866:


jamshale (approved):
I've been trying to replicate the issue consistently and haven't been able to really reproduce it at all. Also the scenario tests haven't failed in a couple days. This change does look similar to the others which has mostly alleviated the problem. 

I think we should just merge it and see if failures happen any more for a week or so.

It's a good thing the scenario tests found this problem. It was probably something which would have become a bigger problem if it got to a heavy use production deployment.

jamshale:
@swcurran I'm going to merge this. Maybe if it's stable for a week or so we can create a new minor release for it.

swcurran:
Are the tests reactivated?  Were you able to adjust the environment parameters to cause the error to be consistent? No problem is not, just wondering.

jamshale:
Yes, the tests are activated so we'll know if the problem happens again.

I tried restricting resources and running the tests locally as well as with my forks github actions runner. Was unable to get the problem to happen.

PR #3963, opened 2025-11-25
fix(kanon):updated connection cleanup to share 1 thread and added logging to detect connection leakage 

vinaysingh8866:


jamshale:
Although this doesn't seem to completely resolve the problem (failed on the 7th run in github actions), it looks like there's a lot of improvement here, especially around logging and helping to better determine the root of the problem when it occurs.

@vinaysingh8866 I think we should merge it, but I'm unable to update the branch.

vinaysingh8866:
@jamshale it seems the params on the docker file are incorrect. 
I have updated it another dev from our team is looking at the test configuration.

jamshale:
I'll merge this if we get it updated with main again. There's enough useful debugging and connection management improvements that it's worth it even if there's another source of the failures.

vinaysingh8866:
@jamshale we need to add logs to the docker container, when a test fails it should show all the logs of the container.

jamshale:
Ok. Lets merge this so it doesn't get behind and I can do that.

PR #3964, opened 2025-11-25
Disable kanon profile scenario test

jamshale:
Disables the intermittent kanon profile scenario tests.

Also cleaned up a duplicated grammar mistake in the comments.

### DIFF
--- a/scenarios/examples/anoncreds_issuance_and_revocation/example.py
+++ b/scenarios/examples/anoncreds_issuance_and_revocation/example.py
@@ -133,7 +133,7 @@ async def main():
             schema_name=schema_name,
         )
 
-        # Present the the credential's attributes
+        # Present the credential's attributes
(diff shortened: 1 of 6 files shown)


swcurran:
Is it possible to keep the test running, but accept the fail?  That way the test keeps running via GHA (were it seems to crop up more) and not just locally, and we can see if there is progress with PRs that target this issue.  My concern is that we won't learn enough if we are only testing potential fixes locally.

jamshale:
Yes, I can disable the restriction in gitbhub actions instead. That's a good idea.

jamshale:
Actually. The scenario tests aren't required to merge the PR. I'm going to close this. There's nothing to do. We can just make sure it's the kanon test failing for now and ignore it if it's unrelated.

FYI, the settings to change this are repo level and I don't have the permissions if we do want to make it required in the future. I think maybe just the unit tests are required atm.

PR #3983, opened 2025-12-12
feat: Add logging to scenario tests

jamshale:
Partly to help diagnose the kanon profile intermittent failures with github actions but also just help debugging failing scenario tests quicker. This logs output from the test docker containers.


jamshale:
Ugh. This ended up being way simpler than I was making it. Before cleanup, simply run the docker compose -f <file> logs command and it will get the logs of the failed test.

Issue #3986, opened 2025-12-15
[BUG] - Creating a cred def using the askar-anoncreds wallet and a sov did will return the wrong schemaId value


PatStLouis:
When I create a anoncreds schema and a cred def using an upgraded wallet + a did sov, the cred def record stores the wrong schemaId value.

It will store the sequence number as the schemaId:
```json
{
  "credential_definition_id": "ECG9NyqU3J7hpX1BrSPUpm:3:CL:3032693:test",
  "credential_definition": {
    "issuerId": "ECG9NyqU3J7hpX1BrSPUpm",
    "schemaId": "3032693",
    "type": "CL",
    "tag": "test"
}
```

I believe the schemaId should return the actual schemaId, not the seqNo.

swcurran:
Fix merged, test added.

PR #3988, opened 2025-12-16
Fetch schemaId on legacy cred def

PatStLouis:
When upgrading a wallet from askar to askar-anoncreds, a cred def created with a legacy did sov will return a seqNo instead of a schemaId. See #3986

This PR will resolve the schema from the seqNo and fetch the correct value.


jamshale:
I am wondering what happens if the controller provides the actual schema ID for lookup instead of the seq no. Maybe there's a switcharoo that can happen before fetching the cred def as well? I think the lookup is specific for using seq no (without looking at the code)

PatStLouis:
> I am wondering what happens if the controller provides the actual schema ID for lookup instead of the seq no. Maybe there's a switcharoo that can happen before fetching the cred def as well?

The controller already needs to provide the schemaId when doing a looking, providing the seqNo won't return any results.

This is about the GET `/anoncreds/credential-definition/{credDefId}` endpoint not returning the proper schemaId value. All other places seems to return / process the correct id value.

PR #3991, opened 2025-12-19
Fix issues with anoncreds upgrade

jamshale:
I don't know how both Ian and I missed this but I've found 2 issues so far. Luckily they aren't very complicated.

1. The rev_reg_def wasn't getting the issuer_id field set which cause serialization to fail when looking it up.
2. The upgrade process/script was deleting issuer_cred_rev objects when it shouldn't have. I think this was just a lack of knowledge on my part and thinking they weren't needed.

I'll continue looking for any other issues and add some scenario test coverage for these fixes.

These should be back-ported and have patch releases when they get merged.

### DIFF
--- a/scenarios/examples/restart_anoncreds_upgrade/example.py
+++ b/scenarios/examples/restart_anoncreds_upgrade/example.py
@@ -110,7 +110,7 @@ async def connect_agents_and_issue_credentials(
     )
     print(">>> Done!")
 
-    return (inviter_conn, invitee_conn)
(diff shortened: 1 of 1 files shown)


jamshale:
@PatStLouis A couple things I found. I think this should fix both of the revocation issues you were experiencing.

jamshale:
Interop tests will be repaired by https://github.com/openwallet-foundation/owl-agent-test-harness/pull/920

Issue #3845, opened 2025-07-30
:bug: When auto-storing a credential fails, the cred ex record remains in state: done


ff137:
I'm encountering an issue where I can reliably cause auto-storing of credentials to fail. (Related to revocation registry refactoring.)

The failure is not the issue here, but rather how acapy handles the auto-storing failure.

There are some auto-store failures that are unrecoverable (e.g. "Invalid state: Issuer is sending incorrect data", an error message that propagates from the `anoncreds-rs` and `anoncreds-clsignatures-rs` packages). Retrying to store this credential will continue to fail. Scenarios such as this should have the cred-ex-record never enter a `done` state (since receiving a webhook event indicating done would make one believe the credential was stored), and I believe it should instead go to an `abandoned` state.

Then there are other scenarios where auto-storing failed due to an error with the DB. These cases should in fact be retryable. Currently, it would not be possible for a holder to manually call the `/issue-credential-2.0/records/{cred_ex_id}/store` for this record, because current logic for `store_credential` requires the record to be in state: credential-received.

So there are a few issues here:
- an auto-store failure should not have the record transition to a "done" state, since that would usually indicate the credential is successfully stored in the holder's wallet.
- an auto-store failure that is unrecoverable should transition to `abandoned` (my proposal).
- an auto-store failure that _is_ recoverable should:
  - perhaps automatically be retried, up to 3-5 times with some delay (to account for DB errors);
  - and/or, remain in state `credential-received` so that it can be manually stored at a later time.

___
Notes:
- `V20CredIssueHandler` is the relevant handler that does the `if context.settings.get("debug.auto_store_credential")` check.
- This handler first calls `V20CredManager.receive_credential`, which ultimately calls `receive_credential` for the relevant `V20CredFormatHandler` (be it anoncreds, indy, ld_proof, vc_di).
  - All of those format handlers (apart from the ld_proof one) have a `receive_credential` method that looks like this:
  ```py
      async def receive_credential(
        self, cred_ex_record: V20CredExRecord, cred_issue_message: V20CredIssue
    ) -> None:
        """Receive anoncreds credential.

        Validation is done in the store credential step.
        """
   ```
  - That's it ... no actual code. The receive credential method leaves the validation to be done at the `store` step, meaning that errors are only caught once storage is attempted. That's how it was done for indy, and that's presumably how it was copy-pasted for anoncreds and vc_di.

[...]

PR #3999, opened 2025-12-24
Issue #3845 fix :  auto-store failure going to done state

sonivijayk:
## Issue Summary

1. Applied check to prevent auto-store failures from moving v2.0 credential exchanges to `done`
2. Added storage failures logic to retry by keeping state `credential-received` and saving `error_msg` 
3. Abandon unrecoverable auto-store failures and log proper message under exception
5. Updated test

## Testing
- python -m pytest acapy_agent/protocols/issue_credential/v2_0/handlers/tests/test_cred_issue_handler.py


jamshale:
There is some co-pilot suggestions. If you could go over them quickly and decide if they are worth addressing, please. The fix looks good to me.

PR #4005, opened 2025-12-30
feat: Only log failing scenarios

jamshale:
The logging for the scenario tests is quite extensive. It would be better for diagnosing the reason for failures if it only displayed the container logs of the failing scenario.


esune (approved):
Good idea, I did not think about over-logging problems, this is definitely welcome so we won't have to be digging in neverending logs

PR #4015, opened 2026-01-08
Fix plugin installer hardcoded version

PatStLouis:
Dynamically get the source default version instead of hard coding it.


swcurran:
@PatStLouis -- can you please update to the latest base so we can merge this?  Thanks!

PR #4019, opened 2026-01-12
DIF presentation debug logging

nodlesh:
### Summary

- Kept a compact verification summary at DEBUG and gated detailed verification results behind debug.presentations.
- Added a sanitizer (_summarize_doc_result) that omits full documents/proofs.
- Added a unit test to ensure sensitive fields are not logged.

### Key changes

- DEBUG‑level logs for create/search/verify for DIF presentation‑exchange
- Detailed verification results only when debug.presentations is enabled
- New test: test_summarize_doc_result_redacts_sensitive_fields

These logs are useful when debugging DIF verification (verified=false with no detail). This change keeps the signal while respecting sensitive data and existing debug.presentations behavior.

Detailed logs only emitted when --debug-presentations is enabled.


swcurran:
@nodlesh -- can you please address the Ruff / Lint checks?  Should be easy enough.

@jamshale -- are the PR tests a real issue?

PR #4028, opened 2026-01-22
Synchronously store the revocation registry

PatStLouis:
This is to address failing integration tests for revocation registry rotation:
https://github.com/openwallet-foundation/acapy/actions/runs/21226131058/job/61073454321?pr=4026

Some race conditions were causing the agent to try and set an active registry before it was stored locally.


swcurran:
@ff137 -- don't know if you have time to look at this, but I know you have done a lot of work on revocation registry creation handling and errors.  Your feedback would be appreciated.

PR #4030, opened 2026-01-23
Add aipv2 migration doc

PatStLouis:


loneil:
Actually @PatStLouis might be nice to have a Table of Contents since it's pretty long with lots of sections. (like in https://github.com/openwallet-foundation/acapy/tree/main/docs/demo)

PatStLouis:
@loneil agreed for the table of content, I will look into it. I also want to extend the credential management section about guidelines and suggestion for cases where you have existing issued credentials / credential exchanges.

swcurran:
On the https://aca-py.org site, the table of contents is free.

I'm thinking that on the site, we might have a separate top level category -- "Upgrading" and have all the articles that talk about that in that category.  Currently, they are (mostly?) under "Deploying", which is OK but not great.

Not to say a ToC is bad for when browsing the repo, but we should keep pushing those looking for docs to the aca-py.org site.

PR #4034, opened 2026-01-27
Gracefully handle non-LTS releases

### LINKED ISSUE #4029: Fix logic in tag-recreate-lts.yml for situation where tag is an RC
The current logic is that there is a step that determines the tag for the release and another step that retrieves it. The first step skips setting the tag when it is an RC, and the latter step fails when no tag is set.

Help is needed to rationalize the handling when the tag is for an RC so that the GHA doesn't fail.

esune:
Resolves #4029 

The issue was that the step determining whether the current branch is an LTS was hard-exiting instead of letting the next step set the skip flag and skip the rest of the steps correctly.

### DIFF
--- a/.github/workflows/tag-recreate-lts.yml
+++ b/.github/workflows/tag-recreate-lts.yml
@@ -130,8 +130,8 @@ jobs:
           
           # Validate semantic versioning format (semver 2.0)
           if ! echo "$RELEASE_TAG" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'; then
-            echo "Error: Release tag '$RELEASE_TAG' does not match semantic versioning format (X.Y.Z[-prerelease][+build])"
-            exit 1
+            echo "Warning: Release tag '$RELEASE_TAG' does not match semantic versioning format - skipping"
+            exit 0
           fi
           
           echo "release_tag=$RELEASE_TAG" >> "$GITHUB_OUTPUT"


.github/workflows/tag-recreate-lts.yml
```diff
@@ -130,8 +130,8 @@ jobs:
           
           # Validate semantic versioning format (semver 2.0)
           if ! echo "$RELEASE_TAG" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'; then
-            echo "Error: Release tag '$RELEASE_TAG' does not match semantic versioning format (X.Y.Z[-prerelease][+build])"
-            exit 1
+            echo "Warning: Release tag '$RELEASE_TAG' does not match semantic versioning format - skipping"
```
swcurran, line 133: Will this work if the RC is an LTS release -- e.g. 1.3.4rc0?

esune: It looks like this will work fine (skip LTS release creation) for rc releases due to semver validation failing. I think this is desirable since we do not want an LTS tag to point to a rc, but rather point to the final release for that version.

When running the workflow manually (i.e.: need to manually re-create a release image) it *could* attempt to create the release if semver checks passed, but that would require using the hyphen we are not using (had to ask for the help of AI to parse all the regexp scenarios again, but I am fairly confident the output is correct).

Trigger | RC Format | Result
-- | -- | --
workflow_run | 1.3.4rc0 | ✅ Skipped gracefully (RC check)
workflow_dispatch | 1.3.4rc0 | ❌ Fails at semver validation
workflow_dispatch | 1.3.4-rc0 | ⚠️ Would pass semver, attempt LTS creation

I can add extra checks and handling for the `workflow_dispatch` scenarios if deemed necessary/desirable.

swcurran: You are right -- no such tag for any RC.  Thanks!

PR #4046, opened 2026-02-11
feat: add list endpoint for out-of-band records

### LINKED ISSUE #4044: Add list endpoint for out-of-band records
## Summary

The out-of-band protocol currently has no endpoint to list OOB records. Unlike DID Exchange, which shares `ConnRecord` with the connections API and is therefore queryable via `GET /connections`, OOB maintains its own separate record types (`OobRecord` and `InvitationRecord`) — an appropriate design given connectionless use cases. However, this means OOB records are not discoverable through any existing list endpoint.

## Current OOB Endpoints

| Method | Endpoint | Description |
|--------|----------|-------------|
| `POST` | `/out-of-band/create-invitation` | Create an invitation |
| `POST` | `/out-of-band/receive-invitation` | Receive an invitation |
| `GET` | `/out-of-band/invitations?oob_id=...` | Fetch a **single** record by `oob_id` |
| `DELETE` | `/out-of-band/invitations/{invi_msg_id}` | Delete records by `invi_msg_id` |

Both the fetch and delete endpoints require knowing the specific record ID (`oob_id` or `invi_msg_id`). There is no way to list or query all existing OOB records.

## Problem

If the `oob_id` or `invi_msg_id` is lost (e.g., due to operational errors, application restarts, or logging gaps), there is no way to:

1. **Discover** which OOB records exist in the wallet
2. **Clean up** stale or orphaned OOB records accumulated from failed or abandoned exchanges

This makes it difficult to manage garbage data that builds up over time, particularly in production environments with high invitation volume.

## Comparison with Connections

The connections API provides `GET /connections` with rich filtering (by state, alias, protocol, etc.), making it easy to audit and clean up connection records. OOB records lack equivalent functionality.

## Proposed Solution

Add a `GET /out-of-band/records` (or similar) endpoint that lists `OobRecord` entries with support for:

- **Pagination** — `offset` and `limit` parameters to prevent out-of-memory errors when the record count is large
- **Filtering** — optional query parameters such as:
  - `state` (initial, await-response, done, etc.)
  - `role` (sender, receiver)
  - `connection_id`

This would bring OOB record management in line with the patterns established by other protocols in ACA-Py (e.g., `GET /connections` supports pagination and filtering).

kukgini:
Add GET /out-of-band/records endpoint to query OobRecord entries with pagination and filtering support. This addresses the lack of any way to discover or list existing OOB records in the wallet.

Supports filtering by state, role, connection_id, and invi_msg_id, plus standard pagination parameters (limit, offset, order_by, descending).

Closes #4044


kukgini:
I created this PR by referencing the connection route implementation. However, someone pointed out the following. Could you please check?:

> `alt=True` is incorrect with scalar string `post_filter` values. The `match_post_filter` method in `base_record.py` uses Python's `in` operator when `alt=True`, which performs substring matching on strings rather than equality checking. For example, filtering by `state="initial"` with `alt=True` would cause `record.get("state") in "initial"` to match any substring (e.g., `"i"`, `"init"` ). The fix is to remove alt=True (defaulting to `False` ), which uses direct equality -- consistent with the pattern in `issue_credential/v2_0/routes.py` and `present_proof/v2_0/routes.py`.

jamshale:
> I created this PR by referencing the connection route implementation. However, someone pointed out the following. Could you please check?:
> 
> > `alt=True` is incorrect with scalar string `post_filter` values. The `match_post_filter` method in `base_record.py` uses Python's `in` operator when `alt=True`, which performs substring matching on strings rather than equality checking. For example, filtering by `state="initial"` with `alt=True` would cause `record.get("state") in "initial"` to match any substring (e.g., `"i"`, `"init"` ). The fix is to remove alt=True (defaulting to `False` ), which uses direct equality -- consistent with the pattern in `issue_credential/v2_0/routes.py` and `present_proof/v2_0/routes.py`.

I'm not sure about this...

swcurran:
@PatStLouis -- can you please backport this to the 1.3 and 1.2 (if applicable) LTS branches, so we can include it in the next releases?  I assume that is a quick task.

PR #4047, opened 2026-02-12
Fix revocation setup

PatStLouis:
There were inconsistencies with the new revocation setup when used in conjunction with the author/endorser model. This also affects similar interactions such as webvh witnessing.

It was boiled down to the rotation steps, expecting the new revocation registry to be available right away. This fix ensures the current active registry is properly decommissioned and sets the backup as active, while creating a new backup registry.


swcurran:
Does this need to be backported to the LTS branches -- either or both?

PatStLouis:
@swcurran I think the LTS branch won't need this as the revocation update itself wasn't backported (which I think is okay)

Issue #3995, opened 2025-12-23
Using the --no-transport flag disables webhooks


timbl-ont:
My understanding is that --no-transport enables aca-py to start without transports started in support of scenarios where didcomm is not required - e.g. lightweight modular plugin deployment

It appears that this also disables webhooks which are still needed in support of plugins. FYI @PatStLouis

PatStLouis:
@timbl-ont that was the intention, I would have to review how the implementation of this flag was made, its been a while. I know webhooks are often used in tandem with exchange states, but there are also other use cases. I can review this as it shouldn't affect anything other than protocol registration. Thanks for pointing this out.

PatStLouis:
@timbl-ont can you confirm if #4050 solved this issue?

swcurran:
Fixed in #4050 - Closing

PR #4050, opened 2026-02-17
Enable outbound transport in `--no-transport` mode, only disable inbound

PatStLouis:
Thanks @timbl-ont for raising this [issue](#3995). The `--no-transport` mode was added a while back to disable the need to configure outbound and inbound transport if a deployment didn't need to setup didcomm channels. By disabling oubound transport, webhooks were no longer being sent. This PR keeps the oubound transport enabled for http messages.


jamshale:
@PatStLouis Unable to update the branch with main. Could you do this and we'll get it in.

PR #4056, opened 2026-02-19
fix: handle revocation registry race conditions for BDD integration tests

PatStLouis:
## Summary

Fixes revocation-related BDD test failures (`features/revocation-api.feature`) by addressing timing issues in the test helper and race conditions in the core revocation registry decommission flow.

### Changes

**BDD test helper (`demo/runners/support/agent.py`)**
- Added polling loop in `register_schema_and_creddef_anoncreds` that waits up to 60s for the revocation registry to become active before returning, replacing reliance on a static 2s sleep that was insufficient on external ledgers like BCovrin.

**Revocation decommission flow (`acapy_agent/anoncreds/revocation/revocation.py`)**
- Fixed race condition in `decommission_registry` where both the direct `store_revocation_registry_definition` call and the async event chain tried to INSERT the same registry, causing `AskarError: Duplicate entry`. The direct store is now wrapped in a try/except that handles the duplicate gracefully.
- Fixed `TypeError: 'IterEntryList' object is not iterable` in `decommission_registry` by wrapping `fetch_all` result in `list()` for Python 3.13 compatibility.

**Active registry endpoint (`acapy_agent/anoncreds/routes/revocation/registry/routes.py`)**
- `GET /anoncreds/revocation/active-registry/{cred_def_id}` now returns HTTP 404 instead of an unhandled 500 when no active registry exists, reducing noisy error stack traces during expected polling.

### DIFF
--- a/acapy_agent/anoncreds/revocation/tests/test_revocation.py
+++ b/acapy_agent/anoncreds/revocation/tests/test_revocation.py
@@ -1047,6 +1047,87 @@ class TestAnonCredsRevocation(IsolatedAsyncioTestCase):
         assert result[0].tags["state"] == RevRegDefState.STATE_DECOMMISSIONED
 
     @mock.patch.object(AskarAnonCredsProfileSession, "handle")
+    async def test_decommission_registry_duplicate_store_handled(self, mock_handle):
+        """When event chain already stored the registry, duplicate is handled gracefully."""
+        mock_handle.fetch_all = mock.CoroutineMock(
+            side_effect=[
+                [
+                    MockEntry(
+                        name="backup-reg-reg",
+                        tags={
+                            "cred_def_id": "test-rev-reg-def-id",
+                            "state": RevRegDefState.STATE_FINISHED,
+                            "active": "false",
+                        },
+                    ),
+                ],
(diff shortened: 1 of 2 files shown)


acapy_agent/anoncreds/revocation/revocation.py
```diff
@@ -1531,7 +1531,13 @@ async def decommission_registry(self, cred_def_id: str) -> list:
...
         if new_backup_reg and not isinstance(new_backup_reg, str):
             new_rev_reg_def_id = new_backup_reg.rev_reg_def_id
-            await self.store_revocation_registry_definition(new_backup_reg)
+            try:
+                await self.store_revocation_registry_definition(new_backup_reg)
+            except AnonCredsRevocationError:
+                LOGGER.debug(
+                    "Registry %s already stored by event chain",
+                    new_rev_reg_def_id,
+                )
```
ff137, line 1540: It's a lil bit dubious to suppress an exception without checking what the exception is. The assumption here is that it can only fail in 1 way, but there should be some error detail that one can check to validate that assumption, and if something else went wrong, then raise or log at higher level.

PR #4059, opened 2026-02-19
docs: add migration guide for Indy to did:webvh AnonCreds issuance

### LINKED ISSUE #3885: Add document for how to move an Issuer from an Indy-based AnonCreds credential to one rooted in did:webvh

PatStLouis:
Adds a strategic guide for issuers transitioning from Indy-based AnonCreds credentials to did:webvh-rooted ones. Covers prerequisites, dual-issuance transition strategy, step-by-step setup, and impact on controllers and verifiers.

Closes #3885

### DIFF
--- a/docs/deploying/IndyToWebVHMigration.md
+++ b/docs/deploying/IndyToWebVHMigration.md
@@ -47,9 +47,17 @@ For technical background on how ACA-Py supports multiple AnonCreds methods, see
 
 ## Key Differences at a Glance
 
+> **Note:** The did:webvh column below describes the **did:webvh Server AnonCreds
+> Method** as implemented by ACA-Py and the WebVH plugin—with namespace (`ns`),
+> identifier (`id`), and a defined path to the resource. The did:webvh method
+> itself is flexible: it can support "plain" did:webvh DIDs (no extra path
+> elements) and other DID URL path schemes. This guide and the table refer to
+> the convention used by this implementation, not to a requirement of did:webvh
+> in general.
+
 | Aspect | Indy | did:webvh |
 |---|---|---|
-| **Identifier format** | `WgWx...:2:name:1.0` (legacy) or `did:indy:sovrin:WgWx...` | `did:webvh:{SCID}:domain:ns:id/resources/{digest}` |
+| **Identifier format** | **Schema (legacy):** `{ledgerPrefix}:2:{name}:{version}` (e.g. `WgWx...:2:ExampleSchema:1.0`). **Cred def (legacy):** `{ledgerPrefix}:3:CL:{seqNo}:{tag}`. **DID:** `did:indy:{namespace}:{id}` (e.g. `did:indy:sovrin:WgWx...`). | `did:webvh:{SCID}:domain:ns:id/resources/{digest}` |
 | **Where objects live** | Indy ledger transactions | WebVH server-hosted resources with Data Integrity proofs |
 | **Trust model** | Endorser/Steward signs ledger transactions | Witness attests resource publications |
 | **Revocation tails files** | Uploaded to a tails server | Uploaded to a tails server (same approach) |
@@ -131,7 +139,11 @@ flowchart LR
     subgraph phase3 ["Phase 3: Switch to did:webvh Issuing"]
         Cutover["Stop Indy issuance, issue only did:webvh"]
     end
-    phase1 --> phase2 --> phase3
+    subgraph phase4 ["Phase 4: Remove Indy"]
+        Notify["Issuer notifies parties"]
+        Remove["Remove Indy code and config"]
+    end
+    phase1 --> phase2 --> phase3 --> phase4
 ```
 
 ### Phase 1 -- Prepare
@@ -153,10 +165,44 @@ did:webvh side without changing issuance yet:
 
 ### Phase 2 -- Verifier readiness
 
-Give verifiers time to update their systems to accept credentials rooted in
-either the existing Indy identifiers or the new did:webvh identifiers. Once
-verifiers can resolve did:webvh and include the new credential definition IDs
-in their proof requests, you can schedule the cutover.
+Verifier readiness requires two things. First, verifiers need **updated
+libraries** that support resolving credentials via the did:webvh AnonCreds
+method (so they can resolve and validate did:webvh-rooted credentials).
+Second, they must **update their presentation requests** to add the
+equivalent did:webvh identifiers as an alternative (OR) to the current Indy
+identifiers—so that a wallet receiving the request can respond with whichever
+credential it is holding (Indy-rooted or did:webvh-rooted). Give verifiers
+time to do both. A presentation request can restrict on **issuer (DID)**,
+**schema**, and/or **credential definition**; in each case, add the
+corresponding did:webvh identifier alongside the existing Indy one so that
+either type satisfies the request. Once verifiers can resolve did:webvh and
+have updated their presentation requests, you can schedule the cutover.
+
+**Example — presentation request restrictions:**
+
+Before cutover (Indy only), a verifier might restrict on a single credential
+definition:
+
+```json
+"restrictions": [
+  { "cred_def_id": "WgWxqztrNooG92RXvxSTWv:3:CL:20:tag" }
+]
+```
+
+After updating for the transition (accept Indy or did:webvh), the verifier
+includes both identifiers so that credentials from either VDR are accepted:
+
+```json
+"restrictions": [
+  { "cred_def_id": "WgWxqztrNooG92RXvxSTWv:3:CL:20:tag" },
+  { "cred_def_id": "did:webvh:zQ3sh...:example.com:my-org:issuer-01/resources/abc123" }
+]
+```
+
+Alternatively, restrictions can be expressed by **issuer** (`issuer_id`) or
+**schema** (`schema_id`); in each case, add the corresponding did:webvh
+identifier so that credentials rooted in either Indy or did:webvh satisfy the
+request.
 
 ### Phase 3 -- Switch to did:webvh Issuing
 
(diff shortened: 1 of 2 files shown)


docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,311 @@
...
+
+**No data loss:**
+
+- Your wallet retains all Indy objects. ACA-Py supports multiple AnonCreds
+  registries simultaneously -- the correct registry is selected automatically
+  based on the identifier pattern of the object being accessed.
+
+## Migration Strategy: Dual-Issuance Transition
+
+The recommended approach is a phased transition where Indy and did:webvh
```
swcurran: I disagree with this.  I think the recommended approach is an abrupt issuance transition from Indy to did:webvh, and a transitioned holding period where some holders have Indy-rooted credentials from before the transition, and some have did:webvh-rooted from after.  I can't see a case where you would want to issue the same credentials at the same time.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,311 @@
...
+### Phase 1 -- Current State
+
+You are issuing Indy-based AnonCreds credentials. No changes yet.
+
+### Phase 2 -- Dual Issuance
+
+1. Set up the did:webvh infrastructure (server, witness, plugin) alongside
+   your existing Indy configuration.
+2. Create a did:webvh DID for your issuer.
+3. Re-register your schemas and credential definitions under the new DID.
```
swcurran: I don't think the "Re-" is needed -- these are new objects.  Picky, picky...

docs/deploying/IndyToWebVHMigration.md
PatStLouis: this makes the assumption that **all** wallets already support webvh. If we are ok with this assumption, then we can change it. If holders don't support webvh and you abruptly stop issuing indy, this can have adverse effects.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,311 @@
...
+4. Start issuing new credentials using the did:webvh credential definitions
```
swcurran: I think we need to call out here that the new identifiers be shared with verifiers as much as possible for a period before issuance of did:webvh-rooted credentials begins. This allows verifiers to update their presentation requests to accept credentials rooted in either VDR.  During this period -- the did:webvh objects exist, but are not being used for issuance.  We might leave an AB testing exception to verify being able to issue a credential or two for testing, but not general use.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,311 @@
...
+   for new holders or use cases.
+5. Continue issuing Indy credentials for existing holders and verifiers who
+   have not yet added did:webvh resolution support.
+6. Communicate the transition to your verifier ecosystem so they can prepare
+   to resolve did:webvh identifiers.
+
+### Phase 3 -- did:webvh Only
```
swcurran: "Phase 3 - Switch to did:webvh Issuing"

PatStLouis:
@swcurran I think a lot of this comes down to, how much should the issuer be limited by it's ecosystem? If an issuer needs to wait for all holders and verifiers to support their updated components, this creates a limitation on the least up to date entity.

Its hard to describe this in an upgrade guide with a decentralized ecosystem in mind. An issuer might not have a way to contact all verifiers. I think a lot of this wording can be massaged in the "transition" phase. Issuers might not know which wallet software stores their credentials, same for verifiers. Issuers can publish a notice of their upcoming changes, but may not have direct line of communication with wallet / service providers.

PatStLouis:
They might have "critical" lines of business where they just must keep issuing indy until all parties have updated their software, which may or may not be aligned with the issuer's technological choices.

swcurran:
Let's discuss.  We can call these out, as you say, but what is the "recommended" way?   I think the recommended way to get the holders/verifiers software updated/ready (as best you can in a decentralized world), and then do the transition.

If the criteria for keeping to issue the Indy-rooted credentials, how do you know what holder software has been updated?  AFAIK the "discover features" protocol -- as useful as it is -- is not broadly implemented, and doesn't cover (for example) Indy to did:webvh transitions.

docs/deploying/IndyToWebVHMigration.md
esune: I evidently didn't pay enough attention on the first pass, I agree with @swcurran on the fact that while Indy and WebVH credentials can coexist, the recommended approach should be to start issuing on WebVH and only support revocation of Indy credentials while phasing out, at least when considering a creddef that is migrated from one technology to the other.

PatStLouis:
Addressed review feedback: reframed the guide around the full issuance process and a planned cutover (not dual-issuance), added the assumption that issuers/holders/verifiers already support did:webvh, and called out sharing new identifiers with verifiers before cutover and the need to keep managing already-issued Indy credentials (e.g. revocation). Also applied the copy edits (Phase 3 title, "register" not "re-register") and noted the pattern can apply to other VDR transitions.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,353 @@
...
+  servers.
+- **Interoperable** -- Credentials issued with did:webvh use the same AnonCreds
+  cryptography; verifiers process them identically to Indy-based credentials.
+
+For technical background on how ACA-Py supports multiple AnonCreds methods, see
+[Publishing AnonCreds Objects To Other Ledgers/VDRs](../features/AnonCredsMethods.md).
+
+## Key Differences at a Glance
+
+| Aspect | Indy | did:webvh |
```
swcurran, line 58: A relative minor issue, but let me know what you think -- maybe a note early in the document?  What you have in this table the "did:webvh Server AnonCreds Method" where the implementation is opinionated -- there is two extra elements in the DID (`ns` and `id`), plus a defined path to the resource. `did:webvh` itself can support any scheme -- "plain" did:webvh DIDs (no extra elements), any DID URL path.  So we at least need to call that out vs. implying that this is what did:webvh requires.

It's too late for this, but I've wondered silently if the path should contain metadata such `/resources/anoncreds-schema/{digest}`.  Did you consider that?  Or the same mechanism that Indy used -- with the object type enumerated in the identifier.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,353 @@
...
+|---|---|---|
+| **Identifier format** | `WgWx...:2:name:1.0` (legacy) or `did:indy:sovrin:WgWx...` | `did:webvh:{SCID}:domain:ns:id/resources/{digest}` |
```
swcurran: Might want to expand the Indy examples to include a the elements of a identifier data type.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,353 @@
...
+   resulting identifiers change).
+4. **Share the new did:webvh identifiers with verifiers** (credential
+   definition IDs, schema IDs, etc.) so they can add them to their
+   presentation requests. During this period, the did:webvh objects exist but
+   are **not** used for general issuance. Limited testing (e.g. issuing a
+   credential or two to verify the pipeline) is acceptable.
+
+### Phase 2 -- Verifier readiness
+
+Give verifiers time to update their systems to accept credentials rooted in
```
swcurran: I would expand this a bit more, to indicate that requires two things.  Updated libraries that support resolving via the did:webvh AnonCreds Method, and updating the presentation request to add the equivalent did:webvh identifiers as an `OR` to the current Indy identifiers so that a wallet receiving the request can respond with whichever they are holding.

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,353 @@
...
+either the existing Indy identifiers or the new did:webvh identifiers. Once
+verifiers can resolve did:webvh and include the new credential definition IDs
```
swcurran: May not be a credential definition -- a presentation request can restrict on DID (issuer), schema and/or credential definition.

Might be worth putting in an example here -- before just indy and after Indy or did:webvh

docs/deploying/IndyToWebVHMigration.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Phase 3 -- Switch to did:webvh Issuing
+
+1. On the chosen date, **stop issuing new Indy credentials**. All new
+   issuance uses did:webvh credential definitions only.
+2. Previously issued Indy credentials remain valid and verifiable for their
+   lifetime. They do not need to be re-issued. The issuer must **continue to
+   manage** those Indy credentials (e.g. process revocation and other status
+   updates) for as long as they are in use.
+3. You can remove the Indy ledger connection only when all Indy credentials
```
swcurran: Perhaps make this Phase 4?  Issuer notifies all parties (especially verifiers) that the Indy credentials have all been revoked. Issuers and verifiers (at their leisure) can then remove both the code and the presentation request configuration that uses/references Indy rooted-credentials. Verifiers only do that after all of the credential types they accept have been converted.

mkdocs.yml
```diff
@@ -147,6 +147,7 @@ nav:
     - Databases: deploying/Databases.md
     - Persistent Queues and Caching: deploying/RedisPlugins.md
     - The askar-anoncreds Wallet Type: deploying/AnonCredsWalletType.md
+    - Migrating Issuance from Indy to did:webvh: deploying/IndyToWebVHMigration.md
```
swcurran: Change to "Migrating AnonCreds Issuance..."

docs/deploying/IndyToWebVHMigration.md
PatStLouis: > It's too late for this, but I've wondered silently if the path should contain metadata such /resources/anoncreds-schema/{digest}. Did you consider that? Or the same mechanism that Indy used -- with the object type enumerated in the identifier.

We can explore adding a resource type slug to the url, and require all attested resource to include this slug in their id. Since this is not currently the case, lets leave it out of the doc for now.

PatStLouis:
@swcurran I think I've addressed all your suggestions, aside from including a resource type slug in the uri!

PR #4067, opened 2026-02-24
Consolidate snyk scanning GHA

### LINKED ISSUE #4021: Look into GHA for Scanning LTS branches
- 1.2 GHA is failing
- Need to add the GHA for 1.3

esune:
The snyk scanning GHA was failing for LTS branches, presumably due to a change in behaviour in the cli after a recent update. We also had two different scan actions in the repo doing the same, so this PR consolidates the workflow to be run on both `main` and `lts` branches.

Previously, the `lts` branch only used the `monitor` task, however it seems like it may be good practice to have test + monitor on both `main` and `lts` branches so I proceeded in this direction. If this is not desirable we can scope the execution of the monitor task just to `lts` branches.

Resolves #4021


.github/workflows/snyk.yml
```diff
@@ -3,7 +3,9 @@ on:
   push:
     branches:
       - main
+      - '**.lts'  # LTS release branches (e.g., 0.12.lts, 1.2.lts)
```
swcurran, line 6: Do we want to include only active LTS branches?  For example, 1.2.lts and 1.3.lts are still supported, 0.12.lts is not. 1.2.lts support will be dropped in April.  We won't do anything (AFAIK) if a vulnerability is found in unsupported LTS branches.

If we dropped 0.X.lts, it would mean we wouldn't need to add aries_cloudagent below. Not a big deal, but there you go...

esune: If we are not going to support `0.12.lts` any longer then the `aries_cloudagent` path becomes obsolete, yes. Happy to remove it if we are not going to release on that branch anymore, it shouldn't hurt leaving it in otherwise.

swcurran: But if you are going to run on all **.lts branches, we need to keep it. Only drop it if you make the branch selector more precise.

esune: Then I think I'd leave it so we don't have to continuously update the GHA to pick the right branch - the action will trigger if we push to an LTS branch, which would mean we need a new release for it.

PR #4078, opened 2026-03-05
chore: bump jsonpath-ng to 1.8.0 and fix full_path format

PatStLouis:
- Update jsonpath-ng from ^1.7.0 to ^1.8.0
- Normalize str(full_path) in reveal_doc: jsonpath-ng 1.8+ uses parentheses (e.g. ((a.b).c)); strip them for unflatten keys
- Use maxsplit=1 in re.split to fix deprecation warning


jamshale:
This one should be back ported to the LTS branches if possible. I hope to have the additional anoncreds upgrade revocation fixes done soon which should also go into the LTS branches.

PR #4080, opened 2026-03-06
fix: Revocation errors in the anoncreds upgrade

jamshale:
I'm honestly still not entirely sure why the problem with the accumulator value didn't occur during the existing test. My guess is that it was getting synced before attempting a revocation through the issuance/presentation flow???

It did help me find an additional problem with the indexing of the revocation list. The 0 index is reserved for the anoncreds/credx authentication algorithm. With the old record format the list was just a list of cred_rev_id's so the 0 index thing didn't make a big difference, it was just added before using the credx library. However the new record type is an array of 0's and 1's for the revocations. We need to start at index 1 and make the list one index larger for the conversion to line up correctly.

I did this in other places but forgot to go back and fix the upgrade script and it wasn't found with subsequent testing.

Updated the tests to be more focused on the revocation lists being what is expected and that all the credentials can be revoked correctly after the upgrade.

### DIFF
--- a/acapy_agent/wallet/anoncreds_upgrade.py
+++ b/acapy_agent/wallet/anoncreds_upgrade.py
@@ -246,7 +246,7 @@ async def get_rev_reg_def_upgrade_object(
             CATEGORY_REV_REG_DEF_PRIVATE, rev_reg_def_id
         )
         accum_record = await storage.get_record(CATEGORY_REV_REG, rev_reg_def_id)
-        acccum_value = json.loads(accum_record.value)["value"]["accum"]
+        accum_value = json.loads(accum_record.value)["value"]["accum"]
 
     revoc_reg_def_values = json.loads(askar_issuer_rev_reg_def.value)
 
@@ -270,7 +270,7 @@ async def get_rev_reg_def_upgrade_object(
         rev_reg_def,
         askar_reg_rev_def_private.value,
         is_active,
-        acccum_value,
+        accum_value,
     )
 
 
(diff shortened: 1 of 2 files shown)


jamshale:
This seems to be working but there are still problems when deploying it into traction I'm trying to figure out.

acapy_agent/anoncreds/default/legacy_indy/registry.py
```diff
@@ -827,6 +827,8 @@ async def _revoc_reg_entry_with_fix(
             txn_record_type=GET_REVOC_REG_ENTRY,
         )
 
+        ledger.profile = profile
```
jamshale: @PatStLouis Bizarrely this fixes the `Unknown Did` error after upgrade. For some reason I don't understand yet the ledger class was loading the `root_profile` instead of the `tenant` profile. It does this same process in various other places without issue.

I'm not super happy with this fix even though it should be fine. I'll try and figure out why this is happening a bit longer but my testing so far seems like everything is working better.

acapy_agent/wallet/anoncreds_upgrade.py
```diff
@@ -277,18 +286,19 @@ async def get_rev_list_upgrade_object(
...
 
     rev_list = RevList(
         issuer_id=rev_reg.issuer_id,
         rev_reg_def_id=rev_reg_def_upgrade_obj.rev_reg_def_id,
         revocation_list=revocation_list,
-        current_accumulator=json.loads(
-            rev_reg_def_upgrade_obj.askar_issuer_rev_reg_def.value
-        )["revoc_reg_entry"]["value"]["accum"],
+        current_accumulator=rev_reg_def_upgrade_obj.accum,
     )
```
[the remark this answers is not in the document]
jamshale: This is covered well by the scenario tests so I'm going to mark this as resolved.

acapy_agent/anoncreds/default/legacy_indy/registry.py
jamshale: This is removed from this PR and will be addressed in subsequent PR.

jamshale:
I don't think the unit test coverage is too important here. If I had more time I'd add some coverage but I think the scenario test covers the changes well enough.

jamshale:
This PR is ready for the accumulator value being set correctly and the indexing problem which was caused by the change it data structure which hadn't been accounted for correctly.

PR #4119, opened 2026-04-17
[WIP] Remove legacy indy

jamshale:
This is an in progress PR that removes all indy and ledger components/modules from the core of acapy in favour of a plugin that has patches for all the the indy usage that was baked into the core handlers and modules.

It's very much still work in progress. There is still places that the patches haven't been created like all issuance and presentation protocol managers/handlers.

Most of the tests will fail...

I'll open the corresponding plugin PR and link the two and hopefully we can work on pushing this over the line in the next several months with acapy getting less regular updates currently.


jamshale:
### From claude helping to summarize the approach and current status...

**What's being removed from core acapy (PR #4119)**

The file tree shows 247 Python files touched, all deletions or stripping of Indy-specific code. The key modules being gutted or removed entirely are:

- `acapy_agent/indy/` — the whole indy package (holder, issuer, models, credx implementations, verifier, constants)
- `acapy_agent/anoncreds/default/legacy_indy/` — the legacy Indy AnonCreds registry, author, routes, and revocation recovery logic
- `acapy_agent/config/ledger.py` — ledger configuration wiring
- `acapy_agent/config/default_context.py` — where Indy components get injected into the DI container at startup
- `acapy_agent/did/indy/` — the `did:indy` manager and routes
- `acapy_agent/commands/provision.py` and `upgrade.py` — Indy-specific provisioning and wallet upgrade paths
- Various protocol handlers under `connections/`, `holder/`, `anoncreds/revocation/`, and the admin server

In short, it's stripping out the Indy ledger pool, the Indy credential/presentation format handlers, the `did:indy` DID method, endorsement flows, and schema/cred-def publishing — all things that were historically hardwired into core.

---

**How the plugin (PR #3034) monkey-patches it back in**

The plugin's file tree mirrors the deleted modules almost exactly — `indy_ledger/indy_ledger/anoncreds/`, `credential_definitions/`, `anoncreds/routes/`, etc. The approach ACA-Py plugins use is a `setup` function in `__init__.py` that gets called at agent startup. Based on what's visible in the tree and how other ACA-Py plugins work, the plugin:

1. **Re-registers Indy components into the DI container** — things like the `IndyLedger`, `IndyHolder`, `IndyIssuer`, and `IndyVerifier` classes that core used to inject automatically, the plugin now injects conditionally at load time via `context.inject_or` / `context.settings`.

2. **Patches protocol handler modules** — for places where Indy logic was baked directly into core message handlers (e.g. credential issuance managers, presentation managers), the plugin uses Python's module-level attribute replacement — essentially `some_core_module.SomeClass = PluginReplacementClass` — to swap in Indy-aware versions without modifying core.

3. **Re-registers routes** — the plugin adds back the `/ledger/`, `/schemas/`, `/credential-definitions/`, `/revocation/` and related endpoints that were stripped from core, via ACA-Py's standard `register` route hook.

4. **Carries the AnonCreds Indy registry** — the `anoncreds/registry.py` in the plugin re-registers `LegacyIndyRegistry` as the AnonCreds method for resolving and writing schemas and cred defs to an Indy ledger.

---

**The tricky part — and what's still unfinished**

The monkey patching works cleanly for the ledger/schema/cred-def surface, but it gets hairy in the issuance and presentation *protocol managers* — the V1/V2 credential exchange and present-proof handlers. Those managers have Indy-specific branching logic deeply embedded (checking wallet type, falling back between Indy and AnonCreds format handlers, etc.), and the plugin hasn't fully patched all of those yet. That's the explicit "still WIP" gap jamshale calls out.

PR #4163, opened 2026-07-02
Fix anoncreds upgrade failure in-memory in-progress check

jamshale:
We were upgrading a traction tenant that was in a corrupted state and failed. The in memory, in progress check was never set off which caused the api endpoints for this tenant to be stuck with 503 errors until pods were restarted.


jamshale:
This sonarcloud fail isn't valid for the test file imo.

swcurran (approved):

PR #4164, opened 2026-07-06
fix: complete stalled setup when activating a posted revocation registry

kukgini:
## Description

`IndyRevocation.get_or_create_active_registry` promoted the oldest `posted` registry to `active` with a bare wallet state change (`_set_registry_status`).

A registry is only ever left in a lasting `posted` state when its setup pipeline stalled between `send_def` and `send_entry` — most commonly a failed tails file upload (`on_revocation_registry_init_event`: `send_def` → `upload_tails_file` → `send_entry`). Promoting such a registry by flipping the wallet state skips the missed steps, so the "recovered" registry is broken:

- its ledger `tailsLocation` points at a file the tails server never received — holders get a 404 and can never build non-revocation proofs, and
- its initial accumulator entry (`REVOC_REG_ENTRY`) is missing from the ledger.

Every credential subsequently issued against it is silently unverifiable. If the agent's local tails copy was also lost (ephemeral storage), issuance instead wedges on `get_or_fetch_local_tails_path` until manual intervention.

This PR resumes the stalled pipeline instead of bare-flipping the state:

- `upload_tails_file()` — idempotent against the reference tails server, since `put_file` already treats HTTP 409 "already exists" as success — covers the case where the original failure happened after the upload;
- `send_entry()` — publishes the missed initial entry and itself performs the `posted -> active` transition (it also already contains the `fix_ledger_entry` correction path for entry conflicts).

If either step fails, the registry stays `posted`, an error is logged, and promotion is retried naturally on the next issuance attempt (callers already treat a `None` result as retry-after-delay). This is deliberately conservative: the worst case becomes "issuance paused with loud errors" instead of "broken credentials issued".

Note on endorser-mediated deployments: the promotion path has no endorser context, so the resume performs a direct ledger write; for endorser authors it will fail and leave the registry `posted`, which is no worse than the previous behavior (a broken activation).

**Tests**: reworked `test_get_or_create_active_registry_has_no_active_with_posted` to assert the missed steps run and no bare state change occurs; added `test_get_or_create_active_registry_posted_setup_incomplete` asserting that a registry whose setup cannot be completed is not activated. (The 3 `TestDeleteTails` failures reproduce on a pristine `main` checkout and are unrelated.)

**Backport request**: this code path ships in the 1.3 LTS line (observed on a 1.3.4 deployment; the function and the `put_file` 409 handling are identical on `1.3-lts`). We'd appreciate a backport to `1.3-lts` — happy to open that PR ourselves if preferred.

## Related Issue

N/A — full analysis included above.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Chore / maintenance

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)

### DIFF
--- a/acapy_agent/revocation/indy.py
+++ b/acapy_agent/revocation/indy.py
@@ -7,6 +7,7 @@ from uuid_utils import uuid4
 
 from ..core.profile import Profile
 from ..ledger.base import BaseLedger
+from ..ledger.error import LedgerError
 from ..ledger.multiple_ledger.ledger_requests_executor import (
     GET_CRED_DEF,
     GET_REVOC_REG_DEF,
@@ -260,6 +261,7 @@ class IndyRevocation:
         except StorageNotFoundError:
             pass
 
+        posted_registries = []
         async with self._profile.session() as session:
             full_registries = await IssuerRevRegRecord.query_by_cred_def_id(
                 session, cred_def_id, None, IssuerRevRegRecord.STATE_FULL, 1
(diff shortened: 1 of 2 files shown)


jamshale (approved):
Thanks for the contribution. This is a good PR. 

My only concern with it is the indy module is legacy code at this point. There is a desire to remove everything into a legacy plugin in favor of the anoncreds implementation.

Basically, I'm wondering if the same issue exists in the anoncreds implementation of indy `acapy_agent/anoncreds/default/legacy_indy/recover.py` and if this needs to be applied there as well. I don't have time for a deep dive into it atm.

kukgini:
I checked the anoncreds implementation on current `main` for the same defect — it doesn't need this fix, because the failure window this PR closes doesn't exist there:

1. **Ordering**: `create_and_register_revocation_registry_definition` uploads the tails file *before* registering the def on the ledger, inside a single try block — an upload failure aborts the step (failure event + retry), so a `finished` def implies its tails file is already on the server. The legacy indy module writes to the ledger first, which is what creates the stalled-POSTED window this PR completes.
2. **No bare promotion on the issuance path**: the anoncreds `get_or_create_active_registry` only fetches the `active=true` registry and raises `No active registry` otherwise. Rotation (`handle_full_registry_event`) activates only `finished` backups, and if a rev list were somehow missing, issuance fails loudly in `_create_credential` rather than silently issuing credentials against a broken registry — which was the actual harm here.
3. **Stall recovery exists there by design**: the event-driven setup pipeline persists in-progress steps and re-emits them on startup (`revocation/auto_recovery/`), and `_create_credential` already lazily re-uploads tails for rev lists left in `failed` state — the same complete-on-issuance idea as this PR.

`default/legacy_indy/recover.py` addresses a different problem (rebuilding a lost ledger accumulator from wallet state), so nothing to apply there either.

PR #4167, opened 2026-07-08
fix: Check RECORD_TYPE_ACAPY_UPGRADING type for clustered upgrade check

jamshale:
During a multi-tenant and clustered deployment with anoncreds upgrades noticed a problem where the non cached upgrade check was the wrong status. This was because the check wasn't checking for the actual in progress status. Meaning the finished upgrades would still trigger this upgrade in progress status one time for clustered tenants.


swcurran:
Looks reasonable to me, but I should not be the approver.  Please update the branch to the latest so that once we get an approval we can merge this.

Thanks!

acapy_agent/admin/server.py
```diff
@@ -214,28 +217,30 @@ async def upgrade_middleware(request: web.BaseRequest, handler: Coroutine):
...
-            in_progress_upgrades.set_wallet(context.profile.name)
-            is_subwallet = context.metadata and "wallet_id" in context.metadata
-
-            # Create background task and store reference to prevent garbage collection
-            task = asyncio.create_task(
-                check_upgrade_completion_loop(
-                    context.profile,
-                    is_subwallet,
+            # Check if the upgrade is actually in progress (not finished)
+            if upgrade_initiated[0].value == UPGRADING_RECORD_IN_PROGRESS:
```
ff137:
```py
if upgrade_initiated and upgrade_initiated[0].value == UPGRADING_RECORD_IN_PROGRESS:
```

Two ifs -> one, because nothing else runs only in the 1st block. Removes extra nesting / more readable

ff137 (dismissed):
1 nit

esune (dismissed):

swcurran (approved):

PR #4171, opened 2026-07-16
fix: bound memory usage of record queries and maintenance/upgrade scans

esune:
## Description

This is an attempt to mitigate/fix very high memory usage for ACA-Py when an elevated number of records is stored in the database. I am not 100% sure of the implications of the change to the upgrade script (c.c.: @jamshale ) since that is a special case, but I *think* the change should not affect how things work other than by processing smaller batches each time.

Tests pass locally, I have not yet gone through a full round of testing on a deployed instance, but can attempt to coordinate one if deemed necessary.

## Summary

ACA-Py could consume very large amounts of memory (12–16 GB observed) on
instances with high record counts. The root cause is a pattern where entire
record categories are loaded into Python memory at once instead of being
paginated at the storage layer. This PR removes those unbounded loads across
the query path, the admin list endpoints, and the anoncreds maintenance/upgrade
sweeps.

## Root cause

`BaseRecord.query()` fell back to `find_all_records` (loading the whole record
category into memory) whenever a `post_filter` was supplied — even when
`limit`/`offset` were requested. The results were then held again during
serialization, so a single request could hold the full dataset several times
over. On agents with many connection / credential-exchange / presentation /
route records (active and stale), this ballooned memory, multiplied by
concurrent or polled requests (e.g. monitoring hitting `GET /connections?state=active`).

## Changes

- **Storage query core:** `BaseRecord.query()` now pages through storage in
  bounded batches when a post-filter is combined with pagination, so peak memory
  is bounded to a single page. Unpaginated post-filter queries keep their
  existing "return all matches" behavior, so internal callers are unaffected.
- **Connections:** `state` and `their_role` (indexed tags) are pushed into the
  storage tag query via `$or` instead of being post-filtered, so the common
  `?state=active` query filters and paginates at the database.
- **Mediation / Revocation admin endpoints:** the mediator keylist endpoint and
  the revocation registry listing now support `limit`/`offset`/`order_by`/
  `descending`; the revocation `STATE_INIT` exclusion moved into the indexed tag
  query (`$not`).
- **AnonCreds auto-recovery:** the in-progress / failed / completed-event sweeps
  page through storage; cleanup deletes per page and advances the offset past
  kept records so deletions don't skip rows.
- **AnonCreds upgrade:** `get_rev_list_upgrade_object` pages through per-registry
  credential-revocation records (up to tens of thousands) and retains only the
  maximum `cred_rev_id` instead of the full record list.

## Behavior changes (please review - will need to be listed in changelog)

- The **mediator keylist** and **revocation registry listing** endpoints are now
  paginated by default (max 100 results unless `limit` is supplied). Clients that
  relied on receiving every record in a single call must page. All other list
  endpoints were already paginated.

## Testing

- Updated and added unit tests for: batched paging in `BaseRecord.query()`,
  `$or`/`$not` tag pushdown, pagination parameters on the mediation/revocation
  endpoints, batched recovery sweeps (including cleanup offset accounting), and
  the upgrade cred-rev paging.
- Affected unit suites pass locally (storage/messaging models, connections,
  issue-credential v2.0, present-proof v2.0, out-of-band, coordinate-mediation,
  routing, revocation, anoncreds revocation, wallet upgrade).

## Notes

- The inbound message connection-resolution hot path was already using indexed
  tag lookups plus the verkey cache and is unchanged.
- `RoutingManager.get_routes` was intentionally left unchanged — it is
  connection-scoped internally; the unbounded risk was the admin keylist route,
  which is now paginated.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Chore / maintenance

## Checklist

- [X] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [X] My changes follow the existing code style
- [X] I have added/updated tests where applicable
- [X] I have updated documentation if needed


jamshale (commented):
I think most of this was before we had askar pagination and this stuff was never updated to use it. 

Because all the tests are passing I'm pretty confident this was done correctly. It be nice to see some real metrics on the improvements. 

I'll go over it in more detail but this looks like a solid solution.

I was also think cached did records may be contributing to excessive memory usage but haven't got around to a proper investigation.

jamshale (approved):
I'll approve this because it didn't break any tests and has coverage for the changes. It makes sense that this would cause major memory usage with fetch all.

If this memory usage is blowing up on agent startup I don't really understand which of these areas would be causing it. I don't think anything is calling these queries. Maybe a another service is triggering them?

esune:
> I'll approve this because it didn't break any tests and has coverage for the changes. It makes sense that this would cause major memory usage with fetch all.
> 
> If this memory usage is blowing up on agent startup I don't really understand which of these areas would be causing it. I don't think anything is calling these queries. Maybe a another service is triggering them?

I agree with you, I guess we'll keep en eye open and see if we can identify other causes, but hopefully this would mitigate the issue.

PatStLouis (approved):

PR #4172, opened 2026-07-16
feat: use DELETE for multitenancy subwallet removal

### LINKED ISSUE #3135: Multitenancy endpoint to remove sub-wallet should use crud DELETE
ACA-Py currently exposes the `POST /multitenancy/wallet/XXXXX/remove` to delete a sub-wallet.
Using `POST` for this endpoint is inconsistent with the crud operation being performed and the endpoint should be updated to use `DELETE /multitenancy/wallet/{wallet_id}` instead.

esune:
## Description

This PR Replaces **POST** `/multitenancy/wallet/{wallet_id}/remove` with **DELETE** `/multitenancy/wallet/{wallet_id}` for REST consistency. Updates docs and regenerated OpenAPI/Swagger specs.

The change is mostly an idiomatic nitpick, but describes better the destructive operation of the endpoint and is more consistent with REST specifications.

This will require to be noted in the next release changelog.

## Related Issue

Closes #3135 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [X] Chore / maintenance

## Checklist

- [X] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [X] My changes follow the existing code style
- [X] I have added/updated tests where applicable
- [X] I have updated documentation if needed


esune:
I can fix the maintainability issue, but I believe this is the current pattern for endpoints so I am a bit hesitant to go ahead and implement it just for this set of endpoints...

acapy_agent/multitenant/admin/routes.py
```diff
@@ -691,7 +691,7 @@ async def register(app: web.Application):
             web.get("/multitenancy/wallet/{wallet_id}", wallet_get, allow_head=False),
             web.put("/multitenancy/wallet/{wallet_id}", wallet_update),
             web.post("/multitenancy/wallet/{wallet_id}/token", wallet_create_token),
-            web.post("/multitenancy/wallet/{wallet_id}/remove", wallet_remove),
```
jamshale, line 694: Should we instead mark this a depreciated for backwards compatibility for a release or two?

jamshale:
> I can fix the maintainability issue, but I believe this is the current pattern for endpoints so I am a bit hesitant to go ahead and implement it just for this set of endpoints...

We can ignore it for this PR. It's pretty aggressive for using constants.

swcurran:
So DELETE is common enough that we can use it freely?  I recall the day we had to remove from some app because something in the handling only allowed GET and POST.  BTW -- did you see that QUERY is now allowed?  It was added by IETF a month or so ago.  So instead of using a POST for a complex/large query to do the equivalent of a GET, you can use QUERY.  Well, you can once it is supported in the wild...

esune:
> So DELETE is common enough that we can use it freely? I recall the day we had to remove from some app because something in the handling only allowed GET and POST. BTW -- did you see that QUERY is now allowed? It was added by IETF a month or so ago. So instead of using a POST for a complex/large query to do the equivalent of a GET, you can use QUERY. Well, you can once it is supported in the wild...

`DELETE` is standard REST specification so it *should* be widely supported without issues. The only edge-case is with the body parameter, in this case `wallet_key` for unmanaged wallets: it is discouraged to have body parameters for `DELETE` requests since the operation is by definition destructive, however I think in this specific case it is acceptable since it covers a small gap that requires the key to be able to proceed, and it explicitly defines the operation as destructive by REST standard.

WadeBarnes:
> So DELETE is common enough that we can use it freely? I recall the day we had to remove from some app because something in the handling only allowed GET and POST. BTW -- did you see that QUERY is now allowed? It was added by IETF a month or so ago. So instead of using a POST for a complex/large query to do the equivalent of a GET, you can use QUERY. Well, you can once it is supported in the wild...

I believe what you are recalling harkens back to our work on CCSS a decade ago.  Site Minder did now allow DELETE as a verb, so our workaround for delete was to POST with `httpmethod=DELETE`.

acapy_agent/multitenant/admin/routes.py
esune: Sorry, I completely missed this comment. This is a good suggestion, I pushed a small update to have the old endpoint marked as deprecated alongside the new one.

jamshale (approved):

loneil (approved):
Traction references subwallet deletion in some Innkeeper functionality but just as `BaseMultitenantManager.remove_wallet()` directly so the route change won't affect that.

jamshale:
This failure is just an intermittent Interop test. I'm confident in the rest of the tests.

Issue #3300, opened 2024-10-21
Error with out-of-band protocol while self-issuance 


nb-vivek-bodar:
I am experiencing an error with out-of-band protocol while self issuing a credential.

### Step to reproduce the error: 

1. Create credential offer for out-of-band.
2. Create out-of-band invitation.
3. Receive the above  out-of-band invitation on the same agent.
4. The error occurs.

- Error with Handshake Protocol (did-exchange/1.0): When trying to complete the handshake, I receive the following error:

`Traceback (most recent call last):
  File "/home/aries/.local/lib/python3.12/site-packages/aries_cloudagent/protocols/didexchange/v1_0/handlers/complete_handler.py", line 24, in handle
    await mgr.accept_complete(context.message, context.message_receipt)
  File "/home/aries/.local/lib/python3.12/site-packages/aries_cloudagent/protocols/didexchange/v1_0/manager.py", line 1101, in accept_complete
    raise DIDXManagerError(
aries_cloudagent.protocols.didexchange.v1_0.manager.DIDXManagerError: No corresponding connection request found
[...]

PR #4173, opened 2026-07-17
fix: self-issuance/self-connection OOB and DID exchange failures (#3300)

nb-vivek-bodar:
## Description

Fixes self-issuance / self-connection failures when an agent creates an out-of-band invitation and receives it back into its own wallet:

- **`StorageDuplicateError` in `find_oob_record_for_inbound_message`**: a self-connection produces two OobRecords (role=sender and role=receiver) sharing one `invi_msg_id`. Now disambiguated by the recipient verkey the message was delivered to, with a message-direction heuristic as fallback (didexchange request / handshake-reuse always belong to the role=sender record).
- **`DIDXManagerError` ("No corresponding connection request found") in `accept_complete`**: the OOB connection-reuse cleanup deleted the wallet's own mirror ConnRecord (reciprocal DIDs) while it was still awaiting its own DIDXComplete. The reciprocal-DID case is now detected and skipped; genuine stale-connection-reuse cleanup is unchanged.
- **ConnRecord state regression in self-connections**: removed two redundant post-send `ConnRecord.state` saves that could overwrite a state the (synchronous, self-addressed) delivery had already advanced.
- **`record-not-found` problem report for the holder's `request-credential`** (invitation with both a handshake protocol and an attached offer): after the handshake, the receiver-role OobRecord is deleted, so the inbound attached offer's pthid lookup matched and consumed the wallet's own role=sender OobRecord. A receiver-bound message-type guard now prevents attached offer / presentation-request messages from ever consuming a role=sender OobRecord. Additionally, `V20CredManager.receive_request` falls back to a thread/role-only lookup when the connection-filtered lookup misses (the issuer record was created connectionless), while rejecting records already bound to a different connection.

Normal two-party flows are unaffected: the OOB guard only fires when an inbound offer / presentation request resolves to the wallet's own role=sender record (impossible in a two-party exchange), and the `receive_request` fallback only runs where the previous code already failed with `StorageNotFoundError`.

Regression tests are included for every failure point, parametrized across askar and askar-anoncreds, including guards that normal two-party OOB / did-exchange flows still behave as before. Verified manually end-to-end in a multitenant deployment (invitation with handshake + attached offer received into the same wallet, through to the credential stored and both exchange records reaching `done`).

This is the `1.6.lts` backport of #4174 (same commits, cherry-picked; the fix was developed and manually verified end-to-end against 1.6).

## Related Issue

Relates to #3300 (closed by #4174)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Chore / maintenance

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] I have updated documentation if needed

### DIFF
--- a/acapy_agent/core/oob_processor.py
+++ b/acapy_agent/core/oob_processor.py
@@ -35,6 +35,20 @@ _SENDER_ROLE_MESSAGE_TYPES = {
     MESSAGE_REUSE,
 }
 
+# Message types that are always sent by the *sender* of an invitation to its
+# *receiver* (the attachable messages). An inbound message of one of these types can
+# therefore never belong to our own role=sender OobRecord: it can only match our
+# role=receiver record. This matters for self-connections (one wallet playing both
+# roles for one invitation): once the receiver-role record has been consumed, an
+# inbound attached offer/present-request must not consume the sender-role record,
+# which is still needed to process the receiver's reply (e.g. request-credential).
+_RECEIVER_ROLE_MESSAGE_TYPES = {
+    CRED_20_OFFER,
+    PRES_20_REQUEST,
+    "issue-credential/1.0/offer-credential",
+    "present-proof/1.0/request-presentation",
+}
+
 
 class OobMessageProcessorError(BaseError):
     """Base error for OobMessageProcessor."""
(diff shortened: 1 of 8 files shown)


nb-vivek-bodar:
Closing in favor of #4174 (same fix targeting main, per the standard contribution flow). Happy to re-open or re-submit this as a 1.6.lts backport once #4174 is merged, if the maintainers want it.

PR #4194, opened 2026-08-13
fix: fall back to default write ledger when persisted write_ledger is gone

### LINKED ISSUE #4193: Multitenant: tenant with persisted ledger.write_ledger for a removed ledger becomes unusable and undeletable
## Summary

In a multi-ledger, multitenant deployment, a subwallet's write-ledger choice is **persisted into its wallet record settings** as `ledger.write_ledger` when the tenant calls `PUT /ledger/{ledger_id}/set-write-ledger` ([`base_manager.py#L82-L95`](https://github.com/openwallet-foundation/acapy/blob/a24e86a3a6080f85cbbd38fdcc708ae85e1550f4/acapy_agent/ledger/multiple_ledger/base_manager.py#L82-L95)).

If an operator later removes that ledger from `--genesis-transactions-list` (or leaves the entry but clears its `is_write` flag without setting `read_only: true`), the stored value no longer matches any entry in the write-configurable pools, and `get_write_ledger_config_for_profile` raises `ProfileError` ([`multi_ledger.py#L36-L47`](https://github.com/openwallet-foundation/acapy/blob/a24e86a3a6080f85cbbd38fdcc708ae85e1550f4/acapy_agent/utils/multi_ledger.py#L36-L47)).

Because this runs inside `AskarProfile.__init__` → `bind_providers` ([`profile.py#L134`](https://github.com/openwallet-foundation/acapy/blob/a24e86a3a6080f85cbbd38fdcc708ae85e1550f4/acapy_agent/askar/profile.py#L134)), **every** profile construction for that tenant fails, which permanently bricks the tenant:

- every request with the tenant's token 500s (profile is built per request);
- the wallet **cannot be deleted**: `remove_wallet` builds the tenant profile first ([`base.py#L260`](https://github.com/openwallet-foundation/acapy/blob/a24e86a3a6080f85cbbd38fdcc708ae85e1550f4/acapy_agent/multitenant/base.py#L260)), so `DELETE /multitenancy/wallet/{id}` (and e.g. Traction's innkeeper hard delete) hits the same `ProfileError`;
- there is no API path to amend the stale setting: `PUT /multitenancy/wallet/{wallet_id}` only accepts whitelisted `extra_settings`, which do not include `ledger.write_ledger`.

The only ways out are a config rollback (re-adding the ledger id) or direct wallet-store surgery.

## Steps to reproduce

1. Run a multitenant agent with `--genesis-transactions-list` containing ledgers `A` (`is_write: true`) and `B` (`is_write: true`).
2. As a tenant, call `PUT /ledger/B/set-write-ledger` (persists `ledger.write_ledger: B` plus the endorser alias/DID into the wallet record settings).
3. Remove `B` from the genesis list (or set `is_write: false` without `read_only: true`) and restart the agent.
4. Any call with that tenant's token — including `DELETE /multitenancy/wallet/{wallet_id}` from the admin — fails with:

```
acapy_agent.core.error.ProfileError: ledger.write_ledger in profile settings does not
correspond to a write configurable ledger provided with --genesis-transactions-list
```

Trimmed traceback of the delete path:

```
  File ".../acapy_agent/multitenant/base.py", line 260, in remove_wallet
    profile = await self.get_wallet_profile(...)
  File ".../acapy_agent/multitenant/single_wallet_askar_manager.py", line 119, in get_wallet_profile
    return AskarProfile(...)
  File ".../acapy_agent/askar/profile.py", line 52, in __init__
    self.bind_providers()
  File ".../acapy_agent/askar/profile.py", line 134, in bind_providers
    write_ledger_config = get_write_ledger_config_for_profile(settings=self.settings)
  File ".../acapy_agent/utils/multi_ledger.py", line 47, in get_write_ledger_config_for_profile
    raise ProfileError(error_message)
```

Observed in production (BC Gov Traction, single-wallet-askar multitenancy) with tenants stranded on decommissioned test ledgers; code is unchanged on current `main` (a24e86a).

## Suggested fix

Treat the persisted `ledger.write_ledger` as a preference rather than an invariant: when it doesn't match any write-configurable ledger, log a warning and fall back to the existing default selection (first production / non-production write ledger — the `else` branch at [`multi_ledger.py#L48-L54`](https://github.com/openwallet-foundation/acapy/blob/a24e86a3a6080f85cbbd38fdcc708ae85e1550f4/acapy_agent/utils/multi_ledger.py#L48-L54)) instead of raising. That single change unbricks affected tenants for both normal use and deletion.

At minimum, `BaseMultitenantManager.remove_wallet` should be able to delete a wallet whose stored ledger selection is no longer configured.

Happy to submit a PR for the fallback approach if maintainers agree with the direction.

## Workaround for operators

Temporarily re-add the removed ledger id to `--genesis-transactions-list` with `read_only: true` (sufficient, since entries with `is_write` *or* `read_only` are considered write-configurable — [`multi_ledger.py#L25-L28`](https://github.com/openwallet-foundation/acapy/blob/a24e86a3a6080f85cbbd38fdcc708ae85e1550f4/acapy_agent/utils/multi_ledger.py#L25-L28)); only the genesis URL needs to resolve at startup since pool connections are lazy. Restart, then delete or re-point the affected tenants, then revert.

loneil:
When a subwallet selects a write ledger via PUT /ledger/{id}/set-write-ledger, the choice is persisted into its wallet record settings. If that ledger is later removed from --genesis-transactions-list (or loses its is_write / read_only flags), get_write_ledger_config_for_profile raised ProfileError on every profile construction, leaving the tenant unusable and undeletable (remove_wallet builds the profile first) with no API path to amend the stale setting.

Treat the persisted value as a preference instead of an invariant: log a warning and fall back to the default write ledger selection. A configuration with no write-capable ledgers at all still raises ProfileError.

Adds unit tests for get_write_ledger_config_for_profile (previously uncovered).

Fixes #4193

### DIFF
--- a/acapy_agent/admin/tests/test_route_auth_coverage.py
+++ b/acapy_agent/admin/tests/test_route_auth_coverage.py
@@ -82,9 +82,7 @@ def _collect():
                 route = (
                     path_arg.value if isinstance(path_arg, ast.Constant) else "<dynamic>"
                 )
-                registrations.append(
-                    (handler.id, node.func.attr.upper(), route, path)
-                )
+                registrations.append((handler.id, node.func.attr.upper(), route, path))
 
     return registrations, defs_by_name
 
--- a/acapy_agent/utils/multi_ledger.py
+++ b/acapy_agent/utils/multi_ledger.py
@@ -39,13 +39,14 @@ def get_write_ledger_config_for_profile(settings: BaseSettings) -> dict:
         elif write_ledger in non_prod_write_ledger_pool:
             write_ledger_config = non_prod_write_ledger_pool.get(write_ledger)
         else:
-            error_message = (
-                "ledger.write_ledger in profile settings does not correspond to a "
-                "write configurable ledger provided with --genesis-transactions-list"
+            LOGGER.warning(
+                "ledger.write_ledger %s in profile settings does not correspond to "
+                "a write configurable ledger provided with "
+                "--genesis-transactions-list; falling back to the default "
+                "write ledger",
+                write_ledger,
             )
-            LOGGER.error(error_message)
-            raise ProfileError(error_message)
-    else:
+    if write_ledger_config is None:
         if len(prod_write_ledger_pool) >= 1:
             LOGGER.debug("Using first production write ledger")
             write_ledger_config = (list(prod_write_ledger_pool.values()))[0]
--- a/acapy_agent/utils/tests/test_multi_ledger.py
+++ b/acapy_agent/utils/tests/test_multi_ledger.py
@@ -1,0 +1,65 @@
+from unittest import TestCase
+
+from ...config.settings import Settings
+from ...core.error import ProfileError
+from ..multi_ledger import get_write_ledger_config_for_profile
+
+PROD_WRITE = {"id": "prod-write", "is_production": True, "is_write": True}
+PROD_READ_ONLY = {"id": "prod-read-only", "is_production": True, "read_only": True}
+NON_PROD_WRITE = {"id": "non-prod-write", "is_production": False, "is_write": True}
+NON_PROD_NEITHER = {"id": "non-prod-neither", "is_production": False}
+
+
+def make_settings(ledger_config_list, write_ledger=None) -> Settings:
+    settings = Settings({"ledger.ledger_config_list": ledger_config_list})
+    if write_ledger:
+        settings["ledger.write_ledger"] = write_ledger
+    return settings
+
+
+class TestGetWriteLedgerConfigForProfile(TestCase):
+    def test_write_ledger_in_prod_pool(self):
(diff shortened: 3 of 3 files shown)


esune (commented):
I wonder if we should provide an API to clear the current write ledger (perhaps the settings API?) for scenarios where the agent does not have another write ledger configured that can be used as default?

The scenario I am thinking is: agent is set to write LedgerA, LedgerA is decommissioned/removed and there is no other write ledger in the configuration. As-is we are forced to add a write ledger to the configuration to have the wallet fall-back to it. Maybe this is the desired pattern anyway though... 🤔

jamshale (approved):
It makes sense to fall back to the default and not break a tenant due to this. There's been a bit of history in acapy where errors were chosen instead of logging and recovery.

esune (approved):

PR #4196, opened 2026-08-16
Fix DIF holder presentation signing for did:key

### LINKED ISSUE #4195: [BUG] DIF PE (ld_proof) holder always signs with auto-detected DID; explicit issuer_id in send-presentation is ignored
## Describe the Bug
When responding to a DIF Presentation Exchange request for a `ld_proof`-format credential, `DIFPresFormatHandler.create_pres` always calls `DIFPresExchHandler.create_vp(..., is_holder_override=True)`. This forces the `if self.is_holder or is_holder_override:` branch inside `create_vp`, which unconditionally calls `get_sign_key_credential_subject_id()` to auto-detect the signing DID — completely ignoring `self.pres_signing_did`, i.e. any explicit `issuer_id` passed via `POST /present-proof-2.0/records/{id}/send-presentation`. For a `did:key` credential subject (the norm in `--no-ledger` setups), the auto-detection fails to select a signing key, so the resulting VP is returned with no `proof` at all, and the exchange completes with no error on the holder side. Separately, `issuer_id` also gets silently reset to `None` in `create_pres` unless `presentation_definition` is re-supplied in the same manual `send-presentation` call.

## Steps to Reproduce
1. Run two ACA-Py agents with `--no-ledger`, connected via OOB + DID Exchange (`did:peer`).
2. On each agent, create a `did:key` (`POST /wallet/did/create {"method":"key","key_type":"ed25519"}`).
3. Issue a JSON-LD (`ld_proof`, `Ed25519Signature2018`) credential from agent A to agent B's `did:key` via Issue Credential 2.0.
4. From agent A, send a DIF PE presentation request (`POST /present-proof-2.0/send-request`) for that credential.
5. On agent B, respond via `--auto-respond-presentation-request`, or manually call `POST /present-proof-2.0/records/{id}/send-presentation` with `{"dif": {"issuer_id": "<the did:key>"}}`.

## Expected Behavior
The presentation is signed using the credential's `did:key` subject (either auto-detected correctly, or using the explicitly supplied `issuer_id`), and verification on agent A succeeds (`verified: true`).

## Actual Behavior
The submitted VP has no `proof` field. Agent A's verifier logs:
`DIF-PRES verify_pres result: verified=False errors=[LinkedDataProofException('presentation must contain "proof"')]`
No error or warning is logged on the holder (agent B) side — it returns 200 and completes the exchange as if nothing went wrong.

## Environment
- ACA-Py version: `py3.13-1.6-lts` (`ghcr.io/openwallet-foundation/acapy-agent:py3.13-1.6-lts`)
- Python version: 3.13
- OS: Docker containers (Linux), orchestrated from Windows host

## Additional Context
Reproduced consistently across multiple fresh wallets/connections while building an integration project with home assistant.

jacksonriding:
Fixes #4195.

Preserves explicit issuer_id when the original presentation definition is reused; honors explicit pres_signing_did before holder auto-detection; skips non-local subject DIDs while discovering a local did:key; and fails explicitly when holder signing cannot determine a local DID instead of returning an unsigned VP.

Validation: 74 affected DIF tests passed, 46 optional BBS tests skipped; Ruff passed; git diff --check passed.


jacksonriding:
Forgot to sign :P

swcurran:
The two failing tests look real -- could you @jacksonriding take a look at them?  The ruff one should be trivial, but not sure about the PR Tests -- certainly seems to be in the area that you are updating.

@jacksonriding -- I'm assuming that once this PR is merged, it should be cherry-picked and applied to the 1.6.lts and 1.3.lts branches as well?  Are you able to do that or should I plan to do it?

jacksonriding:
Thanks @swcurran.  I tracked down and fixed both failures. Ruff was a formatting issue in the new test. The PR test
failures were caused by an incorrectly configured requirement mock and an existing suite-mismatch test that still expected an unsigned VP instead of the new explicit error behavior.
 
DIF suite now passes all tests with BBS. Ruff also passes.

Will look into backporting 1.6 and 1.3 lts after merge

swcurran (approved):

PR #4217, opened 2026-09-09
fix: backport DIF holder presentation signing to 1.6 LTS

jacksonriding:
Backports #4196 to `1.6.lts`, as requested in https://github.com/openwallet-foundation/acapy/pull/4196#issuecomment-5531736939.

Preserves an explicit `issuer_id` when reusing the requested presentation definition, honors an explicit `pres_signing_did` in holder mode, and skips non-local subject DIDs during local signing-key discovery. Holder mode now raises `DIFPresExchError` when no local signing DID is available, instead of returning an unsigned presentation.

The four-file patch is identical to the net change in merge commit c148063f307ba33a41402ed1b73e241c07dd55c4 (matching stable patch ID), including the corrected regression tests. No dependency or unrelated main-branch changes are included.

Validation on Python 3.13, using this LTS branch's locked dependencies


swcurran (approved):
LGTM - thanks

